PR #1254, opened 2015-01-19
fix git commit newline

agrajaghh:


agrajaghh:
wow, that was fast :smile:

PR #1368, opened 2015-04-28
Wix installer

laurensb:
I reworked the build process so that is now uses the standard WiX utils in stead of WixProjBuilder and WixIncludeMake, this allows for building through the standard msbuild tool (assuming WiX is installed) and should hopefully be future-proof.

I have tested this on my own PC and as far as I can tell everything works. The uninstall is not as smooth as it can be when Duplicati is still running but it seems to work. There are also some warnings when building but these seem non relevant.

As discussed Duplicati 2 will install alongside Duplicati 1, the installer is compatible with Vista and up, only the x64 installer will be build for now.

Please take a look at the changes and let me know what you think.


kenkendk:
Very nice work!

PR #1381, opened 2015-05-16
update year to 2015

trampi:
Probably not the most important pull request. I have incremented 2014 to 2015 because it did not seem right to me that the webinterface still displayed '2014'. :smile:


kenkendk:
Thanks! :)

PR #1385, opened 2015-05-22
Google Cloud Storage backend

thekrugers:
I've written a backend for Google Cloud Storage as I was not able to get the S3 Compatibility stuff to work properly with my Google Cloud.


kenkendk:
Great initiative!

As you have seen, the Google API is not really playing nice and needs assembly redirects, lacks signing etc.

I will be looking at implementing the Cloud Storage support from the API directly:
https://cloud.google.com/storage/docs/concepts-techniques?csw=1

I cannot merge the pull request as it conflicts, I will take the source code from this pull request and implement it directly.

kenkendk:
I have added the Cloud Storage backend without using the Google API classes.

It does not support the X509 certificate though, do you need that?
Instead it uses OAuth as recommended.

thekrugers:
OAuth is even better, the only reason I did the X509 stuff is because OAuth is not one of my strong coding points.

Thanks

Issue #1449, opened 2015-09-01
Backup run failed on windows with "Persian Calendar" as system default


aghajani:
System.FormatException: The DateTime represented by the string is not supported in calendar System.Globalization.PersianCalendar.
at System.DateTimeParse.ParseExact(String s, String format, DateTimeFormatInfo dtfi, DateTimeStyles style)
at Duplicati.Library.Main.Volumes.VolumeBase.ParsedVolume.Parse(String filename, IFileEntry file)
at Duplicati.Library.Main.Operation.FilelistProcessor.<RemoteListAnalysis>m__0(IFileEntry n)
at System.Linq.Enumerable.WhereSelectListIterator`2.MoveNext()
at System.Linq.Enumerable.WhereSelectEnumerableIterator`2.MoveNext()
at System.Collections.Generic.List`1..ctor(IEnumerable`1 collection)
at System.Linq.Enumerable.ToList[TSource](IEnumerable`1 source)
at Duplicati.Library.Main.Operation.FilelistProcessor.RemoteListAnalysis(BackendManager backend, Options options, LocalDatabase database, IBackendWriter log)
at Duplicati.Library.Main.Operation.FilelistProcessor.VerifyRemoteList(BackendManager backend, Options options, LocalDatabase database, IBackendWriter log)
at Duplicati.Library.Main.Operation.BackupHandler.Run(String[] sources, IFilter filter)
at Duplicati.Library.Main.Controller.<Backup>c__AnonStorey0.<>m__0(BackupResults result)
at Duplicati.Library.Main.Controller.RunAction[T](T result, String[]& paths, Action`1 method)
at Duplicati.Library.Main.Controller.Backup(String[] inputsources, IFilter filter)
at Duplicati.Server.Runner.Run(IRunnerData data, Boolean fromQueue)

kenkendk:
I have merged in 9ae7ea2 to master and tester tested with the fa-IR locale, which I assume has Persian calendar.

kenkendk:
Could you try the build here:
http://updates.duplicati.com/preview/duplicati-2.0.0.90_preview_2015-09-15.zip

aghajani:
Yep it works and the issue is resolved. 
But why you just didn't merge my codes and re-coded all the changes again?

kenkendk:
I was worried that there were other places where the time formating was used, which would cause similar issues.
I was also seeking to consolidate all time-strings that were being serialized into the same format.
[...]

PR #1456, opened 2015-09-07
Solving datetime culture specific bug

aghajani:
When you change your system default calendar, the previous backups will cause exceptions.
After these changes backups will be culture independent.
It will resolve issue #1449


kenkendk:
Great find!

What are the ```#if DEBUG``` statements used for?

I will have a look through the code to see if I can find more incorrect uses of `DateTime.ToString()` and `DateTime.Parse()`.

aghajani:
The `#if DEBUG` statements around try-catch expressions are for debugging only.
The line with `base.Dispose(disposing)` raises exceptions (Object already disposed exception) all the time while debugging and actually not allowing you to debug your code.
So I added the try-catch around it (just for debugging) to stop throwing exceptions while debugging.

kenkendk:
I have implemented your fix in 9ae7ea2, could you verify that it works for you?

Issue #1448, opened 2015-09-01
Unable to restore anything without local db


dannysu:
Version: Duplicati 2.0.0.87- 2.0.0.87_preview_2015-07-03
Operating System: Windows 10, Arch Linux

Preparation:
 - Run Duplicati.GUI.TrayIcon.exe
 - Make 3 folders: Test_Source, Test_Backup, Test_Restore
 - Make a text file with some text and save it to Test_Source folder

Restore Successful Case:
 - Use the Duplicati web interface, create a backup configuration
  - Select Test_Backup folder for "Backup to..."
  - Select no encryption
  - Select Test_Source folder as the source data
  - Deselect repeat schedule
 - Manually trigger a backup
 - Verify duplicati files show up in Test_Backup
 - From the UI, do a restore and choose Test_Restore as the new location
 - Notice that the restore is successful and the text file created earlier is restored

Restore Unsuccessful Case:
 - Use the Duplicati web interface, create another backup configuration
  - Select Test_Backup folder for "Backup to..."
  - Select no encryption
  - Select Test_Source folder as the source data
  - Deselect repeat schedule
 - Don't do a backup, immediately do a restore with this new configuration
 - Got error message: "Failed to get list of backup times: No filesets found on remote target"
 - Do a repair database, then try restore again
 - Got error message: "Failed to get list of backup times"

kenkendk:
There is a parsing issue that has not been fully identified.
See issue #1437 for more details.

If you can shed further light on the problem, it would be much appreciated.

Until the issue is resolved, you can use the 2.0.0.89 version:
  http://updates.duplicati.com/preview/duplicati-2.0.0.89_preview_2015-08-24.zip

This version includes a simpler restore tool:
  https://github.com/duplicati/duplicati/blob/master/Duplicati/CommandLine/RecoveryTool/help.txt

dannysu:
I know for certain that the bug was introduced in 2.0.0.86 and was not there in 2.0.0.85.

dannysu:
This is the culprit: 1be8d16e22711a509be69c5eaf5e35c049c3a224

The last time restore was for sure still working was at commit b08c6df11640dae9491aee32b38318b44e9799cf. Starting from 14413de2c6e6831de6f393e93320402ff352b96d there was several compile errors that were introduced, but they don't seem to be related to the restore bug. I manually fixed the compile errors and narrowed things down to 1be8d16e22711a509be69c5eaf5e35c049c3a224.

[...]

PR #1457, opened 2015-09-08
Default 'time' option to 'now' when not set

dannysu:
This fixes problems with restore that complains about no fileset. The
reason there is no fileset is because 'time' was defaulting to beginning
of time, so no backup files are ever before the beginning of time.

### DIFF
--- a/Duplicati/Library/Main/Options.cs
+++ b/Duplicati/Library/Main/Options.cs
@@ -592,10 +592,12 @@ namespace Duplicati.Library.Main
         {
             get
             {
+                string time;
                 if (!m_options.ContainsKey("time") || string.IsNullOrEmpty(m_options["time"]))
-                    return new DateTime(0, DateTimeKind.Utc);
+                    time = "now";
                 else
-                    return Library.Utility.Timeparser.ParseTimeInterval(m_options["time"], DateTime.Now);
+                    time = m_options["time"];
+                return Library.Utility.Timeparser.ParseTimeInterval(time, DateTime.Now);
             }
         }
         


kenkendk:
Thanks for the PR!

That `Time` option is used multiple places, I will need to test if something is affected by this change.

dannysu:
Yeah, the `Time` code has been like that forever, so more testing is needed.
Speaking of tests, do you have test data you can share out?
On Linux with MonoDevelop, all of the existing tests fail for me.

kenkendk:
The default value of "Time" is checked various places.
If `Ticks == 0` then it is interpreted as "omitted".

Based on your discoveries, it seems that this needs to be checked somewhere.

The filelistfilter is applied here:
https://github.com/duplicati/duplicati/blob/master/Duplicati/Library/Main/Operation/RecreateDatabaseHandler.cs#L135

I cannot reproduce your issue, but if I ommit --time, then I correctly get `Ticks == 0` in:
https://github.com/duplicati/duplicati/blob/master/Duplicati/Library/Main/Operation/RestoreHandler.cs#L72

And all files are returned.

Where do you see the evaluation that causes an empty list?

kenkendk:
I use two sets of test data.

One set is checkouts from:
https://code.google.com/p/dsmcbe/source/checkout

I put r50, r100, r150, r200, r250 - r259 into ~/testdata/DSMCBE and run SVNCheckoutTests.

The other set is a bunch of mp3 files and images, but they are copyrighted so I cannot share them.
The CommandLineOperationsTest should run if you put a set of folders inside the the "~/testdata/data" folder.

kenkendk:
Could it be that you are in a non-gregorian calendar zone?

That could cause this line to break things:
https://github.com/duplicati/duplicati/blob/master/Duplicati/Library/Main/Operation/RestoreHandler.cs#L46

In that case it would be related to:
https://github.com/duplicati/duplicati/pull/1456

kenkendk:
Does this single-line commit fix the issue:
https://github.com/duplicati/duplicati/commit/eee0dccf91753f40bd1d9c1db3da7487f86ff706

If it has to do with calendar settings, then this should fix it here:
https://github.com/duplicati/duplicati/commit/9ae7ea233aca059405e06a3f52639cd89ab95ec9

dannysu:
eee0dccf91753f40bd1d9c1db3da7487f86ff706 does fix repro steps I outlined in #1448, which involve very simple data as I didn't put anything realistic in there during the test. However, when I tested it with my actual backup data things still don't work, so it looks like there are multiple issues and more debugging is needed.

dannysu:
oh and my issue shouldn't be calendar related

kenkendk:
Ok, I'll merge my single-line fix into master then.

Issue #1404, opened 2015-07-03
Improve file picker


kenkendk:
The source folder picker should be changed into a file explorer tree.

At a minimum the tree should have filled and half-filled checkboxes to indicate what should be included.

koobs:
Also, selecting multiple folders is possible (allowed), but doesn't work correctly (only the first path selected is added). 

I suggest we make this issue a meta issue with a list of usability enhancements that can be checked off

- [ ] Fix selecting multiple paths/folders in file picker
- [ ] Support HTML5 Drag'n'Drop of paths into File Browser label dropzone

kenkendk:
I have not attempted to make a "meta issue", but it sounds like a good idea. There are discussions to be had on all issues I suppose :).

The HTML5 Drag-n-drop is unfortunately not possible. The reason is that the drop delivers the files and folders to the app, but does not reveal the real source. That means that Duplicati can read the file/folder names (but not path information) and the file data.

koobs:
@kenkendk Would you like me to create a separate "Improve Web UI usability (meta issue)" issue, or refactor this one to make it cover more than the original description?

renestach:
We have a new file picker now in NGAX

Issue #1067, opened 2014-08-05
Selecting multiple source paths in 'Browse'-window not possible


kenkendk:
_From [cedric.s...@googlemail.com](https://code.google.com/u/103030783990590450414/) on July 30, 2014 09:20:49_

What steps will reproduce the problem? 1. Create a new backup job
2. Switch to 'Source Data'-tab and click 'Browse'
3. Select multiple directories on the same level by pressing 'Ctrl' while clicking
4. Click 'OK' What is the expected output? What do you see instead? I expect all selected Paths to appear in the list of sources - but just the path clicked as the last one is listed. What version of Duplicati are you using? On what operating system? Win8.1 64bit, Duplicati 2.0.0.53- 2.0_preview_2014-07-25 What backend (destination) are you using? Issue is the same with all backends.

_Original issue: http://code.google.com/p/duplicati/issues/detail?id=1040_
##

kenkendk:
I think this was from the greeno theme which is not longer used.

PR #1474, opened 2015-10-07
minor UI improvements for file picker

msilvers:
a few minor UI improvements for file picker.
- disabled selecting multiple folders for backup or restore destinations; multiselect is only allowed when selecting data files
- added checkboxes with tri-state support to show which files are and are not included. 

These issues are mentioned on issue #1404 

filemultiselect
filesingleselect


kenkendk:
Nice!

There seems to be missing some handling of the multiselect in the `select_node` and in `browsePath` callback method? Currently it only returns a single node, but with multiselect it should return a list?

msilvers:
Yes...I think the current UI allows multi select (via Ctrl+Click) but has the same issue.  (seems like the same behavior as reported in issue #1067?)
Either way...I'm working on a fix for that and will submit it soon.

kenkendk:
OK.

I am working on a new UI based on Angular, which already supports filters and multi-select for source folders:
https://github.com/duplicati/duplicati/tree/angular_ui/Duplicati/Server/webroot/ngax

It will not be the default UI right away though, there are many loose ends that needs fixin.

kenkendk:
I merged your changes, if you get the multiselect fixed, you can either do it here or in a new PR.
If you would like credit for the changes, could you please add yourself to the "acknowledgements.txt" file and send a new PR?

Issue #1532, opened 2016-01-06
Can't backup unix root filesystem: "Path cannot contain a zero-length string"


mach-o:
When you try to back up "/", duplicati seems to strip away the slash and the backup fails.

`# mono Duplicati.CommandLine.exe backup $root_target "/."`
`Backup started at 2016-01-04 3:47:47 PM`
`Checking remote backup ...`
`  Listing remote folder ...`
`Scanning local files ...`
`Failed to process path: / => Path cannot contain a zero-length string`
`Parameter name: path`
`Compacting remote backup ...`
`Compacting not required`
`Checking remote backup ...`
`  Listing remote folder ...`
`  Duration of backup: 00:00:05`
`  Files added: 0`
`  Files deleted: 0`
`  Files changed: 0   Data uploaded: 0 bytes`
`  Data downloaded: 0 bytes`
`No files were processed. If this was not intentional you may want to use the "test-filter" command`
`Backup completed successfully!`

See [my post to the mailing list](https://groups.google.com/forum/#!topic/duplicati/DAxQ2UOqKvY) for more details.

mach-o:
So the error happens at [BackupHandler.cs:116](https://github.com/duplicati/duplicati/blob/a51399145b85e8c4cdee835ec852e11a6fd72cec/Duplicati/Library/Main/Operation/BackupHandler.cs#L116) (I'm not sure how to make that link revision-independent, sorry).

I'd be happy to fix this (and other duplicati bugs) if I could get a mono debugger to work properly on Ubuntu. I installed MonoDevelop, which compiles duplicati without issue, but I wasn't able to get its debugger working, and wasn't able to find any documentation explaining how to do so, [as I mention in this mailing list post](http://lists.ximian.com/pipermail/monodevelop-list/2016-January/016445.html). If anyone has pointers to a Ubuntu-compatible mono debugger tutorial, please let me know.

kenkendk:
The BackupHandler.cs link you point to is calling "IsBlockDevice", which is the one stripping the slash:
https://github.com/duplicati/duplicati/blob/a51399145b85e8c4cdee835ec852e11a6fd72cec/Duplicati/Library/Snapshots/LinuxSnapshot.cs#L482

[...]

Issue #1039, opened 2014-08-05
exclude is not supported in 2.0 UI


kenkendk:
_From [bob.hami...@gmail.com](https://code.google.com/u/117560194729930362967/) on June 25, 2014 04:29:25_

What steps will reproduce the problem? 1. Create backup
2. Options, add --exclude=*.mp3 3. What is the expected output? What do you see instead? should be excluding extensions as listed
log showing:
The supplied option --exclude is not supported and will be ignored What version of Duplicati are you using? On what operating system? Duplicati 2.0.0.7- 2.0_CLI_experimental_2014-06-23 - DEBUG What backend (destination) are you using? file & webdav Please provide any additional information below.

_Original issue: http://code.google.com/p/duplicati/issues/detail?id=1012_

kenkendk:
The warning has been removed

PR #1542, opened 2016-01-12
Root unix filesystem as a source path

mach-o:
I believe these changes should fix the problems with unix root filesystems identified in [issue #1532](https://github.com/duplicati/duplicati/issues/1532). I essentially largely eliminated the recurring DIR_SEP variable and substituted its use with calls to System.IO.Path.GetFullPath.

I also renamed a LICENSE.txt file to license.txt because it caused compile problems on linux due to case sensitivity. Since a number of files reference it as lower case, I thought it easiest to make one single change, but if you prefer to keep the capital filename, you could sort through the references.


mach-o:
I'm having some cold feet about this, excludes aren't working properly for me. For example: `backup $target / --exclude=/sys` still tries to backup the tree below /sys.

Not sure whether this is a result of these changes or an unrelated problem.

kenkendk:
Great work!

For the UpdateVersionStamp tool, I don't think there needs to be any fixes. It is only used to set the version number, so it should not be used on `/` at all.

The license renaming is fine.

If you remove the TODO line in UpdateVersionStamp I will merge it.

mach-o:
Thanks! I've removed the line. Please check whether --exclude works for you after (and perhaps also before) merging.

kenkendk:
Your update comment got lost in the stream :)

What should be the issue with --exclude? This only affects the OS specific calls?

mach-o:
I now suspect that the exclude issue is completely unrelated, and is the one already being tracked in [issue 1039](https://github.com/duplicati/duplicati/issues/1039). Is it possible that --exclude simply isn't implemented at all? If so, a warning should appear to the user whenever a user uses the flag that it's being ignored until it's fixed.

mach-o:
Or is Issue 1039 only referring to the form --exclude=*.mp3 rather than --exclude=/boot ?

In any case, neither option currently works for me, but in any case I don't think it has anything to do with this patch.

kenkendk:
I would be surprised if it was related to this patch.
Issue #1039 is just that --exclude gives a warning, not that it does not work. I will close it.
If you need to exclude a folder, you need to specify the path with a trailing slash:
`--exclude=/boot/`

This is slightly confusing as many other tools allow you to use either format, but the format allows you to exclude all files called boot with `--exclude=*/boot` or all folders with `--exclude=*/boot/` at the expense of having to write `--exclude=[.*/boot/?]` to exclude both types.

steilerDev:
@kenkendk Maybe unrelated, but whats your process on getting new versions out? Saying when will this feature be shipped to clients?

mach-o:
kenkendk: Unless I am doing something wrong, this doesn't work for me:

`$ mkdir -p ~/testdir/testsubdir-to-exclude`
`$ touch ~/testdir/testsubdir-to-exclude/exclude_file`
`$ duplicati-cli test-filters ~/testdir --exclude=~/testdir/testsubdir-to-exclude/`
`Including folder: /home/macho/testdir/`
`Including folder: /home/macho/testdir/testsubdir-to-exclude/`
`Including file: /home/macho/testdir/testsubdir-to-exclude/exclude_file (unknown)`
`Matched 1 files (0 bytes)`

This is version 2.0.0.91-2.0.0.91_preview_2015-11-18

mach-o:
@steilerDev : There's no release schedule. On the duplicati Google group, Kenneth has said that "[We do not have an ETA and do not put up timelines](https://groups.google.com/forum/#!searchin/duplicati/release/duplicati/yA5phuhUDSs/hnR2dq_Aqi0J)" at least in part because "[There are a relatively low number of contributors, and I don't know why that is](https://groups.google.com/forum/#!searchin/duplicati/release/duplicati/fFu8p4pA4uY/TGlmut1Hha8J)".

@kenkendk : maybe duplicati.com could be updated to reflect this state of affairs? Currently the prominent "Download duplicati" button does not make clear that it is linking to an unsupported version (1.3.4), which I have the sense is deprecated. Perhaps the 2.0 preview should be the most prominent release. The 2.0 preview, meanwhile, was last updated in July 2014, which will make visitors reasonably start to wonder whether this project has simply been abandoned. Perhaps it could also be made clear that development continues, but without a concrete release schedule, and that further contributors are welcome.

steilerDev:
I see....thats a shame..

It's a great tool. Is there anyway to support you @kenkendk?

kenkendk:
@mach-o: I can reproduce the filter issue. For some reason the `~` is not expanded in the filters part. I think the shell is supposed to expand it, but for some reason it does not.

Try this instead:
```
duplicati-cli test-filters ~/testdir --exclude=/home/macho/testdir/testsubdir-to-exclude/
```

That works for me.

kenkendk:
@steilerDev: Good question. The best way to help out, is to start fixing things or adding features. Just grab whatever issue seems manageable to you. There is also the google groups forum which could use a few responders. If you are more into release planning, UX or similar you can contact @renestach.

mach-o:
Thanks, @kenkendk, your modified example works for me too! Unfortunately, it doesn't fix the underlying problem that I was trying to illustrate with the example I concocted, which is that an `--exclude` of a root directory still doesn't seem to work:

`$ duplicati-cli test-filters / --exclude=/var`
`Including source path: /`
`Including folder: /`
`Including file: /core (285.37 MB)`
`Storing symlink: /initrd.img`
`Storing symlink: /initrd.img.old`
`Storing symlink: /vmlinuz`
`Storing symlink: /vmlinuz.old`
`Storing symlink: /initrd.img`
`Storing symlink: /initrd.img.old`
`Storing symlink: /vmlinuz`
`Storing symlink: /vmlinuz.old`
`Including folder: /var/`
`Including folder: /var/www/`
`Including folder: /var/www/html/`
`Including file: /var/www/html/index.html (11.06 KB)`
`Including folder: /var/tmp/`
<etc...>

Any further thoughts on why this might be?

kenkendk:
Yes, you need to add the trailing slash to target a folder in a filter: 
```
duplicati-cli test-filters / --exclude=/var/
```

mach-o:
Perfect! Thanks so much. I thought I had already tried this and had it fail, but it seems to be working.

mrhahaz:
I to am having an issue like mach-o, not sure if something has changed since it was working for that user. What I am trying:

test-filters /home/kevin --exclude=/home/kevin/VirtualBox\ VMs/ --exclude=/home/kevin/GoogleDrive/ --exclude=/home/kevin/.thunderbird/ --exclude=/home/kevin/SharePoint/ --exclude=/home/kevin/.backups/ --exclude=/home/kevin/OneDrive/ --exclude=/home/kevin/nextcloud/ --exclude=*/.cache/ --exclude=*/cache/ --exclude=*/*cache/ --exclude=/home/kevin/scripts/ --exclude=/home/kevin/scriptsfiles/ --exclude=/home/kevin/.logs/ --exclude=/home/kevin/.wine/ --exclude=/home/kevin/.mozilla/ --exclude=*/gvfs-metadata/ --exclude=*/syncthing/ --exclude=*/.mono/ --exclude=*/evolution/ --exclude=*/chromium/ --exclude=*/Application\ Cache/ --exclude=/home/kevin/installs/


Correctly excludes the files specified in the exclude with a calculated size of ~900mb, I initially was using /VirtualBox\ VMs/ instead of the full path and the test was still including those files. So after some trying things ended up with the options above that WORK when I use test-filters. However, when I copy/paste those exact filters behind my backup command:

backup "file:///mnt/backup/home" "/home/kevin/" --log-level="Information" --log-file="/tmp/duplicati.log" --backup-name="home_backup" --passphrase="************" --dbpath="/home/kevin/.config/Duplicati/TUVQZAKXST.sqlite" --encryption-module="gpg" --compression-module="zip" --dblock-size="20gb" --keep-time="3M" -exclude....

It now begins creating backups in excess of 40Gb (most of that size is due to my VMs). Is there a different syntax I need to be using? Or am I missing something? I have tried creating this with the GUI and moved to using the command line after that didn't work. I tried changing my backup to path, database, job name and removing all temp files I could find.

Backup jobs where I don't need to exclude anything are working AWESOME! Once again in summary, when testing the filters my excludes work but when doing an actual backup either through the GUI or cli they do not.

Issue #1391, opened 2015-05-31
Restoring without local database takes ages


jarmo:
I tried to restore a single ~500kB file from a ~500GB backup without a local database. It takes hours and hours and i can see a lot of `Downloading file (27,22 KB) ...` messages. As soon as i specify local database with `--dbpath` then the restore is done in minutes.

It does take hours even if the backup destination is local disk.

There seems to be something fundamentally wrong when restoring a single file with exact path from local destination takes more than few minutes.

Why does it take that long?

Using latest duplicati 2.x version.
##

jarmo:
It finally finished. Pay attention to the duration.

```
Checking remote backup ...
  Listing remote folder ...
Searching backup 0 (25.05.2015 1:44:00) ...
Checking existing target files ...
  1 files need to be restored (563,18 KB)
Scanning local files for needed data ...
1 remote files are required to restore
  Downloading file (49,95 MB) ...
  0 files need to be restored (0 bytes)
Verifying restored files ...
Restored 1 (563,18 KB) files to c:\Users\Jarmo\Desktop\restore
Duration of restore: 15:24:03
```

jarmo:
Restoring now all files from a 70 and 700 GB backup without local database files. Wish me luck, because this is a real situation this time and not just a test. Currently seeing "Downloading file (47.78 MB) ..." appearing REALLY slowly although the backup location is a local disk.

kenkendk:
If you do not have a local database, Duplicati will build one before continuing.
There is some logic in the latest (2.0.0.90) version that checks if it can avoid to restore the entire database, which should speed things up. If something is not matching correctly, Duplicati will start looking for the needed information in potentially all files.

If you need to perform the restore, I recommend that you run the "repair" command first to generate the local database. This will make it possible to stop and resume the restore process without needing to generate the database again.

The 2.0.0.90 version also features a recovery tool, which works with a simple text file instead of a database.

jarmo:
@kenkendk thank You for your reply.

I can now tell more about the whole situation, described in the order of steps done:

1) HDD failed on 15th Oct;
2) Made clean install of Windows 10 at Oct 16th;
3) Started two simultaneous restore processes at Oct 16th of ~70GB and ~700GB backups - both without having a local database present;
4) At the evening of Oct 18th I saw that restore process for ~70GB was successfully completed, but there were no files restored from my ~700GB backup due to error similar to #1143 - not a single file - same problem can be reproduced on a Windows and Linux environment;
[...]

PR #1577, opened 2016-02-14
Fix for performance issue on restore.

FootStark:
Massively reduces restoration time caused by inefficient query in `BlockMarker.UpdateProcessed`.


kenkendk:
What was the motivation for removing the two other index entries? Are they not used?

kenkendk:
Discussion is in #1391

PR #1584, opened 2016-02-18
Performance improvement on restore

FootStark:
Improves performance when restoring files by fixing some DB index and SQL related issues .


kenkendk:
I am impressed that you were able to unfold the queries and optimize them :)

Awesome work!

FootStark:
I will do something about the updateProcessed on the weekend. Learning SQLite -  It's actually kind of fun.

Issue #57, opened 2014-08-05
Files with errors are marked as deleted


kenkendk:
_From [kenneth....@gmail.com](https://code.google.com/u/117596759558990029378/) on April 20, 2009 13:48:33_

If a file is locked or unreadable for some reason, it is skipped.
Unfortunately this means that the file is marked as if it was deleted.
When restoring the backup, the file will be missing rather than include the
version that was avalible previously.

The user must then restore the file from the previous backup manually.
There is no data loss, but it is not the intuitive way to handle the problem.

_Original issue: http://code.google.com/p/duplicati/issues/detail?id=57_

PR #1590, opened 2016-02-27
Better performance on restore from XXL DataSet

FootStark:
Updated BlockMarker to work on the Blocks table directly
New update statistics leveraging SQLite Triggers
Prepared for file completion watcher to trigger file verification as soon as restoring a single complete file is done.


FootStark:
Hmm, there's something strange with the Travis-CI auto build:
When I started the request it tried to build and failed with:
`Mono.Data.Sqlite.SqliteException : Abort due to constraint violation columns Name, State are not unique` called from `LocalDatabase.UpdateRemoteVolume`
As I didn't touch anything near that place, I now commited a comment (no real code changed) and the build ran again and it passed. 
Another curiosity: Before first try started there was a build from branch `concurrent_processing #57` shown as active for more than 24 hours, and that is now marked as passed.
I thought you might want to know.

kenkendk:
There is a race somewhere with the the backup for really short runs (1 or 2 files) that causes the `Constraint Violation` error. I usually re-run the test when that happens, but I think only I can see the re-run button. 
I am working on the `concurrent_processing` branch which will fix such races (among other nice things).

There are other issues with Travis, but overall I like the idea of running the checks on each commit. `build #57`does not appear to have lasted for 24 hours on my end, it says 47 minutes.

For this PR, I will look it through Wednesday, but it looks very nicely done :)

FootStark:
Yes, it jumped to 47 Minutes after the build for the commit failed, from being yellow and "active for 24 hours" before. But I like the auto check, too, it is just good to know that a rerun can help and keeps one from serching non-existent errors to long.

I'm already excited how `concurrent_processing` will perform. I also introduced some changes to put basic multithreading to work in another branch, and with the to-come changes to SharpAESCrypt I got download and decryption during restore to run at 90% CPU and 200GB/hour on a quad core. I hope I get it ready to commit this week.

kenkendk:
That looks really nice, and it passes the tests, so I assume it handles metadata correctly as well.

There is a comment here that suggests otherwise:
https://github.com/duplicati/duplicati/pull/1590/files#diff-c362e5f0eb09ac0715893cf68628cce8R55

Any reason for that ?

FootStark:
Actually this whole tracking thing is only used to display progress update right now, so no real harm can be done. I was not sure whether to keep a counter for metadata blocks during restore, so I just excluded them for now. If the file verification switch should rely on it, it can be added.
I what I did not figure out yet: Is there always exactly zero or one metadata block per file?

kenkendk:
In principle there could be multiple metada blocks pr. file, but the implementation does not support it yet. 
The backup process will give a warning and skip it if you attempt to store metadata that exceeds a single block. The restore, compact and repair processes assume that there is only a single block for metadata.

PR #1596, opened 2016-03-02
Download piping - Multithreading on Download

FootStark:
Parallelized HMAC and decryption using DirectStreamLink (as in SharpAESCrypt)
I forked DoGet it in two methods to make it easier to verify (new: coreDoGetPiping, old: coreDoGetSequential)
Introduced an option "enable-piped-downstreams" to activate new behaviour (I trust my changes, but let's consider it experimental right now)


kenkendk:
Ok, this looks fine.

I agree with the conservative approach, lets introduce the feature, but keep it turned off by default.
Brave users can enable it and we get some real-world feedback :)

kenkendk:
Btw, you should add a pull request where you add yourself to the acknowledgements file:

FootStark:
Ok, I want to add another small change (order of volumes) and integrate the new SharpAESCrypt-Library after your approval (also with option and disabled by default), then we should ask in #1391 to test a new preview.

PR #1601, opened 2016-03-05
Reorganized `ScanForExistingTargetBlocks`

FootStark:
This change tackles several challenges:
1. It fixes a bug where duplicati truncates forthwritten files backuped
with a size being a multiple of blocksize instead of renaming them.
2. It tries to verify file data on beginning when scanning for target blocks. If successful, marks files to
not be reverified in the end. This helps to speed up restore-into-existing scenarios.
3. On overwrite restore, removes readonly attribute from files having to
be modified. Also only uses read mode, allowing scanning on Read-only files. This does not include necessary metadata changes, as they are not detected in `ScanForExistingTargetBlocks`. If this is correct actually depends on the overall strategy for how to handle Read-Only files on restore., which has to be discussed. When applying metadata, the attribute would be reset.
4. A speedup by computing block and file Hashes in parallel when scanning.

To accomplish this, I added a Field `"DataVerified"` to the temp file tables. This can also be used when incorporating verification into the block restore procedure.
I had to put the function `SetFileDataVerified`to mark files in the BlockMarker because of the Transaction, so it is actually now a FileAndBlockMarker.
I decided to adjust file sizes already when scanning, not in the very end after restore (where it actually takes up a pretty long time). The downside of this is: If the backup is not available later on, all files larger than in the backup have already been truncated. But I do not think that is a problem: If someone decides to restore the files, he should not want the old ones around anyway.

As this change goes a little deeper into the whole restore process, you might verify thoroughly, but my tests have been successful yet. I hope you agree on the read-only attribute handling.


kenkendk:
I agree with the handling strategy for read-only.
I see that you have identified another potential weirdness with restoring metadata, but your changes work as good as the current implementation so I think we can deal with that seperately.

I will re-submit to Travis as it hit the race for the tiny backup, and once it passes I will merge it.

FootStark:
The key to Metadata handling will be to compare it. When tackling this, it should also be discussed where and how to store it. When testing, I had to download 20gb of  a 90gb backup only for Metadata, with zero data blocks to be restored. I will open an issue to discuss this when I am back in a week.

Issue #1582, opened 2016-02-15
"Deleting unwanted files" takes ages. Duplicati busy since 5 days.


deHoeninger:
Hi,
last Thursday, I changed the max time to keep from 90 to 70 days. The Backup had a size of round about 950GB on GoogleDrive. Round about 200 GB were deleted on Thursday from GDrive as well, which makes sence. No other changes on Gdrive since then.
My problem is, that Duplicati is still busy and doing stuff for that cleanup task "deleting unwanted files". According to the Live Log its deleting stuff from different tables and that since 5 days now. Its constantly consuming 100% CPU and there is no progress bar or any other information that reflects the current progress.
I'm using version: Duplicati 2.0.0.98- 2.0.0.98_preview_2016-01-27

Why is it taking so long?
What can I do?
What will happen, if I kill duplicati in that state?

According to the live logs it's doing DB stuff, but the throughput seems to be very slow, according to the timestamps:
```
2016-02-15 10:38: Starting - ExecuteNonQuery: DELETE FROM "File" WHERE "BlocksetID" IN (SELECT DISTINCT "BlocksetEntry"."BlocksetID" FROM "BlocksetEntry", "Block" WHERE "BlocksetEntry"."BlockID" = "Block"."ID" AND "Block"."VolumeID" = ? UNION SELECT "BlocksetID" FROM "BlocklistHash" WHERE "Hash" IN (SELECT "Hash" FROM "Block" WHERE "VolumeID" = ?)) OR "MetadataID" IN (SELECT DISTINCT "BlocksetEntry"."BlocksetID" FROM "BlocksetEntry", "Block" WHERE "BlocksetEntry"."BlockID" = "Block"."ID" AND "Block"."VolumeID" = ? UNION SELECT "BlocksetID" FROM "BlocklistHash" WHERE "Hash" IN (SELECT "Hash" FROM "Block" WHERE "VolumeID" = ?))
2016-02-15 10:38: ExecuteNonQuery: DELETE FROM "Fileset" WHERE "VolumeID" = ? took 00:00:00.000
2016-02-15 10:38: Starting - ExecuteNonQuery: DELETE FROM "Fileset" WHERE "VolumeID" = ?
2016-02-15 10:38: ExecuteNonQuery: DELETE FROM "FilesetEntry" WHERE "FilesetID" IN (SELECT "ID" FROM "Fileset" WHERE "VolumeID" = ?) took 00:00:00.000
[...]

Issue #1283, opened 2015-02-09
Local database should store file paths in a space-optimized fashion


bks:
Using Duplicati 2.0.80, I have a ~75 GB backup set, but the local database is taking up 1.999 GB itself. 

From poking at `%APPDATA%\Duplicati\LRELCBKGXP.sqlite` with sqlite3, it appears that a great deal of that space is used storing the full path for every file in the database. Running `SELECT sum(length(Path)) FROM File;` reports that the actual path strings occupy ~572 MB not counting overhead. Running `DROP INDEX FilePath;` and comparing the `.sqlite` file size before and after indicates that the index uses ~730 M, so the total space cost of the path strings is probably closer to 600 MB in the main table and then the same amount again in the index.

I did a bit more analysis of the paths using python:

``` python
tot, paths = 0, 0
for name in db.execute('SELECT Path FROM File'):
     tot += name[0].rindex('\\')
     if name[0].endswith('\\'): # this File is for a directory
         paths += len(name[0])
print 'Total redundant path components: %d, total unique path components %d' % (tot, paths)
```

It tells me that the leading path components contribute ~418 MB of data for only ~36 MB of unique leading paths; the 418 MB is then doubled by the FilePath index.

So, I suggest optimizing the storage of file paths. One simple way is to partially dematerialize the full path into `dirname` and `basename` parts:

``` sql
CREATE TABLE "Dirname" (
      "ID" INTEGER PRIMARY KEY,

 -- this creates an implicit index, but the UNIQUE constraint is important
      "Dirname" TEXT UNIQUE NOT NULL
);

CREATE TABLE "File" (
 (
        "ID" INTEGER PRIMARY KEY,
        "Dirname" INTEGER NOT NULL,
        "Filename" TEXT NOT NULL,
        "BlocksetID" INTEGER NOT NULL,
        "MetadataID" INTEGER NOT NULL
);

-- If the UNIQUE constraint is really only (Dirname, Filename), it would be nice to take the other
-- columns out of the index. But if not, this is still a lot smaller than before.
CREATE UNIQUE INDEX "FilePath" ON "File" ("Dirname", "Filename", "BlocksetID", "MetadataID");
```

This fairly simple change will probably shrink the local database by 50+%.
##

kenkendk:
Great investigation!

I had considered doing a full split of the path, but that gives very large chained lookups.
Your solution would work with minimal overhead, given that most folders contains more than a single file.

rocifier:
[...]

PR #1620, opened 2016-03-13
Mitigate slow volume deletion

FootStark:
This fix should mitigate the slow deletion of volumes (as explained in #1582 ) a little bit by caching a slow subquery in a a temporary table.
It is still far away from being optimal, but already helps by speeding up to about 2.5x speed.

Best way to speed things up is to process `RemoveRemoteVolume` for all removed volumes at once. This should be easy, as it seems the db-ops are collected anyway until `FlushDbMessages` is called, prior to being written to DB. I could implement a grouping with execution as whole there, but it would change the order of db-ops, which are alternating deletes and writes to dbLog in the list. Even though they are enclosed in a transaction anyway (all or nothing), it would change the timestamps in the db-log, which would also be written all at once. If you do not see a problem with this, I would implement the change.

Another small optimization would be adding an index for VolumeId to the database.
I added this with two other proposals (improve `BlocksetEntry`, delete `BlocklistHash` ) as comments to the `schema.sql` file. I was not sure how and when the database schema can be changed, so I didn't want to break anything. About the `BlocklistHash` table: I could not find a practical use and believe it is a remainder of some old schema. If so, it should be removed from code to not slow things down.

### DIFF
--- a/Duplicati/Library/Main/BackendManager.cs
+++ b/Duplicati/Library/Main/BackendManager.cs
@@ -364,18 +364,30 @@ namespace Duplicati.Library.Main
 	        			entries = m_dbqueue;
 	        			m_dbqueue = new List<IDbEntry>();
 	        		}
-	        	
+
+                // collect removed volumes for final db cleanup.
+                HashSet<string> volsRemoved = new HashSet<string>();
+
 	        	//As we replace the list, we can now freely access the elements without locking
 	        	foreach(var e in entries)
 	        		if (e is DbOperation)
 	        			db.LogRemoteOperation(((DbOperation)e).Action, ((DbOperation)e).File, ((DbOperation)e).Result, transaction);
-	        		else if (e is DbUpdate)
-	        			db.UpdateRemoteVolume(((DbUpdate)e).Remotename, ((DbUpdate)e).State, ((DbUpdate)e).Size, ((DbUpdate)e).Hash, transaction);
+                    else if (e is DbUpdate && ((DbUpdate)e).State == RemoteVolumeState.Deleted)
+                    {
+                        db.UpdateRemoteVolume(((DbUpdate)e).Remotename, RemoteVolumeState.Deleted, ((DbUpdate)e).Size, ((DbUpdate)e).Hash, true, transaction);
+                        volsRemoved.Add(((DbUpdate)e).Remotename);
+                    }
+                    else if (e is DbUpdate)
+                        db.UpdateRemoteVolume(((DbUpdate)e).Remotename, ((DbUpdate)e).State, ((DbUpdate)e).Size, ((DbUpdate)e).Hash, transaction);
                     else if (e is DbRename)
                         db.RenameRemoteFile(((DbRename)e).Oldname, ((DbRename)e).Newname, transaction);
-	        		else if (e != null)
-	        			m_stats.AddError(string.Format("Queue had element of type: {0}, {1}", e.GetType(), e.ToString()), null);
-	        			
+                    else if (e != null)
+                        m_stats.AddError(string.Format("Queue had element of type: {0}, {1}", e.GetType(), e.ToString()), null);
+
+                // Finally remove volumes from DB.
+                if (volsRemoved.Count > 0)
+                    db.RemoveRemoteVolumes(volsRemoved);
+
 	        	return true;
 	        }
         }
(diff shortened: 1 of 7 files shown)


kenkendk:
The whole `FlushDbMessages` call is a bit of a mess. It is a workaround to ensure that there is only a single thread that writes to the database, as some SQLite builds do not support multithreading support.
In my concurrent_processing branch I have fixed this by have a single-threaded interface instead. In that version, the operations on the database will be more immediate and more mixed than the current.

This will also ensure that the database is committed at then end, unlike now, where a rollback can cause an error log entry to be lost.

I think there should be no issues with grouping the deletevolume calls if that works. You just need to consider that the program can stop after having deleted some remote files, but before having written the desire to delete them into the database. That would cause it to try to recreate the files. If they can be marked for deletion early on, but not execute the removal until later that should be fine. The remote verification process could check for such volumes and execute the cleanup there to solve the issue.

To update the schema:
1) Edit the schema.sql to the new format (new databases are created with this).
2) Bump the version number near the bottom of schema.sql
3) Add a new file with the version number prefixed that upgrades from the earlier version.
4) Make sure the new file also updates the version number so it is not executed again.
5) Set the build-type to "embedded resource" for the new file

kenkendk:
I think you are correct about the blocklisthashes. It is used during restore (iirc), but there is no need to store it explicitly.

kenkendk:
It looks like even Debian has picked up on SQLite 3.8, so if the WITHOUT ROWID improves things, we can include that as well.

kenkendk:
Now that you are looking deeply into SQL statements and storage, have you seen this idea:
#1283

kenkendk:
Could you drop me an email so I can add you to our Slack channel?
It is easier to discuss things there I think.

FootStark:
I have never used slack before, do I have to create an account first? 
You can mail me here:
use.01f9.once@lueck-senses.de

The address is temporary (public space here), but I will respond with a permanent one if you mail me.

FootStark:
Hmm. Travis has an very old SQLite Version as stated and discussed [here](https://github.com/travis-ci/apt-package-whitelist/issues/368).
I could dynamically check the SQlite-Version in use and adapt the schema to either use `WITHOUT ROWID` or not, but I think the complexity is not worth it, so maybe I better revert... Suggestions?

kenkendk:
Not sure how much work to put into it, but it would be nice if autodetection was done in the schema file.

Yyou can query the sqlite version with this:
```
select 
substr(sqlite_version(), 0, instr(sqlite_version(), ".")) as "Major",
substr(substr(sqlite_version(), instr(sqlite_version(), ".") + 1), 0, instr(substr(sqlite_version(), instr(sqlite_version(), ".") + 1), ".")) as "Minor"
```

But then you need to conditionally create the table, which I don't know how to do.

How much space does it save to use `WITHOUT ROWID` ?

FootStark:
Advantages of `WITHOUT ROWID` should be:
- It saves a separate rowid (long?, so 8 Bytes per row)
- It saves the complete Forward index as it is replaced with PK(again long?,3 x 8 Bytes per row)
- With PK as rowid and the backwards index covering now, it saves a row lookup for Forward and Backward joins.

With a 100GB backup and 100KB blocks we have at least 1M blocks, so we should spare a minimum of around 30 MB per 100GB backup size. plus a lookup. With 1TB backup, could be around 300MB. I can check with a rebuild on my 90GB test backup tonight.

FootStark:
Ok, I was curious now:
Test results (all sizes with `VACUUM` before):
Backup ~90GB (DVD-images), BlockSetEntry-COUNT= 971.020

- DB-size after normal rebuild: 193MB
- DB-size with BlockSetEntry converted to `WITHOUT ROWID`: 160MB
- DB-size with BlockSetEntry removed completely: 146MB

That means, it saves about 2/3rd of BlockSetEntry size (14MB vs 47MB). I would say that matters...

kenkendk:
If Travis does not support that version of SQLite, there are probably a lot of boxes out there that does not support this either (Synology, QNAP, etc).

Would you modify the parser to just remove the `WITHOUT ROWID` statements from the SQL files?

FootStark:
I thought about something like a general condition's tag, maybe:
`{#ifver >=3.8.2)} WITHOUT ROWID {#fi}`
That can be used for broader cases (like whole statements), too.

kenkendk:
Fine by me. If you think it can be done without too much effort.

kenkendk:
But I don't think it should get too much attention. After all, we will likely not want drastically different schemas such that the rest of the code needs special care to handle all versions of the schema.
In this case it is simple because it works the same with and without the extra directive.

FootStark:
I agree on that the code should not have to handle different db-schemes. I woudn't want to support that, too. But I still wrote a small preprocessor to allow user's with more modern libraries to take advantage of that. I see that as coding sport... ;)

kenkendk:
Neat :)

PR #1621, opened 2016-03-13
Added option aes-set-threadlevel

FootStark:
You might have to readd the SharpAESCrypt.exe. I just copied it, but github thinks it's changed?!?


kenkendk:
Yep, I added the newest build again.

PR #1638, opened 2016-03-19
new tray icons

### LINKED ISSUE #1630: tray icon is disappearing on windows
Running duplicati 2.0.1.4 on Windows 10:

* after startup the tray icon is grey.
* during the backup process the tray icon is green
* when the backup process is finished, the tray icon gets black/transparent: 

tray

agrajaghh:
fixes #1630 

Made the play button one pixel smaller (@FootStark was right, a bit smaller looks better) and darkened the grey a bit more:
tray_icons_all

I added:
* the illustrator file
* exported svg files for each icon
* batch file to create the *.ico files 
* the created ico files

I removed the old icons, but I can leave them in of course if you prefer that?

The *.ico files include 16, 24, 32, 48 and 256px sizes, just like the "ICO for Windows 7, Windows 8, Vista and XP" file from http://icoconvert.com . I didn't use this website because the resize algorthim used there is really shitty and doesn't produce nice small versions. Instead I created a windows batch file which creates the ico files from the svg files using ImageMagick and Inkscape. I guess it would be easy to create a OSX or linux shell script if needed.

Do you need something else?


kenkendk:
Great work!

PR #1647, opened 2016-03-22
Use SmallIconSize from SystemInformation in Win-Tray

### LINKED ISSUE #1646: distorted display of the tray icons on windows
since #1630 is closed (and has a different topic)

quote from there:
hmm, nitpick but the ico I created looks like that (1600%)
ico

but its displayed for me on windows 10 like that:
tray

It's displayed as exactly 16x16px for me in the taskbar. And there is a 16x16 version included in the ico file. Maybe windows is using 24x24 (or another version) and is resizing it, but that wouldn't make sense :grin:

I'm using version 2.0.1.8

FootStark:
Fixes #1646


FootStark:
@kenkendk: Just saw you already commited something fixing the icon issue. Shall I close this PR?

kenkendk:
I will merge this and update my own patch. I was working on the OSX icons, and then I implemented your  suggestion with SmallIconSize as well.

Issue #1244, opened 2015-01-14
Support for Office365 OneDrive accounts


kenkendk:
The Office365 accounts with OneDrive needs a special OAuth flow, going through Azure AD.

TimoD:
Hello,

I receive an exception trying to backup on OneDrive:

Failed: Der Remoteserver hat einen Fehler zurückgegeben: (413) Request Entity Too Large.
Details: System.Net.WebException: Der Remoteserver hat einen Fehler zurückgegeben: (413) Request Entity Too Large.
   bei Duplicati.Library.Main.Operation.BackupHandler.HandleFilesystemEntry(String path, FileAttributes attributes)
   bei Duplicati.Library.Main.Operation.BackupHandler.Run(String[] sources, IFilter filter)
   bei Duplicati.Library.Main.Controller.<Backup>c__AnonStorey0.<>m__0(BackupResults result)
   bei Duplicati.Library.Main.Controller.RunAction[T](T result, String[]& paths, Action`1 method)

Is this because of the mentioned OAuth-handling for Office365-accounts? I thought about, because the error occured when my free space is full and only 1 TB is left on my OneDrive now. If so, when will there be a solution?

Thanks,
Timo

TimoD:
Hm, after I read elsewhere that OneDrive has a 100 MB file size limit, I tested with 99 MB file size instead of 100 MB which I choose first. The backup is working now.

wadami:
I'm also interested in the support for Office365 accounts, since the company I work for is providing OneDrive for Business (OD4B) account to store backups.

m1r4c:
Also voting for this feature, especially now that microsoft is offering unlimited storage for office365 customers.

FootStark:
A OneDrive for Business backend has now been added and is available to test in build [Canary_v2.1.0.9](https://github.com/duplicati/duplicati/releases/tag/v2.0.1.9-2.0.1.9_canary_2016-03-26).

PR #1648, opened 2016-03-22
A SharePoint / OneDrive for Business backend

FootStark:
A SharePoint / OneDrive for Business backend using the Microsoft SharePoint client side object model.
Some remarks:
* OneDrive for Business is actually implemented as a SharePoint site, thus the handling is identical from a technical point of view. I still added two backends to supply separate protocol keys if this should change. From the history of OD4B this is entriely possible some day.
* I picked the protocol keys "`mssp`" and "`od4b`" for the backends. I didn't know of a better naming convention, but they can of course be easily changed.
* The SP.CSOM libraries are the official ones from the [MS NuGet-package](https://www.nuget.org/packages/Microsoft.SharePointOnline.CSOM/) (v16.1.x.x). Yet I decided to just copy the two dll's needed to `thirdparty` and remove the nuget from the config, as this seems to be how the other components did it too.
* I hope I updated the license project correctly. MS provides an rtf with the EULA (included) and I copied it manually to a license.txt and added this too. Not sure if this can be removed, please check.
* I couldn't find out how to put the new backends under `Propietary` section in the UI instead of the `Others` section. This might fixed.

This feature solves #1244


kenkendk:
If the EULA is included that should be fine too.

The proprietary label is set by the GUI so the new backend names need to be added there. I will do that.

Issue #1659, opened 2016-03-26
2.0.1.7 System.Globalization Error 


luni64:
After updating to 2.0.1.7 I get the following Error when I start a Backup: 

![image](https://cloud.githubusercontent.com/assets/12611497/14058255/f3e3ea48-f31a-11e5-8751-f0b91f8f95f5.png)

Thanks
Lutz

renestach:
The green bottom bar looks like the old user interface. Does it work if you switch to the new one (which should be default now)?

luni64:
Oh, the new interface looks great, didn't notice this. 

Anyway, same bug: 
![image](https://cloud.githubusercontent.com/assets/12611497/14059065/8b0f120c-f335-11e5-87ee-e4ece549c75b.png)

FootStark:
Same as #1606. Which environment?
Please install .NET Framework 4.5

luni64:
Does it neet to be exactly 4.5? I have installed 4.0 and 4.6.1. 
Environment is Win10, Browser Chrome V49. Duplicati Server runs on a Suse Linux system. 

![image](https://cloud.githubusercontent.com/assets/12611497/14059744/fb475bbc-f34c-11e5-826c-0192fc4c8632.png)

FootStark:
No, 4.6.1 is fine. But the error is happening on the server, so it's the MONO environment which would need an update. Sadly, I am not sure what mono version complies to what .NET Framework version and what is currently available as a package for Suse. We tried to fix the next release, that locales are not set if they are not available. So if an Mono-update does not help, you can try this.

luni64:
Tried to update mono. I now have Version 3.0.6-3.1.3. Error still there.

If I understand you correctly I'll have to wait for the next duplicati update? No problem for me. Any idea when the update will be available?

FootStark:
Not sure if this is compatible, but there are mono versions > 4... ;http://www.mono-project.com/docs/getting-started/install/linux/#opensuse-and-sles
For duplicati there are canary-builds (=experimental nightlys) pretty often that incorporate the latest changes. There might be even one tonight.

renestach:
I think at least 3.10 is required.

kenkendk:
There is a new canary update available now (which should fix the problem):
https://github.com/duplicati/duplicati/releases

luni64:
[...]

PR #1660, opened 2016-03-26
Hopefully fixes NoMethodImplementation on .NET40

FootStark:
The set was already wrapped, now the get is too.
As in #1659.

### DIFF
--- a/Duplicati/Library/Main/Controller.cs
+++ b/Duplicati/Library/Main/Controller.cs
@@ -382,6 +382,16 @@ namespace Duplicati.Library.Main
 		}
 
         /// <summary>
+        /// Attempts to get the locale, but delays linking to the calls as they are missing in some environments
+        /// </summary>
+        [System.Runtime.CompilerServices.MethodImpl(System.Runtime.CompilerServices.MethodImplOptions.NoInlining)]
+        private static void DoGetLocale(out System.Globalization.CultureInfo locale, out System.Globalization.CultureInfo uiLocale)
+        {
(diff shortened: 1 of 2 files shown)


Duplicati/Library/Main/Controller.cs
```diff
@@ -533,8 +543,8 @@ private void SetupCommonOptions(ISetCommonOptions result, ref string[] paths, re
             if (m_options.HasForcedLocale)
             {
                 var locale = m_options.ForcedLocale;
-                m_resetLocale = System.Globalization.CultureInfo.DefaultThreadCurrentCulture;
-                m_resetLocaleUI = System.Globalization.CultureInfo.DefaultThreadCurrentUICulture;
+                // Wrap the call to avoid loading issues for the getLocale method
+                DoGetLocale(out m_resetLocale, out m_resetLocaleUI);
```
kenkendk: I think there should be a try/catch around here also?

FootStark:
Hmm, I am actually not sure when and how the JIT throws an exception.
But It's worth a shot. I'll add it to the pull request.

kenkendk:
The error happens when the method is executed. The loader attempts to fix the call address of the methods, but fail to find them. So the exception is throw where the method is invoked. By wrapping like this it it defers the execution to the `DoGetLocale` and `DoSetLocale` method calls (but the NoInline option is required to avoid the compiler inlining it and canceling the encapsulation.

My comment was a bit premature because the idea was that the exception is thrown if the user attempts to override the local on .Net 4.0. If the user does not set the override, it does not care that the method is missing. But the extra safeguards cause no harm.

Issue #1657, opened 2016-03-24
V 2.0.1.7 Local database not created properly from backup data during Restore


chris-cashin:
When restoring using 2.0.1.7- without the backup database files are randomly excluded.
An error message indicating that the metadata could not be applied and that the file is not found is generated. If the backup database is provided the process works correctly and no files are excluded.

This seems to be a fairly consistent issue.
No back-end was involved in testing just local process
Sample Command line without --dbpath that fails follows: 
"C:\duplicati\Duplicati.CommandLine.exe" restore "file://C:\SOURCE\\" --passphrase="1234" --overwrite="TRUE" --time=now --restore-path=D:\RESTORED --dbpath="" --Prefix="PPL" --full-block-verification="true" --no-local-blocks="TRUE" --disable-module="console-password-input"

FootStark:
@chris-cashin That seems strange. Can you provide some more information on your issue:

* Can you give an example of the filenames and the file sizes of the files that failed to restore? 
* Did you check that the files were really missing on the HD, or were they just not accessible to Duplicati? 
* Have there already been files present in the target folder? 
* Do the folders backed up contain files with long filenames and some that correspond to the old (8+3)-DOS-Format (I assume you're on Win)?
* Furthermore, did you try it again with an older version of Duplicati?
* And, is there a log available / can you reproduce the error with a verbose log to make sure there are no other errors involved?

chris-cashin:
Stefan the environment in server 2012 R2
The source and destination are local
I did a simple test to replicate the problem
The file names are simple names like GLTRX or CPORDL with no extensions.
[log.txt](https://github.com/duplicati/duplicati/files/190396/log.txt)

Files are large 500MB+
There is only one backup and restore cycle to empty directories
If I use the database the restore works with 2.0.1.7
If I use the 2.0.0.91 version restore works correctly without database.
Log snippet attached in log file.txt and follows:

2016-03-26 01:19:15Z - Information: Backend event: Get - Completed: TEST-bfb299c5e711e43adacbcf596d4685627.dblock.zip.aes (2.51 MB)
2016-03-26 01:19:15Z - Profiling: RemoteOperationGet took 00:00:00.074
2016-03-26 01:19:15Z - Information: Recording metadata from remote data: D:\RESTORE\
2016-03-26 01:19:15Z - Information: Recording metadata from remote data: D:\RESTORE\CPORDL
2016-03-26 01:19:15Z - Profiling: Starting - CommitBlockMarker
[...]

PR #1661, opened 2016-03-26
Hotfix/recreate database error

FootStark:
Fixed SQL used for restoring BlockSetEntries from Blocklists.


FootStark:
Should fix #1657

kenkendk:
The SQLite engine seems to correctly do integer math, but no harm in being explicit.

PR #1666, opened 2016-03-29
Update SystemSettingsController.js

mylemans:
Fixed javascript triggered when trying to save settings while remote access is allowed


kenkendk:
Thanks for the fix!

If you would like to be mentioned in the acknowledgements, feel free to create a pull request that updates `License/acknowledgements.txt`

PR #1669, opened 2016-03-30
Fixed: Multiple possible "Collection was modified" exceptions by swit…

mylemans:
…ching to ConcurrentDictionary

Changed: Switched to using ConcurrentDictionary to prevent "Collection
was modified" from occurring
Added: Simulated the 'old' race conditions so that unexpected situations
regarding unique keys not being unique act the same as they did before
Removed: Pointless ToList() calls as they do NOT prevent "Collection was
modified" exceptions from occurring in a multi threaded situation (it
just makes the 'window' smaller)

### DIFF
--- a/Duplicati/Server/WebServer/AuthenticationHandler.cs
+++ b/Duplicati/Server/WebServer/AuthenticationHandler.cs
@@ -16,6 +16,7 @@
 //  License along with this library; if not, write to the Free Software
 //  Foundation, Inc., 59 Temple Place, Suite 330, Boston, MA 02111-1307 USA
 using System;
+using System.Collections.Concurrent;
 using System.Linq;
 using HttpServer;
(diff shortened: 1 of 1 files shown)


mylemans:
Sorry about the conflicts, don't know how to deal with it, hope you can?

I actually ran into this with opening the site in a 2 second browser and refreshing it

kenkendk:
I have fixed the conflict and will commit once it passes the automatic build.

The "ToList()" where just meant to guard against the code deleting from itself, it was not meant to help with multithreadiing. I think the code there was never designed to be multithreaded; not sure how I could have overlooked that.

kenkendk:
To fix the merge conflict, you simply have to pull the duplicati master into your own master and merge the changes. With sourcetree, you should be able to right click the master branch and say "merge remote/master into master".

PR #1686, opened 2016-04-07
Type field inadecuate

ics79:
The field with id "generic_password" is the type "text" and password paraphrase is visible is better change to type "password".


kenkendk:
Yes, thanks!

ics79:
by nothing. Is your job, it's great! Thanks!

PR #1690, opened 2016-04-09
Alternative ftp backend

mylemans:
Implemented an alternative FTP backend using System.Net.FtpClient


kenkendk:
Overall: good job!

Minor thing: There is a lot of whitespace changes that makes it hard to review the changes. What editor have you used?

I am not sure what the changes to the solution file formatting options do?
Also, is there a reason that the solution file format version was updated?

The `LICENSE.TXT` is called  `license.txt` in the other projects, I can't remember if the name is used for anything. The `Debug` folder under the thirdparty library is not required. For debugging, developers should just download it.

The code looks like a copy of the original FTP implementation, would it make sense to re-use some things, e.g. The List command parsing logic?

mylemans:
@kenkendk Take a closer look, it does its own parsing (not custom)

And I used VS2015 Community Edition.

kenkendk:
Yes, I see, I remembered the list parsing wrong.
I have cleaned a few things, and will merge from this branch:
https://github.com/duplicati/duplicati/tree/feature/mylemans-alternative_ftp_backend

PR #1722, opened 2016-04-23
Upload files in chunks to prevent timeouts on large files

ReneHezser:
Used example from https://msdn.microsoft.com/en-us/library/office/dn904536.aspx to prevent timeouts on uploading the backup to OneDrive for Business.


FootStark:
Thanks for the solution. I didn't check it with larger block-files, and wasn't aware of possible timeouts. Do you know the timeout values for SharePoint and can specify with which file sizes the timeouts happen? I will review, test and merge the code on the weekend.

ReneHezser:
Hi Stefan,
I can't tell you about the specific size for files that cause timeouts. My backup is several GB and it didn't upload to OneDrive for Business (SharePoint Online). With the changes, I was able to upload a 5.x GB file.
René

ReneHezser:
Any chance the pull request will be accepted? The current implementation doesn't let me perform successful uploads to OneDrive for Business in SharePoint Online.

renestach:
Just one hint: The "max backup files size" has been set to 50MB for some reasons. You can increase this to 100MB or 200MB but everything above is not recommended. 

Why did you choose to have single files of 5GB?

ReneHezser:
I didn't. At least not that I know. I've configured the backup through the web interface of duplicati.

renestach:
What value does the "max backup file size" on step 4 of your backup have? Default value is 50MB and then no uploaded file should be larger than 50MB.

FootStark:
I am at it. I have to change something to allow non seekable streams. Should be done today/tomorrow.

FootStark:
BTW: It would still be great to know what exactly causes your timeouts and where they occur. Maybe it's related to another environment condition; I still couldn't reproduce the timeout. MS states [timeout to be 30 min here](https://msdn.microsoft.com/en-us/library/office/dn904536.aspx), so that would be an awefully slow internet connection not transferring 50MB/30min...

ReneHezser:
@FootStark With the default settings the files are 50MB and not timeout occured

FootStark:
Will be merged with added control-options via #1771.

PR #1743, opened 2016-04-30
Added duration to last successful run on home page

trampi:
This small pull request adds the duration of the last successful run to the home page.

It looks like this:
duration

Thanks for your great work!
Fabian


trampi:
Hm, the travis build failed.
From other builds with this error I conclude that it does not seem to be my fault:
`No output has been received in the last 10 minutes, this potentially indicates a stalled build or something wrong with the build itself.`

Any idea what caused this?

kenkendk:
I think the stalling issue has something to do with the Travis environment. It started happening after some changes that are not even included in the unittest. I have restarted the build.

PR #1750, opened 2016-05-05
Suggested fix for lookup of file's last backup entry

FootStark:
- Fixed SQL to create temp table for lookup of last backup's file
entries
- Removed temp table by writing directly to internal lookup struct.
- Direct lookup of single file entries from DB
- Added reverse lookup index to table FilesetEntry, also making it WITHOUT_ROWID.

Would still need an actual performance comparison.

### DIFF
--- a/Duplicati/Library/Main/Database/Database schema/6. Optimize FileSetEntry-Table.sql
+++ b/Duplicati/Library/Main/Database/Database schema/6. Optimize FileSetEntry-Table.sql
@@ -1,0 +1,29 @@
+﻿BEGIN TRANSACTION;
+
+DROP INDEX "FilesetentryIndex";
+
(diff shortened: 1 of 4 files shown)


kenkendk:
Do you have a test setup for it? 

I do not have enough files to emulate the problem, but the unittests seems to take appx the same time so this will have little impact on small backup sets.

FootStark:
I only tested for correctness (best I could) and checked the execution plans (on which base I updated the db schema). All seems fine as far as I could see. Yet, I also did not have a bigger test backup to check actual performance.
BTW: I did not do a real test with the internal lookup tables. But I think they were broken before, as it tried to read a field (BlocksetId) that wasn't in the temp table before.

Issue #1744, opened 2016-05-02
Bug when backup folder does not exist (Amazon Cloud Drive)


tophee:
As suggested by Kenneth, I am posting [this](https://groups.google.com/d/msg/duplicati/7_IsqJIIO70/tC-TSegPEwAJ) as an issue here.

I encountered the following issue when trying to setup duplicati with ACD:
1. set the "Path on server" to "duplicati test" (a directory that does not yet exist on my ACD)
2. Get the AuthID
3. Click "Test connection"
4. Popup asks if the directory "duplicati test" should be created, answer "yes"
5. The same popup comes up again. Answer yes again. This leads to the following error: "Failed to connect: The remote server returned an error: (409) Conflict."
6. steps 3-5 can be reproduced multiple times (except that now step 4 is ommited)

Unfortunately, the problem doesn't stop there. When I actually try to run the backup, I get an error "Unable to find folder duplicati test in /" But when I check my cloud drive, the folder is there.

I then tried a folder name without a space. It also produces the same error message when testing the connection. But when I run the job, it actually does write a couple of files to the cloud but then stops with an error that said something like "a file is missing". I cannot reproduce it because when I try to run the same job again, it claims to run successfully (within a few seconds) and then stops, waiting for the next scheduled time. However, during that supposedly successful run, it did not add any further files to the cloud (and there are gigabytes left to do in the folder it is supposed to backup). 

So it looks like something is seriously not working between duplicati and ACD...

Christoph
##

Xelio:
I've encounter the same problem and found the source of the problem.

In documentation of Amazon cloud API, filtering values of Nodes API need space or special characters escaped by adding a backslash (\) before it. 
[https://developer.amazon.com/public/apis/experience/cloud-drive/content/nodes#Filtering](url)

But in Duplicati.Library.Backend.AmazonCloudDrive.AmzCD, space is escaped by Utility.Uri.UrlEncode to a plus sign (+), which cause the API return nothing and throw a FolderMissingException
[https://github.com/duplicati/duplicati/blob/master/Duplicati/Library/Backend/AmazonCloudDrive/AmzCD.cs#L163](url)

kernel-sanders:
Having the same issue. @Xelio have you found a way to patch it yourself?

warwickmm:
Closing this since the Amazon Cloud Drive API has been discontinued.

https://forum.duplicati.com/t/amazon-cloud-drive-discontinued/7391

PR #1779, opened 2016-05-21
Fix the escaping of special characters of Amazon Cloud Drive API filtering

Xelio:
For filtering parameters, prepend \ to special characters + - & | ! ( ) { } [ ] ^ ' " ~ * ? : \ and space according to Amazon Cloud Drive API documentation, and encode space as %20 instead of + to avoid mistaking space as plus sign in filtering parameter.

Amazon Cloud Drive API documentation on escaping special characters of filtering parameters:
[https://developer.amazon.com/public/apis/experience/cloud-drive/content/nodes#Filtering](https://developer.amazon.com/public/apis/experience/cloud-drive/content/nodes#Filtering)

These change will fix the problem in #1744


kenkendk:
Thanks for finding and fixing this!
You can make a new PR to insert yourself in the Acknowledgement file:

PR #1802, opened 2016-06-10
Recovery recompress

mnaiman:
Adds ability to recompress to another type (eg 7z->zip) whole remote backend.


kenkendk:
Hi.

You need to add the "Recompress.cs" file to the .csproj file, otherwise it will not compile.

mnaiman:
Youre right, Im new to Git from SVN and little bit fighting with commits.

kenkendk:
Feel free to submit a new PR with your name added to the `acknowledgements.txt` file:

kenkendk:
And thanks for the work, it looks very nice.

Issue #1800, opened 2016-06-09
SSH clients is not handling connections well


mnaiman:
From log of my SFTP server can be seen how number of sessions is increasing with each file. At the end there are hundreds of them.

```
ACCEPT|09.06.2016 22:26:44|There are now 3 active connections
ACCEPT|09.06.2016 22:26:44|There are now 4 active connections
ACCEPT|09.06.2016 22:26:44|Client is attempting none authentication
ACCEPT|09.06.2016 22:26:44|Client is attempting password authentication
ACCEPT|09.06.2016 22:26:44|[SFTP:2406720_0:username:IP_ADDRESS] READ: *Verifying password for username.*
USER|09.06.2016 22:26:44|[SFTP:2406720_0:username:IP_ADDRESS] READ: *USER username*
USER|09.06.2016 22:26:44|[SFTP:2406720_0:username:IP_ADDRESS] READ: *CONNECT username*
MDTM|09.06.2016 22:26:44|[SFTP:2406720_0:username:IP_ADDRESS] READ: *getFileAttributes /backup/*
MDTM|09.06.2016 22:26:44|[SFTP:2406720_0:username:IP_ADDRESS] READ: *getFileAttributes /backup/Backup_Name/*
CWD|09.06.2016 22:26:44|[SFTP:2406720_0:username:IP_ADDRESS] READ: *CWD /backup/Backup_Name/*
STOR|09.06.2016 22:26:44|[SFTP:2406720_0:username:IP_ADDRESS] WROTE: *250 "/Backup_Name/" CWD command successful.*
MDTM|09.06.2016 22:26:44|[SFTP:2406720_0:username:IP_ADDRESS] READ: *getFileAttributes /backup/Backup_Name/*
CWD|09.06.2016 22:26:44|[SFTP:2406720_0:username:IP_ADDRESS] READ: *CWD /backup/Backup_Name/*
STOR|09.06.2016 22:26:44|[SFTP:2406720_0:username:IP_ADDRESS] WROTE: *250 "/Backup_Name/" CWD command successful.*
LIST|09.06.2016 22:26:44|[SFTP:2406720_0:username:IP_ADDRESS] READ: *150 LIST /backup/Backup_Name/*
LIST|09.06.2016 22:26:44|[SFTP:2406720_0:username:IP_ADDRESS] READ: *226 LIST /backup/Backup_Name/*
[...]

PR #1803, opened 2016-06-11
fix 1341

### LINKED ISSUE #1341: System.Net.Sockets.SocketException (0x80004005)
Hi, im receiving this regularly, what problem could this indicate?

Warning: Operation Put with file Backup_Name-b3a1326343abf427abe6db0236e72b7b2.dblock.zip.aes attempt 1 of 5 failed with message: An established connection was aborted by the software in your host machine
System.Net.Sockets.SocketException (0x80004005): An established connection was aborted by the software in your host machine
   at Renci.SshNet.Session.SocketWrite(Byte[] data)
   at Renci.SshNet.Session.SendMessage(Message message)
   at Renci.SshNet.Channels.Channel.SendMessage(ChannelDataMessage message)
   at Renci.SshNet.Sftp.SftpSession.RequestRealPath(String path, Boolean nullOnError)
   at Renci.SshNet.Sftp.SftpSession.GetCanonicalPath(String path)
   at Renci.SshNet.SftpClient.InternalUploadFile(Stream input, String path, Flags flags, SftpUploadAsyncResult asyncResult, Action`1 uploadCallback)
   at Renci.SshNet.SftpClient.UploadFile(Stream input, String path, Boolean canOverride, Action`1 uploadCallback)
   at Renci.SshNet.SftpClient.UploadFile(Stream input, String path, Action`1 uploadCallback)
   at Duplicati.Library.Main.BackendManager.DoPut(FileEntryItem item)
   at Duplicati.Library.Main.BackendManager.ThreadRun()

mnaiman:
Fix #1341 maybe partially #1800 but that need changes in recreatedatabasehandler.cs as well

### DIFF
--- a/Duplicati/Library/Backend/SSHv2/SSHv2Backend.cs
+++ b/Duplicati/Library/Backend/SSHv2/SSHv2Backend.cs
@@ -27,7 +27,7 @@ using Renci.SshNet.Common;
 
 namespace Duplicati.Library.Backend
 {
-    public class SSHv2 : IStreamingBackend, IRenameEnabledBackend
+    public class SSHv2 : IStreamingBackend, IRenameEnabledBackend, IBackend, IDisposable
     {
         public const string SSH_KEYFILE_OPTION = "ssh-keyfile";
         public const string SSH_KEYFILE_INLINE = "ssh-key";
@@ -41,8 +41,8 @@ namespace Duplicati.Library.Backend
         private string m_password;
 
         private int m_port = 22;
-		
-		private SftpClient m_con;
+
+        private SftpClient m_con;
 
         public SSHv2()
         {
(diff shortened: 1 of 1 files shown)


kenkendk:
Can you see from your server log that this fixes the issue? I cannot see which of your changes would cause less sessions to be created.

Previously, the CreateConnection(true) calls would return the connection with CWD set to the folder.
In your fix, you set the folder on each call, but does that prevent a new session from being created?

mnaiman:
It is primary fix for #1341, but i did some refactor during debugging, because 
if (m_con != null && !m_con.IsConnected) it created whole new connection instead of just connecting.
Same for changing dir, sometimes it was difficult to read when directory was about to change.
But it is just my opinion :)

Issue #1812, opened 2016-06-15
SFTP/SSH connections should verify the server's fingerprint


HRomie:
To eliminate the possibility of Man-in-the-middle attack, it is required for the client to verify the servers fingerprint with the pre-defined hash value provided by the user. This should happen before sending the login credentials and the client must disconnect if it could not validate the server. right now, without this verification, Duplicati is ignoring a very important part of the SSH protocol and exposing the user to the possibility of man-in-the-middle attacks.

kenkendk:
I agree, this should be added.

mnaiman:
Submitted pull request #1823

kenkendk:
This was fixed with 4ac829147e47a2a5782d2c29581643559f0dd1ef

PR #1823, opened 2016-06-24
Fix for issue #1812 - SSH fingerprint verification

mnaiman:
add ability to check server fingerprint


kenkendk:
Great fix.

I will see if I can update the UI to also grab the host key and insert it.

mnaiman:
Web UI can work with these new switch, it was checked.
Only thing Im not sure is approach.
Current approach is, that if fingerprint is not specified it allows to connect to server.
Better approach could be to refuse to connect to server if not specified or to add new switch like --shh-allow-all-fingerprints.
This could be secure by design. I can alter commit to this if we agree.

kenkendk:
Yes, that is how the other SSH clients are doing, so we should do the same.
Then the WebUI can just connect without a fingerprint and the error message will contain the fingerprint.

mnaiman:
Yeah, error contains fingerprint right now, so it is easy change. Do you want me to do it or you are working on it?

agrajaghh:
just to clairfy, you want to accept the (unknown) fingerprint on the first connection and refuse the connection if it changes, right? so TOFU (trust on first use)? Because `--shh-allow-all-fingerprints` sounds different/dangerous

kenkendk:
@agrajaghh My suggestion was to not trust anything, but instead report the fingerprint in an error message. On the CLI the message would be something like `please add --ssh-fingerprint=xyz to trust this host key`, in the WebUI it would be something like `Would you like to trust the SSH Host key xyz ?`.

To get the TOFU style, we would need to store the fingerprint somewhere, either in some local storage folder, or in the local database.

mnaiman:
Updated to not trust anything and reporting server fingerprint in error message.

Duplicati/Library/Backend/SSHv2/SSHv2Backend.cs
```diff
@@ -31,6 +31,8 @@ public class SSHv2 : IStreamingBackend, IRenameEnabledBackend, IBackend, IDispos
     {
         public const string SSH_KEYFILE_OPTION = "ssh-keyfile";
         public const string SSH_KEYFILE_INLINE = "ssh-key";
+        public const string SSH_FINGERPRINT_OPTION = "ssh-fingerprint";
+        public const string SSH_FINGERPRINT_ACCEPT_ANY_OPTION = "shh-accept-any-fingerprints";
```
kenkendk: Small typo: should be `ssh-accept-fingerprints` instead of `shh...`

Issue #1817, opened 2016-06-18
SMTP(TLS) Authentication is not working


HRomie:
I have setup Duplicati to send email to my webserver after each backup. The server requires authentication. however, no matter what i do, Duplicati does not seem to be authenticating.
Here's the email setting that I've used in the web gui (my real domain name has been replaced with example.org):

```
--send-mail-from=Duplicity Test <test@example.org>
--send-mail-level=All
--send-mail-password=pass123
--send-mail-to=admin@example.org
--send-mail-url=tls://smtp.example.org
--send-mail-username=user123
```
 Since, this is my own server, i am able to provide you with the actual smtp conversation:

```
 Out: 220 mail-smtp.example.org ESMTP ready
 In:  EHLO theclient
 Out: 250-mail-smtp-tls.example.org
 Out: 250-PIPELINING
 Out: 250-SIZE 20971520
 Out: 250-STARTTLS
 Out: 250-ENHANCEDSTATUSCODES
 Out: 250-8BITMIME
 Out: 250 DSN
 In:  STARTTLS
 Out: 220 2.0.0 Ready to start TLS
 In:  EHLO theclient
 Out: 250-mail-smtp-tls.example.org
 Out: 250-PIPELINING
 Out: 250-SIZE 20971520
 Out: 250-AUTH PLAIN
 Out: 250-ENHANCEDSTATUSCODES
 Out: 250-8BITMIME
 Out: 250 DSN
 In:  MAIL FROM:<test@example.org>
 Out: 553 5.7.1 <test@example.org>: Sender address rejected: not logged in
```

Please note that it is perfectly legal to advertise 'AUTH PLAIN' only after TLS negotiation. It however does not seem to be the problem here anyway, as I've activated it also before TLS, or even tried to use the normal non-TLS connection and still Duplicity did not authenticate. 

I am testing Duplicity v2.0.1.17_canary_2016-05-14 on windows 8.1 x86

mnaiman:
Why your server accept only AUTH PLAIN and on AUTH LOGIN? I have AUTH LOGIN and it works fine. Can you create test account and send me connection info to debug against your server?

HRomie:
Hi mnaiman,

[...]

PR #1827, opened 2016-06-25
Fix for issue #1817 - SMTP TLS Authentication

mnaiman:
fixed using external library


HRomie:
Does this still do certificate validation? Or is it lost in the process?

kenkendk:
@mnaiman Thanks for merging the updates.
I saw you were using NuGet and I wanted to update the rest to also use NuGet.

@HRomie The MailKit library is doing the SSL verification, and I assume it uses the default .Net SSL certificate handling.

PR #1829, opened 2016-06-26
Fix for issue #1198 - Support for SSL certificates in WebServer

### LINKED ISSUE #1198: https for webinterface
Hi, it is possible to add support for https and custom certificates? It transfers passwords like webinterface, or passphrases for backups in clear text.
Thanks

mnaiman:
Fixes #1198


mnaiman:
Ultimate solution is to add file picker on web settings to specify certificate file and update it in database and show certificate "valid to" in settings.
But it require angular skill.

HRomie:
Great work. thanks
where does it save the certificate and the private key? (I'm guessing the database?)
And would it be possible to update them (specially the cert) via command line? cause it would be useful in case someone uses it with Let's Encrypt cert and wants to update the cert via a script.

kenkendk:
@mnaiman: Great work!
Not sure that the `Verify` call is what is needed. From my understanding it checks that the certificate has a valid chain. This means that you cannot use a self-signed certificate, which would be required if you want to use a "localhost" server. Correct me if I misunderstood.

But I think you can change the calls to check if the certificate has a valid private key (otherwise you cannot use it for a server anyway). The property is called `HasPrivateKey`.

Also, if the certificate is invalid, the server starts without a certificate. I would think that the server should fail to start if the user specified a certificate and the certificate is invalid?

@HRomie: The webserver needs to be restarted for the new certificate to work, and since you specify the certificate on the commandline, you can update that way. The certificate is stored in the database unencrypted, which is problematic, but otherwise we would need to store the password as well. Generally I recommend using the OS disk/folder encryption tools to protect the database.

Would it make sense to not store the certificate and just require that it is specified on the commandline?

mnaiman:
@kenkendk 

- Youre

 right about validity of certificate - fixed in latest commit.

- If

 the certificate is invalid it is better to start server on non encrypted interface to give admin possibility to choose certificate via web admin interface (needs to be programmed as I suggested). 
If server failed to start admin needs to connect  directly to server and correct from command line.

- certificate is stored in db unencrypted because if attacker gets access to database he has access to username/password/url of remote storage, AES password of backup so in my point of view more critical things than certificate which could be invalidated in case of breach easily.

kenkendk:
Loading a certificate over non-ssl is dangerous because you then send the certificate and password in clear. I will create a new issue for loading the certificate.

Issue #1470, opened 2015-09-25
Support DropBox


kenkendk:
There is an API here:
https://www.dropbox.com/developers/core/docs
##

gagarine:
This wasn't fixed in https://github.com/duplicati/duplicati/pull/1835 or I'm missing something?

warwickmm:
Fixed in pull request #1835.

duplicatibot:
This issue has been mentioned on **Duplicati**. There might be relevant details there:

https://forum.duplicati.com/t/dropbox-business-integration-with-a-team-folder/11445/5

PR #1835, opened 2016-06-29
Implements issue #1470 - Add DropBox backend

wvc:


kenkendk:
Thanks for this.
It looks clean.
I will merge asap.

kenkendk:
@wvc: Feel free to submit a new PR with your information added to the acknowledgements file:

wvc:
@kenkendk : I saw that the dropbox backend was added to the latest release, but it wasn't possible to select it in the web-interface. Is there a way I can setup a backup with dropbox, perhaps via the commandline?

kenkendk:
It should just show up in the UI, but I think the build is missing the actual dropbox implementation.

wvc:
@kenkendk : it seems the dropbox backend is not yet functional in the latest release (Duplicati - 2.0.1.20_canary_2016-09-01 )   .  I'm eager to get it working in one of the next releases, but the following things need to be fixed first, I think:

- the UI: it allows you to select dropbox as a backend when setting up a new backup,  but the form fields are not yet correct. It basically shows the form of the previous backend that was selected.  I think the form should look similar as the one for Google Drive.   Could you point me to the right direction where I can create this?

- AccesToken. The DropboxHelper I implemented subclasses OAuthHelper. In my original version, I made sure the accesToken was included in the request header 'Authorization' (bearer ...) , but this seems broken after the style changes you committed.   To be able to work, I should update OAuthHelper such that the AccesToken property also has a setter and that the token never expires. This way I can set the AccesToken when the DropboxHelper is constructed. This is the way the Dropbox api works: the accestoken remains valid until it is revoked by the user from the dropbox website. Do you think I could change OAuthHelper in this way?    Or should I switch to subclassing JSONWebHelper and do my own custom work there?

wvc:
@kenkendk : concerning the accestoken. I tested subclassing JSONWebHelper instead of OAuthHelper and this works fine. This is probably the best approach since the dropbox-api doesn't use the refresh token approach anyway.

kenkendk:
The reason it fails to load the UI is because there are no options in the backend, and the UI expects it.

I have fixed the UI to support non-option backends and I have added the options to the backend:
https://github.com/duplicati/duplicati/tree/feature/fixes_to_dropbox

I will merge those into master as soon as the tests pass.

kenkendk:
AccessToken: It is a bit hard to see the changes, due to the line-feed change, but I can see that I changed the base class from `JSONWebHelper` to `OAuthHelper`, which probably is the cause.

The OAuthHelper ensures that all calls to the API adds the `Authorization: Bearer ...` header, it does this in:
https://github.com/duplicati/duplicati/blob/master/Duplicati/Library/Backend/OAuthHelper/OAuthHelper.cs#L69

I think the DropBox backend just needs to set `AutoAuthHeader = true;` for it to work, and I have added that as well.

kenkendk:
For the UI to look correct, wit needs to be hooked up to the OAuth html/js and I have added that too.

I think that should be it :)

wvc:
@kenkendk : thanks for the changes. However, I think there is still one issue before the backend will work.

The OAuthHelper  sets  m_tokenExpires to DateTime.UtcNow upon construction. This causes the accesToken to immediately expire and it will call GetTokenResponse. However the GetTokenResponse does not work in the case of Dropbox. That is why I think it might be better to just use JSONWebHelper

kenkendk:
Yes, you are right, I forgot that detail.

I would like to use some of the code in `OAuthHelper` (the url stuff), so I have added an option to the `OAuthHelper` that toggles this "access-token-only" behavior.

I am running the unittests now, and it seems to work, except for a bug with small files, that I will fix.

wvc:
Great, looking forward to it.

kenkendk:
There is still an error with large files, I will see if I can fix it later today so it can be part of the experimental build.

kenkendk:
Added here (minus the large-file fix):
https://github.com/duplicati/duplicati/tree/feature/more_dropbox_fixes

wvc:
@kenkendk : I tested the latest release (2.0.1.21_experimental_2016-09-09) and everything worked properly on my test backup, including the Web-UI, getting the authid, ... . Thanks, good job!

Only one remaining issue: the upload speed was slow (100 KiB/s perhaps). I'm at my work office and uploads should be fast here. I'll look into it, because I've heard things like this about the dropbox API elsewhere. It may be related to either the size of the upload chunks, or the current App license being a developer license.

agrajaghh:
> update: I think the upload speed problem boils down to Utility.Utility.DEFAULT_BUFFER_SIZE being only 64 KiB. According to this post: https://github.com/ncw/rclone/issues/103#issuecomment-134688176 , at 64 KiB chunks we can only expect roughly 80 KiB/s uploads, due to the api-overhead.

@wvc could you please open a seperate issue for that so its not getting lost?

wvc:
Ok , will do.

kenkendk:
@wvc: The `Utility.Utility.DEFAULT_BUFFER_SIZE ` is used to copy from source(-file) to (network-)stream only. The mentioned size on the link you posted relates to chunk sizes, where the chunk size is 10MB in the current implementation.

The point of chunking is to protect agains upload failures, where just the last chunk can be re-uploaded instead of the entire file (although this is not implemented in the backend yet).

wvc:
Hmm, you are right. I will look into it further. But I suggest continuing any other discussion here : #1929

kenkendk:
Agreed :)

wvc:
Just as a note: I've been using the dropbox backend since 2.0.1.21_experimental_2016-09-09 released and it works as expected (it's uploading everything). The initial upload took a while (17GB at 2 - 3 Mbps), but the incremental uploads were quite quick to finish.

Good work, hope to see more people using it. Another note, when we reach 50 users, we need to apply for production status, otherwise dropbox won't allow more users to connect.

kenkendk:
Yes, good point about the user-count.
I have upgraded to 500 users, they will not review until there are 50 users on the service, but now I punched in the details so they should start the review as soon as there are 50 users.

PR #1852, opened 2016-07-13
Fix a few minor GUI issues

Vaporizer8:


kenkendk:
Looks good.

Did you check that the extra plugins to activate validation is required (from commit 7127ebba382f0c6b04c734d2cb77879443dbb49b)?
The parser should remove the 's' from the protocol name and continue, such that they are not needed.

Maybe you need this fix as well 6c2636f2311d2234a720527715a74106c52bff1f ?

Vaporizer8:
You are indeed correct, those changes are no longer needed since the fix from 6c2636f. I've reverted the changes from 7127ebb

kenkendk:
Super, I am merging then.
Feel free to create a new PR to add yourself to acknowledgements.txt:

Issue #1828, opened 2016-06-26
When backup is deleted option remain in db


mnaiman:
When I deleted backup in web UI, options for that backup remain in server db Options table (not consistent db). Maybe when new backup is created it inherits same settings, which is not intended.

4m1g0:
Hello.

The problem is that the backup id is not autoincremental, so when you delete the last backup and add a new one, it obtains the same id and all the configuration is attached to it magically :)

http://www.sqlite.org/faq.html#q1

> Note that the integer key is one greater than the largest key that was in the table just prior to the insert. The new key will be unique over all keys currently in the table, but it might overlap with keys that have been previously deleted from the table. To create keys that are unique over the lifetime of the table, add the AUTOINCREMENT keyword to the INTEGER PRIMARY KEY declaration.

I added the AUTOINCREMENT manually in the database and works perfectly, the problem is that when I try to add it to the Schema.sql so that duplicate makes it autoincremental on creation, the adapter crashes...

It seems that the version of the sqlite library we are using does not support autoincrement, at least I've tried upgrading to version 1.0.102.0 and it works... but I'm new with this and I'm not very sure if this is the actual problem...

If I can do something to help with this, tell me, I want to contribute to the project.

Btw, this is somehow related: https://github.com/duplicati/duplicati/issues/1201

4m1g0:
It was my fault, autoincrement is supported by our version of sqlite.

So #1856 fix the problem, but I'm not sure if that's the way you handle db updates.

Tell me If I should do it in another way. 

Thanks

kenkendk:
This was closed with d63430e

Issue #1201, opened 2014-11-14
advanced deleting of backup configs


donestefan:
When selecting "delete Configuration" in the GUI  it might be a good idea to use the following Popup to 

- offer exporting the configuration to a file before deleting
- offer deleting the local database in the same step (to avoid cluttering the duplicati config folder with old configurations). Maybe one should even point out that deleting the database does not render the backup data useless, since the local database can easily be rebuild from the Backup destination when re-importing a configuration.
- even offer deleting the actual backup data 

Quite a few users might not be aware that all this is worth considering... 
And when you already deleted a Backup Configuration it's very difficult to find out which database was associated with it (so you can delete it later). It might be worth stating that clearly before the user deletes the config.

kenkendk:
I have implemented this in 5a74a6e22e201bc6fa0d75942796d079232c8e41.

Perhaps I am being paranoid, but I really don't like the idea that there is a single web-api call that can erase your backup. If I merge this in, then ransomware can simply make a http request to the local webserver and this will cause Duplicati to erase the remote backup.

I think we should add something like a CAPTCHA or similar defense, if the user has checked the "delete all remote data" option.

agrajaghh:
> I think we should add something like a CAPTCHA or similar defense, if the user has checked the "delete all remote data" option.

or ask for the remote password and/or encryption password again?

I agree, this option is dangerous... But I guess every other backup solution has a way of deleting the backup as well no?

kenkendk:
We can ask for the password again, but it is trivial to bypass this from outside the GUI.

I think other backup solutions are not as exposed if they do not have a web-api. You can automate some GUI clicks, but sending a web-request is far easier.

agrajaghh:
true, but you could also delete all local files, set "keep backups" to 1 second and run compacting via the web-api or command line. The only safe thing would be not to save the backend password I guess, but this would make background backups impossible 😏

kahomono:
I have been testing a LOT of large backups.

I **had** been going to my SSH login to delete the last backup before making the next one. 

Then I tried the "Delete" checkbox and got the captcha.  I can only describe my feeling at seeing a captcha sully the UI of Duplicati as, a shock of disappointment.

I'll be back to SSH when the next backup is ready to go.

PR #1856, opened 2016-07-15
Fix for issue: #1828 - Add autoincrement atribute to backup id

4m1g0:


kenkendk:
Looks fine, but why is it needed?

4m1g0:
Because if it is not autoincremental, the ID is unique over all keys currently in the table, but if you delete one backup and create another one, the new one will likely get the same id and all the information related to the old backup will be attached to it (options, errors, logs... referencing the old id) I've tried to explain it in detail here: https://github.com/duplicati/duplicati/issues/1828

Foreign tables are not deleted when removing a backup. This is intentional as discussed here https://github.com/duplicati/duplicati/issues/1201 but we are overriding it when new backups are created.

kenkendk:
Arh, I see.
Thanks for the fix, feel free to create a new pull request where you add yourself to acknowledgements:

PR #1886, opened 2016-08-12
add issue template

agrajaghh:
This adds an issue template so future bugreports have hopefully more information.
What do you think? Is it too much like that?

you can read more about issue templates here:
https://github.com/blog/2111-issue-and-pull-request-templates

**The template looks like that:**

---

I have:
- [ ] searched open and closed issues for duplicates

### Version info

**Duplicati Version:** 0.0.0.0
**Operating System:**
**Backend:**

### Bug description
Describe here the issue that you are experiencing.

### Steps to reproduce
- using hyphens as bullet points
- list the steps
- that reproduce the bug

**Actual result:** Describe here what happens after you run the steps above (i.e. the buggy behaviour)
**Expected result:** Describe here what should happen after you run the steps above (i.e. what would be the correct behaviour)


kenkendk:
Nice! I have wanted this for a long time, but did not know it could be done.

Thanks!!

PR #1896, opened 2016-08-19
DuplicatiVerify powershell

mnaiman:
Converted python script to powershell for machines where is not powershell installed


mnaiman:
It is possible to bundle these scripts to installation folder, eg. tools, or scripts?

kenkendk:
Thanks for this!

kenkendk:
I think these tools should be bundled, but currently they are not.

Right now the "commandline" project is pulling in the "win-tools" folder.
We could tack on the scripts to this, or create an empty project that has installer items?

mnaiman:
I vote for empty project, not to bundle with commandline

Issue #1875, opened 2016-08-04
can't restore


jvook:
I have a OneDrive backup and a local hard disk backup.  When I try to restore from local HDD the process runs but the files never appear in their correct location:  C:\Users\<name>\Desktop, C:\Users\<name>\Documents, C:\Users\<name>\AppData

I used default restore options (restore to same place as original location) and from what I can see in logs Duplicati finds and verifies the files, but they never appear in the location and aren't actually restored.

I imported the backup from json.
I then repaired local database before restoring as I saw in another thread. (I received an error when trying to restore without repair, something about Local File Already Exists or something)

After running a while with scanning local disk etc. it goes back to "No Scheduled Tasks" but files are not restored.

I also tried with option --no-local-blocks as suggested in another thread since data was local.

Using latest experimental.

ETA:  When restoring from a Macrium Reflect image I made, I saw that one single file was restored but no others. (C:\Users\<name>\Documents\Dolphin Emulator\MemCard.raw)

jvook:
I tried to decrypt the AES file and just restore manually but the files are full of random characters and I can't make heads or tails of them.

This seems to run counter to this quote from the Duplicati homepage:

> Duplicati is built using standard tools and formats. Un-encrypted archives are simple .zip archives. Encrypted archives are .zip archives that can be decrypted with AES Crypt. So, even without Duplicati all your data are belong to you :-)

I can't restore these from standard tools at all.

jvook:
WinRAR could not extract the duplicati zip files.  THe following appeared ad nauseum:

```
    J:\backup\1\duplicati-b8fcd5b5a34c44492801d8ea39e375ccf.dblock.zip: Cannot create J:\backup\1\1\KhqlmCGHttLssyd998T8hM24HZGr64oByTCiW-nUjKA=
!   The directory or file cannot be created.
    J:\backup\1\duplicati-b8fcd5b5a34c44492801d8ea39e375ccf.dblock.zip: Cannot create J:\backup\1\1\ODOnvsdEnNxaW_W30AZzpEYuqCzlnXdDR8dsqrsBjfM=
!   The directory or file cannot be created.
    J:\backup\1\duplicati-b8fcd5b5a34c44492801d8ea39e375ccf.dblock.zip: Cannot create J:\backup\1\1\R0FyFlBZ9uyY5kDb2MUDZbbcr2QCbtDJPaWWpRN3OVI=
!   The directory or file cannot be created.
[...]

PR #1930, opened 2016-09-12
Fix restore remote backup, bug introduced in: 6096d8379bca4acb16b52a8…

4m1g0:
…4e5f18108c63e8643

### DIFF
--- a/Duplicati/Server/WebServer/RESTMethods/Backup.cs
+++ b/Duplicati/Server/WebServer/RESTMethods/Backup.cs
@@ -172,16 +172,7 @@ namespace Duplicati.Server.WebServer.RESTMethods
         {
             var input = info.Request.Form;
 
-			string[] filters;
-            var rawpaths = (input["paths"].Value ?? string.Empty).Trim();
-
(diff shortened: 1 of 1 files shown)


kenkendk:
Thanks for spotting this!

4m1g0:
Restore is not working at all, when restoring a remote backup (not an existing one in the computer)

This is caused by this commit: 6096d8379bca4acb16b52a84e5f18108c63e8643

When restoring the local database to create a temporary backup, we are not taking into account that the path is in JSON. https://github.com/duplicati/duplicati/blob/master/Duplicati/Server/WebServer/RESTMethods/Backup.cs#L249

We should do the same in this method, to parse it from JSON.

Steps to reproduce it:

- Use Ngax theme
- Backup some folder
- Now try to restore it through the "restore backup" option in the main menu (not through the backup line itself)
- When restoring, it calls /restore with paths in a json array, but this endpoint expects a coma separated string.

I think some error from #1875 can be related to this.

kenkendk:
I have put up a new experimental build with the fix

PR #1938, opened 2016-09-22
fix commandline help

agrajaghh:


agrajaghh:
also removed  `--backend-log-database`, see #1941

PR #1982, opened 2016-10-05
Hyper-v fixes

mnaiman:
fixes for hyper-v
remove very dangerous parameter hyperv-backup-no-merge which was not btw default
added basic logging
added safe checks


mnaiman:
Please do not merge yet, need to test on Server 2008, 2012, 2016

Duplicati/Library/Snapshots/WindowsSnapshot.cs
```diff
@@ -108,59 +105,94 @@ public WindowsSnapshot(string[] sourcepaths, Dictionary<string, string> options)
...
+
+                    if (productType != 1) // Hyper-V writer is present only on Server version of Windows
                     {
-                        //Find sourcepaths in Hyper-V WriterMetaData
-                        var m_backup_wmd = m_backup.WriterMetadata.FirstOrDefault(o => o.WriterId.Equals(HyperVWriterGuid));
-                        if (m_backup_wmd != null)
-                            foreach (var component in m_backup_wmd.Components)
+                        try
+                        {
+                            m_backup.GatherWriterMetadata();
```
kenkendk: It looks like this call is now only made if `productType != 1`, where it was previously executed always.
I think this should be called regardless of the existence of Hyper-V?

Duplicati/Library/Snapshots/WindowsSnapshot.cs
```diff
@@ -108,59 +105,94 @@ public WindowsSnapshot(string[] sourcepaths, Dictionary<string, string> options)
...
-                                    m_sourcepaths.Add(file.Path + file.FileSpecification);
+                                    m_sourcepaths.Add(path);
+                                    Logging.Log.WriteMessage(string.Format("For VM {0} - adding {1}.", hyperVGuest.ID, path), Logging.LogMessageType.Profiling);
                                 }
-                            }
                     }
 
-                    m_backup.FreeWriterMetadata();
+                    m_sourcepaths = m_sourcepaths.Distinct(StringComparer.CurrentCultureIgnoreCase).OrderBy(a => a).ToList();
                 }
```
kenkendk: This should use `Library.Utility.Utility.ClientFilenameStringComparer` to be consistent, but since Hyper-V is Windows-only it is not terribly important

Duplicati/Library/Snapshots/WindowsSnapshot.cs
```diff
@@ -108,59 +105,94 @@ public WindowsSnapshot(string[] sourcepaths, Dictionary<string, string> options)
...
-                {
-                    try { m_backup.FreeWriterMetadata(); }
-                    catch { }
 
-                    throw;
-                }
+                m_backup.StartSnapshotSet();
 
                 //Figure out which volumes are in the set
                 m_volumes = new Dictionary<string, Guid>(StringComparer.InvariantCultureIgnoreCase);
```
kenkendk: This should also use `Library.Utility.Utility.ClientFilenameStringComparer`

Duplicati/Library/Snapshots/WindowsSnapshot.cs
```diff
@@ -108,59 +105,94 @@ public WindowsSnapshot(string[] sourcepaths, Dictionary<string, string> options)
...
 
-                    //Option: Create-new-vm-after-restore, 
-                    //hyperVUtility.CreateHyperVMachine("TestVM", xmlpath);
-                    #endregion Testing Features
-                }
+                    Logging.Log.WriteMessage(string.Format("Found {0} virtual machines on Hyper-V.", hyperVGuests.Count), Logging.LogMessageType.Information);
 
-                m_backup.StartSnapshotSet();
+                    bool bNotFound = false;
+                    List<string> notFoundVM = new List<string>();
```
kenkendk: Perhaps use var?

kenkendk:
Good work, I have only a single important question, which is if the `GatherWriterMetadata` should be called even if the OS is not a server OS?
There can be other writers than Hyper-V (i.e. Outlook, MSSQL, etc).

mnaiman:
Tested on Win 2008, 2012, 2016
@kenkendk pls, can you merge?

mnaiman:
I would be nice to choose hyper-v machines from web GUI, but I have no clue how to do that. I could write REST methods, but not in angular.

kenkendk:
For picking from the WebUI, there needs to be somewhere where this is exposed (i.e. a REST method).
If it is inside the advanced options, there is no hook to put in custom stuff there atm, but we could add that.
Otherwise it is fairly easy to populate a drop-down or list with strings from a REST call in Angular.

PR #1992, opened 2016-10-09
Hyper-V REST API

mnaiman:
Added REST API for allow to choose Hyper-V Guests from GUI during adding hyperv-backup-vm option.


kenkendk:
Next step is then to determine how/where to use expose this in the UI. 
@renestach: do you have an opinion?

PR #1999, opened 2016-10-10
small adjust vss

mnaiman:
Changed order of steps according to https://msdn.microsoft.com/en-us/library/aa384630(v=VS.85).aspx


mnaiman:
One think I'm curious is Component mode. If IVssBackupComponents::AddComponent should be called. Like in here https://hypervbackup.codeplex.com/SourceControl/latest#HyperVBackup/HyperVBackup/BackupManager.cs

kenkendk:
From what I can read, we should call `AddComponents`, but at the same time it states that writers are implicitly added.
I'd say we call `AddComponents` just to be safe.

mnaiman:
I think that components are different from writers

PR #2002, opened 2016-10-11
small fix

mnaiman:
set current directory to allow .\path\ expansion


Duplicati/Server/Program.cs
```diff
@@ -257,6 +257,7 @@ public static void RealMain(string[] args)
                 {
                     //Portable mode uses a data folder in the application home dir
                     Environment.SetEnvironmentVariable(DATAFOLDER_ENV_NAME, System.IO.Path.Combine(StartupPath, "data"));
+                    System.IO.Directory.SetCurrentDirectory(System.IO.Path.GetDirectoryName(StartupPath));
```
kenkendk: The variable `StartupPath` should be the folder name.
If you call `GetDirectoryName(StartupPath)` you get the parent directory?

mnaiman:
youre right, pushing fix

PR #2020, opened 2016-10-16
Amazon cloud drive consistency delay

AndersGron:
I have no idea if this is correct or not, but it seems to fix my issues with consistency.

### DIFF
--- a/Duplicati/Library/Main/Operation/BackupHandler.cs
+++ b/Duplicati/Library/Main/Operation/BackupHandler.cs
@@ -1183,8 +1183,6 @@ namespace Duplicati.Library.Main.Operation
                         
                         m_blockvolume.Dispose();
                         m_blockvolume = null;
-                        m_indexvolume.Dispose();
-                        m_indexvolume = null;
                     }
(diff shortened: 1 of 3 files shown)


Duplicati/Library/Main/Operation/BackupHandler.cs
```diff
@@ -1183,8 +1183,6 @@ private bool AddBlockToOutput(BackendManager backend, string key, byte[] data, i
                         
                         m_blockvolume.Dispose();
                         m_blockvolume = null;
-                        m_indexvolume.Dispose();
```
kenkendk: Why are these lines removed?

kenkendk:
It looks good, it adds delays to all the operations, and I think it is consistent.

Duplicati/Library/Main/Operation/BackupHandler.cs
AndersGron: Calling Dispose() at that point is always an error because the if block above guarantees that m_indexvolume is null at that point.

PR #2029, opened 2016-10-18
Fix of find-volume.sh script

a1ezzz:
The current script can't find volume groups, that were created during Debian 8.1 installation. The reason is that $VOLUME variable has "dm-<number>" value, which can't be resolved to $LVSID. So, in this patch, I check mounts to determine alternate $VOLUME name.

Also, some checks are added for empty commands result.

### DIFF
--- a/Duplicati/Library/Snapshots/lvm-scripts/find-volume.sh
+++ b/Duplicati/Library/Snapshots/lvm-scripts/find-volume.sh
@@ -20,7 +20,7 @@ NAME=$1
 # Get the reported mount point for the current folder
 #
 VOLUME=`df -P "$NAME" | tail -1 | awk '{ print $1}'`
-if [ "$?" -ne 0 ] || [ "$VOLUME" == "" ]
+if [ "$?" -ne 0 ] || [ -z "$VOLUME" ]
 then
 	[[ "$?" -ne 0 ]] && EXIT_CODE=$? || EXIT_CODE=-1
 	echo "Error: unable to determine device for $NAME"
@@ -28,7 +28,7 @@ then
 fi
 
 MOUNTPOINT=`df -P "$NAME" | tail -1 | awk '{ print $NF}'`
-if [ "$?" -ne 0 ] || [ "MOUNTPOINT" == "" ]
+if [ "$?" -ne 0 ] || [ -z "$MOUNTPOINT" ]
 then
 	[[ "$?" -ne 0 ]] && EXIT_CODE=$? || EXIT_CODE=-1
 	echo "Error: unable to determine mount point for $NAME"
@@ -55,14 +55,13 @@ get_lvmid $VOLUME
 #
 # Get the LVM path for the mapped volume (second try)
 #
-if [ "$LVMID" == "" ]
+if [ -z "$LVMID" ]
 then
-	CMD="mount | awk '(\$3 == \"$MOUNTPOINT\") {print \$1}'" 
-	VOLUME=` eval $CMD`
+	VOLUME=`mount | awk '($3 == "'$MOUNTPOINT'") {print $1}'`
 	get_lvmid $VOLUME
 fi
 
-if [ "$LVMID" == "" ]
+if [ -z "$LVMID" ]
 then
 	EXIT_CODE=-1
 	echo "Error: unable to determine volume group (VG) for $NAME"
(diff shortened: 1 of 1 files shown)


Duplicati/Library/Snapshots/lvm-scripts/find-volume.sh
```diff
@@ -20,33 +20,56 @@ NAME=$1
 # Get the reported mount point for the current folder
 #
 VOLUME=`df -P "$NAME" | tail -1 | awk '{ print $1}'`
-if [ "$?" -ne 0 ]
+if [ "$?" -ne 0 ] || [ "$VOLUME" == "" ]
```
kenkendk: Should be ` || [ -z "$VOLUME" ]`

Duplicati/Library/Snapshots/lvm-scripts/find-volume.sh
```diff
@@ -20,33 +20,56 @@ NAME=$1
...
 then
-	EXIT_CODE=$?
+	[[ "$?" -ne 0 ]] && EXIT_CODE=$? || EXIT_CODE=-1
 	echo "Error: unable to determine device for $NAME"
 	exit $EXIT_CODE
 fi
 
 MOUNTPOINT=`df -P "$NAME" | tail -1 | awk '{ print $NF}'`
-if [ "$?" -ne 0 ]
+if [ "$?" -ne 0 ] || [ "MOUNTPOINT" == "" ]
```
kenkendk: Missing a `$` and should be: ` || [ -z "$MOUNTPOINT" ]

Duplicati/Library/Snapshots/lvm-scripts/find-volume.sh
```diff
@@ -20,33 +20,56 @@ NAME=$1
...
+	fi
+	export LVMID
+}
+
+get_lvmid $VOLUME
+
+#
+# Get the LVM path for the mapped volume (second try)
+#
+if [ "$LVMID" == "" ]
```
kenkendk: Should be `if [ -z "$LVMID" ]`

Duplicati/Library/Snapshots/lvm-scripts/find-volume.sh
```diff
@@ -20,33 +20,56 @@ NAME=$1
...
-LVMID=`lvs "$VOLUME" --options vg_name,lv_name --noheadings | tail -1 | awk '{ print $1 "/" $2}'`
-if [ "$?" -ne 0 ]
+
+function get_lvmid {
+	LVMID=`lvs "$1" --options vg_name,lv_name --noheadings 2>/dev/null | tail -1 | awk '{ print $1 "/" $2}'`
+	if [ "$?" -ne 0 ] 
+	then
+		EXIT_CODE=$?
+		echo "Error: Unable to determine volume group (VG) for mapped volume $VOLUME"
+		exit $EXIT_CODE
```
kenkendk: If you exit here, you cannot do the second try?
Is it because the lvs call returns zero, but a non-matched result?

Duplicati/Library/Snapshots/lvm-scripts/find-volume.sh
```diff
@@ -20,33 +20,56 @@ NAME=$1
...
+
+get_lvmid $VOLUME
+
+#
+# Get the LVM path for the mapped volume (second try)
+#
+if [ "$LVMID" == "" ]
+then
+	CMD="mount | awk '(\$3 == \"$MOUNTPOINT\") {print \$1}'" 
+	VOLUME=` eval $CMD`
```
kenkendk: Why do you need eval?

Can you write it like:
```
VOLUME=`mount | awk '(\$3 == \"$MOUNTPOINT\") {print \$1}'`
```

Duplicati/Library/Snapshots/lvm-scripts/find-volume.sh
a1ezzz: No, lvs returns non-zero value, but last pipe command "overrides" this. For example, the next bash command "`asdasdasdasd | cat`" will have zero exit code because "cat" return zero value.

kenkendk: I see :)

PR #2032, opened 2016-10-19
Fix debian dependencies

a1ezzz:
In Debian 8.1, Duplicati.Server.exe generates many error messages in console without installed libmono-system-numerics4.0-cil package.

Web page request generates attached errors.
[error.txt](https://github.com/duplicati/duplicati/files/538821/error.txt)


kenkendk:
Super :)

PR #2045, opened 2016-10-22
Windows service fix

mnaiman:
Prevent flooding EventLog with: The Duplicati service service has reported an invalid current state 0.

Bad pinvoke - long => unit (it is DWORD based on MSDN)

http://stackoverflow.com/questions/30299600/service-causes-scm-error-reported-an-invalid-current-state-0
https://msdn.microsoft.com/cs-cz/library/windows/desktop/ms685996(v=vs.85).aspx
http://stackoverflow.com/questions/6725547/should-dword-map-to-int-or-uint


kenkendk:
It seems the MS example code (where I got the struct definition from) is wrong :)

Your version looks correct, compared to this:
http://pinvoke.net/default.aspx/advapi32/SERVICE_STATUS_PROCESS.html

PR #2082, opened 2016-11-03
Fix not storing metadata for symlinks that point to nonexistent files.

gerco:
On Mac OS X, llistxattr doesn't exist and Mono doesn't allow us to pass XATTR_NOFOLLOW to listxattr. As a result of that we always try to follow the link, which fails with ENOENT if the target file doesn't exist.

The net result of this is that no user/group/perm is stored for a symlink that points to a nonexistent file.

### DIFF
--- a/Duplicati/Library/Snapshots/SystemIOLinux.cs
+++ b/Duplicati/Library/Snapshots/SystemIOLinux.cs
@@ -170,10 +170,22 @@ namespace Duplicati.Library.Snapshots
             var f = NoSnapshot.NormalizePath(file);
             var dict = new Dictionary<string, string>();
 
-            var n = UnixSupport.File.GetExtendedAttributes(f);
-            if (n != null)
-                foreach(var x in n)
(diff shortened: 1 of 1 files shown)


kenkendk:
Actually, now that I am reading up on it... if `llistxattr` does not exist (which it does not on OSX), we should not read extended attributes for a symlink, unless we have `--symlink-policy=follow` as we will otherwise read the extended attributes of the target file, not the source.

kenkendk:
I've made a fix that I think has more correct handling of the symlink metadata:
https://github.com/duplicati/duplicati/commit/1f7110503a25572a5070eb4913d9a485a6f44052

Can you have a look and see if it makes sense? (Or even try it).

For OSX (which does not support `llistxattr`) it just returns no metadata for symlinks.
If you think it makes more sense to try to read the target file metadata, we can change to that.

gerco:
I'm closing this one. We'll continue with #2085, I'll test and comment on that.

PR #2085, opened 2016-11-03
Store correct symlink metadata

kenkendk:
This is an alternative to #2082


gerco:
I tested this fix on Mac OS X and it works for me.

kenkendk:
@gerco Thanks for testing!

I think the error if you use `--symlink-policy=follow` is the correct result, as you are specify that you want to ignore the symlink part and use whatever the symlinks point to as the target, and since it does not exist, it is an error.

PR #2092, opened 2016-11-04
Add support for lvm snapshots of XFS filesystems

kakoskin1:
On Linux XFS filesystem checks that there are no duplicate uuids on mount. Snapshots, being copies of original filesystems, have the same uuid as the original filesystems. To disable this check you need to use option nouuid when mounting snapshots of an XFS filesystem. Other filesystems do not support option nouuid.


kenkendk:
Thanks for the patch!

PR #2121, opened 2016-11-13
Service for duplicati web server

a1ezzz:
This is a systemd service that can be used for starting the duplicati web-server. Service is disabled by default.


kenkendk:
Nice! That has been on my TODO for a while now :)

PR #2130, opened 2016-11-18
fix  schedule caluclate next time

tianqiaoxiangzhe:
the schedule can not run on the  caluclated next time use
'datetime.now'  ,    because my time zone is several hours higher than
utc


kenkendk:
Thanks for catching

PR #2141, opened 2016-11-22
Fix #2132 by testing on null

### LINKED ISSUE #2132: Cannot change schedule if previously empty
- [ x] searched open and closed issues for duplicates

### Version info

**Duplicati Version:** 32

### Bug description
If Allowed days were previously empty, it behave like lst = null in EditBackupController:
$scope.toggleArraySelection = function (lst, value) {}
Than it throws error in screenshot and will not save allowed days to database.

### Screenshots
error

<bountysource-plugin>
---
Want to back this issue? **[Post a bounty on it!](https://www.bountysource.com/issues/39401136-cannot-change-schedule-if-previously-empty?utm_campaign=plugin&utm_content=tracker%2F4870652&utm_medium=issues&utm_source=github)** We accept bounties via [Bountysource](https://www.bountysource.com/?utm_campaign=plugin&utm_content=tracker%2F4870652&utm_medium=issues&utm_source=github).
</bountysource-plugin>

tudacs:
This fix might not be the 'angular'-way but it's the most straigthforward fix.

I think there should be check to deny the user from disabling all days in the first place, as I see no use case for disabling all days, but I'm too unfamiliar with angular to implement that :(.

### DIFF
--- a/Duplicati/Server/webroot/ngax/scripts/controllers/EditBackupController.js
+++ b/Duplicati/Server/webroot/ngax/scripts/controllers/EditBackupController.js
@@ -148,6 +148,9 @@ backupApp.controller('EditBackupController', function ($scope, $routeParams, $lo
     };
 
     $scope.toggleArraySelection = function (lst, value) {
+        if (lst === null)
+          lst = [];
+
         var ix = lst.indexOf(value);
 
         if (ix > -1)


agrajaghh:
> I think there should be check to deny the user from disabling all days in the first place, as I see no use case for disabling all days, but I'm too unfamiliar with angular to implement that :(.

What if you have e.g. just monday and want to move it to tuesday, then you can't uncheck monday first. I think its fine like that. But there should maybe be an error message if you save and have no day selected... Or do we already have one? (can't test right now)

tudacs:
You're right: The check should be performed just before saving (like checking the passphrase, if i remember that right).

There had been no issue with unchecking all checkboxes and then checking some again without saving and reloading in between, as the javascript object then (correctly) is an empty array. The bug just jumps in if you save with no weekdays checked and then enter the configuration again.

agrajaghh:
oh okay, I should really test it before writing ;)

kenkendk:
It might not make sense if the user chooses "no weekdays", but that case is handled as "all days allowed" in the code.

Issue #2033, opened 2016-10-19
update dialog too small


agrajaghh:
It look's like the update dialog is too small for some languages:

![dialog](https://cloud.githubusercontent.com/assets/5296073/19517687/fca5dcdc-9602-11e6-8eb5-fb017dde8e33.png)
##

tudacs:
I feel a bit sorry for hijacking your self-assignment and hope that's okay for you :).

agrajaghh:
@tudacs dont worry 😊 I had no time yet to fix it, help is always welcome 😊 

but please keep in mind that the layout should work for small screens (like smartphones) as well. I think the design goal was to have it from screensizes like iphone4 upwards

kenkendk:
This was fixed with 3208a1d

PR #2142, opened 2016-11-22
Improve layout of notification area, fixing #2033

tudacs:
- increase width to 480px
- make responsible with breakpoint at 492px (due to paddings and borders)
- reorganize buttons to allow height of notifiaction responding to buttons
  in multiple rows in very narrow viewports:
    - replace paddin-bottom by adding a float-clear element
    - move top-padding from div.buttons to top-margin of each button

### DIFF
--- a/Duplicati/Server/webroot/ngax/less/style.less
+++ b/Duplicati/Server/webroot/ngax/less/style.less
@@ -224,7 +224,7 @@ ul.notification {
     left: 0px;
     right: 0px;
     margin: auto;
-    width: 300px;
+    width: 480px;
 }
 
 .notification {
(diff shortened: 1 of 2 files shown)


agrajaghh:
please keep in mind that the layout should work for small screens (like smartphones) as well. I think the design goal was to have it from screensizes like iphone4 upwards.

I'm not sure yet how to fix this dialog for small screens

tudacs:
That's why I added the layout breakpoint with a media query for the notification area. Below the 480px (492px, due to padding and borders) it switches to 100% width. I tested it (with Firefox) down to a viewport width of 285px, which worked quite well.

duplicati-small

agrajaghh:
cool, sorry about that one as well 😏

kenkendk:
Thanks for fixing!

PR #2156, opened 2016-12-04
Fixes

mnaiman:
added additional checks


kenkendk:
I think it is possible to have permissions to deal with HyperV and MSSQL without being Administrator? You check is just seeing if the user has Administrator privileges?

mnaiman:
To properly enumerate and then backup user needs to have Admin rights to run VSS snapshot or SeBackupPrivilege. https://msdn.microsoft.com/en-us/library/windows/desktop/bb530716(v=vs.85).aspx
nonVSS backup is not supported, because files are locked(SQL) and result would be inconsistent(hyper-v)

kenkendk:
I have been looking for a more accurate permission check, and came up empty.
Lets use the version you have, and then someone who needs a particular permission check can ask for it, and maybe then we can figure out what it is.

Issue #2157, opened 2016-12-06
Dependency is not satisfiable: libappindicator0.1-cil on Debian 9 Stretch (Testing)


k3vmcd:
I have:
- [x] searched open and closed issues for duplicates

----------------------------------------

### Version info
**Duplicati Version:** 2.0.1.30-1
**Operating System:** Debian 9 "Stretch" (Testing) - all packages current as of 12-5-2016
**Backend:** N/A

### Bug description
Dependency is not satisfiable: libappindicator0.1-cil

Debian Stretch does not include this specific version of the libappindicator package ("libappindicator0.1-cil"): https://packages.debian.org/source/stretch/libappindicator

It appears this package is replaced by libappindicator3-0.1-cil.

### Steps to reproduce
- Download `duplicati_2.0.1.30-1_all.deb` from the Duplicati website
- Ensure mono dependency satisfied (e.g., `mono --version`; if not installed then `sudo apt-get install mono-runtime`)
- Attempt install of .deb file with gdebi using `sudo gdebi duplicati_2.0.1.30-1_all.deb`

**Actual result:** gdebi install fails with ```This package is uninstallable
Dependency is not satisfiable: libappindicator0.1-cil```
**Expected result:** Duplicati package installs without unmet dependencies.

k3vmcd:
Also affects latest Canary release `v2.0.1.34-2.0.1.34_canary_2016-12-05`

kenkendk:
@k3vmcd Do you know how to fix this without needing two different .deb files?

kenkendk:
Maybe we can use the `Recommends` type:
https://www.debian.org/doc/debian-policy/ch-relationships.html

Then have:
```
Recommends: libappindicator3-0.1-cil libappindicator0.1-cil
```

Unfortunately, without that package, the tray-icon will not appear, making it hard to use the app.

TediouslyBored:
I think recommended packages are treated like dependencies by default in Debian, and thus usually installed. If you want to keep it as a dependency, the page you linked says you can separate alternatives with a vertical bar.

> In the Depends, Recommends, Suggests, Pre-Depends, Build-Depends and Build-Depends-Indep control fields of the package, which declare dependencies on other packages, the package names listed may also include lists of alternative package names, separated by vertical bar (pipe) symbols |. In such a case, that part of the dependency can be satisfied by any one of the alternative packages.

k3vmcd:
[...]

PR #2168, opened 2016-12-10
Add libappindicator3-0.1-cil dependency as substitute for libappindicator0.1-cil to allow installation on Debian Stretch

k3vmcd:
Bug fix for https://github.com/duplicati/duplicati/issues/2157


kenkendk:
Great fix! Thanks!

PR #2212, opened 2016-12-28
Use CWD to change to path in AlternativeFTP client connection

seaeagle1:
This makes the alternative FTP client behave a bit more 'stateful' by changing the working directory to the target path upon connection instead of prepending it before every command.

This fixes an issue I encountered with my NAS where writing to the FTP by absolute path was denied because the FTP root is non-writable. Probably has something to do with the root only listing several 'shared folders'. Hence using CWD and relative paths.


kenkendk:
Knowing FTP, there are likely servers that fail with CWD too, but works with the other approach :).

I will merge it now, but I expect users to report that they need the non-CWD version.

PR #2242, opened 2017-01-09
Revert "Update Strings.cs"

agrajaghh:
This reverts commit 4266ed9d4a6de5d04289d00b74c85c18dd389463.

Previous version was correct, right?


kenkendk:
I think so, yes

PR #2291, opened 2017-01-29
Add pagination to backend List

### LINKED ISSUE #2129: Amazon Cloud Drive: Error 429, "Too Many Requests"
I have:
- [x] searched open and closed issues for duplicates

### Version info

**Duplicati Version:** 2.0.1.30
**Operating System:** Windows 10 64-bit, version 1607
**Backend:** AmzCD

### Bug description
Occasionally, over the past couple of weeks, when running a backup targeting the Amazon Cloud Drive backend, the backup will fail with an error message:

`System.Net.WebException: The remote server returned an error: (429) Too Many Requests.`

I'm sorry to say that I don't have a repro for this, since it's only happened twice in as many weeks and I can't discern the conditions that led to it. Also, I've found that waiting until the next day is usually enough to get the backup to finish properly. I wanted to report it anyway, though, just to get some more information if possible. From what I've read, error 429 might apply to all users of the application using the Amazon API, in this case Duplicati. Is that correct, or is it an error that's exclusive to me, the individual user? As you can see, the error returned by the "remote server" doesn't provide much detail.

Furthermore, is there any set of best practices for Amazon Cloud Drive (or other backends, for that matter) that would help alleviate this? Setting a bandwidth throttle, or perhaps changing the block size? I left the backup settings at their defaults, including the block size of 50MB, and I'm considering increasing that to 200MB (the backup in question is ~370GB, with a great number of large files), would that in fact quarter the number of requests to the remote server as I expect it to?

I have no development experience with Amazon's APIs, and although I'll be happy to try the change myself, a backup that large may take a while to complete, so I thought it would be wise to ask the actual project developers for some feedback in the meantime. :)

<bountysource-plugin>
---
Want to back this issue? **[Post a bounty on it!](https://www.bountysource.com/issues/39307684-amazon-cloud-drive-error-429-too-many-requests?utm_campaign=plugin&utm_content=tracker%2F4870652&utm_medium=issues&utm_source=github)** We accept bounties via [Bountysource](https://www.bountysource.com/?utm_campaign=plugin&utm_content=tracker%2F4870652&utm_medium=issues&utm_source=github).
</bountysource-plugin>

klammbueddel:
Draft to fix #2129.

### DIFF
--- a/Duplicati/Library/Backend/AmazonCloudDrive/PageResult.cs
+++ b/Duplicati/Library/Backend/AmazonCloudDrive/PageResult.cs
@@ -1,0 +1,63 @@
+﻿#region Disclaimer / License
+// Copyright (C) 2015, The Duplicati Team
+// http://www.duplicati.com, info@duplicati.com
+// 
+// This library is free software; you can redistribute it and/or
+// modify it under the terms of the GNU Lesser General Public
+// License as published by the Free Software Foundation; either
+// version 2.1 of the License, or (at your option) any later version.
+// 
+// This library is distributed in the hope that it will be useful,
+// but WITHOUT ANY WARRANTY; without even the implied warranty of
+// MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the GNU
+// Lesser General Public License for more details.
+// 
+// You should have received a copy of the GNU Lesser General Public
+// License along with this library; if not, write to the Free Software
+// Foundation, Inc., 51 Franklin Street, Fifth Floor, Boston, MA  02110-1301  USA
+// 
+#endregion
+using System;
+using System.Linq;
+using System.Collections.Generic;
+using Duplicati.Library.Interface;
+
+namespace Duplicati.Library.Backend.AmazonCloudDrive
+{
+    internal class PageResult : IPageResults
+    {
+        private string nextToken;
+        List<FileEntry> files;
+
+        internal PageResult(List<FileEntry> files, string nextToken)
+        {
+            this.files = files;
+            this.nextToken = nextToken;
+        }
+
+        internal string NextToken
+        {
+            get
+            {
+                return nextToken;
+            }
+        }
+
+        public IEnumerable<IFileEntry> Items
+        {
+            get
+            {
+                return files;
+            }
+        }
+
+        public bool More
+        {
+            get
+            {
+                return nextToken != null;
+            }
+        }
+
+    }
+}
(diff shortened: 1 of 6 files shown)


kenkendk:
The build is failing with a compile error:

```
	PageResult.cs(25,10): error CS1014: A get or set accessor expected
	PageResult.cs(35,15): error CS1525: Unexpected symbol `bool', expecting `class', `delegate', `enum', `interface', `partial', or `struct'
	PageResult.cs(37,12): error CS1525: Unexpected symbol `get', expecting `class', `delegate', `enum', `interface', `partial', or `struct'
	PageResult.cs(39,16): error CS1525: Unexpected symbol `return', expecting `class', `delegate', `enum', `interface', `partial', or `struct'
	PageResult.cs(40,13): error CS1514: Unexpected symbol `}', expecting `.' or `{'
```

Duplicati/Library/Backend/AmazonCloudDrive/PageResult.cs
```diff
@@ -0,0 +1,44 @@
...
+            this.nextToken = nextToken;
+        }
+
+        internal string NextToken
+        {
+            get
+            {
+                return nextToken;
+            }
+        }1
```
kenkendk: There is a stray `1` character here :)

Duplicati/Library/Interface/IPagingBackend.cs
```diff
@@ -0,0 +1,19 @@
...
+namespace Duplicati.Library.Interface
+{
+    public interface IPagingBackend : IBackend
+    {
+        IPageResults List(IPageResults previous);
+    }
+
+    public interface IPageResults
+    {
+        IEnumerable<FileEntry> Items { get; }
```
kenkendk: This should be `IEnumerable<IFileEntry>` right?

Duplicati/Library/Backend/AmazonCloudDrive/PageResult.cs
klammbueddel: whoops.. was late that day ;)

Duplicati/Library/Interface/IPagingBackend.cs
klammbueddel: sure, fixed that

klammbueddel:
Still this change does not fix the issue. How do you think to split the List operation into chunks? I am not sure where to test for the Backend to implement IPagingBackend and where to inject the paging.
Would appreciate some more hints to finish this task.

kenkendk:
Sorry for the delay :/.

I think the right place to check for pagination support would be here:
https://github.com/duplicati/duplicati/blob/master/Duplicati/Library/Main/BackendManager.cs#L1066

This is where all code ends up when fetching the list. This could then use the pagination, retry delay etc.

PR #2294, opened 2017-01-30
Add HTTP reporting module

vatara:
I added an HTTP reporting module. It was based off the existing XMPP module and basically has the same functionality.


kenkendk:
Nice :)

I think some extra care is required for the extra parameters, as you could send something like `a=1&b=2` which will be URL encoded, but we can look at that if there is a request.

PR #2315, opened 2017-02-12
Upgrade .NET framework to 4.6.2

mnaiman:


mnaiman:
Support for .NET Framework 4, 4.5, and 4.5.1 ended on January 12, 2016.

mnaiman:
.NET framework 4.6.2 is supported on all OS that have support from MS https://msdn.microsoft.com/en-gb/library/8z6watww(v=vs.110).aspx

mnaiman:
Added features:
https://en.wikipedia.org/wiki/.NET_Framework_version_history#.NET_Framework_4.6

kenkendk:
I am a bit worried that this will cause problems with Mono installations.
They are not always updated, and especially Ubuntu/Debian is always way behind.
Are there new features that we need from 4.6.2?

mnaiman:
Latest Mono is installable easily from official repository http://www.mono-project.com/docs/getting-started/install/linux/
Main reason is 4.5,x brach is obsolete without any support

Some features applicable to Duplicati:
- Support for paths longer than 260 characters
- Many other performance, stability, and reliability related fixes in RyuJIT, GC
- Enhanced support for Elliptic Curve Digital Signature Algorithm (ECDSA) X509 certificates (HTTPS)

PR #2331, opened 2017-02-21
Make progress bar more responsive when backing up large files

asselin:
When backing up large files, the progress bar appears to get stuck for long periods of time.  This PR adds in the progress of the current file to make it more responsive.


kenkendk:
Great idea!

PR #2344, opened 2017-03-01
Update localization-fr.po

glannes31:
First part of additions to the missing french translation . Others parts are coming later.


glannes31:
@kenkendk  Yes, but there are a lot of not reviewed words. It's normal ?

kenkendk:
@glannes31 Yes, I think the review process is not so well implemented at this point.

PR #2352, opened 2017-03-03
Update login.html

johenning:
Made login page title more distinct, helpful for auto-type features of password managers.


kenkendk:
Thanks!

Issue #1945, opened 2016-09-24
JottaCloud backend request


soniariv:
I have:
- [x ] searched open and closed issues for duplicates

---
### Version info

**Duplicati Version:** 2.0.0.0
**Operating System:** Windows 7
**Backend:** JottaCloud
### Bug description

I would like some support for Jottacloud, please. https://www.jottacloud.com/
It is a Norwegian cloud with good privacy support and one of the few cloud services with servers in Europe. It already provides AES-256 of data traffic but apparently not the encryption on-site. So Duplicati could be an interesting companion to secure all data before it is uploaded.
Thank you!
##

jkaberg:
@kenkendk You can close this one :smile: (its implemeneted, ref #2355)

kenkendk:
Thanks!

PR #2355, opened 2017-03-04
Jottacloud backend

### LINKED ISSUE #756: Add JottaCloud.com + more
_From [morten.v...@gmail.com](https://code.google.com/u/115576868845018120957/) on October 13, 2012 23:34:43_

I would like to see Google Drive and Jottacloud as own tabs under Options. I would also for both of these like to have an option for mounting Jotta/Google as a driveletter.

_Original issue: http://code.google.com/p/duplicati/issues/detail?id=729_
## <bountysource-plugin>

Want to back this issue? **[Post a bounty on it!](https://www.bountysource.com/issues/3562496-add-jottacloud-com-more?utm_campaign=plugin&utm_content=tracker%2F4870652&utm_medium=issues&utm_source=github)** We accept bounties via [Bountysource](https://www.bountysource.com/?utm_campaign=plugin&utm_content=tracker%2F4870652&utm_medium=issues&utm_source=github).
</bountysource-plugin>

albertony:
Hi.
I'm throwing myself into my very first pull request, so please be gentle if I'm doing something terribly wrong..

I've made an effort these last couple of nights to implement a backend for Jottacloud. This (norwegian by the way, like me) cloud provider offer different functionality, such as regular sync, its own backup utility (not with client-side encryption unfortunately), and then it has this "archive" feature which (if I understand it correctly) is intended for upload of data for off-site storage that you don't want to sync with your devices. This archive feature seems to be a good match for uploading Duplicati backups, and also combined with the unlimited plans that Jottacloud offers. Jottacloud are also offered via several white-label products, such as "Get Sky", "Elkjøp Cloud" etc. I have tested my backend implementation with "Get Sky" and that works.

My backend implementation is using a REST api that Jottacloud has made available, but it is not really official. Official employees have written some short introduction to it on their [forum](http://forum.jotta.no/jotta/topics/api_http), but thats it. Also they were just talking about downloading files via GET-requests, not uploading. But some people have figured out how to upload via "reverse engineering", I used the [jottalib](https://github.com/havardgulldahl/jottalib) Python implementation from @havardgulldahl as a very good source. The upload is done via a post request containing multipart form data, with MD5 hash, file size, timestamps etc. in both the form data and in headers, so it is perhaps not the most stream-lined approach for uploading large files etc 

This should fix #756 and #1945, I guess.


albertony:
I have some more progress on figuring out how to handle the mount points properly, so I will do some additional commits probably later today.

Duplicati/Library/Backend/Jottacloud/Jottacloud.cs
```diff
@@ -0,0 +1,405 @@
...
+        {
+            if (!stream.CanSeek)
+            {
+                throw new System.Net.WebException(Strings.Jottacloud.FileUploadError, System.Net.WebExceptionStatus.ProtocolError);
+            }
+
+            // Pre-calculate MD5 hash, we need it in query parameter, in HTTP header and in POST message data!
+            string md5Hash;
+            using (System.Security.Cryptography.MD5 md5 = System.Security.Cryptography.MD5.Create())
+                md5Hash = BitConverter.ToString(md5.ComputeHash(stream)).Replace("-", string.Empty);
```
kenkendk: The stream you get here can be throttled, which will make the hash computation "throttled".
There is another backend that needs the same, and it "walks" the streams, until it gets the underlying FileStream and computes the hash on that, such that it can be done without the throttle.

albertony: OK, thanks, I'll see if I can find that..

albertony: Could it be the Backblaze backend you are refering to, with the "A bit dirty, but we need the underlying stream to compute the hash without any interference" section?

kenkendk: Yes, just found it: https://github.com/duplicati/duplicati/blob/master/Duplicati/Library/Backend/Backblaze/B2.cs#L161

kenkendk: I have been struggling with that for CloudFiles and S3 too:
https://github.com/duplicati/duplicati/blob/master/Duplicati/Library/Backend/CloudFiles/CloudFiles.cs#L322

albertony: I was looking at the CloudFiles first, but it seems to me that it sends the data to the RequestStream while calculating the hash. I need the MD5 in advance, in some headers and multipart items, before writing the data to the RequestStream. So if I'm correct I need something more like the B2 variant.

kenkendk:
Looks good, is this ready to be merged?

Can you somehow unwrap all the auto-gen changes to the `.sln` file?
Otherwise, I might be able to do it after the merge.

albertony:
Oops, did not pay attention to all the changes in the sln file. Should be fixed now, with the recent commit.
I don't know how thorough it is expected of me to test the functionality.. So far I have only been running it for a simple setup. It seem to work fine, but I could run a larger test this evening if we want to be more confident about it.

albertony:
Done some more testing now: Used it for backup of some 30GB of files. Ran initial backup first, then incremental update after changing some files, then restore of latest version of some files, and the restore of previous version. No problems. Did not push limits like unplugging network during backup etc, but I guess its ready to be included in next pre-release.

kenkendk:
There is a tool called `Duplicati.CommandLine.BackendTester.exe` which is what I usually use to test the backends. Network issues etc should be handled centrally, with retries etc.

I am merging now, and then we can see if there are users reporting issues.

PR #2357, opened 2017-03-06
tray fixes

mnaiman:


mnaiman:
auth_token = authquery["auth-token"].Value;  returning always null
When tray icon run in non hosted mode, click open did not end up with logged on user because of this bug

kenkendk:
The default url has been wrong for so long....
Good find!

Issue #2266, opened 2017-01-19
AutoUpdater causes slowdown


kenkendk:
When starting through the autoupdater, it needs to verify the update packages have not been tampered with. This is not a problem for the tray icon or the server, but for the commandline it is a noticeable delay.

The scanning of all files takes a while, especially since there are many `html` and `js` files in `webroot` which is not used by the commandline.

Maybe the commandline can pass a flag indicating that the autoupdater can skip checking the webroot folder.

PR #2362, opened 2017-03-08
Update manager : Ignore webroot Folder

frelis:
Added a property to ignore, if set, the webroot folder, when checking if
the update packages  are not tampered

fixes duplicati/duplicati#2266

### DIFF
--- a/Duplicati/CommandLine/Program.cs
+++ b/Duplicati/CommandLine/Program.cs
@@ -51,8 +51,6 @@ namespace Duplicati.CommandLine
                 var options = tmpparsed.Item1;
                 var filter = tmpparsed.Item2;
 
-                Duplicati.Library.AutoUpdater.UpdaterManager.IgnoreWebrootFolder = true;
-
                 verboseErrors = Library.Utility.Utility.ParseBoolOption(options, "debug-output");
                 verbose = Library.Utility.Utility.ParseBoolOption(options, "verbose");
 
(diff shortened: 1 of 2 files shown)


Duplicati/CommandLine/Program.cs
```diff
@@ -50,6 +51,8 @@ public static int RealMain(string[] args)
                 var options = tmpparsed.Item1;
                 var filter = tmpparsed.Item2;
 
+                Duplicati.Library.AutoUpdater.UpdaterManager.IgnoreWebrootFolder = true;
```
kenkendk: Is this required?

kenkendk:
Looks nice!

I guess the other commandline tools could use the same flag?

Did you notice any speedups from this?

frelis:
Yes i an attribute of the class if you set tit in other tools it will work.
I was not able to test it, sorry, i am new on this, i was not able to generate the manifest file.

Duplicati/CommandLine/Program.cs
frelis: I think so, because lower in the code, the Updater is called again,

kenkendk: Yes, but the static variable is shared, so if you set it once, it stays set?

frelis: You are right, I already changed.

PR #2367, opened 2017-03-10
Fixed an error in renaming absolute Windows paths.

Danstahr:
Previously, the output of the first replace was never used and therefore the absolute Windows paths were not restored correctly.

The two patterns (i.e. ?: and \\\\) are disjoint so overwriting the previous result in case a match is found
for the second one is OK.


kenkendk:
Good catch!

PR #2370, opened 2017-03-12
Jottacloud backend: Minor adjustment in the verification of file state

albertony:
I have been thinking about the file states in the Jottacloud API since #2355 was merged, and came to the conclusion that we could do a minor adjustments to make it more "bullet proof". Jottacloud tracks file revisions with states, but since its completely undocumented its not that easy to figure out exactly how the logic is. With this commit we are a bit more on the safe side, throwing FileUploadError if there is the slightest possibility the file upload did not go as expected, and making sure we are not listing files which is not in expected state.

### DIFF
--- a/Duplicati/Library/Backend/Jottacloud/Jottacloud.cs
+++ b/Duplicati/Library/Backend/Jottacloud/Jottacloud.cs
@@ -204,16 +204,20 @@ namespace Duplicati.Library.Backend
                 System.Xml.XmlNode xRevision = xFile.SelectSingleNode("currentRevision");
                 if (xRevision != null)
                 {
-                    System.Xml.XmlNode xNode = xRevision.SelectSingleNode("size");
-                    long size;
-                    if (xNode == null || !long.TryParse(xNode.InnerText, out size))
-                        size = -1;
-                    DateTime lastModified;
-                    xNode = xRevision.SelectSingleNode("modified"); // There is created, modified and updated time stamps, but not last accessed.
-                    if (xNode == null || !DateTime.TryParseExact(xNode.InnerText, JFS_DATE_FORMAT, System.Globalization.CultureInfo.InvariantCulture, System.Globalization.DateTimeStyles.AdjustToUniversal, out lastModified))
-                        lastModified = new DateTime();
-                    FileEntry fe = new FileEntry(name, size, lastModified, lastModified);
-                    files.Add(fe);
+                    System.Xml.XmlNode xNode = xRevision.SelectSingleNode("state");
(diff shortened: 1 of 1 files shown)


jkaberg:
I'm sorry for beeing offtopic but wanted to ask @albertony: could you implement jotta support in [rclone](https://github.com/ncw/rclone)? I'd love to see it there aswell.

Ontopic: awesome work! Can't wait to get jotta up and running here 😄

kenkendk:
Ok, I am merging in. The canary build (2.0.1.52) does not include this change, in case there is some debugging later.

albertony:
Perfect, thanks.

albertony:
@jkaberg : Did not know about rclone before, but it looks interesting so maybe I will have a look at it. But not today! ;)

jkaberg:
@albertony @kenkendk I guess Jottacloud should be included by now? I installed 2.0.1.61_experimental_2017-05-30 just now, and I can't seem to find Jotta when selecting backend from the Web UI

albertony:
Yes it should. I am using it daily myself, but then I am building Duplicati from source so maybe it is somehow missing from the release/installer version. Could it be that the recent commit for #2535 solves it?

jkaberg:
@albertony just tried with latest canary build (2.0.1.66-1) which has Jotta enabled, but I couldn't get it to connect to Jotta. Any requirements out of the ordinary?

albertony:
Good you got it working, but it should have created the destination folder as well. What is your destination path?

jkaberg:
@albertony <hostname>/<folder>/.. didn't alter the mountpoint (using default Archive)

albertony:
So: From WebUI, what did you write as the folder path? Did you mean you wrote "//"? I can see that paths such as this slips through the verification of folder path.. The basic use case is that you write a folder name/path (which is required input), for example "MyBackup" and then this folder will be created in the built-in "Archive" mount point of the built-in "Jotta" device. When you click the "Test connection" button it should ask you if it should create the folder if it does not exist. Sorry I'm still a bit confused about your exact case.

jkaberg:
@albertony I wrote "test/folder/" as path, when I hit Test connection I get asked if I want to create the folder where Duplicati after a while reponds with an timeout

Previous path in previous path got truncated due to formatting

albertony:
> Previous path in previous path got truncated due to formatting.

Ah, that explains it!

> I wrote "test/folder/" as path, when I hit Test connection I get asked if I want to create the folder where I hit yes. Duplicati then after a while reponds with an timeout.

Strange, this always works for me..

PR #2385, opened 2017-03-17
tahoe_lafs_uri_message_fix

koo5:
not sure if intended behavior, but without the leading '/' is how it actually works


kenkendk:
Yes, that is how it works, must have been a text that was not updated from some earlier experiment.

PR #2388, opened 2017-03-18
Google Oauth integrated in application

easysoft2k15:
I would like to being able to integrate _OAuth2_ directly in the application without using the external web service You provide.
This is my implementation ( for Google Drive) that somehow plugin in into the framework.
I thought it may be of interest for the community as well.
There are a couple of gotcha:
1. the Client Id and Secret are hard coded so You need to recompile to use Your own. This should not be a problem as probably whoever will use this features will likely want to recompile is own version. Don't know if there may be security concern.
2. I use my own file for saving needed data. This file is stored on the same directory as the application

If You would like to merge into master branch but need some "adjustment" please let me know.
Alessandro

### DIFF
--- a/Duplicati/Library/Backend/GoogleServices/Duplicati.Library.Backend.GoogleServices.csproj
+++ b/Duplicati/Library/Backend/GoogleServices/Duplicati.Library.Backend.GoogleServices.csproj
@@ -1,4 +1,4 @@
-<?xml version="1.0" encoding="utf-8"?>
+﻿<?xml version="1.0" encoding="utf-8"?>
 <Project DefaultTargets="Build" ToolsVersion="4.0" xmlns="http://schemas.microsoft.com/developer/msbuild/2003">
   <PropertyGroup>
     <Configuration Condition=" '$(Configuration)' == '' ">Debug</Configuration>
@@ -36,6 +36,7 @@
     </Reference>
   </ItemGroup>
   <ItemGroup>
+    <Compile Include="GoogleOAuthHelper.cs" />
     <Compile Include="Properties\AssemblyInfo.cs" />
     <Compile Include="GoogleCloudStorage.cs" />
     <Compile Include="Strings.cs" />
--- a/Duplicati/Library/Backend/GoogleServices/GoogleDrive.cs
+++ b/Duplicati/Library/Backend/GoogleServices/GoogleDrive.cs
@@ -55,7 +55,8 @@ namespace Duplicati.Library.Backend.GoogleDrive
             if (options.ContainsKey(AUTHID_OPTION))
                 authid = options[AUTHID_OPTION];
 
-            m_oauth = new OAuthHelper(authid, this.ProtocolKey) { AutoAuthHeader = true };
+            m_oauth = authid==null || authid == ""? new GoogleOAuthHelper():
+                new OAuthHelper(authid, this.ProtocolKey) { AutoAuthHeader = true };
             m_filecache = new Dictionary<string, GoogleDriveFolderItem[]>();
         }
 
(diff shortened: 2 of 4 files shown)


easysoft2k15:
One thing I forgot to mention.
The code that is actually making the OAuth authentication is taken from Google repository and I believe it has a _Apache 2 License_.
So before You accept the pull request I'll probably need to split this part of the code on a different file in order to complete it with the license.

kenkendk:
I like the idea that you can authenticate without a third party server.
Unfortunately, this was not a use-case for the OAuth 2.0 spec so we are stuck with needing a WebUI somewhere.

Your version is a bit like "run-your-own-server", so it will not work for most users, notably because it requires some port-mapping to be reachable from the internet.
On top of this, you need to set up OAuth authentication to get the secrets, and lock down the allowed urls.

In other words: this is for advanced users :).

Since you need the secret values somewhere, would it be better if all information was stored in a file?

Maybe it would be simpler and more consistent to simply allow the user to set these as options, something like:
```
--google-oauth-client-secret=123
--google-oauth-client-id=xyz
--google-oauth-scope=openid https://www.googleapis.com/auth/drive
--google-oauth-port=...
--google-oauth-ip=...
--google-oauth-browser-command=firefox
```

These can then be set in the `--parameterfile` and re-used.

But since this relies on a system with a web-browser, it cannot be used from the commandline, so maybe it would be better to have this as a more generic "OAuth self-hosting" tool?

I think it could work with the other OAuth systems with a few tweaks here and there.

This would allow system admins (and power-user friends) to generate the required tokens and pass them to the end users?

What are your thoughts on this?

Btw: the build fails because you need to add the `GoogleOAuthHelper.cs` file to the Google `.csproj` file.

easysoft2k15:
Hi Kenneth,
thank You for the quick replay.

The main reason behind this would be to allow people (or Companies for that matter) that would like to use Duplicati to package up their own product (of course following the requested license agreement). So the scenario I see for this is for example a Company that wants to build a software package based on Duplicati to  distribute to their users in order to sell some kind of "crypto-backup" service for the cloud.
Of course I don't know if this is a direction You'd like to go.

On the technical side, my understanding of _OAuth 2_ is that it should allow also client application to being able to authenticate with an authentication provide like Google and alike.
The code I committed bring up a browser and ask the user for permission to his Google Drive files even if the user is using the CommandLine version of Duplicati.

For the aforementioned reasons having the Client ID and Secret inside the code would make sense: the Company that will provide the "Crypto-backup" service would be responsible for using their own credential and the end user will just use the final product without care about _Oauth 2_ or whatever is needed to make the service working.
In any case have the credential compiled into IL code is not ideal as someone may just decompiled it but probably there's some way around it using some kind of encrypted keystore.
I'd like to know You thoughts on the general idea.

Btw: I committed also the _.csproj_ file

Alessandro

kenkendk:
I have no problem with other people using Duplicati for various commercial or personal purposes.

But, embedding the OAuth secrets is problematic for at least three reasons:

1) It is easy to steal. If someone steals it they can disrupt service for many.
2) Many OAuth providers (rightfully) disallow distributing the tokens embedded in code
3) Distributing modified code triggers the LGPL license and requires you make the modified code publicly available

I was thinking that the changes make more sense for an end user that knows how to get an OAuth secret and want to use Duplicati without relying on the OAuth service. Such a user would benefit more from providing the OAuth values to the unmodified binary, such that updates etc. works without hassle.

I do not like that a component within the main Duplicati requires that you have a listening webserver that is publicly reachable. 

I also dislike the idea of opening a browser window, as it is not guaranteed that this can work (i.e. using SSH, or using the WebUI from another machine, or running Duplicati as a service).

For all the above reasons, a better approach would perhaps be to use the code you have already, to run an external OAuth service. The Company can then either run this service on their own machine, or set it up so that each client runs this service as well. If this approach is used, the Company can just change the default `--oauth-url` to point to the server, and everything works the same way.

This new binary can have another license (MIT?) and allow embedding stuff more easily.

Not sure you are aware, but the code use to run the current OAuth server is here:
https://github.com/duplicati/oauth-handler

Issue #2066, opened 2016-10-30
create-report appends zip even if output-file name already ends with .zip


kakoskin1:
I have:
- [x] searched open and closed issues for duplicates

---
### Version info

**Duplicati Version:** 2.0.1.29_canary_2016-10-27
**Operating System:** Fedora 24
**Backend:** File
### Bug description

Command "duplicati-cli create-report <output-file> appends .zip to output-file name even if output-file name already ends with .zip
### Steps to reproduce
- duplicati-cli create-report <path-to-some-backup> output-file.zip

**Actual result:** 
Duplicati produces .zip file named output-file.zip.zip.

**Expected result:** 
Duplicati produces .zip file named output-file.zip.
##

albertony:
This issue could be closed, I guess?

ts678:
Can this be closed based on above fix? 2.0.5.111 on Windows tests fine using `.zip`, `.ZIP`, and no suffix. You get a single suffix.

PR #2394, opened 2017-03-20
Fixed issue #2066: The command create-report appends zip even if outp…

albertony:
…ut-file name already ends with .zip. Reason is that System.IO.Path.GetExtension returns extension including the '.' character, so that needs to be excluded when comparing it to the compression module name. Also made the comparison case-insensitive so that MYFILE.ZIP will still be matched to compression module "zip" without making the file name MYFILE.ZIP.zip.


kenkendk:
Arh, I was wondering why it added an extra extension :)

PR #2401, opened 2017-03-23
Improved database performance and logging

Danstahr:
- Don't use Vacuum() just before PurgeLogData (which performs vacuum() already)
 - Removed unncessary poorly performning query (after optimizing it I found out it's unused)
 - Correctly log calls for ExecuteScalarInt64

### DIFF
--- a/Duplicati/Library/Main/Database/ExtensionMethods.cs
+++ b/Duplicati/Library/Main/Database/ExtensionMethods.cs
@@ -87,7 +87,7 @@ namespace Duplicati.Library.Main.Database
                     self.AddParameter(n);
             }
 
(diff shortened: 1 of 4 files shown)


kenkendk:
The query used to toggle how the cache is used, in that missing hashes means that the cache would be incomplete. But since I removed the cache, this is not used anymore.

Danstahr:
Thanks for the merge!

The problem with the query was that sqlite performed full scan on the
blocks table and then searched the remotevolume one. With blocks being the
largest table this performed poorly, especially since the usual list of
erratic remote volumes was short.

Forcing the RemoteVolume table to be scanned first by using a subquery
solved the problem.

PR #2421, opened 2017-04-06
Translations updated

ponsfrilus:
* `xxx ...` -> `xxx...`
* some translations added
* some spelling corrected


kenkendk:
@ponsfrilus We use Transifex to manage the translations:
https://www.transifex.com/duplicati/duplicati/languages/

If you log in there, you can edit the strings and discuss with the other translators what the best fixes are.

ponsfrilus:
@kenkendk I haven't seen that. I suggest that you add the link to transifex in the *.po files. I do not want to do the work twice, so I hope someone will use that PR. If not, never mind !

PR #2424, opened 2017-04-09
restore_from_python.py

downpoured:
by Ben Fisher, https://github.com/downpoured
a Python script to restore Duplicati files
similar to Duplicati.RecoveryTool, but with no dependencies on Mono/.NET
uses streaming apis to restore a large number of files and use limited RAM.
supports backups using AES encryption (.aes) or No Encryption (.zip),
if data uses GPG/other encryption, decrypt files to .zip before running this tool.

A tool like this is beneficial from a digital preservation point of view, which has been mentioned on the mailing lists a few times, 
https://groups.google.com/forum/#!topic/duplicati/yk4Vvr0KRfE
https://groups.google.com/forum/#!topic/duplicati/uRAcHaBQDfA


kenkendk:
Awesome!

Why is the `ijson.py` file required, is it not enough to use the plain `json` module?

Should we perhaps put the files into its own folder, such that it is clear what belongs together?
Also, maybe we should put in a README in the folder that explains that there might be an updated version in your repo?

downpoured:
Those are good suggestions.

The built-in json module works, but it constructs the entire object in memory, which would lead to heavy memory usage for the dlist in a backup with many files. I guess it's not strictly needed if the machine has enough RAM.

kenkendk:
I see, loading the full json is likely very memory demanding.

Issue #2302, opened 2017-02-02
Performance issues when recreating the local database [$25]


kenkendk:
This issue is a collection issue for a few issues related to recreating the database.

If you add comments to this issue, make sure to include what version you are using, and make sure you are using at least 2.0.1.44. This issue is ONLY about the performance aspect of the recreate, if you have errors on recreate, please open a new issue.

The related issues are:
#1699
#1391
#2048 
#2050
#2140 
#2178

emaktech:
All I can say is that I'm experiencing the same issue with a backup that is around 60GB and contains around 50,000 files.

I'm on the latest build 2.0.1.46_canary_2017-02-03 and the backup was restored on another system via JSON file.

The operation has been running for the last 4 hours and is less than halfway complete. There seems to be one CPU core that is used fully. I will report back if this operation ever completes successfully or not. At the moment it seems to take at least a full day to recreate a local database.

I also am experiencing very slow browsing of subdirectories on the client PC that backed up the files (about 15 minutes to navigate into each subdirectory through the GUI).

basrieter:
Same issue here. 200 GB backup for which browsing took for ages. I decided to recreate the DB. It uses my SSD and has been running for 12 hours now. 

```
[...]

PR #2434, opened 2017-04-15
Index on BlocklistHash

zed9h:
Related to the performance issues on database recreation, as commented on #2302 

This is my first contribution, see if this is all right.


kenkendk:
Looks fine. Do you have some numbers on how much faster it is?

kenkendk:
I think I will merge it with: https://github.com/duplicati/duplicati/tree/feature/guard_against_blocklisthash_duplicates

So we only get a single database update.

PR #2451, opened 2017-05-02
Tray readpassword

mnaiman:
Added ability to read web password for trayicon from server DB


Duplicati/GUI/Duplicati.GUI.TrayIcon/HttpServerConnection.cs
```diff
@@ -439,14 +438,8 @@ public string StatusWindowURL
         {
             get 
             { 
-                try
```
kenkendk: If you remove the try/catch here, the trayicon could crash if `GetAuthToken()` fails?

Duplicati/GUI/Duplicati.GUI.TrayIcon/HttpServerConnection.cs
```diff
@@ -305,28 +306,25 @@ private T PerformRequest<T>(string method, string urlfragment, Dictionary<string
...
-                        try 
-                        { 
-                            var c = httpex.Cookies[XSRF_COOKIE];
-                            if (c != null)
-                                t = c.Value;
-                        }
-                        catch
-                        {
-                        }
+                        var t = httpex.Cookies[XSRF_COOKIE]?.Value;
```
kenkendk: Are you sure there is no exception if you access a cookie that is not there?

Duplicati/GUI/Duplicati.GUI.TrayIcon/Program.cs
```diff
@@ -133,73 +137,256 @@ public static void RealMain(string[] _args)
...
+                        Console.WriteLine("Request error: " + ex.Message);
+
+                        reSpawn++;
                     }
-                }
+                } while (reSpawn < 3);
+            }
+        }
+
+        private static Server.Database.Connection GetDatabaseConnection(string[] args)
```
kenkendk: Why not make `GetDatabaseConnection` public in `Duplicati.Server` and then call it there?

Duplicati/GUI/Duplicati.GUI.TrayIcon/Program.cs
```diff
@@ -133,73 +137,256 @@ public static void RealMain(string[] _args)
...
-                        // as it would be disposed by now
-                        if (hosted != null)
-                            hosted.InstanceShutdown -= shutdownEvent;
+        /// <summary>
+        /// Helper method with logic to handle opening a database in possibly encrypted format
+        /// </summary>
+        /// <param name="con">The SQLite connection object</param>
+        /// <param name="useDatabaseEncryption">Specify if database is encrypted</param>
+        /// <param name="password">Encryption password</param>
+        private static void OpenDatabase(System.Data.IDbConnection con, bool useDatabaseEncryption, string password)
```
kenkendk: Why not make `OpenDatabase` public in `Duplicati.Server` and then call it there?

Duplicati/Server/Database/ServerSettings.cs
```diff
@@ -299,9 +303,41 @@ public void SetWebserverPassword(string password)
...
+
+        public void GenerateWebserverPasswordTrayIcon()
+        {
+            var password = "";
+            var pwd = "";
+
+            if (!string.IsNullOrEmpty(m_values[CONST.SERVER_PASSPHRASE]))
+            {
+                password = Guid.NewGuid().ToString();
+                var buf = Convert.FromBase64String(m_values[CONST.SERVER_PASSPHRASE_SALT]);
```
kenkendk: I don't think there is a guarantee that the `SERVER_PASSPHRASE_SALT` value is not empty?

Duplicati/Server/WebServer/AuthenticationHandler.cs
```diff
@@ -179,14 +179,19 @@ public override bool Process(HttpServer.IHttpRequest request, HttpServer.IHttpRe
                         return true;
                     }
 
+                    var password = Program.DataConnection.ApplicationSettings.WebserverPassword;
+
+                    if (request.Headers["User-Agent"] != null && request.Headers["User-Agent"].StartsWith("Duplicati TrayIcon Monitor"))
```
kenkendk: We need a better way to do this. Relying on the `User-Agent` header seems dodgy.

We can either have an alternative `LOGIN_SCRIPT_URI`, or use a custom header like `X-TrayIcon-Login`.

Duplicati/Server/Program.cs
```diff
@@ -380,6 +380,8 @@ public static void RealMain(string[] args)
                 if (commandlineOptions.ContainsKey("webservice-password"))
                     Program.DataConnection.ApplicationSettings.SetWebserverPassword(commandlineOptions["webservice-password"]);
 
+                Program.DataConnection.ApplicationSettings.GenerateWebserverPasswordTrayIcon();
```
kenkendk: This would mean that the TrayIcon login feature is enabled for all users.
Can the TrayIcon generate the password instead? That way it will only be generated for users who want this feature.
We then need to make sure the login does not accept an empty password, so we do not open all machines.

Duplicati/GUI/Duplicati.GUI.TrayIcon/Program.cs
```diff
@@ -133,73 +137,256 @@ public static void RealMain(string[] _args)
...
                     {
-                        int port = Duplicati.Server.Program.ServerPort;
-                        url = "http://127.0.0.1:" + port;
+                        System.Diagnostics.Trace.WriteLine("Request error: " + ex.Message);
+                        Console.WriteLine("Request error: " + ex.Message);
+
+                        reSpawn++;
                     }
-                }
+                } while (reSpawn < 3);
```
kenkendk: Why 3?

Duplicati/GUI/Duplicati.GUI.TrayIcon/HttpServerConnection.cs
mnaiman: I have tested it. Result:

httpex.Cookies[XSRF_COOKIE]?.Value
"Vu1CieBQLjIR3QqXExXkN6siIcgQeMPUbqnTts5yBcc%3D"
httpex.Cookies["test"]?.Value
null
httpex.Cookies["test"]
null

Duplicati/GUI/Duplicati.GUI.TrayIcon/Program.cs
mnaiman: Fixed in next commit

Duplicati/GUI/Duplicati.GUI.TrayIcon/Program.cs
mnaiman: Fixed in next commit

Duplicati/Server/Program.cs
mnaiman: Whole feature logic is:
1) When server is started, it generate random password for tray icon (separate from password for login form). Because user dont need to know password (it is only for confirmation, that tray know secret from database) it is not problem that is generated for all users (of course when user disable password at all, even tray password will be empty)
2) SERVER_PASSPHRASE_SALT is set by SetWebserverPassword. So when password is empty, salt is empty.
3) TrayIconPassword share SALT with FormPassword

Duplicati/Server/Database/ServerSettings.cs
mnaiman: Comment is lower

Duplicati/Server/WebServer/AuthenticationHandler.cs
mnaiman: Yeah X-TrayIcon-Login could be option. Logic was that if client is not browser (user), but TrayIcon it will not check FormPassword, but TrayIconPassword.

Duplicati/GUI/Duplicati.GUI.TrayIcon/HttpServerConnection.cs
kenkendk: Yes, I did not see the `?` null-chain operator.

Duplicati/GUI/Duplicati.GUI.TrayIcon/Program.cs
```diff
@@ -133,73 +137,95 @@ public static void RealMain(string[] _args)
...
-                        tk.Init(_args);
-
-                        // Make sure that the server shutdown does not access the tray-icon,
-                        // as it would be disposed by now
-                        if (hosted != null)
-                            hosted.InstanceShutdown -= shutdownEvent;
+                        reSpawn++;
                     }
-                }
+                } while (reSpawn < 3);
```
kenkendk: Why is this required? And why is it `3` ?

Duplicati/Server/WebServer/AuthenticationHandler.cs
kenkendk: Yes, that would be great.

Duplicati/GUI/Duplicati.GUI.TrayIcon/HttpServerConnection.cs
mnaiman: Generally is not good to have empty catches, this is why I removed these try-catch and allow Exception to be bubble to RealMain, where WebException is caught.  There is this reSpawn < 3 - idea was to give some option to absorb eg network problems, but after 3 tries to fail. This will not happen now, because empty catch will suppress this fail.
Im thinking about add some code to inform user, or add logging to TrayIcon, because no info is passed to user.

Duplicati/Server/WebServer/AuthenticationHandler.cs
mnaiman: No problem, done.

PR #2455, opened 2017-05-04
Russian localization and ClearSans font without subsetting

vnau:
Russian localization on transifex now good enough to add to Duplicati.

Currently Duplicati use subsetted Clear Sans webfonts so some chars in polish and most russian chars not covered. Fonts replaced to ones without subsetting.


agrajaghh:
I just updated the strings at transifex and included russian. Could you please rebase this PR to just contain the added strings and the full webfonts? 😊

vnau:
Just rebased, leaving two commits with strings and fonts.

PR #2460, opened 2017-05-06
Improved BackupHandler performance

Danstahr:
For backends with sufficiently high throughput (e.g. local NAS), hashing the files / blocks is the most time consuming part. Performing the two hashes in parallel saves approximately 15% of the backup time alone.

There's more room for parallelization in he backup handler - I'll send more pull requests revolving around this in near future.


kenkendk:
I have a fully parallel processing system ready here:
https://github.com/duplicati/duplicati/tree/concurrent_processing

It does as many parallel hashings as you like, and also performs compressions in parallel.

I am holding back merging the changes until we have a beta release.

Danstahr:
Nice! Is there anything I can help with?

kenkendk:
Sure!

Right now it does not merge because of many updates to the master branch.
I think it is easier if I do that once I have some time, so we avoid loosing some fixes in the process.

In the `concurrent_processing` branch, there are two `BackendManager` classes, the old one and the new concurrent one. The backup process is using the new one, but all the other operations are using the old one. We should delete the old one and then use the new one in all the other operations.

Also, the backup operation is encapsulating access to the database to force single-threaded access to it. We should wrap all the database operations from the other operations as well.

Not sure it it makes sense to work on any of this before I clean up the merge conflicts though.

Danstahr:
Attempted to do the merge in #2485.

PR #2485, opened 2017-05-18
Merged the concurrent_processing branch with master

Danstahr:
I have successfully done a basic functionality test (backup / restore cycle), but this will need more testing before going to master.


kenkendk:
@Danstahr Kudos for the hard work!
This is a really hard merge, as much of the code in `BackupHandler.cs` has been split to multiple files, so som manual tracking has to be done to make sure everything from the 500+ commits is merged in.
I hand-tracked all the files with conflicts and have updated the concurrent_processing branch now.
It is possible that I broke something in the process, but lets see.

Issue #2473, opened 2017-05-15
Feature request: Test write permissions on 'SFTP Test connection'


Dobatymo:
I am using 2.0.1.53_experimental_2017-03-13.

Recently my backups failed with a ftp 550 error. But the 'test connection' button said everything is fine. Later i found out my write permissions were removed on that ftp server. It would be useful if the ftp 'test connection' button would test for correct write permissions and tell the users if any permissions are lacking.

kenkendk:
Yes, should be simple to do too.

duplicatibot:
This issue has been mentioned on **Duplicati**. There might be relevant details there:

https://forum.duplicati.com/t/ftp-vs-ftp-alternative/8690/6

ts678:
It seems like posting URLs to the forum no longer adds mentions, so doing by hand:

https://forum.duplicati.com/t/sftp-connection-issue-to-pi4-solved/17644/6?u=ts678

was an SFTP failure that would have gotten simpler with this Test connection change.

kenkendk:
This is actually implemented for the FTP backend, so I have updated the description to mention SFTP

PR #2500, opened 2017-05-23
Test write permissions on 'FTP Test connection'

vnau:
Check access permissions (Read/Write/Delete) for AlternativeFTP Backend using test file during connection test (Issue: #2473).

The same checks may be applied to FTP backend and other backends with possible access restrictions.


kenkendk:
Looks good!

Maybe we should move it later, so this can be done for all backends.

PR #2510, opened 2017-05-29
Enable Caching for static webroot files

mknjc:
Change the FileModule constructor of "/" to enable "Last-modified" header which allows webbrowsers to cache static files.


kenkendk:
Thanks!

Issue #2440, opened 2017-04-24
Include/Exclude filters not available to run-script


nsrussell:
I have:
- [x] searched open and closed issues for duplicates

----------------------------------------

### Version info

**Duplicati Version:** 2.0.1.53
**Operating System:** Windows 10 64-bit
**Backend:** Any

### Bug description
Include/Exclude filters not available to run-script

### Steps to reproduce
- Create a backup job and add an exclude filter, eg -[.AppData.]
- Add a run-script-before batch file that redirects 'set' to a file, eg set >C:\Temp\debug.log
- Execute the job

**Actual result:** Gets lots of environment variables including many DUPLICATI__* ones, however not DUPLICATI__FILTER as per the run-script-example or any other variable containing the exclude filter.
**Expected result:** A DUPLICATI__FILTER environment variable containing "-[.AppData.]" for the example filter mentioned above.

kenkendk:
Oh, that is an omission on my part :/

It also does not allow you to modify the filters.

The output sent to the script should be added here:
https://github.com/duplicati/duplicati/blob/master/Duplicati/Library/Modules/Builtin/RunScript.cs#L185

And the readback here:
https://github.com/duplicati/duplicati/blob/master/Duplicati/Library/Modules/Builtin/RunScript.cs#L267

We might want to do something like the parsing done from the parameters file (prependfilter and appendfilter):
https://github.com/duplicati/duplicati/blob/master/Duplicati/CommandLine/Program.cs#L349

IGTHORN:
Hi @kenkendk 

I think I have solved this issue. However, when performing a restore operation from the GUI, I get an error saying that filters are not supported. It seems that during restore, it also runs the startup script and reads in the supplied filters. Should I prepare a pull request so you can see the issue? I think that another change would be need to either not add filters from the startup script during restore; or, not run the startup script at all during restore (though that would change existing behaviour and is probably less desirable).

nsrussell:
Getting an error when setting the filter for a restore operation would seem to be correct behaviour to me. The run script should be using the DUPLICATI__OPERATIONNAME environment variable to determine what to do and only setting filters when this is "Backup".

kenkendk:
This is closed with e88afa4

PR #2528, opened 2017-06-07
Bug fix for "Include/Exclude filters not available to run-script" #2440

IGTHORN:
Fix for "Include/Exclude filters not available to run-script" #2440. Passing the "filter" option to the Configure method so that the before-script can alter its value. Removing "filter" option afterwards because it is not a supported command line option, merely a placeholder for the before-script to modify the filters.


Duplicati/Library/Main/Controller.cs
```diff
@@ -671,21 +671,18 @@ private void SetupCommonOptions(ISetCommonOptions result, ref string[] paths, re
...
+            var pristinefilter = string.Join(System.IO.Path.PathSeparator.ToString(), FilterExpression.Serialize(filter));
+            m_options.RawOptions["filter"] = pristinefilter;
+            
             var qp = new Library.Utility.Uri(m_backend).QueryParameters;
             foreach(var k in qp.Keys)
-                conopts[(string)k] = qp[(string)k];
-
-            // Make the filter read-n-write able in the generic modules
-            var pristinefilter = conopts["filter"] = string.Join(System.IO.Path.PathSeparator.ToString(), FilterExpression.Serialize(filter));
+                m_options.RawOptions[(string)k] = qp[(string)k];
```
kenkendk: I think this will mean that all connection options (i.e. those passed as the query string) will overwrite basic options, like:
```
duplicati_cli backup "xyz://test?dblocksize=1mb" /sourcedata --dblock-size=2mb
```

Is there a reason for doing that?

IGTHORN: I wanted to put all the options into one dictionary so that the rest of the configuration code could treat them the same.

I wasn't sure whether URL options should override command line parameters or vice-versa. I'm not sure what the use case would be to supply an option in both the URL and as a command line parameter. I can switch it around so that the URL options override command line parameters rather?

kenkendk: The intention with the URL parameters is that all connection related options can be stored in a single URL.
The `IConnectionModule` is an interface I introduced to mark that the module needs the url arguments.

So the logic is that the URL parameters should override the commandline arguments, but _only_ for the connection modules.

IGTHORN: Hmm OK, is there ever an instance where you would want an IConnectionModule to be able to modify the options dictionary that is passed to it in Configure()? Because the problem I'm seeing is that RunScript takes a reference to the options dictionary that is passed to it in the Configure() method and then modifies this dictionary based on the script output. The modifications that RunScript makes must be visible to the rest of the process, so they're stored in m_options. The previous code would pass a copy of the dictionary to IConnectionModule instances, which means if they were to modify the dictionary (in a similar way to RunScript) those changes are not reflected in m_options and would not be visible to the rest of the process.

I have updated the code with the assumption that there is no need for an IConnectionModule to change the options. I have brought back the conopts dictionary so that options supplied on the URL would only be visible to modules that implement IConnectionModule. I thought maybe the reason to prevent other modules from seeing these options is that they contain usernames and passwords (eg?user=x&pwd=y). However, that is visible in the remoteurl variable of Execute() and visible to batch scripts in DUPLICATI__REMOTEURL so I'm actually not sure. Is there some other reason for wanting to separate them? I don't see an issue with letting URL options override command line options _always_, or vice versa, and it would eliminate the extra dictionary in the code.

IConnectionModule marks the module as wanting to see the URL connection options, right? Is there a downside to having other modules see those options?

If you want IConnectionModule instances to be able to change m_options, I'll need to make some additional changes.

kenkendk: Hmm, good question...
Generally, the code should not change the m_options dictionary, except for the special case where user-provided code is added (the script module).

The only class currently implementing the `IConnectionOptions` interface is `HttpOptions`. It needs the URL stuff because it could include allowed certificate hashes etc., but it does not need to modify the options.

I think the updated version achieves the same as before, but with the script problem fixed.

As for not exposing the password, I agree we should fix that, but it is currently leaking from other places as well (into the WebUI). We should revisit that idea once we shield it from the WebUI.

The other question "...letting URL options override command line options always...".
I would find it surprising if I pasted a URL, and that changed some of my non-connection settings. I don't think we can use just "override" because options have default values, and having the non-overriding side set it can change the value, except if it is also set by the overriding side. 
With the current setup, the url-parameters always override the commandline options, but the options are only seen in the backend (and the `HttpModule`).

PR #2529, opened 2017-06-08
Added support for executing non-SQL code when performing a database upgrade

Danstahr:
Some of the database migrations I'm planning to perform are very slow if using SQL only
or entirely impossible using the standard SQLite toolchain. I added a simple interface
with "before SQL" and "after SQL" hooks so that a database upgrade can perform custom
code operations as well.


kenkendk:
Ok, looks good.

I think you can remove the comment about the transaction, as the SQL upgrades also explicitly handles the transactions.

PR #2534, opened 2017-06-10
Fixed a condition in ListChangesHandler.

Danstahr:
The compare command was failing with ArgumentNullException.


kenkendk:
Thanks!

PR #2535, opened 2017-06-10
Make Jottacloud available to public.

Danstahr:


kenkendk:
Thanks, I completely overlooked that part...

PR #2558, opened 2017-06-22
added console support on windows in trayicon

mnaiman:
added console support on windows
HttpOptions to choose https parametrs to connect to server (like https tls versions)


Duplicati/GUI/Duplicati.GUI.TrayIcon/Program.cs
```diff
@@ -174,7 +177,10 @@ public static void RealMain(string[] _args)
 
             if (options.TryGetValue(HOSTURL_OPTION, out url))
                 serverURL = new Uri(url);
-            
+
+            System.Net.ServicePointManager.SecurityProtocol = SecurityProtocolType.Tls12 | SecurityProtocolType.Tls11;
+            new Duplicati.Library.Modules.Builtin.HttpOptions().Configure(options);
```
kenkendk: The `Duplicati.Library.Modules.Builtin.HttpOptions` class is meant to be disposed.
It is a bit of a dangling setup to use it here, as it may be overridden later by a backup.

What about setting the protocol on the request here instead:
https://github.com/duplicati/duplicati/blob/master/Duplicati/GUI/Duplicati.GUI.TrayIcon/HttpServerConnection.cs#L357

Duplicati/GUI/Duplicati.GUI.TrayIcon/HttpServerConnection.cs
```diff
@@ -259,6 +259,9 @@ private string GetAuthToken()
             var sha256 = System.Security.Cryptography.SHA256.Create();
             var password = m_password;
 
+            if (m_password == null)
```
mnaiman: null is sufficient. Main reason was to not pass null there System.Text.Encoding.UTF8.GetBytes.
If password is "", it will go ok whole function, of course server will return unauthorized.
IsNullOrWhiteSpace will work too.

Duplicati/GUI/Duplicati.GUI.TrayIcon/Program.cs
```diff
@@ -64,6 +64,9 @@ public static int Main(string[] args)
         
         public static void RealMain(string[] _args)
         {
+            if (Duplicati.Library.Utility.Utility.IsClientWindows)
+                Duplicati.Library.Utility.Win32.AttachConsole(Duplicati.Library.Utility.Win32.ATTACH_PARENT_PROCESS);
+
```
mnaiman: thx

Duplicati/GUI/Duplicati.GUI.TrayIcon/Program.cs
mnaiman: Yeah, you are right, I had same worries, especially with override in backup.
Same implies for backup, that this settings could influence backup options.

Issue #2578, opened 2017-06-30
Don't use `VACUUM` automatically


kenkendk:
The current implementation uses `VACUUM` freely in an attempt to reduce space used by the local database. This works well for small backups, but the database can grow quite large (+60GiB) and then it becomes a problem, as the `VACUUM` step writes a new clean copy of the database.

To fix this we need multiple steps:
  - Remove all uses of `VACUUM` in the code
  - Add a commandline `vacuum` command that allows manual activation
  - Add a "Reclaim space" button in the UI that calls the `vacuum` operation
  - Add a `--auto-vacuum` option that calls the `vaccum` command after an operation

saviodsouza:
+60GB database ???

kenkendk:
Yes, this is reported in #2382

kenkendk:
I will close this even though the "Add a 'Reclaim space' button ..." is not implemented, as it is possible to execute this from the "Commandline" part of the UI, and the operation is not expected to be used by most users.

PR #2626, opened 2017-08-04
Progress on #2578

Danstahr:
Progress on the individual tasks of the issue:

- [x] Remove all uses of VACUUM in the code
- [x] Add a commandline vacuum command that allows manual activation
- [ ] Add a "Reclaim space" button in the UI that calls the vacuum operation
- [x] Add a --auto-vacuum option that calls the vaccum command after an operation

### DIFF
--- a/Duplicati/CommandLine/Commands.cs
+++ b/Duplicati/CommandLine/Commands.cs
@@ -1041,6 +1041,24 @@ namespace Duplicati.CommandLine
 
             return 0;
         }
+
+        public static int Vacuum(
+            TextWriter outwriter, 
+            Action<Duplicati.Library.Main.Controller> setup,
+            List<string> args, Dictionary<string, string> options, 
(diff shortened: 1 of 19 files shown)


kenkendk:
Looks great!

The backup does not call vaccum directly, but it calls it through this method:
https://github.com/Danstahr/duplicati/blob/58a95e03e606894deea83ca3e4365e76c410e9ea/Duplicati/Library/Main/Database/LocalDatabase.cs#L1307

Danstahr:
Good catch! I did search for all vacuum references but missed this one.

Fixed.

kenkendk:
The Travis error is from SharpCompress. Could you pull in master and see if it fixes the build?

Danstahr:
Done.

kenkendk:
It is strange that the fix is required, as it just changes the folder name to camel-case, instead of lowercase. Tests have been working without that fix for months.

PR #2627, opened 2017-08-06
Fixed some of the failing tests in concurrent_processing branch

Danstahr:


Danstahr:
For the record, the remaining failing tests for concurrent_processing have to do with handling of empty files / blocks. The backup part tries to explicitly store them in volumes, but if it's the only block stored in that volume the size of its data is zero and the compaction kicks in, deleting valid entries. I don't fully understand the code just yet but I'll get the fix done soon.

PR #2637, opened 2017-08-12
Fixed empty block handling in concurrent_processing branch

Danstahr:
Makes all the tests pass!

### DIFF
--- a/Duplicati/Library/Main/Database/LocalDatabase.cs
+++ b/Duplicati/Library/Main/Database/LocalDatabase.cs
@@ -789,16 +789,21 @@ ON
                 if (itemswithnoblocklisthash != 0)
                     throw new InvalidDataException(string.Format("Found {0} file(s) with missing blocklist hashes", itemswithnoblocklisthash));
 
-                if (cmd.ExecuteScalarInt64(@"SELECT COUNT(*) FROM ""File"" WHERE ""BlocksetID"" != ? AND ""BlocksetID"" != ? AND NOT ""BlocksetID"" IN (SELECT ""BlocksetID"" FROM ""BlocksetEntry"")", 0, FOLDER_BLOCKSET_ID, SYMLINK_BLOCKSET_ID) != 0)
-                    throw new Exception("Detected file entries with not associated blocks");
+                if (cmd.ExecuteScalarInt64(@"SELECT COUNT(*) FROM ""Blockset"" WHERE ""Length"" != 0 AND ""ID"" NOT IN (SELECT ""BlocksetId"" FROM ""BlocksetEntry"")") != 0)
+                {
+                    throw new Exception("Detected non-empty blocksets with no associated blocks!");
+                }
 
+                if (cmd.ExecuteScalarInt64(@"SELECT COUNT(*) FROM ""File"" WHERE ""BlocksetID"" != ? AND ""BlocksetID"" != ? AND NOT ""BlocksetID"" IN (SELECT ""ID"" FROM ""Blockset"")", 0, FOLDER_BLOCKSET_ID, SYMLINK_BLOCKSET_ID) != 0)
+                    throw new Exception("Detected files associated with non-existing blocksets!");
 
                 if (verifyfilelists)
                 {
                     using(var cmd2 = m_connection.CreateCommand(transaction))
                     foreach(var filesetid in cmd.ExecuteReaderEnumerable(@"SELECT ""ID"" FROM ""Fileset"" ").Select(x => x.ConvertValueToInt64(0, -1)))
                     {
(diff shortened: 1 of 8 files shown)


kenkendk:
Wow!

Duplicati/Library/Main/Database/LocalDatabase.cs
```diff
@@ -955,40 +960,41 @@ System.Collections.IEnumerator System.Collections.IEnumerable.GetEnumerator()
...
-            AND ""G"".""Index"" = 0
-            AND ""I"".""BlocksetID"" = ""E"".""ID""
-            AND ""I"".""BlockID"" = ""H"".""ID""
-            AND ""I"".""Index"" = 0
+          ""A"".""BlocksetId"" >= 0 AND
+          ""D"".""FilesetID"" = ? AND
+          (""I"".""Index"" = 0 OR ""I"".""Index"" IS NULL) AND  
+          (""G"".""Index"" = 0 OR ""G"".""Index"" IS NULL)
         ) J
     LEFT OUTER JOIN
```
kenkendk: You have rewritten the crazy join in a cartesion-product-style syntax with left-joins?

How sure are you that your rewrite is correct?
I find it hard to follow the rewrite :/

Duplicati/Library/Main/Operation/RestoreHandler.cs
```diff
@@ -778,8 +784,6 @@ private static void PrepareBlockAndFileList(LocalRestoreDatabase database, Optio
                 {
                     // Find the largest common prefix
                     var largest_prefix = options.DontCompressRestorePaths ? "" : database.GetLargestPrefix();
-                    if (options.DontCompressRestorePaths)
```
kenkendk: Dooh!

Duplicati/Library/Main/Operation/Backup/StreamBlockSplitter.cs
```diff
@@ -103,13 +103,10 @@ public static Task Run(Options options, BackupDatabase database, ITaskReader tas
                             var lastread = 0;
                             var buf = new byte[blocksize];
                             var lastupdate = DateTime.Now;
-                            var firstBlock = true;
 
                             // Core processing loop, read blocks of data and hash individually
-                            while (((lastread = await stream.ForceStreamReadAsync(buf, blocksize)) != 0) || firstBlock)
+                            while (((lastread = await stream.ForceStreamReadAsync(buf, blocksize)) != 0))
```
kenkendk: The `firstblock` was used to ensure that at least one block was emitted, also for empty files.
Now that it is removed, it is possible to have files with zero blocks, and I could imagine that one or more queries fail because of this assumption.

Not sure how to find these though...

Duplicati/Library/Main/Operation/RestoreHandler.cs
Danstahr: This is not a big deal, just a tiny bit of redundant work harming readability.

kenkendk: I know... the "Dooh!" was me seeing I wrote stupid code :)

Duplicati/Library/Main/Operation/Backup/StreamBlockSplitter.cs
Danstahr: I only found one query that was failing because of this - the crazy join you commented on above. I did a bit of manual testing and ran all unit tests with positive results. 

I'm still not 100% convinced that nothing blows up with this change in corner cases, but we'll need to test this branch thoroughly before making it available to the general public anyway.

Duplicati/Library/Main/Database/LocalDatabase.cs
Danstahr: Frankly, I often find Duplicati SQL queries hard to follow :).

The rewrite is equivalent to the previous version with the following exceptions:
- Previously it used inner joins, now we're using left joins. This is needed because we might not have a block for all entries anymore. I could combine comma joins with explicit left joins but I find such queries even more cryptic.
 - Added a condition for non-negative blockset IDs. This is to filter out special blockset values (currently -100 and -200).
 - Added IS NULL conditions for possible indices. The purpose of the where clause is to always pick the first block, but what if we don't have any blocks -> allow null as well.

kenkendk: I agree that the SQL queries are hard to follow!

Only way I can think of that makes it easier is to re-organize the data in a non-normalized form, but that would have other downsides, including increased database size :(

Danstahr:
The tests on Travis are flaky. What's annoying is they're flaky **only** on Travis and I'm not able to reproduce the failures locally, which makes debugging pain in the butt. 

My first guess is a race condition or a stream not being properly closed *somewhere*.

Danstahr:
I managed to get the tests consistently passing. The fix is hacky, but does the trick for now.

Issue #2463, opened 2017-05-07
speed upload


mauro-miatello:
enhancement request: add (average) upload speed in status bar and in email notifications

kenkendk:
Yes, I have a mockup for how that would look, and the data is reported, so I just need to add it.

samuel-w:
This should be closed due to #2676.

warwickmm:
Thanks @samuel-w.

PR #2676, opened 2017-08-31
Added - Display of current upload speed

goldbattle:
image
Thought I would give #2463 a go. If the `BackendSpeed` is the default -1 it will just not have the last bit of text. I also widened the progress bar from `545px => 580px` so that the longer text fits.


gorbiWTF:
Awesome! Would it be possible to switch between MBit/s and MByte/s?

kenkendk:
Thanks! Maybe @gorbiWTF's suggestion should be a setting?

Also, maybe we should detect when the processing is stalled due to uploads, and then only display the speed? This would require an extra flag in the status reported from the server.

kenkendk:
Btw, you cannot edit the `styles.css` file directly, as we are using less.

You need to edit the less file, and then use `lessc` to build the `styles.css` file otherwise the changes are lost on next compile.

PR #2677, opened 2017-08-31
Added Sia as a backend provider

fuffenz:
Well, it's a first version.   It does pass CommandLine.BackendTester and I've successfully made some backup and restore via the web GUI.

Tested with Sia-UI  / siad 1.30 on Windows 10.   

duplicati-sia1

I will be doing some more tests and perhaps some refactoring, but I think it's good enough for a first version and would appreciate some comments. :)


Duplicati/Library/Backend/Sia/Sia.cs
```diff
@@ -0,0 +1,433 @@
...
+            return files;
+        }
+
+        public void Put(string remotename, string filename)
+        {
+            string endpoint ="";
+            string siafile = m_targetpath + "/" + remotename;
+
+            try {
+                endpoint = string.Format("/renter/upload/{0}/{1}?source={2}&datapieces=1&paritypieces=12",
```
lukechampine: why are the redundancy settings being specified here? If unspecified, the recommended settings are used.

fuffenz: Oops.  That's some leftovers, I was thinking they should be configurable options but didn't get that far.

zachherbert:
Will test over the next couple days. Could you update to "Sia Decentralized Cloud" instead of "Sia Collaborative Cloud"?

kenkendk:
The build error seems to be related to the project file:
```
/home/travis/build/duplicati/duplicati/Duplicati.sln (default targets) ->
(Build target) ->
/home/travis/build/duplicati/duplicati/Duplicati/Library/Backend/Sia/Duplicati.Library.Backend.Sia.csproj (default targets) ->
/usr/lib/mono/xbuild/14.0/bin/Microsoft.Common.targets (_CheckForInvalidConfigurationAndPlatform target) ->

	/usr/lib/mono/xbuild/14.0/bin/Microsoft.Common.targets: error : 'OutputPath' property is not set for this project. Usually this is caused by invalid Configuration/Platform combination. Original values: Configuration: Release Platform: Any CPU.

	 7 Warning(s)
	 1 Error(s)
```

alexbarnsley:
I don't know if it's related to my setup, but my upload hangs on "Verifying...". I've not changed any Duplicati config. I experienced the same issue when working on the bounty myself, but I managed to do a workaround by using Streams. 

You'll notice the `BackendFileProgress` hangs at zero, despite the file being on the Sia hosts
image
image

Does anyone else experience this? It's all based on the code here (where it deals with updating progress of the uploads, but only if using IStreamingBackend):

https://github.com/duplicati/duplicati/blob/master/Duplicati/Library/Main/BackendManager.cs#L753-L761

I'm happy to make the changes @fuffenz if you like in terms of getting the Streams to work, as I've already had to put those fixes in to get past this stage. If no one else experiences this then maybe it's just my setup? It's been hung for the last 30 minutes or so, despite the progressstate Ajax requests still going.

fuffenz:
@alexbarnsley  Did you set minimum redundancy to 2 for the backup?  My guess is that it's waiting for redundancy to reach 3.  I should probably add some kind of reasonable timeout if the upload is at 100% but havn't reached requested redundancy.

alexbarnsley:
@fuffenz I bypassed the minimum redundancy by hard-coding the minimum from 3 to 2 in the initial SiaBackend declaration, just for my testing. I hit that minimum redundancy problem first, so I reduced the minimum to 2 which go me past that issue, then I got hit with this Verifying hang. I'll try and work out later how to increase redundancy of the antfarm, so I don't have to tweak any code.

fuffenz:
If it hangs on verifying my guess is that there is a pending download from Sia.   I'll try to figure it out later tonight.

alexbarnsley:
@fuffenz yeh I remember debugging it quite a bit which gave me headaches. the Put and Get worked perfectly, but for whatever reason it wouldn't go any further in the progress bar. I then stumbled across the above block of code in `BackendManager.cs` which updates the progress, only if `IStreamingBackend` is implemented.

I implemented streams which resolved the "Verifying..." hang that I was having. I could see the download was working too as I'd placed a tonne of breakpoints and could confirm the file itself downloaded to the temp folder. Let me know how your tests get on and if you need me to try anything!

alexbarnsley:
@fuffenz There's an issue if you have no files hosted on Sia at all. `SiaFileList fl = GetFiles();` results in `fl.Files` being `null`. This throws a `Object reference not set to an instance of an object` exception here `foreach (var f in fl.Files)` on [Line 261](https://github.com/fuffenz/duplicati/blob/b82c8a9d4cae78dde62f31d35338c22590410466/Duplicati/Library/Backend/Sia/Sia.cs#L261)

Do you just want me to report issues or put fixes in for them?

alexbarnsley:
Just tested your last batch of changes @fuffenz - uploading for me now 👍 You must have been right about it waiting for a redundancy of 3

I'll continue to test, see if I can recreate my earlier issue, but I don't think I will - I wasn't able to increase my redundancy to 3 in my antfarm so I'm glad you reduced it!

fuffenz:
@alexbarnsley I think the redundancy setting should be up to the user to decide and not forced, but there could be a warning if lower than X.

alexbarnsley:
I completely agree 👍 

Could you try setting the remote path to something like `/siatest/`. Does it hang at "Waiting for upload" for you with that? It was for me, I changed the path to `siatest` and all worked beautifully.

alexbarnsley:
@fuffenz following on from my last comment, it is an issue because the ending "/" in my example is added to the SiaPath. E.g. Line 163 is doing:

```
if (f.Siapath == siafilename)
```

which equates to

```
siatest/duplicati-b83af79b8b437452e9e2ee4c0c2b32437.dblock.zip == wut//duplicati-b83af79b8b437452e9e2ee4c0c2b32437.dblock.zip
```

This returns false, so the file is never considered to be uploaded.

fuffenz:
@alexbarnsley Nice catch, adding a trim of trailing slashes.  I actually thought it wouldnt be needed since a siafile "path" isn't really a path, but was obviously not thinking that through.

alexbarnsley:
@fuffenz hopefully slashes are enough. I tried lots of non-standard characters earlier and it worked great. Most likely just forward slashes that cause the issues. I'll retest once you've put a fix in 👍

alexbarnsley:
@fuffenz I'm having issues with backslash too (\). I used the path of "\asd\/" which resulted in a path of "\asd\" after your last commit.

This means it's comparing:

"asd/duplicati-20170902T203055Z.dlist.zip" to "\\asd\\/duplicati-20170902T203055Z.dlist.zip"

I think the Sia API removes backslashes from any paths in requests.

kenkendk:
@fuffenz I do not know enough about Sia naming rules, could you comment on @alexbarnsley 's comment about encoding? Otherwise, it looks good, and I will merge once there is a comment on the encoding.

fuffenz:
@kenkendk Sia doesn't really have the concept of file paths, but there could be tools accessing Sia storage that make assumptions and try to parse a path anyway.   I added a simple regexp now that only accepts standard characters and forward slash, should be sufficient.

alexbarnsley:
I'm not able to test the last commit due to not having my machine to hand, but I'm happy that everything else appears to be working (the code for the last commit looks sound 👍)

PR #2696, opened 2017-09-07
Updated the Application's name on the balloon tooltip

CSharpFan:
Balloon tooltip on Windows 10 used to say 'message' and then as the owner 'Duplicati.GUI.TrayIcon'. I changed that to just 'Duplicati'.

And for the Server.Serialization, I just fixed the name, it was wrong.


kenkendk:
Thanks!

Issue #2036, opened 2016-10-20
Feature request: Specify source by drive label


tophee:
In order to be able to backup external drives consistently, it is necessary for duplicati to detect these based on their drive labels rather than the drive letter since the latter can change.
##

tophee:
To anyone who is facing the same problem: you can use USBDLM (USB Drive Letter Manager for Windows)  by Uwe Sieber ([http://www.uwe-sieber.de/usbdlm_e.html](http://www.uwe-sieber.de/usbdlm_e.html)) as a workaround). It allows you to reliably assign the same drive letter to drives based on their label.

kenkendk:
You can use the option --alternate-target-path="*:\folder" and --alternate-destination-marker="markerfile.txt".
Then create an empty file in the destination folder called "markerfile.txt".

Duplicati will then scan all drive letters to find the file, and use the drive where the file is found.

Does that solve your issue?

tophee:
As I understand it, this only looks at the path, not the label of the drive. It could be a workaround, though, if I I created a folder with a unique name on the external drives and put everything on the drive under that  folder. But in that case I think prefer my own workaround posted above cause I'd be artificially making my paths on that drive longer than necessary.

kenkendk:
You specify the target drive as normal, but use `--alternate-target-path=*:\folder` to give the path without the real drive.
Then you put a file (with a name you choose) in the target folder and supply that name as `--alternate-destination-marker`.
This means there is no change in your setup, apart from an extra file.

But this relates to the _destination_ and your request is for the _source_.
A similar scheme could be used for source folders though.

kenkendk:
@tophee Is the fix in #2707 sufficient to support your use case?

tophee:
I'm not sure what exactly the fix includes. @tygill says that 

> it attempts to back up all drives rather than providing a better specification for the specific drive to backup

but I'm not sure what that means in practice. Does it mean that it checks all drives for relevant paths and if it finds them it assumes it is backing up the correct drive? That would sound a bit risky to me. (e.g. what if I have two drives with a similar folder structure?)

tygill:
> Does it mean that it checks all drives for relevant paths and if it finds them it assumes it is backing up the correct drive?

That is what it does, so this behavior does have the possibility of backing up from multiple drives if they have the same file structure. (Whether or not this is a 'risk' depends on your point of view and use case.)

[...]

Issue #1482, opened 2015-10-27
Support marker file for excluding directories


mdekstrand:
Right now, there is the `--exclude-files-attributes` option to exclude files based on attributes.

It would be useful to add a feature present in some other backup systems, such as burp, to exclude directories (and their descendants) based on the presence of a marker file such as `_nobackup`. Then a user could exclude content just by going and creating marker files rather than figuring out how to write exclude lines that match those files.
##

kenkendk:
Yes, that makes sense.

ironjohn:
I would like to add a recommendation to format the marker file specifically for duplicati because something generic like _nobackup could be used by other software.  You could adopt ".dpcignore" (like .gitignore for git) to keep from stepping on the toes of dev's that did not consider the possibility.

ts678:
2.0.3.7 canary change log says "Added option to exclude folders based on the existence of a marker file".

2.0.3.9 help text shows how the issue about the marker name was resolved:

  --ignore-filenames
    Use this option to set a filename, or list of filenames, that indicate
    exclusion of a folder which contains it. A common use would be to have a
    file named something like ".nobackup" and place this file into folders
    that should not be backed up.

seerickcode:
So this is in the main documentation, but not currently implemented ? 

`Warnings: [
    The supplied option --ignore-filenames is not supported and will be ignored,
    Failed to process path: /run/snapd/ns/lxd.mnt => Win32 IO returned 25. Path: /run/snapd/ns/lxd.mnt
]
`

( 2.0.3.3_beta_2018-04-02)

ts678:
I assume it's more like "not currently released in beta" because "( 2.0.3.3_beta_2018-04-02)" is the latest.
Perhaps that will change soon, however until then, anyone who is willing to change to canary can do so.

[Settings in Duplicati](https://duplicati.readthedocs.io/en/latest/03-using-the-graphical-user-interface/#settings-in-duplicati) describes update channels, however downgrading Duplicati (e.g. back to 2.0.3.3) is sometimes complicated by the need to manually downgrade the database format. Using a separate test system for canary can avoid this, as well as keeping any unexpected canary bugs away from production.

Easier still would be to wait for a new beta, however if canary really has an issue it'd be nice to hear now.

jstasiak:
Driveby comment:

[...]

PR #2707, opened 2017-09-11
Add support for wildcards in source paths on Windows.

tygill:
This causes backup to run for the given path on all drives, and can be useful in the case where removable drives may change letters randomly.

This change is a partial resolution for the root problem of issue #2036, though it is not quite as elegant as the suggested solution since it attempts to back up all drives rather than providing a better specification for the specific drive to backup.


kenkendk:
Lets include it, and then we can perhaps implement #1482 to add the features to exclude specific folders?

PR #2713, opened 2017-09-12
Update localization_webroot-fr.po

flexbrane:
Update the french translation, "Vous êtes actuellement entrain" become "Vous êtes actuellement **en train**".


agrajaghh:
Thanks for your fix, but could you please update the string at transifex?
https://www.transifex.com/duplicati/duplicati/dashboard/

If we change the translation here in the .po file, its getting overwritten the next time we pull new translations from transifex. Thanks :)

flexbrane:
Done!

PR #2722, opened 2017-09-15
Decrease server's CPU and IO scheduling priority

z0rc:
Setting low priority helps to run server process truly in background and provides CPU cycles to Duplicati only if no other process requires them.


kenkendk:
Thanks!

PR #2723, opened 2017-09-15
Add support for specifying backup paths on Windows by volume guid

tygill:
These volume guids can be found via the 'mountvol' command, and are reported by the built in help.
This also includes some updates to the documentation and logging when using wildcard drives in source paths.

This change addresses #2036, and allows individual drives to be selected for backup, no matter which drive letter they are mounted to.


kenkendk:
Looks good! Cool that you put the no-inlining options there to prevent loading the functions on non-Windows.

Issue #2437, opened 2017-04-19
computer name as part of title


pmarat-sqrt0:
Would you reconsider to add the computer name to the title element of the web gui.
I remember it was there in the old duplicati 2 interface.
---
[![Bountysource](https://api.bountysource.com/badge/issue?issue_id=44299565)](https://www.bountysource.com/issues/44299565-computer-name-as-part-of-title?utm_source=44299565&utm_medium=shield&utm_campaign=ISSUE_BADGE)

Pectojin:
By "title element" I think you mean so that it shows up in the browser tab?

![image](https://user-images.githubusercontent.com/1561484/30515433-d8e51030-9b27-11e7-80f7-d888e8e852d8.png)

I created a pull request.

PR #2728, opened 2017-09-16
WebUI title reflects machine name

Pectojin:
Changed the webpage title to be "MachineName - AppName" allowing users to discern between multiple tabs connecting to different remote machines.

I had to move the ng-controller tag from "< body>" to "< html>" because otherwise it wouldn't recognize ng-bindings in the header.

Related to duplicati/duplicati#2437


Pectojin:
I was a bit too lazy to make branches for my small fixes, so it seems the 2nd small fix I'm made went into this pull request too...

Second commit isn't related to the title. See commit message for details.

Pectojin:
Looking back this pull request today, it is kind of a mess. I'm making some branches, cherry picking the commits, and submitting some new requests in place of this one.

Issue #2664, opened 2017-08-26
Allow empty password field if SSH key is not password protected


warwickmm:
I have:
- [x] searched open and closed issues for duplicates

----------------------------------------

### Version info

**Duplicati Version:** 2.0.2.1_beta_2017-08-01
**Operating System:** Arch Linux with 4.12.8 kernel
**Backend:**

### Bug description
When a password-less ssh key is provided, the application should not complain that the password field is empty.

### Steps to reproduce
- Add a new backup and choose the SFTP (SSH) storage type
- Go to Advanced Options, and select ssh-keyfile
- Provide the path to an ssh key _that is not password protected_

**Actual result:** When continuing, the application complains that the password field is blank.  If the user then enters anything into the password field, the application continues to work as expected.

**Expected result:** If a password-less ssh key file is provided, the password field should be allowed to be empty.

Clete2:
Was going to propose a PR but then realized that you'd also have to check if the key is password-less as described, so I'll just dump the code I wrote so far:

`EditUriBackendConfig.js` at the bottom:
```
   this.require_username = function(scope) {
        if ((scope.Username || '').trim().length == 0)
            return self.show_error_dialog(gettextCatalog.getString('You must fill in the username'));

        return true;
    };

    this.require_password = function(scope) {
        if ((scope.Password || '').trim().length == 0)
            return self.show_error_dialog(gettextCatalog.getString('You must fill in the password'));

        return true;
    };

    this.require_username_and_password = function(scope) {
        self.require_username(scope);
        self.require_password(scope);

        return true;
    };

    this.require_username_and_password_or_key = function(scope) {
        self.require_username(scope);

        if(scope.AdvancedOptions.find(advancedOption => advancedOption.startsWith('--ssh-keyfile='))) {
            return true;
        }

        self.require_password(scope);

        return true;
    };
```

Line 568 of `EditUriBuiltins.js`:
            ```EditUriBackendConfig.require_username_and_password_or_key(scope);```

At least a first shot will just not check if `--ssh-keyfile=` is present.

kenkendk:
Yes, that would require a callback to the server to have it read the file and guess if it requires a password. I think your solution is good enough, as the "Test" button will give an error like "Password required for encrypted private key".

If you make it a PR as-is, then the user can work with unencrypted private key files.

Pectojin:
I think you might be overcomplicating this slightly... Just remove the password check. 

[...]

PR #2729, opened 2017-09-17
Passwordless sftp

Pectojin:
Fix for duplicati/duplicati#2664

Removing line 107 and 108 allows both passwordless SSH login and SSH login using an SSH private key without a password

### DIFF
--- a/Duplicati/Server/webroot/ngax/scripts/services/EditUriBackendConfig.js
+++ b/Duplicati/Server/webroot/ngax/scripts/services/EditUriBackendConfig.js
@@ -110,4 +110,11 @@ backupApp.service('EditUriBackendConfig', function(AppService, AppUtils, SystemI
         return true;
     };
 
+    this.require_username = function(scope) {
+        if ((scope.Username || '').trim().length == 0)
+            return self.show_error_dialog(gettextCatalog.getString('You must fill in the username'));
+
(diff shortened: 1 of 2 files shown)


kenkendk:
This does fix the issue, but it also disables the password check for all other backends.
The right place to fix this is here:
https://github.com/duplicati/duplicati/blob/master/Duplicati/Server/webroot/ngax/scripts/services/EditUriBuiltins.js#L598

That is the validation routine used for SSH.

You can use something like what is proposed in issue #2664:
```
var haskeyfile = scope.AdvancedOptions.find(advancedOption => advancedOption.startsWith('--ssh-keyfile='));
var res = true;

if (haskeyfile) {
    if ((scope.Username || '').trim().length == 0)
       res = self.show_error_dialog(gettextCatalog.getString('You must fill in the username'));  
} else {
   res = EditUriBackendConfig.require_username_and_password(scope);
}

return res;
```

Pectojin:
I'll take a look at implementing something like that tonight or tomorrow

Pectojin:
I reversed my original commit and provided a new fix at the location you suggested. No "Check username only" method existed so I had to create that helper class too.

Thanks for pointing out the right place to make the change. It's a pretty big code base so I'm still trying to figure it out.

Issue #2617, opened 2017-07-31
duplicati-server truncates log file on restart


jerrac:
I have:
- [x] searched open and closed issues for duplicates

----------------------------------------

### Version info

**Duplicati Version:** Duplicati - 2.0.1.73_experimental_2017-07-15
**Operating System:** Ubuntu 14.04
**Backend:** N/A

### Bug description
I have Duplicati set to output logs to /var/log/duplcati.log via the --log-file flag. Every time the duplicati restarts, the log file is overwritten with a new one.

### Steps to reproduce
- Start Duplicati with the --log-file flag.
- Look at the logs in the file.
- Restart Duplicati with the --log-file flag.

**Actual result:** The logs generated before the restart will be gone.
**Expected result:** The logs generated by the restart should be appended to the end of the file.

```
root@ubuntuhost:/var/log# cat duplicati.log 
2017-07-31 10:20:19Z - Information: Server has started and is listening on 127.0.0.1, port 8200
2017-07-31 10:20:19Z - Information: Server has started and is listening on 127.0.0.1, port 8200
root@ubuntuhost:/var/log# service duplicati status
duplicati start/running, process 944
root@ubuntuhost:/var/log# service duplicati restart
duplicati stop/waiting
duplicati start/running, process 19526
root@ubuntuhost:/var/log# cat duplicati.log 
2017-07-31 11:55:10Z - Information: Server has started and is listening on 127.0.0.1, port 8200
2017-07-31 11:55:10Z - Information: Server has started and is listening on 127.0.0.1, port 8200
```

FYI, my /etc/default/duplicati file:
```
# Defaults for duplicati initscript
# sourced by /etc/init.d/duplicati
# installed at /etc/default/duplicati by the maintainer scripts

#
# This is a POSIX shell fragment
#
DAEMON_OPTS=" \
    --log-file=/var/log/duplicati.log \
    --log-level=information \
    --webservice-port=8200 \
;"
```

Pectojin:
It looks like it might be "intentional"

in `duplicati/server/program.cs` at line 220 and 221 it says:
`if (System.IO.File.Exists(commandlineOptions["log-file"]))
 System.IO.File.Delete(commandlineOptions["log-file"]);`

I don't really have a Linux system to test it on, but the log file persists on my windows machine when I remove those lines, so i guess I'll make a pull request and we can take it from there.

kenkendk:
Fixed with ab9d518

PR #2730, opened 2017-09-17
Fix for duplicati/duplicati#2617

Pectojin:
I moved the "delete log file on service start" functionality into DEBUG mode so that release versions of duplicati won't wipe out it's logs on service restart.

I wanted to retain the functionality for testing (no point having gigantic debug logs when debugging).

### DIFF
--- a/Duplicati/Server/Program.cs
+++ b/Duplicati/Server/Program.cs
@@ -217,8 +217,11 @@ namespace Duplicati.Server
 
                 if (commandlineOptions.ContainsKey("log-file"))
                 {
+#if DEBUG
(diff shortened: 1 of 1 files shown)


kenkendk:
Thanks!

I was thinking that the file would just continue to grow, but I guess that is what you would want when you add that option?

Pectojin:
That's my thought. Debugging always has it enabled, so you definitely don't want that log to remain, but for production use it's a different matter.

It's impossible to know how large an administrator is okay with the file growing when they decided to add it. And if you manually specify a log file then I'd assume you're "confident" enough to implement a cleanup solution yourself.

Alternatively we'd need to add a setting in the settings page and do a file size check on startup (or periodically)

PR #2732, opened 2017-09-17
WebUI title reflects machine name

Pectojin:
Changed the webpage title to be "MachineName - AppName" allowing users to discern between multiple tabs connecting to different remote machines.

I had to move the ng-controller tag from "< body>" to "< html>" because otherwise it wouldn't recognize ng-bindings in the header.

Related to duplicati/duplicati#2437


kenkendk:
Thanks!

Issue #2715, opened 2017-09-12
Cannot select no startup delay


JoeB99:
2.0.2.4_canary_2017-09-09

FreeNAS 11-U2 jail

In the web GUI, once a startup delay has been set, it cannot be set back to "no delay".

I think a new tick box should be added that, when ticked shows the startup delay settings, and when unticked, the startup delay settings are greyed out.

kenkendk:
As a workaround for now, you can set `0s` (0 seconds).

PR #2733, opened 2017-09-17
Allow user to disable startup delay

Pectojin:
This is a two fold "fix" to duplicati/duplicati#2715

Changes in settings.html adds the option to select "0 seconds", which wasn't previously possible.
Changes in SystemSettingsController.js adds logic in the controller that will check if the duration value and duration multipliers are set. If it finds that they are not it will set them to "0" and "s" respectively (instead of leaving them blank)


kenkendk:
Thanks!
This fixes #2715

Issue #1443, opened 2015-08-25
Exclude files Duplicati is writing to from backup


kakoskin1:
Currently Duplicati.Commadline.exe tries to backup also its own logfile issuing a warning.

mono Duplicati/Duplicati.CommandLine.exe backup ~/test/ ~/temp/ --log-file="duplicati.log"
Backup started at 25.8.2015 22:27:10

Enter encryption passphrase: 

Confirm encryption passphrase: 
Checking remote backup ...
  Listing remote folder ...
Scanning local files ...
  236 files need to be examined (32,98 MB)
Failed to process path: /home/huginn/temp/duplicati.log => Sharing violation on path /home/huginn/temp/duplicati.log

Version: 2.0.0.87- 2.0.0.87_preview_2015-07-03
OS: Fedora 22

fredericmohr:
I know this bug is way old, but I'm experiencing the same issue since the last update. Currently running 2.0.1.53_experimental_2017-03-13.

```
]
Warnings: [
    Failed to process path: /home/fmohr/.config/Duplicati/duplicati.log => Sharing violation on path /home/fmohr/.config/Duplicati/duplicati.log,
    Failed to process path: /home/fmohr/.config/Duplicati/duplicati.sqlite => Unexpected result count: 0, expected 1, check log for more messages,
    Failed to process path: /home/fmohr/.config/Duplicati/duplicati.sqlite-journal => Unexpected result count: 0, expected 1, check log for more messages
]
Errors: [
    Checking errors, related to #1400. Unexpected result count: 0, expected 1, hash: 0zkrJWZt+ddSegVstbi4Mqxj145xCSlPrF6Hp8axTec=, size: 29696, blocksetid: 907452, ix: 14127, fullhash: 6y0XtGaQ8p7R3lDH/7G3sglqqEHwTQ0wXll29Oww4Wk=, fullsize: 1446634496,
    Found block with ID 2072207 and hash 0zkrJWZt+ddSegVstbi4Mqxj145xCSlPrF6Hp8axTec= and size 26624,
    Checking errors, related to #1400. Unexpected result count: 0, expected 1, hash: mhzTX2j2/WicQXn7EPUpmkr4XS06XrxvDCcaeB6tC/I=, size: 102400, blocksetid: 907454, ix: 28, fullhash: g7bcsKHo5V8Cd9Y39t7MDjlfu8JM5Nzg0S8l0mW2Udw=, fullsize: 3154864,
[...]

PR #2740, opened 2017-09-19
Add --default-filters parameter

tygill:
This change adds the --default-filters parameter, which adds built-in blacklists for common files / folders that don't need to be backed up. This includes things like temp files/folders, caches for common programs, and operating system files.

These filters are grouped into three sets, one each for Windows, Linux, and OSX. (When --default-filters is used without a value, it will automatically select the filters for the current operating system, or the sets to use can be added explicitly.)

The contents for these filters come from this question, though some have been adapted or removed:
https://superuser.com/questions/443890/what-files-file-types-and-folders-to-exclude-from-user-folder-backup

This change adds help / documentation information, and adds these filters to the web UI.


Duplicati/CommandLine/Help.cs
```diff
@@ -84,6 +84,8 @@ public static void PrintUsage(TextWriter outwriter, string topic, IDictionary<st
                     topic = "exclude";
                 else if (System.Environment.CommandLine.IndexOf("--include", StringComparison.InvariantCultureIgnoreCase) >= 0)
                     topic = "include";
+                else if (System.Environment.CommandLine.IndexOf("--default-filters", StringComparison.InvariantCultureIgnoreCase) >= 0)
```
kenkendk: I don't think this is required as the help routine will look at options if no match is found in the help file.

kenkendk:
Awesome work!

I will merge "as-is" because this is already a huge help.
I think we should add Duplicati's own folders also, as mentioned in #1443.

MM-John:
Looks like there's a bug in the list in [superuser](https://superuser.com/questions/443890/what-files-file-types-and-folders-to-exclude-from-user-folder-backup), repeated here. Under All, this entry `.*/(cookies|permissions).sqllite(-.{3})?` has two `l`'s (lower case `L`'s) in `.sqlite`

MM-John:
Yeah. I was looking in commit:
https://github.com/duplicati/duplicati/pull/2740/commits/d4865eb4cdb564399f8234944a5e33480cd4653e
`/Duplicati/Library/Utility/DefaultFilters.cs`
aka:
https://github.com/tygill/duplicati/blob/d4865eb4cdb564399f8234944a5e33480cd4653e/Duplicati/Library/Utility/DefaultFilters.cs

But it looks like `DefaultFilters.cs` isn't even there anymore.

Apparently I just clicked on:
image
above and didn't really know what I was doing :P

PR #2741, opened 2017-09-19
Replace all instances of InvariantCultureIgnoreCase with OrdinalIgnoreCase

tygill:
Replace all instances of InvariantCultureIgnoreCase with OrdinalIgnoreCase in string comparisons

InvariantCulture is useful when comparing / sorting human language strings in a culturely correct way. It handles things like accented letters in a way that makes sense to humans (e.g., 'a' should be sorted next to 'á', rather than after 'z').
Ordinal looks just at the raw code points of the characters. As such, it is recommended for use in cases when comparing system strings (file paths, command line parameters, config settings, etc.). Since it doesn't need to use the culture specific sorting rules, this method can often be faster.

For more information, see https://stackoverflow.com/questions/492799/difference-between-invariantculture-and-ordinal-string-comparison (and other related questions)


kenkendk:
Thanks! That has been on my low-priority TODO for ages :)

PR #2752, opened 2017-09-22
Add support for the quota reporting and renaming interfaces to the OneDrive backend

tygill:
Add support for the quota reporting and renaming interfaces to the OneDrive backend.

To test the quota support, quota information is reported on the command line if it is available.
However, the rename support hasn't been tested, since I don't see anything using that functionality yet.


kenkendk:
The quota stuff should be displayed in the UI automatically, but there is not yet any logic that uses the quota to make decisions.

The renaming is not used anywhere, but I did make a proof-of-concept sync tool that uses the Duplicati backends, and that works better if the backends support renaming, which is why the interface is there.

PR #2754, opened 2017-09-23
Reference property instead of invoking extension method.

warwickmm:
The `Enumerable.Count` method will attempt to cast the object to an `ICollection`, and if successful, reference the `Count` property.  Since we know that we have a `List`, we can reference the `Count` property directly.


warwickmm:
Let me know if you would prefer not to receive these trivial pull requests.  I would like to be able to contribute if I can, but I'm still familiarizing myself with the codebase.

kenkendk:
Yes, all fixes are appreciated 👍 

I don't think this has much impact, given that it is only invoked once and just calls the property directly.

Issue #2619, opened 2017-08-02
Deal with the tilde character


kenkendk:
The tilde character is used to denote "home folder" on Linux, and is used for the same purpose on Windows by Duplicati.

Unfortunately, `~` is also a valid character in paths, causing some issues.

We should review all the places that the tilde is being expanded and see if we can make some sane logic rule that determines when to expand and when not to expand.
This is a bit tricky as people might have backups set up that depends on the expansion, which we will now remove.

DragonFlux:
I ran into this same issue on CherryTree backup files on Windows that end in a tilde. I think a sensible strategy would be:
1 - If a tilde is at the very start of a path AND is followed by a slash, expand it (home folder)
2 - Tildes elsewhere are kept as is

Of course it would be even better if the code can determine which OS it is running on and only do the expansion in Linux.

Are there not functions/libraries to handle expansion in an OS sensitive way? You'll need other expansions as well like for example %temp% on Windows. There has to be an elegant way of handling this...

warwickmm:
I may be missing some cases, but as of revision 6c921ab3ec878a99262150b80b7f29713700c11e, I can only find the following instances where `~` is explicitly used to denote the home directory in the source code:
* `BasicSetupHelper.cs:32`
* `GeneralBlackBoxTesting.cs:32`
* `SVNCheckoutsTest.cs:105`
* `Utility.cs:986`
* `Utility.cs:1016`

The `Utility` class has a private field `HOME_PATH` that represents the path to the user's home directory.  If we were to use this instead of `~` in the code, then would we be able to avoid performing the expansion on Windows?

kenkendk:
It sounds correct to me. I think only the `BasicSetupHelper` is using the `~`. If we can change that to use `%HOME%` then we should be able to remove the `~` expansion?

Not sure if `~` is saved in the configuration though, but we could make an upgrade script that fixes this.

With tilde expansion away, we would need to re-enable it on Linux? Or is it just too annoying to deal with inside Duplicati so we disable it completely?

DragonFlux:
How about you make the expansion an option on the job? That way Linux users can enable it as needed and the rest can have filenames containing tildes.

But even on Linux, this should surely only be expanded if it is used as a directory name (ie: does not apply to file names and does not apply if it is not the whole directory name).

kenkendk:
I agree on the "only as start of path" idea, but I don't see many issues with pure paths.
The problem is more in filters, that could look like:
```
[...]

PR #2756, opened 2017-09-24
Add method to get fully-qualified path

warwickmm:
This can be used to normalize path strings by removing redundant path separators and resolving relative path specifications to absolute ones.  This will also replace the `~` character with the path to the user's home directory, if applicable.
    
These changes fix an issue on some Linux systems, where several unit tests were failing because `BasicSetupHelper.DATAFOLDER` resolved to something like `/home/username//testdata/backup-data`.  Since paths returned by the filesystem do not contain the redundant `/`, the "Border" unit tests would fail when attempting to restore specified paths (the `LocalRestoreDatabase.PrepareRestoreFilelist` method would not find any paths matching the `//`).

This also extracts the tilde expansion to a method, and the expansion is only performed for Unix-like systems to avoid issues with paths like `c:\Progra~1\` on Windows.

### DIFF
--- a/Duplicati/Library/Utility/Utility.cs
+++ b/Duplicati/Library/Utility/Utility.cs
@@ -984,7 +984,7 @@ namespace Duplicati.Library.Utility
         /// <param name="str">The string to expand.</param>
         public static string ExpandEnvironmentVariables(string str)
         {
-            return Environment.ExpandEnvironmentVariables(str.Replace("~", HOME_PATH));
+            return Environment.ExpandEnvironmentVariables(ExpandTilde(str));
         }
 
         /// <summary>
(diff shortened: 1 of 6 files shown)


Duplicati/Library/Utility/Utility.cs
```diff
@@ -1013,11 +1014,39 @@ public static string ExpandEnvironmentVariablesRegexp(string str, Func<string, s
...
         /// <summary>
+        /// On Unix-like systems, expand the '~' character to the user's home directory.
+        /// </summary>
+        /// <param name="path">The path to expand.</param>
+        /// <returns>An expanded version of <paramref name="path"/>, with '~' replaced by the user's home directory.</returns>
+        public static string ExpandTilde(string path)
+        {
+            // We only replace '~' on Unix-like systems to avoid potential issues
+            // with paths like c:\Progra~1\ on Windows.
+            return IsClientLinux ? path.Replace("~", HOME_PATH) : path;
```
kenkendk: I think this will break some Windows things, where `~` is expanded to the user folder?

warwickmm: Sorry, I'm not as familiar with Windows.  In what contexts does `~` represent the user folder on Windows?

warwickmm: In fact, my implementation here isn't quite correct on Linux either.  In most shells, a single `~` or a path beginning with `~/` is expanded to the current user's home directory.  A tilde followed by a username (e.g., `~user`) is expanded to the specified user's home directory.  Moreover, `~` is a valid character for file and directory names on many filesystems, so `~` is a valid filename and `~user` is a valid directory name.

As such, I think we can reasonably handle the following cases:
1. If `~` is not the first character of the string, do not perform any expansion.
2. If the string begins with `~/`, expand it to the user's home directory.  I'm not sure if this is typical on Windows, but we can certainly adopt this as a convention. 

Any other case seems ambiguous to me.  It's difficult to tell if `~user` refers to `user`'s home directory, or a file or directory named `~user`.  For performance reasons, I don't think we would want to do anything too fancy to resolve these.

What do you think is the best way to proceed?

warwickmm: Good point.  It seems that tilde expansion is a bit more complicated and deserves a separate pull request.  Since I was able to resolve my problem with the redundant path separators (fixed my `HOME` environment variable), I'll create a new issue for the tilde expansion and close this pull request.

warwickmm:
See #2619.

PR #2757, opened 2017-09-24
Add IQuotaInfo interface

tygill:
It also makes quota enabled backends return an instance of this interface instead of the quota components separately.
This way, getting quota stats requires only a single remote request on backends like Google Drive and OneDrive.


kenkendk:
Yes, it makes sense to return it as a single value.

PR #2761, opened 2017-09-26
Use simpler data structures when detecting duplicate filenames

warwickmm:
A `Dictionary` is unnecessary since we don't have any key-value pairs.  Since we just need to keep track of a unique collection of filenames, a `HashSet` will suffice.


kenkendk:
Thanks again!

PR #2762, opened 2017-09-26
Improve construction of ArgumentExceptions

warwickmm:
This makes some trivial modifications to several usages of `ArgumentException`:
1. The argument order was fixed in one case in `BlockingQueue.cs`.
2. The `nameof` operator is used instead of a string literal to provide the second argument (parameter name) to the constructor.  This makes it easier to rename the parameter in the future while keeping the exception messages consistent.
3. There were a few cases where only the parameter name was provided to the constructor.  We added some simple messages to clarify the exception.


kenkendk:
Thanks!

PR #2773, opened 2017-09-27
Convert if statements to switch statements for enums

warwickmm:
When the comparison is performed on an `enum`, using a `switch` statement can allow for additional compiler optimizations.


warwickmm:
Other than declaring the `metadataid` variable outside the `switch` statement in revision 
c082b18, the changes are the result of automatic refactorings performed by the IDE.

Also, I'm not sure if these methods will benefit much from the performance improvement.  If it's not worth the noise, feel free to decline the changes.

kenkendk:
I opted for the if/else approach because it is more compact, but the refactoring is fine as well.

PR #2780, opened 2017-09-30
Add method to get unique items and duplicates from collection

warwickmm:
This adds a method to get the unique elements as well as the duplicates from a collection.  The method does so by enumerating the collection just once, and should be more efficient than implementations using LINQ.

This also adds a few unit tests to verify the behavior, and replaces several cases that can benefit from the new implementation.

### DIFF
--- a/Duplicati/Library/Main/Controller.cs
+++ b/Duplicati/Library/Main/Controller.cs
@@ -288,13 +288,11 @@ namespace Duplicati.Library.Main
                 }
 
                 //Sanity check for duplicate files/folders
-                var pathDuplicates = sources.GroupBy(x => x, Library.Utility.Utility.ClientFilenameStringComparer)
-                      .Where(g => g.Count() > 1).Select(y => y.Key).ToList();
+                ISet<string> pathDuplicates;
+                sources = Library.Utility.Utility.GetUniqueItems(sources, Library.Utility.Utility.ClientFilenameStringComparer, out pathDuplicates).OrderBy(a => a).ToList();
 
                 foreach (var pathDuplicate in pathDuplicates)
                     result.AddVerboseMessage(string.Format("Removing duplicate source: {0}", pathDuplicate));
-
-                sources = sources.Distinct(Library.Utility.Utility.ClientFilenameStringComparer).OrderBy(a => a).ToList();
 
                 //Sanity check for multiple inclusions of the same folder
                 for (int i = 0; i < sources.Count; i++)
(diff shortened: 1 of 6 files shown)


Duplicati/Library/Main/Controller.cs
```diff
@@ -288,14 +288,12 @@ public Duplicati.Library.Interface.IBackupResults Backup(string[] inputsources,
                 }
 
                 //Sanity check for duplicate files/folders
-                var pathDuplicates = sources.GroupBy(x => x, Library.Utility.Utility.ClientFilenameStringComparer)
-                      .Where(g => g.Count() > 1).Select(y => y.Key).ToList();
+                ISet<string> pathDuplicates;
+                sources = Library.Utility.Utility.GetUniqueItems(sources, Library.Utility.Utility.ClientFilenameStringComparer, out pathDuplicates).OrderBy(a => a).ToList();
```
warwickmm: Is the `OrderBy` necessary?  I included it because the original implementation sorted the results.

Duplicati/Library/Snapshots/WindowsSnapshot.cs
```diff
@@ -108,15 +108,12 @@ public WindowsSnapshot(string[] sourcepaths, Dictionary<string, string> options)
                 }
 
                 //Sanity check for duplicate files/folders
-                var pathDuplicates = m_sourcepaths.GroupBy(x => x, Utility.Utility.ClientFilenameStringComparer)
-                      .Where(g => g.Count() > 1).Select(y => y.Key).ToList();
+                ISet<string> pathDuplicates;
+                m_sourcepaths = Utility.Utility.GetUniqueItems(m_sourcepaths, Utility.Utility.ClientFilenameStringComparer, out pathDuplicates).ToList();
```
warwickmm: The original implementation sorted the results, but only if the collection contained duplicates.  I interpreted this to mean that the result does not really need to be sorted, so I removed the `OrderBy` clause.

Duplicati/Library/Main/Controller.cs
kenkendk: I think it is used later to find out if an included folder is a sub-folder of a top-level folder.

Duplicati/Library/Snapshots/WindowsSnapshot.cs
kenkendk: I think that is correct, there is no need for sorting. The paths are only used to figure out which disks to make VSS snapshots from.

Issue #2084, opened 2016-11-03
Retention Policy like Time Machine [$50]


gorbiWTF:
I have a suggestion but I don't know how complicated this would be to implement: A retention policy like Time Machine.

From Wikipedia: 
> Time Machine saves hourly backups for the past 24 hours, daily backups for the past month, and weekly backups for everything older than a month until the volume runs out of space. At that point, Time Machine deletes the oldest weekly backup.

Configurable, of course ;)

jrast:
I was thinking about this as well. This would be a very nice feature and requires a lot less storage than keeping all backups for a given duration.

This strategy is also known as "Grandfather-father-son" strategy: https://en.wikipedia.org/wiki/Backup_rotation_scheme#Grandfather-father-son

Here you can find a implementation in shell script for a database backup strategy: http://processmakerblog.com/backup/the-father-the-son-and-the-grandfather-a-basic-backup-strategy/

gorbiWTF:
The "Tower of Hanoi" method sounds interesting too. Maybe with n targets?

joolsr:
Yes I would find this very useful. I have daily backups going back nearly 9 months which I dont need and daily backups from a few days ago that I do ! Kenneth suggests that you can manually delete versions from the commandline using list to find out what versions you have

extink:
@gorbiWTF I guess the "Tower of Hanoi" would be harder te implement with incremental backups, as it will limit you greatly with the number of backups you can have. Where the "Grandfather-father-son" method has more opportunities to implement incremental backups.

gorbiWTF:
@extink I thought so too. Too complicated to implement and to understand for most users (including myself).

kenkendk:
One of the ideas I discussed with @renestach was to use a "spill over" strategy, where you can say "I want 4 daily backups, 2 weekly, 3 monthly", and then whenever you have 4 daily backups, the oldest "spills" to become a weekly and so on.

Beautifully illustrated with a mechanical clock:
https://www.youtube.com/watch?v=UHBHCsrqYMw

joolsr:
Yes exactly what I am looking for . 😀

Julian

agrajaghh:
I really like the setting of Crashplan, I guess this is the spill over strategy?

![grafik](https://cloud.githubusercontent.com/assets/5296073/22303940/c2b1736c-e335-11e6-9a75-d2e8aafc3d11.png)

1activegeek:
[...]

PR #2781, opened 2017-09-30
Retention Policy

TekkiWuff:
As discussed in #2084:
This adds the new option "--retention-policy" which deletes old intermediate backups as specified in the option value by the user. 
For example the value "7D:0s,3M:1D,10Y:2M" means:
- For 7 days keep all backups ("0 seconds" here basicly means the time between two backups can be infinitely small)
- For 3 months keep one backup every day (=24 hours)
- For 10 years keep one backup every two months (=60days)
- The format for the time range values is the same as in "--keep-time"

Additional notes:
- Backups outside the specified ranges are untouched and can for example be removed with the already existing "--keep-time" option
- Old backups are not modified. This means it's **not ensured** that the last version of a deleted file is always kept, as it was briefly discussed here https://github.com/duplicati/duplicati/issues/2084#issuecomment-330050927
- there is currently no nice GUI for configuring this option
  - the advanced option has to be specified as a (rather complex) string
- there is no way to specify a different retention policy algorithm as suggested here https://github.com/duplicati/duplicati/issues/2084#issuecomment-328310297

### DIFF
--- a/Duplicati/Library/Main/Controller.cs
+++ b/Duplicati/Library/Main/Controller.cs
@@ -865,10 +865,10 @@ namespace Duplicati.Library.Main
             if (!string.IsNullOrWhiteSpace(m_options.Prefix) && m_options.Prefix.Contains("-"))
                 throw new Interface.UserInformationException("The prefix cannot contain hyphens (-)");
 
-            //Check validity of keep-staggered-versions option value
+            //Check validity of retention-policy option value
             try
             {
-                foreach (var configEntry in m_options.KeepStaggeredVersion)
+                foreach (var configEntry in m_options.RetentionPolicy)
                 {
                     if (configEntry.Value >= configEntry.Key)
                     {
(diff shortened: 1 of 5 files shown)


Duplicati/Library/Main/Controller.cs
```diff
@@ -865,6 +865,22 @@ private void ValidateOptions(ILogWriter log)
             if (!string.IsNullOrWhiteSpace(m_options.Prefix) && m_options.Prefix.Contains("-"))
                 throw new Interface.UserInformationException("The prefix cannot contain hyphens (-)");
 
+            //Check validity of retention-policy option value
```
kenkendk: I think we should also check that the user does not have multiple options set here.
The `--keep-versions` option should not be allowed with `--retention-policy` as it is a bit unclear to the user what happens.

EntropiaFox:
Not sure if this is the right place to ask how this works, but how do I make sure this retention policy is being used instead of any other scheme for a given backup? Would it override the default setting of keeping an indefinite amount of backups?

Issue #2666, opened 2017-08-27
Disable on battery


cjustin88:
Hi,

Would it be possible to add an option `disable-on-battery` or `allow-ac-only` that will get the power source from `acpid` or `systemd` and dis-allow the backup from proceeding if the computer is not plugged in? This could be useful for preserving battery life on a laptop. On my Mac, I use Arq, which has this feature and on my Windows machine, I used to use Crashplan, which also has this feature.

It would be best to implement this in a distribution-agnostic way (I think `systemd` has replaced most functions of `acpid` on Fedora, for instance).

Thanks,
Justin

kenkendk:
Yes, I think it would be possible, but it would have to be implemented differently on all platforms.

ccoVeille:
Duplicate for #2464 (or at least related)

warwickmm:
I think Linux exposes enough information through sysfs so that we can detect the power source without assuming the existence of certain user space programs or init systems.  I have some ideas that I might try to mock up and see what people think.

hydrian:
Systemd can definitely stop the service if running the headless daemon. We just need to make sure that duplication stops gracefully from the appropriate SIG* signal. 

Not sure if something more uniform and cross-platform should be thought about too. This is very Linux centric. Not sure how this would be handled in other OSes.

warwickmm:
This was implemented in pull request #3086.

PR #2785, opened 2017-10-02
Add preliminary support for detecting power supply source

warwickmm:
This adds preliminary support for detecting the current power supply source.  This can be useful if we want to later add the ability to disable the backup (or other tasks) when on battery power (see issue #2666).

Currently, the detection is only implemented for Windows (still needs testing) and Linux kernels 2.6.23 and newer.  On Linux, by accessing the information directly via `sysfs`, we should be able to avoid assumptions on the existence of particular user space functions.  If for some reason the required entries are not available (e.g., on custom-built kernels), then the implementation returns `Unknown` as the power supply source.

For Mac OS, the implementation currently returns `Unknown` as the power supply source.

### DIFF
--- a/Duplicati/Library/Utility/Duplicati.Library.Utility.csproj
+++ b/Duplicati/Library/Utility/Duplicati.Library.Utility.csproj
@@ -76,6 +76,7 @@
     <Compile Include="HostKeyException.cs" />
     <Compile Include="CallContextSettings.cs" />
     <Compile Include="HashAlgorithmHelper.cs" />
+    <Compile Include="Power\PowerSupply.cs" />
   </ItemGroup>
   <ItemGroup>
     <None Include="app.config" />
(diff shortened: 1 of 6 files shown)


warwickmm:
I don't have a great vision for how this will be used in the future, but for now there is a single entry point, `PowerSupply.GetSource()`, which returns a value from an `enum` representing the power supply source (`AC`, `Battery`, or `Unknown`).  Let me know if you think this is a reasonable approach or not.

warwickmm:
I added the ability to detect the power supply source on Windows in revision 8573ab2.  Unfortunately, I don't have a Windows machine to test this on.  It would be nice if someone could verify that this works correctly on Windows.

kenkendk:
I think this is a good start to adding the support for various actions based on power supply. I can probably add the functionality for MacOS, and test on Windows.

Issue #2555, opened 2017-06-22
backup does not work with files with a "~" symbol in the file name.


sunwayru:
Duplicati.Library.Interface.UserInformationException: Invalid path: «D:\distr\Quik\C:\Users\DmitryC:\Users\Dmitry.txt» System.NotSupportedException: This path format is not supported..
----------------------------------------
### Version info
2.0.1.61_experimental_2017-05-30

kenkendk:
This happens because Duplicati expands `~` into the "user home" as is custom on Linux-based operating systems.
Where do you enter the path?

sunwayru:
I use Duplicati 2.0.1.61_experimental_2017-05-30 for Windows 10.
File names like 'd:/.../~.txt' valid in OS, but Duplicati can't work with them :(

kenkendk:
Duplicati should not expand the paths inside the backup, it only expands paths from the user.

What backup source paths do you enter?
Can you provide a screenshot of this page (Source Data) on your system:
https://www.duplicati.com/articles/Getting-Started/#lg=1&slide=0

sunwayru:
Source Data:
   D:\distr\Quik\~~.txt
   D:\distr\Microsoft Office\~~.txt
screenshot: http://prntscr.com/fr06bv

throws error: http://prntscr.com/fr07gx

warwickmm:
Related to #2619.

TekkiWuff:
Could this get closed, cause tilde expansion isn't supported anymore since PR #2836?

warwickmm:
Yes.  @kenkendk, can you close this?

kenkendk:
Closed :)

Issue #2325, opened 2017-02-19
Special Characters in source path cause backup failure


rjbarber:
I have:
- [x] searched open and closed issues for duplicates
Not, I cant figure out how to do a search in the issue list. Sorry. 

----------------------------------------

### Version info

**Duplicati Version:** 2.0.1.45_experimental_2017-02-03
**Operating System:** WIn7 Ultimate 64 Updated
**Backend:** ACD

### Bug description
Describe here the issue that you are experiencing.
I use special characters in my file names: ~Virtual Filing Cabinet. Duplicati errors out on this. 
"Invalid path: "D:\Users\Roland\Documents\C:\Users\Roland.RJBATHOMEPaperPort Inbox\" (The given path's format is not supported.)"
It appears Duplicati is droping the ~ character and the preceding \ character.
### Steps to reproduce
- using hyphens as bullet points
- list the steps
- that reproduce the bug
-Create backup with a folder name that has the ~ character.
-Run the backup.

**Actual result:** Describe here what happens after you run the steps above (i.e. the buggy behaviour)
Backup fails with error
**Expected result:** Describe here what should happen after you run the steps above (i.e. what would be the correct behaviour)
Backup should have completed with no errors.
![duplicati error screen shot](https://cloud.githubusercontent.com/assets/10442659/23107456/90edba1c-f6b2-11e6-9212-062015e5b350.jpg)

### Screenshots

Attached.
### debug log

Feb 19, 2017 2:46 PM: Failed while executing "Backup" with id: 1
Feb 19, 2017 2:42 PM: Failed while executing "Backup" with id: 1
Feb 19, 2017 2:41 PM: Failed while executing "Backup" with id: 1
Feb 19, 2017 2:41 PM: Failed while executing "Backup" with id: 1
Feb 19, 2017 2:39 PM: Request for http://127.0.0.1:8200/api/v1/backup/1/log?pagesize=100 gave error
Feb 19, 2017 2:39 PM: Reporting error gave error
Feb 19, 2017 2:39 PM: Failed while executing "Backup" with id: 1

kenkendk:
This happens because Duplicati expands `~` to the users home path.
Not sure how we should fix it.

One idea could be to check if the file/folder with the tilde actually exists before expanding.

TekkiWuff:
Could this get closed, cause tilde expansion isn't supported anymore since PR #2836?

rjbarber:
[...]

PR #2836, opened 2017-10-21
Avoid performing tilde expansion

warwickmm:
The expansion being performed was often incorrect, as it's possible for the tilde character to appear as part of the path where it should not be expanded to the user's home directory.

Properly supporting tilde expansion is complicated and requires additional study.  Instead of supporting it only partially, we have decided to simplify things and just drop tilde expansion for now.  See the discussion in issue #2619, as well as issues #2325 and #2555.


Duplicati/Library/Utility/Utility.cs
```diff
@@ -974,16 +974,16 @@ public static string LocateFileInSystemPath(string filename)
...
-        private static readonly string HOME_PATH = Environment.GetFolderPath(IsClientLinux ? Environment.SpecialFolder.Personal : Environment.SpecialFolder.UserProfile);
+        public static readonly string HOME_PATH = Environment.GetFolderPath(IsClientLinux ? Environment.SpecialFolder.Personal : Environment.SpecialFolder.UserProfile);
 
         /// <summary>
-        /// Expands environment variables, including the tilde character
+        /// Expands environment variables.
         /// </summary>
         /// <returns>The expanded string.</returns>
         /// <param name="str">The string to expand.</param>
         public static string ExpandEnvironmentVariables(string str)
```
warwickmm: This method is no longer necessary, as we can just call the built-in method.  I decided to leave this refactoring for a separate pull request so that we don't pollute the diff in this pull request.

warwickmm:
@kenkendk, you [previously mentioned](https://github.com/duplicati/duplicati/issues/2619#issuecomment-333094921) that `~` characters may be saved in the configuration and that an upgrade script may be needed.  Is this still a concern?

Duplicati/Library/Utility/Utility.cs
warwickmm: Actually, it seems that `ExpandEnvironmentVariables` will only expand Windows-style environment variables (e.g., it will expand `%HOME%` but not `$HOME`).  Do we need to better support Linux and MacOS here?

kenkendk: Yes, I have a comment about that in the code. Until now it has only used the Windows-style environment variable expansion (that is how .Net does) but it is simple to do the Linux style expansion as well.

Should we just drop the old Windows-style environment variable expansion on non-Windows? I am not sure.... I think it may break something for current users expecting the expansion, but new users will be confused if it does not work as all the other stuff on their system....

kenkendk:
Yes, I think it is a potential problem. The source paths for the backup job is saved in the database. If we no longer expand them, we should run through all saved (source)paths and apply the expansion.

It could still be problematic for CLI users, but I expect them to be better at dealing with the issue.

kenkendk:
For the build error, if you merge `master` into your branch, it should fix the build issue on Travis.

Duplicati/Library/Utility/Utility.cs
warwickmm: I'm not sure what the best approach is.  Is the main purpose here to expand environment variables specified in command-line options and filters?  If so, one option is to only support (and document) the Windows-style syntax.  Unless there's a way to support both on non-Windows environments if we want to avoid breaking current functionality?

I'll create a separate issue regarding this matter so we can continue the discussion there.

warwickmm:
> Yes, I think it is a potential problem. The source paths for the backup job is saved in the database. If we no longer expand them, we should run through all saved (source)paths and apply the expansion.

Unfortunately, I'm not familiar enough with the code to do address this.  I'd appreciate any help/hints on this. :confused: 

> For the build error, if you merge master into your branch, it should fix the build issue on Travis.

I merged in 43d9b51743, but 2 of the Travis jobs failed with:

```No output has been received in the last 10m0s, this potentially indicates a stalled build or something wrong with the build itself.```

I'm not sure how to diagnose this...did my changes break the build somehow?

Duplicati/Library/Utility/Utility.cs
warwickmm: Created issue #2838.

kenkendk:
Sorry for the delay.

The build error is a Travis issue, no worries.

I have not merged this yet because I need to find the time to look into the potential storage of paths with environment variables and figure out if/how this should be handled.

warwickmm:
No worries.  Sorry for the incomplete pull request...I'm still slowly finding my way around the code.

kenkendk:
Ok, found it. It actually uses `%HOME%` so no problems with that.
If we fix #2838 we need to take care to update the variables in the database as well.

Issue #1568, opened 2016-02-04
Question about keep-time and keep-versions


albertgouws:
Hi, I have been using Duplicati 2 for a while now and love it, thanks for the hard work!

I have a question about the keep-time and keep-versions command line settings.

Is this what you're talking about which defaults to 1 month by default?

Just to be 100% clear on what this does:
If I set this setting to 2 days, run a backup of 10 files and delete 5 of those files after the backup, will Duplicati delete those 5 files from the backup set when I run a backup after say 3 days?

Can the keep-version setting be used at the same time as the keep-time? e.g. set a keep-versions to 5, and keep-time to 2 months. If I then run a backup every day, it will only keep the last 5 **changes** to my files.  If I deleted my my files, **each version** will then stay in the backup set until that version is 2 months old?

Thanks for your time!
Albert

kenkendk:
If you specify both `--keep-time` and `--keep-versions` you get whichever is smaller.
I think you want the opposite.

The default for `--keep-time` and `--keep-version` is zero, meaning that old backups are never deleted.

Each time you run a backup (and files have changed), a new "dlist" (a snapshot, or a backup description) file is created on the remote store. The two options determine when these dlist files are deleted.

Once a dlist file is removed, that particular backup/snapshot cannot be restored anymore.

kenkendk:
I changed this such that you can no longer specify both, because it was not obvious how it should work with two values.

PR #2851, opened 2017-10-27
Match web UI's default keep time with Library's

ItEndsWithTens:
I'd planned to open an issue to discuss why backups currently default to a 3 month retention time, hoping to convince you of the school of thought that says backup software shouldn't be deleting anything unless users explicitly tell it to. Some digging, however, suggests you already feel that way: your comment on issue #1568 implies new backups are meant to default to "unlimited" keep time, and [this line in Options.cs](https://github.com/duplicati/duplicati/blob/202a6641e568abe53d3d5a13b9d9ea4d8cfd0d89/Duplicati/Library/Main/Options.cs#L797) does in fact return 0 if users haven't specified anything.

On the other hand the web UI uses 3 months as its default, as seen in [newbackup.json](https://github.com/duplicati/duplicati/blob/43d9b51743d0135f2d789b1df58de06e14b6d646/Duplicati/Server/newbackup.json#L7). Since that file hasn't been modified since 2014, I'm assuming your later comments and current code more accurately reflect your design intent, so I took the liberty of making the change myself. Apologies if I've missed something important.

### DIFF
--- a/Duplicati/Server/newbackup.json
+++ b/Duplicati/Server/newbackup.json
@@ -4,7 +4,7 @@
             {Name: "encryption-module", Value: "aes"},
             {Name: "compression-module", Value: "zip"},
             {Name: "dblock-size", Value: "50mb"},
-            {Name: "keep-time", Value: "3M"}
+            {Name: "keep-time", Value: ""}
         ]
     },
     Schedule: {


kenkendk:
@renestach What do you think?

kenkendk:
Thanks for the PR. I think we should look into using the new retention policy such that we can have infinite backups, but have the number of backups increasing slowly.

ItEndsWithTens:
Sure, sounds good to me! I can appreciate the importance of reining in the size of one's backups, considering how easily they can grow out of control depending on the frequency of the schedule and the file changes. I only felt that the UI defaults should match the settings used elsewhere in the software.

You're welcome for the pull request, even though it feels a little silly to open one over a matter of two characters. I'm never sure what to do, honestly. Some developers might feel that such a small PR is vain and better left as an issue in the tracker, where others might feel that if a user knows what to change and how to change it, they should show up with the code ready to go; I chose the latter in this case. I hope that wasn't too presumptuous of me. :)

kenkendk:
I strive for whatever is best and easiest for the end user. Ideally, if the user needs to change a setting, we should fix that. 
Of course there are settings where this is not possible (i.e. what folders to back up, and where to store the backup data).

So, any fix is welcome, even if it is only two characters :)

renestach:
Here my thoughts:

* Unlimited backups are nice. You will never lose anything.
* The downside of unlimited backups is the constantly growing backups size and a potential performance degration caused by this. This is why we decided not to take "unlimited backup size" as a default.
* Now that we have a smarter solution regarding backup retention, I prefer to test the new retention policy of #2084 and if it works without a flaw, we should make it the new default setting. This is still in theory a constantly growing backup but at a much slower pace.

The UI for it could look like this:

backup retention

ItEndsWithTens:
Well now I feel like a fool; I searched for what I could find, but hadn't seen mention of the fact that the 3 month default was an explicit design choice.

As an end user I was alarmed to discover that my backups were being deleted by default after that time, for fear that a file might be deleted accidentally (or maliciously, for that matter) and I wouldn't notice for 3 months and 1 day, only to find it wasn't backed up anymore. Deletion is a sensible solution for the size and performance issues, but it surprised me as a default. Digging around in search of answers, I found mentions of the default keep time being unlimited, but I seem to have misinterpreted that and jumped the gun. Sorry!

No hard feelings if you'd like to revert the merge and just focus on this new and improved "smart backup" retention, which does look appealing.

renestach:
All contributions are welcome especially when they are based on good thoughts and some research. The new retention policy still has to prove that it works without any issues. If it does not, we might start the whole discussion over :-)

PR #2867, opened 2017-11-05
Update EditBackupController.js to fix spelling mistake

davegold24:
changed loose to lose. Still learning how to use github properly, so if I'm doing it wrong, tell me.


kenkendk:
All fine :)

PR #2875, opened 2017-11-07
Update loose to lose

davegold24:


kenkendk:
Closing because I will make the update in a combined commit.

PR #2884, opened 2017-11-07
Change wording from experimental to beta

pietsch:
According to the blog, Duplicati 2 is no longer considered experimental but beta.


kenkendk:
Yes, you are right. When I wrote it I knew I would forget to update it :)

Issue #2852, opened 2017-10-27
Slow folder browsing during restore


dgold9670:
- [X] I have searched open and closed issues for duplicates.

----------------------------------------

## Environment info
- **Duplicati version**: 2.0.2.1_beta_2017-08-01
- **Operating system**: Windows 7 Pro 32-bit; disk is SSD
- **Backend**: Minio
Duplicati is running as a service.

## Description
Browsing for file restores is very slow. 
This was discussed in https://forum.duplicati.com/t/very-slow-folder-browsing-during-restore/430

## Steps to reproduce
1. Open Duplicati, go to restore window
2. Select restore from existing backup job
3. Browse in window. Opening a folder takes a minimum of a minute; watching disk monitor shows very heavy IO.

## Debug log
Pending. Given the large amount of discussion in the link above, the issue is clearly understood by dev.

## Thoughts on solutions
It seems that the current method of creating browsing views is to extract significant portions of the database into a temporary database ("list") and then use that temporary sqllite database. 
Each time a new folder is opened, a separate list operation is run.
I can think of three types of solutions:
1. Create index at end of backup, so that the list does not need to extract significant amounts of data
--possible issue is index corruption and increased IO requirements. Code would have to exist to recover from failures.
2. Reuse the temporary SQL database, so only a single list is required.
3. Extract data from database into a different structure, such as CSV or TSV. This could be ordered by directory structure, and would load into the GUI faster.
--main issue would be scalability

davegold24:
Additional thoughts:

davegold24:
Another few thoughts on options to make the restore faster:
1. After selecting the restore to pull from, start by showing versions only, and not versions and file tree. The amount of database work to pull the entire list of files is significant, but the amount to show restore from versions is not. The next screen could show the file tree.
2. Run the complex query that finds all the files that are part of a give point in time once, and keep those results (temp database) around for the entire restore session. There is a single query that takes most of the time, and after each step its results are deleted, then recreated.
3. Add a depth field and/or parent path to help speed up browsing, and simplify queries. ie use the temp filelist from step 2, and query on either to make the query run faster. 
I think both could be done during step 2, without modifying the backup portion of the database (ultimately modifying the backup is better, but using the step 2 results would be easier, IMHO)

I wish I had the programming chops to write the code. I'd certainly be willing to kick in some funds if that would help (although my sense is that isn't the major motivating factor here)

If someone does want to work on this, I'm glad to help in any way I can though.

[...]

PR #2897, opened 2017-11-15
Changed subqueries to improve folder browsing in restore

RobSauro:
Fix to problem reported in issue #2852. Some subqueries have too wide scope and when joining take minutes to run. With this change the subqueries already come filtered and the joining is way less expensive. A sample restore that usually took 250 seconds to build the file list now takes 5 sec.

### DIFF
--- a/Duplicati/Library/Main/Database/LocalListDatabase.cs
+++ b/Duplicati/Library/Main/Database/LocalListDatabase.cs
@@ -266,8 +266,8 @@ namespace Duplicati.Library.Main.Database
                         //Then we select the matching results
                         var filesets = string.Format(@"SELECT ""FilesetID"", ""Timestamp"" FROM ""{0}"" ORDER BY ""Timestamp"" DESC", m_tablename);
                         var cartesianPathFileset = string.Format(@"SELECT ""A"".""Path"", ""B"".""FilesetID"" FROM ""{0}"" A, (" + filesets + @") B ORDER BY ""A"".""Path"" ASC, ""B"".""Timestamp"" DESC", tbname, m_tablename);
-                                                
-                        var filesWithSizes = @"SELECT ""Length"", ""FilesetEntry"".""FilesetID"", ""File"".""Path"" FROM ""Blockset"", ""FilesetEntry"", ""File"" WHERE ""File"".""BlocksetID"" = ""Blockset"".""ID"" AND ""FilesetEntry"".""FileID"" = ""File"".""ID"" ";
+
+                        var filesWithSizes = string.Format(@"SELECT ""Length"", ""FilesetEntry"".""FilesetID"", ""File"".""Path"" FROM ""Blockset"", ""FilesetEntry"", ""File"" WHERE ""File"".""BlocksetID"" = ""Blockset"".""ID"" AND ""FilesetEntry"".""FileID"" = ""File"".""ID"" AND FilesetEntry.""FilesetID"" IN (SELECT DISTINCT ""FilesetID"" FROM ""{0}"") ", m_tablename);
                         var query = @"SELECT ""C"".""Path"", ""D"".""Length"", ""C"".""FilesetID"" FROM (" + cartesianPathFileset + @") C LEFT OUTER JOIN (" + filesWithSizes + @") D ON ""C"".""FilesetID"" = ""D"".""FilesetID"" AND ""C"".""Path"" = ""D"".""Path""";
                         
                         cmd.AddParameter(pathprefix, "1");                    
@@ -318,7 +318,7 @@ namespace Duplicati.Library.Main.Database
                     //Then we select the matching results
                     var filesets = string.Format(@"SELECT ""FilesetID"", ""Timestamp"" FROM ""{0}"" ORDER BY ""Timestamp"" DESC", m_tablename);
                     var cartesianPathFileset = string.Format(@"SELECT ""A"".""Path"", ""B"".""FilesetID"" FROM ""{0}"" A, (" + filesets + @") B ORDER BY ""A"".""Path"" ASC, ""B"".""Timestamp"" DESC", tmpnames.Tablename, m_tablename);
-                    var filesWithSizes = @"SELECT ""Length"", ""FilesetEntry"".""FilesetID"", ""File"".""Path"" FROM ""Blockset"", ""FilesetEntry"", ""File"" WHERE ""File"".""BlocksetID"" = ""Blockset"".""ID"" AND ""FilesetEntry"".""FileID"" = ""File"".""ID"" ";
+                    var filesWithSizes = string.Format(@"SELECT ""Length"", ""FilesetEntry"".""FilesetID"", ""File"".""Path"" FROM ""Blockset"", ""FilesetEntry"", ""File"" WHERE ""File"".""BlocksetID"" = ""Blockset"".""ID"" AND ""FilesetEntry"".""FileID"" = ""File"".""ID""  AND FilesetEntry.""FilesetID"" IN (SELECT DISTINCT ""FilesetID"" FROM ""{0}"") ", m_tablename);
                     var query = @"SELECT ""C"".""Path"", ""D"".""Length"", ""C"".""FilesetID"" FROM (" + cartesianPathFileset + @") C LEFT OUTER JOIN (" + filesWithSizes + @") D ON ""C"".""FilesetID"" = ""D"".""FilesetID"" AND ""C"".""Path"" = ""D"".""Path""";
                     using(var rd = cmd.ExecuteReader(query))
                         if(rd.Read())
@@ -368,7 +368,7 @@ namespace Duplicati.Library.Main.Database
                     for(var i = 0; i < m_filesets.Length; i++)
                         dict[m_filesets[i].Key] = i;
                     
-                    var summation = @"SELECT ""A"".""FilesetID"" AS ""FilesetID"", COUNT(*) AS ""FileCount"", SUM(""C"".""Length"") AS ""FileSizes"" FROM ""FilesetEntry"" A, ""File"" B, ""Blockset"" C WHERE ""A"".""FileID"" = ""B"".""ID"" AND ""B"".""BlocksetID"" = ""C"".""ID"" GROUP BY ""A"".""FilesetID"" ";
+                    var summation = string.Format(@"SELECT ""A"".""FilesetID"" AS ""FilesetID"", COUNT(*) AS ""FileCount"", SUM(""C"".""Length"") AS ""FileSizes"" FROM ""FilesetEntry"" A, ""File"" B, ""Blockset"" C WHERE ""A"".""FileID"" = ""B"".""ID"" AND ""B"".""BlocksetID"" = ""C"".""ID"" AND ""A"".""FilesetID"" IN (SELECT DISTINCT ""FilesetID"" FROM ""{0}"") GROUP BY ""A"".""FilesetID"" ",m_tablename);
                     
                     using(var cmd = m_connection.CreateCommand())
                     using (var rd = cmd.ExecuteReader(string.Format(@"SELECT DISTINCT ""A"".""FilesetID"", ""B"".""FileCount"", ""B"".""FileSizes"" FROM ""{0}"" A LEFT OUTER JOIN ( " + summation + @" ) B ON ""A"".""FilesetID"" = ""B"".""FilesetID"" ORDER BY ""A"".""Timestamp"" DESC ", m_tablename)))


JonMikelV:
I haven't dug into any Duplicati SQL yet (I really should...) but it looks like you added a FileSet ID filter to the query.  Was this filter previously happening somewhere else later in the process or were the original "too large" results simply incorrect and nobody noticed?

RobSauro:
I noticed that subqueries are returning full set of data to join to the outer query, which will get only filesetID present on the temporary table. 

If it filters the subquery to return only data that's "joinable" it makes way less data return, and that turns the response way faster.

I have one example (unaltered from my debug log):

> SELECT DISTINCT A."FilesetID",
                B."FileCount",
                B."FileSizes"
  FROM [Filesets-7C8CA60FDC5B5544A69A8EBB125B54F3] A
       LEFT OUTER JOIN
       (
           SELECT A."FilesetID" AS FilesetID,
                  COUNT( * ) AS FileCount,
                  SUM(C."Length") AS FileSizes
             FROM FilesetEntry A,
                  File B,
                  Blockset C
            WHERE A."FileID" = B."ID" AND 
                  B."BlocksetID" = C."ID"
            GROUP BY A."FilesetID"
       )
       B ON A."FilesetID" = B."FilesetID"
 ORDER BY A."Timestamp" DESC;

The outer "B" query only needs records that have FilesetEntry.FilesetID in the outer "A" query FilesetIDs, all others aren't joined and result in null values for outer "B". The same occurs in the other queries.

To answer your question I think the join will do the filtering and it could just ignore results that didn't come in the temporary table. I don't know if someone noticed that before but I think the query only has that problem after your database is already very large (like mine, with 2Gb), so, in the initial tests it appeared reasonable. I hope I've been clear as I'm not a native English speaker.

JonMikelV:
Your description makes sense - and don't worry about your English, it's very good (I wouldn't have known you were non-native if you hadn't pointed it out).  :-)

Sorry for asking what are likely silly questions, but...
1. Would it not be slightly more efficient to select COUNT(C."Length") instead of COUNT(*) - particularly since C."Length" is used in the SUM?
2. Am I correct in assuming the purpose of doing the counts in a subquery B is so even if there are no matching records we still get a row from [Filesets-7C8CA60FDC5B5544A69A8EBB125B54F3]?
3. I'm not familiar with SQLite (yet) but are you at risk of getting Null FileCount and FileSizes from subquery B if there are no records matching FileSetID?

RobSauro:
1. As I see, that specific code is used to build the total number of files and size in "ListResultFileSet" (to know how many files belong to each set and how large it is).  
"Count" with a parameter only instructs to count items with non-null values into that column but, since I already removed null values by filtering, the result would be the same and it wouldn't add to total anyway (we are grouping by FilesetID and using sum and count, null items wouldn't matter in my view). 
The count with asterisk just get all rows regardless of their values and seems more straightforward than asking to inspect each row again (I work everyday with Oracle and I imagine SQLite doesn't differ that much in simple optimizations).

2. I think the same. The programmer wants to see the fileset row even if there are no files in it, which I think wouldn't happen anyway (why would it create a fileset to leave it empty?). The filtering won't make those lines disappear, it just "helps" leaving relevant rows to be joined.

3. As I see no. That query particularly is used to retrieve the last Fileset and it's size (count and length). I can't think of a Fileset that has no files, and even if it happens, it wouldn't be a problem. That query specifically don't have a path filter to work on (just the Fileset filter on the temp table creation).

The complex query that worries a bit more is that (without my adjustment), and where we get the best result:

```
SELECT C."Path",
       D."Length",
       C."FilesetID"
  FROM (
           SELECT A."Path",
                  B."FilesetID"
             FROM [Filenames-1EF488B7A6B5254094FA2375C07E3C52] A,
                  (
                      SELECT "FilesetID",
                             "Timestamp"
                        FROM [Filesets-7C8CA60FDC5B5544A69A8EBB125B54F3]
                       ORDER BY "Timestamp" DESC
                  )
                  B
            ORDER BY A."Path" ASC,
                     B."Timestamp" DESC
       )
       C
       LEFT OUTER JOIN
       (
           SELECT "Length",
                  FilesetEntry."FilesetID",
                  File."Path"
             FROM Blockset,
                  FilesetEntry,
                  File
            WHERE File."BlocksetID" = Blockset."ID" AND 
                  FilesetEntry."FileID" = File."ID"
       )
       D ON C."FilesetID" = D."FilesetID" AND 
            C."Path" = D."Path";
```
That returns the details about files and folders. 

The outer "C" query is from a temporary table created to have a group of selected filesets ([Filesets-7C8CA60FDC5B5544A69A8EBB125B54F3]), usually one,  and another temp table ([Filenames-1EF488B7A6B5254094FA2375C07E3C52]) containing filenames matching the filter choosen by user.

If you run it the way it is, you have a giant effort to perform the outer "D" query and joining it, since it doesn't filter anything. It will return data that won't join (FilesetIDs that aren't present in the temp tables). 

You have to remember the "leading" table in "LEFT OUTER JOIN" is outer "C", it means that all rows from "C" will always return and, if it finds something matching in "D" it will put its value in "D".Length. So, it has to compare millions of rows with FilesetIDs not matching just to find the right ones. It means the result is the same (I did test it) since all "C" rows are there as expected and "D" matching values were all found.

I wrote too much but I think the changes just improve the queries, not changing its values.

JonMikelV:
I don't think you wrote too much, and I appreciate your time with this.

Not that it matters, but I'm no expert though - I also work daily on SQL, though for me it's fine tuning MS SQL.  But I agree, that 2nd larger query like an optimization candidate...

I'm not familiar with the SQLite formatting or optimizations, so please correct me if I'm wrong but when a select is written like this:
~~~
SELECT A."Path",
    B."FilesetID"
FROM [Filenames-1EF488B7A6B5254094FA2375C07E3C52] A,
    (
        SELECT "FilesetID",
               "Timestamp"
          FROM [Filesets-7C8CA60FDC5B5544A69A8EBB125B54F3]
         ORDER BY "Timestamp" DESC
    )  B
~~~

Is that a "NATURAL" join that automatically joins on matching field names (in this case FilesetID)?  If so, then I guess I could understand the need for the ORDER BY "Timestamp" DESC in B if there's a one-to-many relationship between A and B.

But if A and B have a 1:1 relationship, then wouldn't the following be more efficient as it filters B down to specific FilesetIDs?
~~~
SELECT A."Path",
    B."FilesetID"
FROM [Filenames-1EF488B7A6B5254094FA2375C07E3C52] A
  INNER JOIN [Filesets-7C8CA60FDC5B5544A69A8EBB125B54F3] B On (B."FilesetID" = A."FilesetID")
ORDER BY A."Path" ASC, B."Timestamp" DESC
~~~

I feel the long query you mentioned could be re-written as a set of INNER and LEFT joins rather than subqueries.  (I just don't know the database schema enough yet to understand how the FilesetEntry table relates to the Blockset and File tables.)
~~~
SELECT B."Path"
  , File."Length"
  , A."FilesetID"
FROM [Filesets-7C8CA60FDC5B5544A69A8EBB125B54F3] A  -- Start here, fewer records (was B & C)
  INNER JOIN [Filenames-1EF488B7A6B5254094FA2375C07E3C52] B On (B."FilesetID" = A."FilesetID")  -- Get paths (was A & C)
  LEFT JOIN FilesetEntry On (FilesetEntry."FilesetID" = B."FilesetID")  -- Get lengths, this & following were D
  INNER JOIN BlockSet On (Blockset."???" = B."???")
  INNER JOIN File On (File."BlocksetID" = BlockSet."ID")
    AND (File."ID" = FilesetEntry."FileID")
    AND (File."Path" = B."Path")
ORDER BY B."Path" ASC, A."Timestamp" DESC
~~~

RobSauro:
I agree with you about room for optimization on 2nd query. That cross-join may sound terrible at first look but my approach was to do a minimal change to get the best improvement. 

In 
```sql
SELECT A."Path",
    B."FilesetID"
FROM [Filenames-1EF488B7A6B5254094FA2375C07E3C52] A,
    (
        SELECT "FilesetID",
               "Timestamp"
          FROM [Filesets-7C8CA60FDC5B5544A69A8EBB125B54F3]
         ORDER BY "Timestamp" DESC
    )  B
```
It does a cross-join just to have all paths mixed with all FilesetIDs existing in the temp table, just to join it later (in most of cases the [Filesets-7C8CA60FDC5B5544A69A8EBB125B54F3] will have just the Fileset selected in restore window, so it's not so demanding). 

The timestamp order was just to be sure the latest Filesets are on top, but it's not a problem also, since it will have few records anyway.

You proposed that:
```sql
SELECT A."Path",
    B."FilesetID"
FROM [Filenames-1EF488B7A6B5254094FA2375C07E3C52] A
  INNER JOIN [Filesets-7C8CA60FDC5B5544A69A8EBB125B54F3] B On (B."FilesetID" = A."FilesetID")
ORDER BY A."Path" ASC, B."Timestamp" DESC
```
But the temp table ([Filenames-1EF488B7A6B5254094FA2375C07E3C52] in that example) is created like that:
```sql
CREATE TEMPORARY TABLE "Filenames-1EF488B7A6B5254094FA2375C07E3C52" ("Path" TEXT NOT NULL);
```
You won't have the FilesetID field to do that, so it's another point to change. 

You'd have to rewrite that method and I don't see much gain if [Filesets-7C8CA60FDC5B5544A69A8EBB125B54F3] is small.

I'm not too familiarized with the schema as well, but as I see the "File" belongs to many "Blocksets" and to many "FilesetEntries" also, in his turn the "FilesetEntry" belongs to a "Fileset".

The last query also supposes the [Filenames-1EF488B7A6B5254094FA2375C07E3C52] temp table has information on Blocksets, but that's not the case, it has just the path info.

To bring it all to a simple query would require rewriting of lots of another points in code and eventually modifications to the schema. I also know that code may have other queries which are candidates to improvement but I only modified the most demanding ones, in my view, so I could reduce their running time to a reasonable figure. I'm sure you could gain a second here and there doing more changes.

I don't know if that's planned in your roadmap (knowing that it would require data migration for current users or a complete database rebuild) but I'm interested in helping if needed.

JonMikelV:
Thanks for the clarifications - I guess I still have a lot to learn about sqlite and the Duplcicati schema. :-)

I believe there are plans to re-structure Path field so it's not a single long string, but as you pointed out that requires a lot of individual procs to be updated so it's not likely to happen any time soon.

Until then individual optimizations such as yours seem to be a great way to improve performance within the limits of the current design.

Another thing I've ask about, but haven't had time to champion, is codeblock performance metrics to help identify the specific areas that are running slowly.  I imagine a central logging object that only gets written to when a particular step takes longer than expected (as a passed in parameter).  This could allow developers to narrow down more closely where the issues are happening.

That reminds me - I set up a test backup job of the same 1k file copied 500k times (into various subfolders).  At the least I was hoping to have a test setup that would allow A/B performance testing with and without your updated SQL and at best I was hoping to identify what actually slows things down (many blocks, many files, long paths, many versions, etc).

But in the end the job creates a 313M sqlite file and the restore folder trees are still opening in under 8 seconds, so I have yet to figure out exactly which scenario(s) make one person's restore browsing slow while another person has no issues.

RobSauro:
I imagine people that have the same problem as me experience a 4 to 5 minutes long delay in initial fetch and the same delay for each folder they click.
My situation, for information, is a 2Gb SQLite file, 380k files with various sizes (lots of small ones but some with multi-Gb) and 125 versions.
My test reduced that 4~5 minutes wait to something like 5 seconds, and each folder I click now takes that same 5 seconds.
The best improvement is in the "search" file in the restore section. Now it's usable. I can search for a specific file and it shows as expected. That turns the restore section better than most online backup solutions I've used until now.
I don't know if other people has different situations but it wouldn't hurt to try, just in a beta release, if you agree that new queries don't affect expected result.

Athep:
RobSauro, this is interesting. I have a rather big dataset to try on. However, i need a ready run compiled version for Windows in order to test.

RobSauro:
Athep. If you really want to try, first I recommend the following:
1. Do a Back-up of your SQLite database;
2. Export your Back-up configuration to a json file, just in case;
3. Do a Restore Point in Windows;
4. Uninstall current Duplicati version;
5. Install the new one and configure properly (it may need to recover your configuration from (2) and redirect your database to the old one.

I've build an installer and shared it [here](https://1drv.ms/u/s!AriumgYj3leDyhkAASv2EiquRKjZ).

I have to remember that it worked for me and I don't know if it will for you, and that it could ruin your present configuration, that's why it's important to back-up every config.

If there's some rule preventing posting such build (I didn't find) just let me know and I'll remove it.

JonMikelV:
Rather than the Windows Restore Point/uninstall/install path, would it be easier to have an unzippable version and run it with the `--portable-mode` parameter?

RobSauro:
Just copied [here](https://1drv.ms/u/s!AriumgYj3leDyhpCavRSbgQJQD2y) a zipped folder. 

It worked but the configuration is missing. I must have done something wrong, but it's simple to restore the json config and redirect that to the proper SQLite database, I think.

Athep:
Wow! Time for the initial fetch/listing is now reduced from about 40 min to just a few seconds in my backup!. Subsequent listings is equally fast! There is 130 000 files, 502 versions and 255 GB data in the backend. 

If this fix not is breaking something else in the code, this is the way to go.

kenkendk:
Great catch!

The `FileID` is unique to the set anyway, so the results should be exactly the same.
Obviously, SQLite cannot guess that this is the case, so adding the information in the query enables it to cut the number of files it checks quite drastically.

RobSauro:
Athep. 
Great to know it worked well for a much larger database (lots of versions and blocks). Thanks for the follow-up!

davegold24:
If you want another tester, let me know. I'm excited and happy to see this coming through.

RobSauro:
davegold24,
You could try the zipped build I left [here](https://1drv.ms/u/s!AriumgYj3leDyhpCavRSbgQJQD2y). You just need to run it with --portable-mode, as cited above.

davegold24:
@RobSauro tested and verified. I stopped the windows service, ran the tray icon with the --server-datafolder option.
My database is about 1GB and is on SSD.
Initial load took 15-30 seconds, after that each folder open took around 10 seconds. Well within "normal" restore browsing speeds. 
My commendations for your work. And appreciation.

kenkendk:
I put up 2.0.1.13 as a canary build that includes this change.

dgold9670:
@kenkendk thanks. I noticed, and upgraded my systems to this version.

PR #2898, opened 2017-11-16
Only try to load Windows registry filters on Windows.

tygill:
The reason for this is that Windows filters may be used on other systems (e.g., to ignore Windows files when backing up a Windows drive)

### DIFF
--- a/Duplicati/Library/Utility/DefaultFilters.cs
+++ b/Duplicati/Library/Utility/DefaultFilters.cs
@@ -382,15 +382,19 @@ namespace Duplicati.Library.Utility
         /// <returns>The list of paths to exclude.</returns>
         private static string[] GetWindowsRegistryFilters()
         {
-            try
+            if (Utility.IsClientWindows)
             {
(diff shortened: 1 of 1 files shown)


kenkendk:
As I understand it, the registry operations are not supported on non-Windows, so I don't think there is any harm in calling it. Besides, the registry method is only called when using Windows filters:
https://github.com/duplicati/duplicati/blob/master/Duplicati/Library/Utility/DefaultFilters.cs#L237

Anyway, I don't see any harm in the change you propose, so I'll merge it :)

tygill:
Mono seems to throw an exception (not supported on platform, or something similar), so it does look like this will only work on Windows.

The reason for the check is that the Windows filters can be used when running Duplicati on another OS. For example, if a Linux machine mounts a Windows drive in order to back it up. There are some filters that won't work in this case (since they look only for files at the root of the drive, which won't be the same style on Unix style paths), but many of the others should still work.

(I've thought about ways to try to get the other paths to work on other OS's, but I'm not sure it's worth the effort.)

PR #2917, opened 2017-12-01
fix contributor gallery link

tooomm:
As spotted and reported to Open Collective here: https://github.com/badges/shields/pull/1300#issuecomment-348608215


kenkendk:
Thanks!

PR #2922, opened 2017-12-10
Fix <param> issues in XML documentation

warwickmm:
This fixes a few issues in the C# XML documentation.  There were several places where a `<param>` element corresponding to a non-existent parameter was removed.  We also fixed some of the parameter names to match the names used in the code.


kenkendk:
Thanks!

Issue #2845, opened 2017-10-24
Restoration state text goes on top of header in restore section


darksworm:
- [x] I have searched open and closed issues for duplicates.

----------------------------------------

## Environment info

- **Duplicati version**: 2.0.2.1
- **Operating system**: Windows 10
- **Backend**: N/A

## Description

Replace me

## Steps to reproduce

1. Start backup restoration
2. Resize browser window (not sure if neccessary)
3. Scroll down

- **Actual result**:
Text shows up on top of header
- **Expected result**:
Text gets hidden under header

## Screenshots

![2017-10-24_17-04-23](https://user-images.githubusercontent.com/9987548/31947680-9252a9c4-b8dd-11e7-9412-e57a09db9339.gif)

kenkendk:
Thanks for the illustration! It leaves no doubt to what the problem is 😄

kenkendk:
This was fixed with 4f1b1a1

PR #2944, opened 2017-12-31
Fix for #2845

Pectojin:
During the restore process the floating progress text is displayed on top of the header due to the "position: absolute" attribute. It doesn't do anything else than cause the issue described in #2845

Specifically I deleted 18 characters on line 4 between column 63092 and 63110 ;)


kenkendk:
That file is generated from the less files:
https://github.com/duplicati/duplicati/tree/master/Duplicati/Server/webroot/ngax/less

I can't merge this change, because it will be lost when we rebuild the styles.css file again.
Do you know which class that has the `position: absolute` attribute that you removed?

Pectojin:
Oh, I didn't realize it was generated from the less file. That explains the compression.

It's the `.progress>span` definition. I can change the PR in the morning.

kenkendk:
Perfect, thanks!

Issue #2945, opened 2017-12-31
Advanced options list display issue


Pectojin:
- [x] I have searched open and closed issues for duplicates.

----------------------------------------

## Environment info

- **Duplicati version**: debug (upstream from github)
- **Operating system**: OSX 10.13.2
- **Backend**: Google Drive and Local storage

## Description

When adding advanced options to a backup I get some pretty weird behavior when I reach 11 options. Until I hit 11 options, they're just listed in the order i add them, but once I add 11 options it'll start acting up. The data will still be in the same order, but the option name and option field type changes.

In my case the retention-policy is suddenly replaced by --send-mail-username (while preserving the retention policy data). Then --retention-policy has replaced --allow-missing-source, still retaining the value "true" but now displaying it as a text field rather than a checkbox.

I've attached two pictures because it's really the easiest way to show it. First picture looks correctly (and has 10 items). Second picture looks all messed up and the only thing I did was add an additional item to the list.

I tried adding various different advanced options and with various data, but it doesn't make any difference. It's just that angular is mishandling the data when it reaches 11+ options.

## Steps to reproduce

1. Edit an existing job or make a new one
2. Add 10 advanced options
3. Take notice of the items in the list and their data
4. Add the 11th option
5. Check for data miss-matching in the list.

In my case the first item was always messed up.

- **Actual result**:
Something happened to the list
- **Expected result**:
Nothing should happen to the list

## Screenshots

<img width="829" alt="screen shot 2017-12-31 at 23 11 25" src="https://user-images.githubusercontent.com/1561484/34464468-b88156a8-ee80-11e7-9633-1325b0e5cadd.png">
<img width="839" alt="screen shot 2017-12-31 at 23 11 10" src="https://user-images.githubusercontent.com/1561484/34464469-b89c0c00-ee80-11e7-8568-d7c9871327ec.png">

Pectojin:
I think I've already found the offending code, but it was a bit difficult for me to explain in a commit message, so created this issue for reference.

I'm just testing my change before making a pull request.

kenkendk:
This was fixed with c05ad79

we9v:
Upgraded from 2.0.2.15_experimental to 2.0.2.17_canary_2018-01-23 today and I'm still having this issue in the Settings>Default Options section. (I haven't checked an individual backup's advanced settings, as I have all of them set globally in Default Options.)

[...]

PR #2946, opened 2017-12-31
Fix for #2945

Pectojin:
The orderBy attribute causes the issue described in #2945 and it doesn't actually do anything. The "$index" is just the order in the config file and they're already loaded in that order before they get sorted.


kenkendk:
Do you know why it messes up with the `$index` tracking ?

I cannot remember why, but I added that tracking to fix another problem. It is possible that whatever I was fixing, was later fixed elsewhere, but I would like to understand why it does not  work now.

Pectojin:
I'm actually not sure. Especially because the $index value is always correct. I printed it out next to the names during my troubleshooting and it never changed order.

I spent a good amount of time trying to figure out if it was caused by the helper functions, that get the name and description from the config string, but I couldn't get the function to print anything wrong (input string always matched the entry retrieved).

As far as I could tell, from my testing after removing the tracking, nothing "obvious" broke... But it's hard to tell when I don't know what to look for :)

kenkendk:
Ok, lets try this and see if others report something broken :)

PR #2951, opened 2018-01-05
AddVerboseMessage() now handles null args

JonMikelV:
Let AddVerboseMessage() handle null / empty args variable & update PurgeFilesHandler to correctly call AddMessage()


Duplicati/Library/Main/Operation/PurgeFilesHandler.cs
```diff
@@ -173,13 +173,13 @@ private void DoRun(Database.LocalPurgeDatabase db, Library.Utility.IFilter filte
                                             if (m_options.Dryrun)
                                                 m_result.AddDryrunMessage(msg);
                                             else if (m_options.Verbose)
-                                                m_result.AddVerboseMessage(msg);
```
kenkendk: Why did you change this call to the non-verbose version?

JonMikelV: That was my original fix before you suggested allow verbose to handle null args.  With the null args fix the non-verbose call could stay as it was.  Do you want me to change it back or "un-commit" that change?

kenkendk: If you can "unchange" it (roll back, reverse commit, or just a new commit), I will merge the fix.

JonMikelV: I expect there is a roll back or reverse commit command around here somewhere, but at the moment I'm spending more time trying to figure out Git than I am tweaking code so I'll just try to do another commit with the old code back in place.  :-)

JonMikelV: OK - I think I've got PurgeFilesHandler.cs back to the way it was (most likely handled in Git via the hardest possible way).  :-)

Do I need to submit a new pull request or something like that?

PR #2956, opened 2018-01-09
Check MSSQL 32bit instances on 64bit system

inat3k:
Some programs create MSSQL 32bit instances on 64bit system. In that case Duplicati can't find them because it checks only 64bit registry. That change make possible to check also 32bit registry.


kenkendk:
Thanks!

PR #2957, opened 2018-01-11
Adding Rclone as backend

wjansenw:
Provides a backend to rclone, allowing to link multiple cloud storage providers and technologies.


wjansenw:
I am testing to recreate a DB on an existing backup, and the listing does not seem to work as it should... Probably the large amount of files is causing a problem somewhere.

Duplicati/Library/Backend/Rclone/Rclone.cs
```diff
@@ -0,0 +1,242 @@
...
+            }
+
+            catch (System.ComponentModel.Win32Exception ex)
+            {
+                throw new Exception(String.Format("Program \"{0}\" does not exist", command), ex);
+            }
+
+            string error = p.StandardError.ReadToEnd();
+            p.WaitForExit();
+            if (error.Length > 0) {
```
kenkendk: Not familiar with rclone output, but if there is stderr data, but no match, no error is returned?

Duplicati/Library/Backend/Rclone/Rclone.cs
```diff
@@ -0,0 +1,242 @@
...
+                if (error.Contains("directory not found"))
+                    throw new FolderMissingException(error);
+                if (error.Contains("didn't find section in config file"))
+                    throw new Exception(String.Format("Missing config file? {0}", error));
+            }
+            return p.StandardOutput.ReadToEnd();
```
kenkendk: Not sure how much output there is here, but if there is a lot of output, the process could deadlock after filling the stdout buffer, causing it to hang and never trigger `WaitForExit` ?

Duplicati/Server/app.config
```diff
@@ -1,7 +1,29 @@
-<?xml version="1.0"?>
+﻿<?xml version="1.0" encoding="utf-8"?>
 <configuration>
   <runtime>
-    <loadFromRemoteSources enabled="true"/>
+    <loadFromRemoteSources enabled="true" />
```
kenkendk: Same, if not required, should be removed.

Duplicati/UnitTest/app.config
```diff
@@ -0,0 +1,27 @@
+﻿<?xml version="1.0" encoding="utf-8"?>
```
kenkendk: This should be removed.

Duplicati/UnitTest/Duplicati.UnitTest.csproj
```diff
@@ -169,6 +169,7 @@
     </ProjectReference>
   </ItemGroup>
   <ItemGroup>
+    <None Include="app.config" />
```
kenkendk: Just remove this

Duplicati/Server/webroot/ngax/styles/style.css
```diff
@@ -1,4 +1,4 @@
...
...after,div.modal-dialog .content.buttons .tooltipped:hover:before{display:inline-block;text-decoration:none}div.modal-dialog .content.buttons .tooltipped-w:after{right:100%;bottom:50%;margin-right:5px;-webkit-transform:translateY(50%);-ms-transform:translateY(50%);transform:translateY(50%)}div.modal-dialog .content.buttons .tooltipped-w:before{top:50%;bottom:50%;left:-5px;margin-top:-5px;border-left-color:rgba(0,0,0,.8)}.addwizard form.styled ul,.restorewizard form.styled ul{margin:20px;margin-left:0}.addwizard form.styled input[type=radio],.restorewizard form.styled input[type=radio]{width:20px;margin-left:5px;margin-right:5px}.addwizard form.styled label,.restorewizard form.styled label{width:auto;line-height:normal}.addwizard form.styled div.subtext,.restorewizard form.styled div.subtext{clear:both;margin-left:30px;padding-top:5px;color:#d3d3d3}.pauseoptions form.styled li{line-height:normal;padding:0}.pauseoptions form.styled li input{height:auto;margin-top:8px;margin-right:8px;width:auto}.progress-bar-striped{background-image:-webkit-linear-gradient(45deg,rgba(255,255,255,.15) 25%,transparent 25%,transparent 50%,rgba(255,255,255,.15) 50%,rgba(255,255,255,.15) 75%,transparent 75
```
kenkendk: You cannot change the `styles.css` file, as it is generated from the `less` files, and will be overwritten once the `less` files are recompiled.

wjansenw:
Hi @kenkendk ,
Thanks for the feedback. 

The most important issue is indeed the deadlock. I have rewritten this to async calls, but have not yet submitted this. This indeed blocks this pull request, as with large lists it will not work.

The other remarks: most of them probably came from VS itself...But I will check it in more detail... If I have some time.

Regards,
Wim

Duplicati/Library/Backend/Rclone/Rclone.cs
wjansenw: Indeed, i think I have improved this part already.

Duplicati/Server/webroot/ngax/styles/style.css
wjansenw: i don't believe I touched this file, so probably VS did it... Didn't even know it was in the request...

kenkendk:
Sounds great! 

I think you can just delete the VS auto-generated redirects. Duplicati does not use the redirects in the way that it is injected. Each build contains all dependencies, linked correctly.

There is 

If the user needs a special version, they can edit the local `app.config`.

There is a version of the non-blocking commandline call here:
https://github.com/duplicati/duplicati/blob/master/Duplicati/Library/Encryption/GPGEncryption.cs#L225

But if you write it with with `async` it can be made much cleaner and less error prone.

wjansenw:
I reverted the files that were changed by VS, and some other changes that came in (don't know how).
Also the asynchronous reading of the standardout and standarderror are included.

Duplicati/Library/Backend/Rclone/Duplicati.Library.Backend.Rclone.csproj
```diff
@@ -0,0 +1,61 @@
...
+  <PropertyGroup Condition=" '$(Configuration)|$(Platform)' == 'Release|AnyCPU' ">
+    <DebugType>full</DebugType>
+    <Optimize>true</Optimize>
+    <OutputPath>bin\Release</OutputPath>
+    <ErrorReport>prompt</ErrorReport>
+    <WarningLevel>4</WarningLevel>
+    <ConsolePause>false</ConsolePause>
+  </PropertyGroup>
+  <ItemGroup>
+    <Reference Include="Newtonsoft.Json, Version=10.0.0.0, Culture=neutral, PublicKeyToken=30ad4fe6b2a6aeed, processorArchitecture=MSIL">
```
kenkendk: It looks like you include NewtonSoft.Json v10.0.0.0 here, which causes everything else to break as it depends on v10.0.3:

https://travis-ci.org/duplicati/duplicati/jobs/330695475#L5910

wjansenw: i have v10.0.3 installed, but it reference v10.0.0.0... Couldn't correct in VS, so did it manually. Builds here without issues.

kenkendk:
Great work!

PR #2962, opened 2018-01-12
Move RunScript module to be configured first

BrianLakstins:
The process for adding configuration to scripts for sending email (like username and password) was not working.  It was altering the configuration, but the configuration was not being loaded for the SendMail module.

This update just re-arranges the order that the modules are loaded in so that the RunScript module is loaded first.

This update also changes the example to not use double quotes.  The double quotes were being added as part of the configuration.


kenkendk:
I think a real solution is to have an "order" for the modules, but for now fixing `runscript` as the first module fixes the immediate problems as no other module alters the options.

Issue #2970, opened 2018-01-18
WebDAV Backend doesn't connect to webdav server


inat3k:
- [x] I have searched open and closed issues for duplicates.

----------------------------------------

## Environment info

- **Duplicati version**: 2.0.2.16 canary
- **Operating system**: Windows 10 x64
- **Backend**: WebDAV

## Description

When i set up a backup with webdav backend and i press "Test Connection" to check the connection and create the new folder into the server, it freezes and returns an error

## Steps to reproduce

1. Create a new backup set
2. Set WebDAV backend and configure it
3. Press "Test Connection"

- **Actual result**:
If i set an existing path it freezes.
If i set a non existing path it gives 404 error.
- **Expected result**:
if i set an existing path it should give "Connection succesful"
If i set a non existing path it should ask me if i would create the new folder.

## Screenshots

![duplicati_webdav_error](https://user-images.githubusercontent.com/20947647/35108029-c776c2ba-fc72-11e7-9148-cc10b536cd53.png)

PR #2971, opened 2018-01-18
Fix for #2970

inat3k:
Fix for Test Connection to WebDAV Server

### DIFF
--- a/Duplicati/Library/Backend/WEBDAV/WEBDAV.cs
+++ b/Duplicati/Library/Backend/WEBDAV/WEBDAV.cs
@@ -246,9 +246,11 @@ namespace Duplicati.Library.Backend
 
                 FileEntry fe = new FileEntry(name, size, lastAccess, lastModified);
                 fe.IsFolder = isCollection;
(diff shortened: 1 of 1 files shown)


Duplicati/Library/Backend/WEBDAV/WEBDAV.cs
```diff
@@ -311,7 +313,7 @@ public string Description
 
         public void Test()
         {
-            this.TestList();
```
kenkendk: In principle, the call should be `this.TestList().FirstOrDefault()` to make sure the code runs, but all the other backends do it this way also (load all files).

kenkendk:
Just running Travis again, before merging

Duplicati/Library/Backend/WEBDAV/WEBDAV.cs
inat3k: I only reverted that commit (38883285eee5598e37d56e6ddc417fad7d872ed3) as indicated in description

Issue #2974, opened 2018-01-23
Wrong field focus when typing


bdillahu:
- [X] I have searched open and closed issues for duplicates.

----------------------------------------

## Environment info

- **Duplicati version**: 2.0.2.16_canary_2018-01-18
- **Operating system**: MacOS High Sierra 10.13.2
- **Backend**: Google Drive

## Description

When entering a new configuration, I added an advanced option for "Retention Policy" - when typing in the field after each character the focus jumps out of the field. To type more you have to re-click in the field.

## Steps to reproduce

1. Add advanced option (retention policy)
2. Type multiple characters in field
3. ...

- **Actual result**:
After each character focus jumps out of field
- **Expected result**:
Ability to type multiple characters

Pectojin:
Hey @kenkendk I think I found that thing that you fixed with the `$index` tracking, which you couldn't remember why you added ;) (#2946).

The issue happens consistently on v2.0.2.16_canary (where that PR was merged into) but not on v2.0.2.15_canary, and that PR was the only change to the [advancedoptionseditor.html](https://github.com/duplicati/duplicati/blame/c05ad799686c75f2c4eabd246f5558d09906c86e/Duplicati/Server/webroot/ngax/templates/advancedoptionseditor.html) file for a full year.

Without the `$index` tracking Angular seems to be replacing the entire `item` HTML element and reinserting it, on every character input, causing the loss of focus. With the `$index` tracking the element isn't replaced.

I think the reason is that when tracking without the index, it'll resort them every time the object is changed (even though index isn't changed). But the index tracking was also causing problems.

I'm looking into some options.

Pectojin:
Okay, I'm getting close to a fix that works for both this issue and my issue #2945 but I wanna be certain it works in all edge cases before I submit a pull request.

Pectojin:
This issue can be closed as #2975 reverses the change that introduced this issue.

Fix should be available in next canary build.

PR #2975, opened 2018-01-23
Advanced options list fixes

Pectojin:
First off, I reverted the original fix 8bed599fcd7116051ed881b0a6b51a41361b9fe8, as it is directly causing #2974

Secondly, `orderBy: 'item' as nn track by $index` caused the index be cast to a string. This caused some weird sorting, which partially explained what I was seeing in #2945 with more than 10 items in the list. Simply removing the quotes around `'item'` fixed the sorting. Now entries are sorted alphabetically (because they're sorted alphabetically in the option map).

After ordering by $index there was a miss match between the model and the names/types, like I was originally describing in #2945. This is because names/types are parsed from functions looking at the `item`. Meanwhile, the model was looking at the $index, of the sorted list, in the unsorted list causing the miss match. I fixed this by looking up the index using `$parent.ngModel.indexOf(item)`. This solution looks a bit weird, but it gets the right index solving the mismatch problem.

### DIFF
--- a/Duplicati/Server/webroot/ngax/templates/advancedoptionseditor.html
+++ b/Duplicati/Server/webroot/ngax/templates/advancedoptionseditor.html
@@ -1,5 +1,5 @@
 <ul class="advancedoptions">
-    <li ng-repeat="item in ngModel" class="advancedentry">
+    <li ng-repeat="item in ngModel | orderBy: 'item' as nn track by $index" class="advancedentry">
         <div>
             <label class="shortname">{{getShortName(item)}}</label>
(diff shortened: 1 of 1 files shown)


kenkendk:
I don't think I noticed the `'item'` being in quotes, but that is probably the reason why it broke initially.

Pectojin:
Well, partially the reason. We still had the problem of mismatching `$index`values in 73bed548170f770e17a9b42c040503c21c627ca4. I just couldn't see one issue because of the other.

PR #2978, opened 2018-01-24
Using separate interfaces to read and write to archives.

vnau:
Using parted IArchiveReader and IArchiveWriter interfaces to access archives instead of the single ICompression interface.

### DIFF
--- a/Duplicati/CommandLine/Help.cs
+++ b/Duplicati/CommandLine/Help.cs
@@ -169,7 +169,7 @@ namespace Duplicati.CommandLine
                     lines.Add("");
                     lines.Add("");
                     lines.Add(Strings.Program.SupportedCompressionModulesHeader);
-                    foreach (Duplicati.Library.Interface.ICompression mod in Library.DynamicLoader.CompressionLoader.Modules)
+                    foreach (Duplicati.Library.Interface.ICompressionInfo mod in Library.DynamicLoader.CompressionLoader.Modules)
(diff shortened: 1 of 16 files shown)


kenkendk:
Looks fine.

Do we still need `ICompression` ? It is now an empty interface, and I think there might be developers trying to use it and not getting why it does not work. If we just stick with `ICompressionInfo` there is less room for confusion, unless I have overlooked something?

vnau:
ICompression is not quite empty. It inherits ICompressionInfo, IArchiveReader and IArchiveWriter interfaces and still have to be implemented in archive modules.

I've slightly modified CompressionLoader so it would not load modules that is not implemented ICompression but only ICompressionInfo.

kenkendk:
I think that is what confuses me, because some places it loads `ICompressionInfo` and others `ICompression`:
https://github.com/duplicati/duplicati/pull/2978/files#diff-7063bc427787fc02b7793add679b631aR112

I was thinking that if you remove `ICompression` and then have the methods `IArchiveReader CreateReader(string path)` and `IArchiveWriter CreateWriter(string path)` on `ICompressionInfo`, then there is no need for having `IArchiveReader`/`IArchiveWriter` on `ICompression` ?

vnau:
I like this approach. The only lack is archive modules behavior will be a little bit more complex.

vnau:
Closing hanged PR. Would create a new one later taking in account the proposals.

PR #2979, opened 2018-01-25
Openstack backend loader changes

### LINKED ISSUE #2977: [Bug] OpenStack AuthAPI value is overridden if you edit configuration in UI
Version : 2.0.2.17_canary_2018-01-23
OS/Browser: Xubuntu 16.04 / Chromium

If you select a custom URL / OpenStack endpoint and go back on edit later the field is
replaced by Rackspace URI.

Pectojin:
This fixes #2977

I replaced some mildly confusing code in the Openstack backend loader. The previous code would always result in  being the default url, regardless of input.

The implementation is identical to the s3 implementation, which was much simpler.


kenkendk:
Yes, it seems simpler. 

I think I was trying to support a case where the default url was not listed, and in that case just use the first url as the default. But I can't remember why I thought that was a good idea :).

Pectojin:
Ah, that makes sense. 

It might be an idea for improvement of all backend loaders to have them fail over like that if the provider is removed from C#

But when fixing bugs I prefer to remove code instead of adding it :)

Issue #2904, opened 2017-11-21
Next scheduled execution can not be changed


johan12345:
- [ x ] I have searched open and closed issues for duplicates.

----------------------------------------

## Environment info

- **Duplicati version**: 2.0.2.1_beta_2017-08-01
- **Operating system**: Windows 10
- **Backend**: SFTP

## Description

My computer's clock was set to a wrong date for a short time a few weeks ago. It seems that Duplicati tried to start a backup during that time, because the backup logs show a corresponding entry.

Since then, my backup that is scheduled to be executed every day shows the next scheduled execution time to be in October 2018, and subsequently the backup does not run automatically every day, I have to start it manually instead.

When I try to change the next execution time to a different date (e.g. tomorrow), this change is not reflected on the home page (however, the chosen date is still shown in the backup settings when I try to edit them again). The backup is also not run at the time that I chose.

- **Actual result**:
It should be possible to change the next exceution time
- **Expected result**:
The next execution time does not change

kees-z:
I'm not 100% sure, but I suppose this is by design. If the latest backup is from a date in the future, all backups (until that last backup date) would become incremental backups that were made earlier than the backups the incremental is based on.

I'd suggest to delete all backup versions that have a timestamp in the future, using the Duplicati `delete` command.

johan12345:
Okay, what you described makes sense, but it seems that there is actually no backup from the future, according to the `list` command. There's just a log entry from 2018 that shows that the backup failed because the PC also had no network connection during that time. Probably there's just something saved in the local database about this unsuccessful backup?

kees-z:
You could try to recreate the database (delete and repair).
If that doesn't help, another option could be to export your backup configuration to a JSON file, delete the backup job (do NOT delete the remote files, but the local database can be deleted). Then import the JSON file.
Hope this helps.

kenkendk:
I think this error is with the scheduler. It should work if you change the time it runs and then save it.
Alternatively, try disabling scheduled backups completely, save, edit again, enable, save.

If this still gives no joy, try exporting to a file, delete the schedule section with a text editor, and import it.

johan12345:
@kenkendk Thanks! Just changing the time and saving did not work (as already said in the first post) but disabling scheduled backups, saving, and then re-enabling did fix the problem.

Pectojin:
I'm curious to hear peoples thoughts on built in handling of this issue. Time drift could realistically happen on any machine that loses connection to it's time server.

[...]

PR #2983, opened 2018-01-27
Scheduler recovery from timedrift

Pectojin:
Inspired by #2904 I tried playing a bit around with handling of timedrift.

I added this little piece of code in the scheduler to try and recover from a similar scenario.

In my tests it works well (I just exported some configs and changed the dates before re-importing), and will recover from scenarios where the last backup was performed in the future.

Of course it won't help if only the start time is in the future, but there just isn't any way to tell if the start time was affected by timedrift or intentionally set a year into the future.

On import, with future timestamps:
screen shot 2018-01-27 at 20 34 39
And after the run (files hadn't changed, so it didn't increment the versions number):
screen shot 2018-01-27 at 20 41 48
For testing purposes I had the interval at 6 min.

### DIFF
--- a/Duplicati/Server/Scheduler.cs
+++ b/Duplicati/Server/Scheduler.cs
@@ -245,6 +245,12 @@ namespace Duplicati.Server
                         
                         try
                         {
+                            // Recover from timedrift issues by overriding the dates if the last run date is in the future.
+                            if (last > DateTime.UtcNow)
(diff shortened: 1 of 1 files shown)


kenkendk:
It is an odd situation, so short of asking the user what to do, I think your fix is probably the most sane way of handling it.

PR #2985, opened 2018-01-27
Change order how backup retention options are applied

TekkiWuff:
This change applies the `--keep-versions` option **after** both the` --keep-time` and `--rentention-policy` option have been applied.

Before this change the `--keep-versions` option might delete more versions than necessary, because it didn't take into account, that the other options might already have deleted some versions.


kenkendk:
Great work on that, so now the rule is:
1. Remove older than `--keep-time` (infinite by default)
2. Apply retention policy
3. Remove if there are more than `--keep-versions`

Right now we block using `--keep-time` and `--keep-versions` at the same time, because some may want "or" while others want (read: expect) "and". 

But I can see that both these options can work well with `--retention-policy`, and it does not cause confusion.

Issue #2994, opened 2018-01-30
System.OverflowException: Negating the minimum value of a twos complement number is invalid. 


OronDF343:
- [x] I have searched open and closed issues for duplicates.

----------------------------------------

## Environment info

- **Duplicati version**: 2.0.2.17_canary_2018-01-23
- **Operating system**: Windows 7 Pro x64
- **Backend**: Local

## Description

Every since updating to the new version, I can't back up anymore. See the first log below. Retrying results in the second log, and resolving that issue (repair/recreate db or delete files) results in the first one again...
I have a feeling that the log isn't telling the full story, but I have yet to find any way to directly read duplicati's debug log anywhere in the documentation.

P.S. Some operations can't be cancelled (such as "create bug report")
P.P.S. Is multi-threading being considered? Duplicati is really slow because it uses just a single thread at 100%...

## Debug log

Log 1:
```
Operation Put with file duplicati-id0c64c64036643f2841f428e855ebbce.dindex.zip.aes attempt 1 of 5 failed with message: Negating the minimum value of a twos complement number is invalid.
System.OverflowException: Negating the minimum value of a twos complement number is invalid.

   at System.Math.AbsHelper(Int64 value)

   at Duplicati.Library.Utility.Utility.FormatSizeString(Int64 size)

   at Duplicati.Library.Main.BackendManager.DoPut(FileEntryItem item)

   at Duplicati.Library.Main.BackendManager.ThreadRun()

Renaming "duplicati-id0c64c64036643f2841f428e855ebbce.dindex.zip.aes" to "duplicati-i8d6f25bbee3541518c42c2adea985544.dindex.zip.aes"

Operation Put with file duplicati-i344e1d7c11a74fe39ddcc3e76bc877c3.dindex.zip.aes attempt 1 of 5 failed with message: Negating the minimum value of a twos complement number is invalid.
System.OverflowException: Negating the minimum value of a twos complement number is invalid.

   at System.Math.AbsHelper(Int64 value)

   at Duplicati.Library.Utility.Utility.FormatSizeString(Int64 size)

   at Duplicati.Library.Main.BackendManager.DoPut(FileEntryItem item)

   at Duplicati.Library.Main.BackendManager.ThreadRun()

Renaming "duplicati-i344e1d7c11a74fe39ddcc3e76bc877c3.dindex.zip.aes" to "duplicati-iaab8e9917cee42179895afbdc532fbd9.dindex.zip.aes"

[...]

PR #2995, opened 2018-01-30
Use double instead of long for Utility.FormatSizeString

TekkiWuff:
This _should_ fix the issue described in #2994.

Using `double` instead of `long` for `FormatSizeString` should help, since floating point numbers have a seperate sign bit, thus being able to represent the same range for positve and negative numbers. And since the method is only used for printing rounded numbers anyway, the potential loss of precisions shouldn't matter.

I couldn't test this myself though, since I cannot reproduce the error, but `Math.Abs(double.MinValue)` works fine.

This does NOT change all the calls to this method, some of which still force a cast to `long` when calling it. Should maybe done in a seperate pull request?

### DIFF
--- a/Duplicati/Library/Utility/Utility.cs
+++ b/Duplicati/Library/Utility/Utility.cs
@@ -586,19 +586,19 @@ namespace Duplicati.Library.Utility
         /// </summary>
         /// <param name="size">The size to format</param>
         /// <returns>A human readable string representing the size</returns>
-        public static string FormatSizeString(long size)
+        public static string FormatSizeString(double size)
         {
-            long sizeAbs = Math.Abs(size);  // Allow formatting of negative sizes
+            double sizeAbs = Math.Abs(size);  // Allow formatting of negative sizes
             if (sizeAbs >= 1024 * 1024 * 1024 * 1024L)
-                return Strings.Utility.FormatStringTB((double)size / (1024 * 1024 * 1024 * 1024L));
+                return Strings.Utility.FormatStringTB(size / (1024 * 1024 * 1024 * 1024L));
             else if (sizeAbs >= 1024 * 1024 * 1024)
-                return Strings.Utility.FormatStringGB((double)size / (1024 * 1024 * 1024));
+                return Strings.Utility.FormatStringGB(size / (1024 * 1024 * 1024));
             else if (sizeAbs >= 1024 * 1024)
-                return Strings.Utility.FormatStringMB((double)size / (1024 * 1024));
+                return Strings.Utility.FormatStringMB(size / (1024 * 1024));
             else if (sizeAbs >= 1024)
-                return Strings.Utility.FormatStringKB((double)size / 1024);
+                return Strings.Utility.FormatStringKB(size / 1024);
             else
-                return Strings.Utility.FormatStringB(size);
+                return Strings.Utility.FormatStringB((long) size); // safe to cast because lower than 1024 and thus well within range of long
         }
 
         public static System.Threading.ThreadPriority ParsePriority(string value)


JonMikelV:
Thanks for the comment on the FormatStringB() long cast being safe!  :-)

kenkendk:
I don't think using a double is the right solution, but it does work with a minor precision loss (53 bits for the actual value).
If the calling code is passing in `long.MinValue` then the output is going to be wildly incorrect (`8192.0 PB`).

Do one of you know why it is passing in that value? 

The method should certainly not crash on invalid input, but is it more "clean" to fix it by checking for `long.MinValue` explicitly?

JonMikelV:
Since I didn't realize there would be a precision loss then I agree we're better off doing the `long.MinValue` check.

As for why it's coming in at all, I'm not sure - though everybody I've asked so far seems to be using "local" NTFS destination.

kenkendk:
The precision loss is probably not visible as the value is condensed anyway.

@TekkiWuff I just noticed why you were going for the `double`, it is because we are trying to format a double anyway (from the speed in BackendManager).

I think I have figure out why we get the error in the first place:
https://stackoverflow.com/questions/22757239/double-maxvalue-to-integer-is-negative

It seems that the typecast returns `long.MinValue` if the value is too small (i.e. less than `0`).

I propose that we fix this by adding an overload like:
```csharp
public static string FormatSizeString(double value)
{
    return FormatSizeString((long)Math.Round(value)));
}
```

That will ensure that it is always an integer value we are converting, and thus never get `long.MinValue`.

Does that sound fine?

kenkendk:
It looks like I did not fully understand how the overflow works:
```
var f = 0.01f;
Console.WriteLine((long)f);
0
Execution Time: 0 second(s)
var f = -0.01f;
Console.WriteLine((long)f);
0
var f = ((float)long.MaxValue)+1 ;
Console.WriteLine((long)f);
-9223372036854775808
```

So it truncates the value automatically, and only when it is outside the range that can be represented with a long, it returns `long.MinValue` (the standard says "unspecified value").

The call is `size / seconds`, where seconds is in the range '[0-n[` and size is `[-1-m[`. 
If the copy operation is really fast, the `seconds` part can be a small fraction of a second, causing the speed to be extremely high. (in theory it could also crash with a "divide by zero" error).

Knowing that, should we perhaps add a "limit" parameter, such that really large values are filtered and displays something like "unknown" ?

TekkiWuff:
Intersting discorvery, that it defaults to `long.minValue` when the cast from `double` to `long` overflows.

But then both the size and the speed have to be very extreme values.
This:
```c#
var oneTeraByte = 1000000000000;
var veryShortPositiveDuration = TimeSpan.FromTicks(2);
Console.WriteLine((long) (oneTeraByte / veryShortPositiveDuration.TotalSeconds));
> 5000000000000000000
```
is still within the range of `long`. Only changing it from 2 to 1 ticks ([the smallest unit of time in C# according to MSDN](https://msdn.microsoft.com/en-us/library/system.timespan.ticks(v=vs.110).aspx)) will finally overflow it.

Interestingly dividing ANY value with `TimeSpan.FromTicks(0).TotalSeconds` results in an overflow as well, and NOT in a divide by zero
```C#
Console.WriteLine((long) (1 / TimeSpan.FromTicks(0).TotalSeconds));
> -9223372036854775808
```

At the moment I have no idea though, where such extreme values would come from.

As @JonMikelV noted, everybody with that problems has a local to NTFS partition as target.
While I also have such a target, mine is on a slow HDD. Maybe not using encryption, having an SSD / Optane drive and having source / temp on the same partition as the target leads to such short durations.

Anyway, as to how to to fix the error in the formatting code: I think your first idea with an explicit check for `long.minValue` should be fine as well, if you don't like the use `double` there. It's probably reasonable to assume, that 8192.0 PB both in size and (actual) speed per second wont be seen any time soon ... but don't quote me on that ;)
I'd probably just set it to `long.maxValue` manually, instead of having to come up an error message that potentially also has to be translated. That one byte difference isn't showing up anyway, when the value formatted. ^^
```C#
if (value == long.minValue) {
    value = long.maxValue; // mitigate error when using Math.Abs on long.minValue
}
var absValue = Math.Abs(value);
...
```

JonMikelV:
Since this is now part of the next Experimental milestone (https://github.com/duplicati/duplicati/milestone/9) where did it leave off - fixed with `double` or still needing an update to check for `long.minValue`?

kenkendk:
We can go for the `double` method as in this PR, but I want to add a check to the downloader as well, such that really fast transfers do not output `8000 TB/s`.

kenkendk:
I probably over-engineered it, but I have a fix in 7cd5af7

TekkiWuff:
Looks good :)
Guess this PR is obsolete then.

Hope the people who had the problem will post some screenshots with their actual size and duration values if possible. I'm really curious, where this overflow came frome.

kenkendk:
Hmm, my fix removes the speed information from the display if the duration is less than 100ms.
I will merge this in, and we can see what people report, and decide if we should exclude it from display or not.

Issue #2961, opened 2018-01-12
Bad handling of "" in commandline arguments in Duplicati.WindowsService


mnaiman:
- [x ] I have searched open and closed issues for duplicates.

----------------------------------------

## Description
Attached screenshot shows process tree when Duplicati started as Windows service with following parameters:
"c:\Software\Duplicati\Duplicati\Duplicati.WindowsService.exe" --portable-mode --webservice-sslcertificatefile="C:\ProgramData\letsencrypt-win-simple\httpsacme-v01.api.letsencrypt.org\Default Web Site-all.pfx" --webservice-sslcertificatepassword=""

During spawning new processes "" are disappering making commandline arguments wrong and Server process failing.

## Screenshots
![image](https://user-images.githubusercontent.com/2264442/34873789-cd0ca5fe-f796-11e7-86aa-ad7b1f8e2c9b.png)

mnaiman:
first and second processes are swapped

mnaiman:
Hi @kenkendk I think that commit https://github.com/duplicati/duplicati/commit/7e466e11cf896868e824811f1eeb10039bb7dff3 di not solve issue.

When I run:
c:\Software\Duplicati\Duplicati\Duplicati.WindowsService.exe install --portable-mode --webservice-sslcertificatefile="C:\Software\letsencrypt-win-simple\certs\Default Web Site-all.pfx" --webservice-sslcertificatepassword=""

Result in registry (Computer\HKEY_LOCAL_MACHINE\SYSTEM\CurrentControlSet\Services\Duplicati) for duplicati service is:
"c:\Software\Duplicati\Duplicati\Duplicati.WindowsService.exe" --portable-mode "--webservice-sslcertificatefile=C:\Software\letsencrypt-win-simple\certs\Default Web Site-all.pfx" "--webservice-sslcertificatepassword="

But it should be:
"c:\Software\Duplicati\Duplicati\Duplicati.WindowsService.exe" --portable-mode --webservice-sslcertificatefile="""C:\Software\letsencrypt-win-simple\certs\Default Web Site-all.pfx""" --webservice-sslcertificatepassword=""""""

mnaiman:
It fixed problem during running service - now it is ok.
Install service is still not providing good amount of double quotes :-)

mnaiman:
[...]

PR #3001, opened 2018-02-02
Fix 2961 - Bad handling of "" in commandline arguments

mnaiman:
This commit https://github.com/duplicati/duplicati/commit/b42ab42e57ac8683422fcfd8f0c96a548caa0a19 introduce new functionality to escaped "". But as mentioned in #2961 arguments like:
--webservice-sslcertificatefile="C:\ProgramData\Default Web Site-all.pfx"
--webservice-sslcertificatepassword=""
are striped from "" completely because of wrapping whole argument to "".

Function ProperParseArgs is handling that in correct way on Windows.

Needs to be tested on other systems @kenkendk please review this PR as you are author of that commit, that I didnt broke any of your intended functionality.

OF-TOPIC:Part of code is marked as changed, but is same, because of Windows/Unix line ends in code.


Duplicati/Library/Utility/Utility.cs
```diff
@@ -1548,11 +1548,8 @@ public static string WrapCommandLineElement(string arg, bool allowEnvExpansion)
                 // and sadly it expands only if the variable exists
                 // making it even rarer and harder to diagnose when
                 // it happens
-                arg = arg.Replace("\"", "\"\"");
```
kenkendk: I think removing this will cause many other issues.

Duplicati/Library/Utility/Utility.cs
```diff
@@ -1572,5 +1569,29 @@ public static string WrapAsCommandLine(IEnumerable<string> args, bool allowEnvEx
...
+                if (argsBuilder[i] == '"')
+                    inQuote = !inQuote;
+                else if (argsBuilder[i] == ' ' && !inQuote)
+                {
+                    argsBuilder[i] = '\n';
+                }
+
+            // Split to args array
+            return argsBuilder.ToString().Split(new char[] { '\n' }, StringSplitOptions.RemoveEmptyEntries).Skip(1);
+        }
```
kenkendk: This will not work, because arguments can contain newlines (on some systems, even filenames can contain newlines and tabs).

Duplicati/Library/Utility/Utility.cs
```diff
@@ -1572,5 +1569,29 @@ public static string WrapAsCommandLine(IEnumerable<string> args, bool allowEnvEx
...
+        {
+            var argsBuilder = new StringBuilder(commandLine);
+            var inQuote = false;
+
+            // Convert the spaces to a newline sign so we can split at newline later on
+            // Only convert spaces which are outside the boundries of quoted text
+            for (var i = 0; i < argsBuilder.Length; i++)
+                if (argsBuilder[i] == '"')
+                    inQuote = !inQuote;
+                else if (argsBuilder[i] == ' ' && !inQuote)
```
kenkendk: Arguments can be separated by multiple whitespace characters, including tabs, so this does not work.

kenkendk:
I have tried to reproduce this, and in my tests, setting something like `--test=""` always parses as `--test=`.

Only if I do `"--test="""` can I get `--test="`, and only in Windows.

But I was thinking that maybe the problem is also this line:
https://github.com/duplicati/duplicati/blob/master/Duplicati/WindowsService/Program.cs#L50

The installer does not do anything clever to escape the commandline arguments here?

mnaiman:
If --test="" is parsed to --test= that is bad, because then if it contains path with space if will not work. (like --webservice-sslcertificatefile="C:\ProgramData\Default Web Site-all.pfx")

Maybe escaping in installer could be way how to solved this problem.

mnaiman:
Ok, I tried with following 
"c:\Duplicati\Duplicati.WindowsService.exe" --portable-mode --webservice-sslcertificatefile="""C:\ProgramData\Default Web Site-all.pfx""" --webservice-sslcertificatepassword=""""""

Was translated to
"c:\Duplicati\Duplicati.Server.exe" --ping-pong-keepalive=true --portable-mode --webservice-sslcertificatefile="C:\ProgramData\Default Web Site-all.pfx" --webservice-sslcertificatepassword=""

kenkendk:
Getting closer, this is also bad:
https://github.com/duplicati/duplicati/blob/master/Duplicati/Service/Runner.cs#L60

kenkendk:
In the commit 7e466e1 I changed the two places where the commandline was assembled by joining the arguments with spaces. In my test, this correctly passes the arguments to both the server (via  `Duplicati.WindowsService.exe install`) and also when just running `Duplicati.Service.exe`.

PR #3016, opened 2018-02-07
Feature: Quota warning/error

tygill:
This change adds a warning when the amount of available quota reaches a configurable percentage of the total backup size, and an error when the available quota reaches 0. By default, the warning is sent when the available quota is less than 10% of the backup size.

It also changes/fixes console logging when using the --verbose flag, so that both the verbose and regular diagnostics info is reported (without this fix, some data which is presented in the regular diagnostics info, e.g., quota info, is not presented when using the --verbose flag).


kenkendk:
Great work!

PR #3017, opened 2018-02-07
Windows default filters don't need to be rooted

### LINKED ISSUE #3015: VSS: access denied for "System Volume Information"

## Environment info

- **Duplicati version**:  2.0.2.17_canary_2018-01-23
- **Backend**: FTP

## Description
I have activated VSS to backup opened files too. My backup source is the whole drive D:\ .
The system folder "System Volume Information" should be ignored.

## Steps to reproduce

- set "snapshot-policy" to "Required".
- Two test iterations - the result is the same!
    - set Default filter to: none (no checkbox is activated) 
   - set Default filter to: all (all checkboxes are activated)
- set backup source Volume Drive D:\ 

- **Actual result**:
Error message:
Failed to process path: D:\System Volume Information\1{xxx8876b-c176-4xxx-b7ae-04046e6ccxxx} => (5) Zugriff verweigert: [\\?\GLOBALROOT\Device\HarddiskVolumeShadowCopy2\System Volume Information\1{xxx8876b-c176-4xxx-b7ae-04046e6ccxxx}],
- **Expected result**:
Directory "System Volume Information" is automatically excluded, as mentioned in https://github.com/duplicati/duplicati/search?utf8=%E2%9C%93&q=%22System+Volume+Information%22&type=

## Workaround
Set own filter "Exclude directories whose name contains" to "System Volume Information", then it works.

tygill:
This change updates the Windows default filters so that none of them assume the folder/file they reference is at the root of a drive.

This should make them more robust both on Windows (e.g., if they are on a network share, or mounted under a different folder).
It should also make them useful on non-Windows (e.g., backing up a Windows drive on Linux or Mac).

This should fix #3015.


tygill:
As was mentioned in #3015, this may not actually be necessary, and might cause some files/folders to be skipped / removed from backups which wouldn't be warned. I might keep some of these changes, but I'll likely revert most of them.

tygill:
I'm closing this pull request, in large part due to the additional discussions raised in the thread linked to #3030.

PR #3022, opened 2018-02-08
made the side menu fixed so it follows as you scroll

Pectojin:
We were discussing this on the forum and it's a super simply change, so here it goes.

I fixed the menu and then added padding to the content wrapper so that it's identical to how it looked previously, except now it scrolls :)

Only full sized settings were changed. It didn't seem to have any impact on the mobile layout.

### DIFF
--- a/Duplicati/Server/webroot/ngax/less/style.less
+++ b/Duplicati/Server/webroot/ngax/less/style.less
@@ -677,6 +677,7 @@ body {
                 width: 260px;
                 padding-left: 40px;
                 float: left;
+                position: fixed;
 
                 > ul {
                     > li {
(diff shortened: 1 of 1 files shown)


kenkendk:
What are the odds... I just noticed that earlier today and commited just now:
https://github.com/duplicati/duplicati/commit/39b429f6fde24fa9ad8b5a2c700633fffb10e6df

Pectojin:
Hehe, they're pretty similar, I should have been quicker ;)

Does the margin 260px produce the same distance from the menu? I picked 300 for the padding because the menu is exactly 260 wide + 40 margin.

kenkendk:
I think 300 is correct, because I forgot to factor in the `padding`, but used the `margin` to move it to the right instead.

Issue #3047, opened 2018-02-19
Encryption (and other) Dropdowns empty when using wrong language


BurninLeo:
## Environment info
- **Duplicati version**: 2.0.2.19
- **Operating system**: Win10
- **Backend**: SFTP

## Description

I have removed the recent version (did not configure any backups there) and installed a fresh 2.0.2.19 canary. My Windows is in German. When clicking "Add backup", I see this:

![grafik](https://user-images.githubusercontent.com/15062876/36393044-37351f94-15ae-11e8-9cd4-50c38429c04c.png)

A repair-installation did not change anything. Then I wanted to watch the live log, and see what a new browser window (right-click on "Add backup" -> open in new window) shows me:

![grafik](https://user-images.githubusercontent.com/15062876/36393086-60d91210-15ae-11e8-8bb8-ca63c8da02ee.png)

So I tried a "reload page" in the original tab, an suddenly it works.

It also works, when I load the URL directly.

My assumption is that the problem relates to the "Missing X... token" that I initially got after opening Duplicati after the installation. I cannot replicate that by closing the tray app, and starting again.

(1) I guess that the error message about the token crashes loading the translations. I also guess that's something easy to fix.

(2) It's quite confusing, having no dropdown options when the "wrong" translation is used. Any idea why? Are they loaded after the rest of the page?

Pectojin:
Hmm, the `Missing XSRF Token` issue should have been resolved in 2.0.2.19 - #2699

BurninLeo:
I had assumed that the page would automatically reload, when I confirmed the "Missing XSRF Token" message?! I didn't have the problem since the reload, so it may be a caching issue from the old version.

To me it looked like if the script for the translation just did not start/run after the token problem, however I was unable to track this in the developer tools. Even if the "Missing XSRF Token" should not appear in the latest update (so often) any more, it's confusing that the system recognizes a problem (shows the message) but does not resolve it (e.g., reload the page after "OK" is pressed), but fails in some parts of the GUI while others seem to work properly.

Pectojin:
Ah, yes it does not reload when clicking OK. In general when that error happens you have to reload manually. I agree it's misleading. It wasn't intentionally designed this way, but that's just how general frontend errors are displayed.

When the token is missing it can't load any new resources, so it won't load the translations either like you saw.

Did this happen on the first page load after the upgrade? I *think* it could happen on the first load after upgrading to the patched version, but it should **never** happen again after. 

[...]

PR #3048, opened 2018-02-19
added clarification to  error to help users resolve the problem

Pectojin:
In accordance with discussion in #3047, I'd like to expand a little on the `Missing XSRF Token` error to improve user experience.


kenkendk:
Not sure why it does not reload automatically, but this is better than the previous message.

Issue #2972, opened 2018-01-19
Problems with name resolution on all backups


jpschewe:
- [x] I have searched open and closed issues for duplicates.

----------------------------------------

## Environment info

- **Duplicati version**: 2.0.2.1_beta_2017-08-01
- **Operating system**: Linux
- **Backend**: Nextcloud/webdav

## Description

I setup a backup job to reference my webdav server and it worked great for a couple of weeks. Starting about a week ago I started getting errors and they all say NameResolutionFailure. I checked and the hostname properly resolves. 

Is there a way to get more details on what host the server is trying to talk to? Perhaps my settings got corrupted? The UI still shows the correct host address.

I'm running duplicati as a service on Linux.

```
Operation List with file  attempt 5 of 5 failed with message: Error: NameResolutionFailure
System.Net.WebException: Error: NameResolutionFailure
  at Duplicati.Library.Main.BackendManager.List () [0x00038] in <118ad25945a24a3991f7b65e7a45ea1e>:0 
  at Duplicati.Library.Main.Operation.FilelistProcessor.RemoteListAnalysis (Duplicati.Library.Main.BackendManager backend, Duplicati.Library.Main.Options options, Duplicati.Library.Main.Database.LocalDatabase database, Duplicati.Library.Main.IBackendWriter log, System.String protectedfile) [0x0000d] in <118ad25945a24a3991f7b65e7a45ea1e>:0 
  at Duplicati.Library.Main.Operation.FilelistProcessor.VerifyRemoteList (Duplicati.Library.Main.BackendManager backend, Duplicati.Library.Main.Options options, Duplicati.Library.Main.Database.LocalDatabase database, Duplicati.Library.Main.IBackendWriter log, System.String protectedfile) [0x00000] in <118ad25945a24a3991f7b65e7a45ea1e>:0 
  at Duplicati.Library.Main.Operation.BackupHandler.PreBackupVerify (Duplicati.Library.Main.BackendManager backend, System.String protectedfile) [0x00066] in <118ad25945a24a3991f7b65e7a45ea1e>:0 
```

## Debug log

Turning on the live logging at level Information shows the following error:
```
[...]

PR #3049, opened 2018-02-19
Dns resolution failure

Pectojin:
Initial solution to #2972

We'll need to create `DNSName` methods for each backend at some point, but no reason to delay this fix until all backends are implemented.

### DIFF
--- a/Duplicati/Library/Backend/AlternativeFTP/AlternativeFTPBackend.cs
+++ b/Duplicati/Library/Backend/AlternativeFTP/AlternativeFTPBackend.cs
@@ -450,6 +450,8 @@ namespace Duplicati.Library.Backend.AlternativeFTP
             }
         }
 
+        public string DNSName => throw new NotImplementedException();
+
         private static System.IO.Stream StringToStream(string str)
         {
             var stream = new System.IO.MemoryStream();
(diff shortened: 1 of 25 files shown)


kenkendk:
Should the non-implemented ones perhaps return `null` instead, so we do not get nonsense warnings when trying to refresh? So if `DNSName == null` we don't try the DNS refresh?

kenkendk:
> GetHostEntry throws an error if it fails to find the DNS name, even when it's valid input.

Yes, but I was thinking that those who do not implement the feature, does not have to raise an additional error?

Specifically, the file backend will never implement it (granted, it might not throw `NameResolutionError`), so if the check was like this it would not risk problems:

```csharp
try
{
    if (!string.IsNullOrWhiteSpace(m_backend.DNSName))
        System.Net.Dns.GetHostEntry(m_backend.DNSName);
}
catch
{
    Logging.Log.WriteMessage("Failed to resolve IP for " + m_backend.DNSName, Logging.LogMessageType.Warning);
}
```

Btw, the error handler crashes as well, if you ever get here as it accesses `DNSName` too.

Pectojin:
Oh... silly me... I didn't think about the error handler.

I'll update them to return null and add some checks.

Pectojin:
Changed from `NotYetImplemented` to return null.

I also removed the logging case because I realized it was totally redundant.
screen shot 2018-02-21 at 22 23 32

Issue #3030, opened 2018-02-13
Revise default backup filter UI


Pectojin:
Now that backup filters actually work when defined in the Web GUI we should revise the default filters to make sure other users aren't hit by similar filter issues as I was:
https://forum.duplicati.com/t/release-2-0-2-19-canary-2018-02-12/2455/6

Additionally I think we should improve the UI to display more information on what the default filters actually do.

I think it's a show stopper for an experimental release to risk many people upgrading and not discovering that parts of their backups are now filtered aggressively.

renestach:
I completely agree that we should explain more what the filters actually do. Do you think a short description is sufficient like "These filters exclude system files from your backup. Please check the content of your backup when one of the filters is activated."

Pectojin:
I'm not sure. It seems to be a little more complicated than just being system files https://forum.duplicati.com/t/release-2-0-2-19-canary-2018-02-12/2455/14
That description wouldn't let me know that all of `/var` including `/var/www` isn't backed up. 

There are multiple rules that I think are questionable in this list and even after fixing those it may still be an issue about what people think of each rule.
![screen shot 2018-02-13 at 20 57 06](https://user-images.githubusercontent.com/1561484/36170776-a4d91e90-1100-11e8-8c4b-5bfa90711795.png)

The UI doesn't have to display them in-line, but I think it's unfair to require a user to make a full backup and check the contents of all folders before they can "know" what the filter does.

renestach:
I cannot remember when we introduced this and what the use case behind those OS-specific filters was. Were the introduced so that a simple backup of / or C:\ would turn into a backup that only contained user data? Or is their purpose to remove sneaky files from the backup like the famous 'desktop.ini' on Windows?

Before we revise the filters, we have to define what the exact use case is. If we have that definition already, please tell me :-)

tygill:
I was the one to initially add it, and my initial goal was mostly to remove things that aren't really important to be backed up (caches, temporary files, etc.), and to prevent errors attempting to back up system files that don't need to be backed up which might not be normally accessible (e.g., System Volume Information, $RECYCLE.BIN). One specific example is things like the Google Chrome cache in the AppData folder - I want to backup my entire user folder, but don't need to keep a running update of what happened to be in my Internet cache that day.

Originally, I was also thinking that system files (OS, program files, etc.) should be skipped, but especially after the recent discussions I'm much more inclined towards not filtering those (and letting the user choose to back them up by selecting those folders as sources instead).

[...]

PR #3052, opened 2018-02-20
Remove some default filters for things that should probably still be backed up

tygill:
This is a temporary improvement to the default filters, at least until further UI improvements (as discussed in #3030) can be made.


kenkendk:
I have made an issue for deciding what to do: #3056
I will keep this PR open, until we branch off the experimental.

PR #3059, opened 2018-02-21
Easier syntax for unlimited timeframes and intervals

TekkiWuff:
As discussed in https://forum.duplicati.com/t/new-retention-policy-deletes-old-backups-in-a-smart-way/2195/65 and before:

This allows the user to specify `U` for both the timeframe and interval to hopefully  make some edge cases easier to understand:
- `1W:U`: "For one week keep unlimited (=all) backups”. Internally it works the same as specifying 0s for the intervall, which is still valid.
- `U:1W`: "For unlimited time keep a backup at the interval of one week". This replaces the need for having to specify a very long timeframe like 99Y if you don't want old backups to get deleted. It is always applied after the last rule, so for example if there is a rule `99Y:2W`, then that one will be used before the `U:1W`.

I also refactored the handling of the retention policy option to use a list of custom objects instead of a dictionary to reduce confusion when working with them.

_Since this change doesn't render the current syntax invalid but only expands it, it should be OK to add this change after the next experimental has been released, in case you don't want to delay that one with more and more features._

### DIFF
--- a/Duplicati/Library/Main/Controller.cs
+++ b/Duplicati/Library/Main/Controller.cs
@@ -919,9 +919,10 @@ namespace Duplicati.Library.Main
             {
                 foreach (var configEntry in m_options.RetentionPolicy)
                 {
-                    if (configEntry.Value >= configEntry.Key)
+                    if (!configEntry.IsKeepAllVersions() && !configEntry.IsUnlimtedTimeframe() &&
+                        configEntry.Interval >= configEntry.Timeframe)
                     {
-                        throw new Interface.UserInformationException("A time frame cannot be smaller than its interval");
+                        throw new Interface.UserInformationException("An interval cannot be bigger than the timeframe it is in");
                     }
                 }
             }
--- a/Duplicati/Library/Main/Operation/DeleteHandler.cs
+++ b/Duplicati/Library/Main/Operation/DeleteHandler.cs
@@ -214,8 +214,8 @@ namespace Duplicati.Library.Main.Operation
         private List<DateTime> ApplyRetentionPolicy(DateTime[] backups)
         {
             // Any work to do?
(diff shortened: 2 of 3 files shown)


Pectojin:
<3 <3 <3 👍  💯  🥇
Can't wait to test this

Pectojin:
By the way, we may need to revise the text for `--retention-policy` in advanced options and for the `Custom backup retention` form with this patch.
screen shot 2018-02-21 at 22 59 27

Pectojin:
So far so good. With `30m:U,U:30m` I currently, after 7 backups within 42 minutes, have 5 backups (1 per 6 min for 30 min) and 1 backup older than 30m (so 1 per 30 min).

I have a script constantly inserting text into a file which is then backed up every 6 minutes, so I'll let it run for a while and check that everything looks good tomorrow.
screen shot 2018-02-21 at 23 48 52

TekkiWuff:
> By the way, we may need to revise the text for `--retention-policy` in advanced options and for the `Custom backup retention` form with this patch.

True! I'd probably simply change the example to include both the `U:xx` and `xx:U` rule and change the explanation accordingly, hoping that people understand it ... but I'm open for better suggestions ;)

> I have a script constantly inserting text into a file which is then backed up every 6 minutes, so I'll let it run for a while and check that everything looks good tomorrow.

Hehe, that is exactly my sophisticated test as well. Just with short retention rules so i don't have to wait weeks.
But I'm aware that actual unit tests would probably be better ...

Pectojin:
>But I'm aware that actual unit tests would probably be better ...
It's been so long since I wrote unit tests. I really don't want to start now ;)

Pectojin:
Everything still looks good. So short of unit testing I'd say this works exactly as I expected 👍

PR #3064, opened 2018-02-22
adds python script to change the encryption

wjansenw:


kenkendk:
You can actually do this already with:
https://github.com/duplicati/duplicati/blob/master/Duplicati/CommandLine/RecoveryTool/help.txt

You need the `--reencrypt --reupload` options set.

wjansenw:
From reading the help file, it does not allow me to go eg. from aes to gpg?
That's the reason I wrote the script, but it is also possible to remove
encryption, change pwd, or go from gpg to aes, ...

Op ma 26 feb. 2018 om 09:08 schreef Kenneth Skovhede <
notifications@github.com>:

Issue #2844, opened 2017-10-24
Setting password for the first time causes never ending failed request loop


darksworm:
- [x] I have searched open and closed issues for duplicates.

----------------------------------------

## Environment info

- **Duplicati version**: 2.0.2.1
- **Operating system**: Windows 7 and Windows 10
- **Backend**: N/A

## Description

When first setting a password, when logging in an error pops up: `Failed to get nonce: Too many active login attempts`. The issue is resolved after restarting Duplicati. 

## Steps to reproduce

1. Install Duplicati (or delete duplicati folder from Appdata)
2. Set password in settings
3. Try to log in

- **Actual result**:
An error pops up: `Failed to get nonce: Too many active login attempts`
- **Expected result**:
User successfully logged in

## Screenshots

image-1
![2017-10-24_12-44-29](https://user-images.githubusercontent.com/9987548/31936174-832513b6-b8b9-11e7-822b-edfced2bcae7.gif)
image-2
![qzcbxji](https://user-images.githubusercontent.com/9987548/31936256-b995c6f2-b8b9-11e7-9a1d-4b8d70e0e251.png)

## Notes
After changing the password there seem to be many failed requests to `/login.cgi` and they don't seem to stop (see image-2) but I cannot find where they originate from.
If you already had a password and you unset it, then set it again - this problem does not occur.

SandiaScalpel:
This happens to me using Linux Mint as well.

janmadera:
Same here using Windows. Just set the password and I'm out. How can I login again?

TheFatherMind:
This bug has been around for a while now.  Considering that it totally breaks the product I am surprised it has not been resolved yet.

janmadera:
I was able to get in again by restarting the application and connection using localhost instead of a hostname url from remote. I went directly to settings and disabled password auhentication - now I‘m fine again. 

This workaround is another bug.

fiver22:
Same; on Ubuntu  16.04. Duplicati version 2.0.2.1_beta_2017-08-01

SandiaScalpel:
I agree with @janmadera that the workaround is a bug as well. You certainly
do not want to install as root and enable remote access in the settings
unless you know that have tight control of your port forwarding and trust
everyone on your local network.

wgstarks:
Ran right into this issue after installing on a mac. Localhost workaround was no good since I was already using localhost to start with. Had to run the uninstall.sh script, reboot and then reinstall the app to get back in. Not good.

[...]

Issue #3035, opened 2018-02-14
Login impossible after setting GUI password


BurninLeo:
## Environment info
- **Duplicati version**: 2.0.2.19
- **Operating system**: Windows 10
- **Backend**: SFTP

## Description

The new version 2.0.2.19 suggests to set a passwort if using a multi-user environment. It opens the appropriate form in the GUI. After setting a password, a "login failed" message is displayed. Then the login form is displayed (with the password already stored?!).

When I confirm the password or re-type and then confirm, again, a message windows is shown: "Failed to get nonce: Too many active login attempts." The error seems to persist for a few minutes, now...

This error occured after I tried to run the auto-update (which failed, I assume due to insufficient access rights), and then running the MSI to install the new canary version.

Pectojin:
I believe #2844 is relevant.
Along with:
https://forum.duplicati.com/t/failed-to-get-nonce-too-many-active-login-attempts-after-setting-user-interface-password/2055
https://forum.duplicati.com/t/login-fails-after-setting-password-mac-osx/1339

It's an issue where the Web UI fails to detect that it needs a password to get resources from the server and instead of then prompting the user to login it just keeps asking the server for the resources until the server locks out the Web UI.

I believe you need to close the open tabs in your browser and restart the Duplicati server.
Alternatively you can start `--webservice-password=newpassword` to set a new password or `--webservice-password=` to disable the password.

BurninLeo:
Closing the browser and restarting Duplicati allowed me to login correctly. After that, a new logout and login worked fine. Good, thanks :)

I would apprechiate this being fixed with the next update. You feel some kind of locked out, otherwise.

Pectojin:
Yep, it's on the map but both @kenkendk  and I had problems reproducing the problem 2 weeks ago when we last looked into it (with 2.0.2.17).

I'll do some tests with 2.0.2.19 when I get home.

BurninLeo:
I also couldn't replicate this within the running 2.0.2.19 (tried to change, remove, and re-set the password).

How can I trigger the message that a password is suggested in a multi-user environment? It may be related to that.

As a workaround, you may consider to add a note: "If this error persists, please quit and restart Duplicati" to the error message "Failed to get nonce: Too many active login attempts." - not a great solution, but extremely helpful for people who are locked out after an update (or whatever conditions).

Pectojin:
Hmm, interesting. You may be right that it's related to that multi-user message.

[...]

PR #3066, opened 2018-02-25
fixes password change in hosted tray

mnaiman:
fixes password change in hosted tray https://github.com/duplicati/duplicati/issues/3035


Pectojin:
This also resolves #2844

Thanks! 👍

Pectojin:
I just ran through the scenario on my machine and can confirm I can no longer reproduce the issue after applying this fix. It does what it says on the tin.

And it looks like the original `NOHOSTEDSERVER_OPTION` and `READCONFIGFROMDB_OPTION` options still override the new default behavior. Although I haven't tested that use case.

PR #3070, opened 2018-02-27
Add a note in the repair-error popup where to find the repair action.

### LINKED ISSUE #2118: "Please run repair" - how do I do that?
I have:
- [X] searched open and closed issues for duplicates

### Version info

**Duplicati Version:** 2.0.1.30_experimental_2016-11-03
**Operating System:** Win10
**Backend:** AWS S3

### Bug description
```
Error while running <MyBackup>
Found 1 remote files that are not recorded in local storage, please run repair
```

I couldn't find out what I should do..

Sauermann:
resolves #2118 
Please double check this commit, since I don't have a development environment installed and I am currently unable to test the change.

### DIFF
--- a/Duplicati/Library/Main/Operation/FilelistProcessor.cs
+++ b/Duplicati/Library/Main/Operation/FilelistProcessor.cs
@@ -96,7 +96,7 @@ namespace Duplicati.Library.Main.Operation
 
             if (extraCount > 0)
             {
-                var s = string.Format("Found {0} remote files that are not recorded in local storage, please run repair", extraCount);
+                var s = string.Format("Found {0} remote files that are not recorded in local storage, please run repair (Backup-Options -> Advanced -> Database -> Repair)", extraCount);
                 log.AddError(s, null);
                 throw new Duplicati.Library.Interface.UserInformationException(s);
             }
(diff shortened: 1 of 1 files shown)


kenkendk:
I like the idea, but if we change that text, it will also show for people using the commandline, and I think this would add confusion for commandline users.

Not sure how we should implement this, but maybe we should add some kind of tag (maybe a special exception type?) such that there is an actual link on the popup that you can click to go to the right place?

Sauermann:
@kenkendk thanks for the response. Just tried a simple solution for the missing repair-hint. Sorry, but I will not be able to dig deeper into this, so feel free to reject the pull request. Best wishes!

Filip9696:
Maybe instead say `"Please run database repair"` It won't confuse CLI people while helping users like me who had to google it.

kenkendk:
> Maybe instead say "Please run database repair" It won't confuse CLI people while helping users like me who had to google it.

Yes, that makes sense. I am in the process of rewriting the log/exception system, and I hope this will turn out such that we can have the appropriate buttons in the error dialogs.

kenkendk:
This was fixed with #3095

PR #3086, opened 2018-03-07
Add option to disable backup when on battery power

warwickmm:
This adds an option to disable the backup when on battery power.  This can be used to preserve battery life by avoiding potentially resource intensive backup operations.  Note that only scheduled backups are skipped.  Manually run backups (e.g., by clicking "Run now" in the UI) will still be run.

To be conservative, if the detected power source is `PowerSupply.Source.Unknown`, the backup will still be run.

This addresses issue #2666.

### DIFF
--- a/Duplicati/Server/Scheduler.cs
+++ b/Duplicati/Server/Scheduler.cs
@@ -34,6 +34,8 @@ namespace Duplicati.Server
     /// </summary>
     public class Scheduler
     {
+        private static readonly string LOGTAG = Duplicati.Library.Logging.Log.LogTagFromType<Scheduler>();
+
         /// <summary>
         /// The thread that runs the scheduler
         /// </summary>
@@ -278,7 +280,18 @@ namespace Duplicati.Server
                                 {
                                     var entry = Program.DataConnection.GetBackup(id);
                                     if (entry != null)
-                                        jobsToRun.Add(Server.Runner.CreateTask(Duplicati.Server.Serialization.DuplicatiOperation.Backup, entry));
+                                    {
+                                        Dictionary<string, string> options = Duplicati.Server.Runner.GetCommonOptions(entry, Duplicati.Server.Serialization.DuplicatiOperation.Backup);
+                                        Duplicati.Server.Runner.ApplyOptions(entry, Duplicati.Server.Serialization.DuplicatiOperation.Backup, options);
+                                        if ((new Duplicati.Library.Main.Options(options)).DisableOnBattery && (Duplicati.Library.Utility.Power.PowerSupply.GetSource() == Duplicati.Library.Utility.Power.PowerSupply.Source.Battery))
+                                        {
+                                            Duplicati.Library.Logging.Log.WriteInformationMessage(LOGTAG, "BackupDisabledOnBattery", "Scheduled backup disabled while on battery power.");
+                                        }
+                                        else
+                                        {
+                                            jobsToRun.Add(Server.Runner.CreateTask(Duplicati.Server.Serialization.DuplicatiOperation.Backup, entry));
+                                        }
+                                    }
                                 }
                             }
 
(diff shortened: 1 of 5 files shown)


Duplicati/Server/Scheduler.cs
```diff
@@ -278,7 +278,19 @@ private void Runner()
...
                                     var entry = Program.DataConnection.GetBackup(id);
                                     if (entry != null)
-                                        jobsToRun.Add(Server.Runner.CreateTask(Duplicati.Server.Serialization.DuplicatiOperation.Backup, entry));
+                                    {
+                                        Dictionary<string, string> options = Duplicati.Server.Runner.GetCommonOptions(entry, Duplicati.Server.Serialization.DuplicatiOperation.Backup);
+                                        Duplicati.Server.Runner.ApplyOptions(entry, Duplicati.Server.Serialization.DuplicatiOperation.Backup, options);
+                                        if ((new Duplicati.Library.Main.Options(options)).DisableOnBattery && (Duplicati.Library.Utility.Power.PowerSupply.GetSource() == Duplicati.Library.Utility.Power.PowerSupply.Source.Battery))
+                                        {
+                                            // TODO: Can we log this somewhere?  It would be nice to provide a way for the
+                                            // user to know why the backup wasn't run.
```
warwickmm: I've made several unsuccessful attempts at logging this case.  Any ideas?  It would be nice to provide a way for the user to know why the backup wasn't run.

warwickmm:
A nice feature to have would be for any skipped backups to run immediately upon switching to AC power.  That's beyond the scope of this pull request, but I'll create a new issue if this gets merged in.

Duplicati/Server/Scheduler.cs
kenkendk: You can write to `Duplicati.Library.Logging.Log.WriteInformationMessage()`, we just need to make sure it doesn't spam the log.

I am rewriting the log system, so we can present it in a nice way later.

kenkendk:
I guess we should also support this from the CLI? Maybe just throw an exception in the `Controller.cs` file if the run is disabled? Alternatively, we should warn if the option is set and not being respected.

Duplicati/Server/Scheduler.cs
warwickmm: Thanks.  I had tried that earlier but was looking at the wrong logs when testing.  Added the logging statement in revision f3ca95763f16dfe939dbfe95fa9baeb1c49db219.

warwickmm:
Sorry, I tried several things in `Controller.cs`, but could not find a way to determine whether the backup was triggered by the scheduler, by the user clicking "Run now", or by the command line.  I was unable to find a way to throw an exception or log something without it affecting the scheduled backups as well.  Any help here would be much appreciated.

My original thinking was that this option should only affect scheduled backups, as clicking "Run now" or using the command line feel more manual and are a stronger indication that the user really does want the backup to be run.  If there's no simple solution, would it suffice to better document that this only affects scheduled backups?  For example, we could make it clearer in `Strings.DisableOnBatteryLong` that backups run by clicking "Run now" in the UI or the command line will not respect this option?

warwickmm:
Sorry, please ignore my previous comment.  I realized that scheduled backups do not run through `Controller.ValidateOptions`.

I added a warning in revision 3256174b915d5286f736543cdf4e5a694c4c2514 to inform users that the `disable-on-battery` option is ignored when the backup is triggered manually or from the command line.  I also clarified the behavior in the documentation.

kenkendk:
Not sure what you are saying... the `Controller.ValidateOptions` should be called for any kind of operation, regardless of what started it.

I was suggesting to move the check out into the `Duplicati.CommandLine` project, so it is checked/warned there.

warwickmm:
Yes, you're right.  I was being a bit sloppy in my testing.  Sorry about that!  :flushed:

I moved the check to the `Duplicati.CommandLine.Commands.Backup` method.  I also rebased this branch onto `master` to pick up the new logging methods and clean up the history a bit.

kenkendk:
👍

PR #3087, opened 2018-03-08
Use nameof operator when constructing ArgumentNullException

warwickmm:
This will make rename refactorings easier in the future.


kenkendk:
Yes, I wrote the code before the `nameof` operator was introduced :)

PR #3092, opened 2018-03-08
Made python script run in parallel

piegamesde:
This made the process a lot faster for me 15h -> 7h.


wjansenw:
Cool! My 2nd open source addition ever already gets improved (and used!!). Love it.

Pectojin:
Travis does that sometimes. It also shouldn't matter because you didn't change any code that has to be compiled :)

wjansenw:
I had the same issue. As it's only a script, all should be fine. You didn't break it :).

Wiki looks good as well! Tnx, should have thought about that.

Maybe we need to add a Windows version.

kenkendk:
Yeah, Travis often has timeouts of various kinds, annoying, but better than not running tests

PR #3095, opened 2018-03-09
added repair button to the 'please repair' notification

### LINKED ISSUE #2118: "Please run repair" - how do I do that?

Pectojin:
As discussed in https://forum.duplicati.com/t/how-to-please-run-repair/2771 it's not very obvious how to repair the database to new users. 

I Added a button to make it easy.
screen shot 2018-03-09 at 20 14 46

This also resolves #2118

### DIFF
--- a/Duplicati/Server/webroot/ngax/scripts/directives/notificationArea.js
+++ b/Duplicati/Server/webroot/ngax/scripts/directives/notificationArea.js
@@ -37,6 +37,11 @@ backupApp.directive('notificationArea', function() {
             );
         };
 
+        $scope.doRepair = function(backupid) {
+        AppService.post('/backup/' + backupid + '/repair');
+        $location.path('/');
+        };
+
         $scope.doInstallUpdate = function(id) {
             AppService.post('/updates/install');
         };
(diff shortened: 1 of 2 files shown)


kenkendk:
With the new logging stuff I added, we can remove the dependency on the string (as it may be localized).

There is now a `MessageID` on the notification, and we should show the button for both `ExtraRemoteFiles` and `MissingRemoteFiles`.

I am merging anyway, because it still works and adds value for users.

Pectojin:
Sounds good. We'll need to go through all the buttons anyway since a couple of them use text matching as well.

As usual I just stole some code and refactored it for this use-case.

Pectojin:
@veleek I'm assuming this is on the beta version? 

The code is in the canary build since version 2.0.3.2 but the only beta released since then was 2.0.3.3 which was released shortly after so the fix didn't make it into that release.

veleek:
Correct, the beta version.  There was some confusion on the other feature request thread so I thought I'd bring it up.  I'll follow up on the original issue.  Thanks!

PR #3096, opened 2018-03-09
added 'UserName' and 'StartedBy' variables to the system info page

Pectojin:
As discussed in https://forum.duplicati.com/t/solved-headless-server-access-to-all-files/2736 I wanted to add a "UserName" variable to make it easier to determine which user the server is running under.

Additionally I wanted to display how the server was started (stand alone or tray icon) to make it easier to identify issues with running standalone servers and tray icon servers at the same time.


kenkendk:
Great idea!

PR #3108, opened 2018-03-18
Make fields containing lock objects readonly

warwickmm:
If one of these fields is accidentally reassigned, it's possible for threads to be oblivious to an existing lock.  By making the fields `readonly`, we will be notified at compile-time if we inadvertently redefine one of these fields.


kenkendk:
Cool! I wrote that code way before I got into using readonly :)

PR #3110, opened 2018-03-18
Fix Appveyor build by removing inlined out variable

warwickmm:
This replaces an inlined `out` variable with the older declaration syntax.  Inlined `out` variables are a newer C# language feature that are not supported by the current Appveyor build, which is currently using MSBuild 14.0.


warwickmm:
Oops, there are a few more C# 7 language features to fix.  I'll try and correct them and re-open this pull request.

Issue #3107, opened 2018-03-17
Some character "{}" in filename trigger  System.NullReferenceException


mneiger:
- [X ] I have searched open and closed issues for duplicates.

## Environment info
- **Duplicati version**: 2.0.3.2
- **Operating system**: Linux Mint 18
- **Backend**: Webdav

## Steps to reproduce
trying to purge a file containing a { or } in filename trigger exception.
example : {94b89ca6-23ec-45c0-a648-0f30d5a2f74d}.xpi

more information at
https://forum.duplicati.com/t/system-nullreferenceexception-and-system-formatexception-trying-to-purge/1892/7
## Debug log

    marc@resh-lakish ~ $ sudo duplicati-cli purge 'webdavs://o*********t.stackstorage.com//remote.php/webdav/resh-lakish/home?auth-username=o********t&auth-password=*************' '--passphrase=*********'  --dbpath=/root/.config/Duplicati/KNQLJCGAWY.sqlite --verbose --dry-run=true  '--include=*/Storage/Users/marc/VirtualBox VMs/*'
            Input command: purge
            Input arguments: 
            	webdavs://o**********t.stackstorage.com//remote.php/webdav/resh-lakish/home?auth-username=o*********t&auth-password=**********************

            Input options: 
            passphrase: *************************
            dbpath: /root/.config/Duplicati/KNQLJCGAWY.sqlite
            verbose: 
            dry-run: true

            Starting purge operation
              Listing remote folder ...
            Replacing fileset duplicati-20171227T180000Z.dlist.zip.aes with duplicati-20171227T180001Z.dlist.zip.aes which has with 16 fewer file(s) (9.79 GB reduction)

            System.NullReferenceException: Object reference not set to an instance of an object
              at Duplicati.Library.Main.Database.LocalDatabase.WriteFileset (Duplicati.Library.Main.Volumes.FilesetVolumeWriter filesetvolume, System.Data.IDbTransaction transaction, System.Int64 filesetId) [0x000b4] in <118ad25945a24a3991f7b65e7a45ea1e>:0 
              at Duplicati.Library.Main.Operation.PurgeFilesHandler.DoRun (Duplicati.Library.Main.Database.LocalPurgeDatabase db, Duplicati.Library.Utility.IFilter filter, System.Action`3[T1,T2,T3] filtercommand, System.Single pgoffset, System.Single pgspan) [0x003d6] in <118ad25945a24a3991f7b65e7a45ea1e>:0 
[...]

PR #3111, opened 2018-03-18
Added "{0}" to log calls in PurgefilesHandler and RestoreHandler.

VennGrayDog:
Bug fix for #3107


kenkendk:
I am not sure how this problem arises. The code does not call `Format` if there are no arguments:
https://github.com/duplicati/duplicati/blob/master/Duplicati/Library/Logging/LogEntry.cs#L115

Usually the problem would be something with a mis-formatted string, but here it is a null exception.

I also just tried to inject the curly brackets like this:
```csharp
Library.Logging.Log.WriteProfilingMessage(LOGTAG, "Test", "My {filename}.txt");
```

And this does not trigger any errors. Do you know where the problem happens?

VennGrayDog:
You are right. It should work even without "{0}". I missed that part of the code.

PR #3112, opened 2018-03-18
Fix Appveyor builds by removing use of C# 7 language features

warwickmm:
This should fix the Appveyor builds by removing the use of C# 7 language features.  The current Appveyor build uses MSBuild 14.0, which does not support these newer C# 7 language features.


warwickmm:
Alternatively, we can configure Appveyor to use MSBuild 15.0, but I wasn't sure how to do that.

kenkendk:
I updated the AppVeyour config to use VS2017, and that seems to have fixed the build issues.

warwickmm:
Thanks!  I'll close this since the changes aren't needed anymore.

Issue #2210, opened 2016-12-28
OneDrive for Business: OAuth support [$10]


xko:
I was happy to learn that duplicati supports od4b. But with requirement to store password, I'm still hesitant to use it (and I believe many others as well). If only duplicati was able to use [OAuth mechanism](https://dev.onedrive.com/auth/aad_oauth.htm), that would truly enable the  feature. 

I think for the beginning there's enough to introduce advanced options for necessary paramenters (auth code, app key etc..). There are [tutorials](https://www.sepago.com/blog/2016/10/05/Access-OneDrive-for-Business-with-PowerShell-as-an-Azure-AD-App) online about generating these without needing 3rd party hosted app.

Thaurin:
I was about to create an issue for this myself, before I found this one. I agree wholeheartedly. The od4b support is a perfect solution for my personal and work backup needs, but I cannot responsibly use it when the password is stored in plain text.

This is an older issue, but this needs to get some love and priority. I'll see if I can up the bounty to raise interest, or I may look into the issue myself when I have time.

xyzlabc:
Sorry if this isn't useful, but rclone has an implementation here that might be useful as a reference:
https://github.com/ncw/rclone/blob/2f63a9f81cd5fae993be6f5fa57c705ddd1da465/backend/onedrive/onedrive.go

I can't code so I don't know if this is useful, so sorry if not.

ts678:
This can be closed as done, right? See @tygill link above. I don't know how bounties are handled though.

warwickmm:
Yes, I believe this is done.  I'm not sure exactly how the bounties work now, but @tygill can claim it.

tygill:
Well, I don't really know how the bounty system works, but I've just tried to submit the claim there...we'll see how it goes...

Issue #3101, opened 2018-03-13
Office 365 and 2FA?


zappeee:
- [x] I have searched open and closed issues for duplicates.

----------------------------------------

## Environment info
- **Duplicati version**: 2.0.2.1
- **Operating system**: Windows 10
- **Backend**: OneDrive for Business

## Description
We recently got 2FA on our Office365 accounts.
Now when I try to backup I just get the error message of "No sharepoint web could be logged in to at..."
I tried to replace my password with an app password but got the same error.
Have anyone got this to work with 2FA?
Is there anywhere I can see in detail what the issue is?
I know there's been issues with O4B before so I'm not sure if this is 2FA or not

## Steps to reproduce
Activate 2FA on Office365, probably

- **Actual result**:
Error message as above
- **Expected result**:
Successful backup

## Screenshots
.

## Debug log
.

xyzlabc:
Would this have anything to do with #2210 ?

tygill:
Both this and #2210 might be solved with #3024, as using the Microsoft Graph APIs consolidates authentication for both personal and business OneDrive / SharePoint (as well as unifying the file API, so a single backend should be able to provide support for all of them). I've started looking at implementing this, and hopefully will have some more time to work on this in the next few weeks.

zappeee:
I see, thanks for the info!

So currently Duplicati is not compatible with Office 365 2FA?

wjansenw:
Maybe if you use rclone as backend? They haven't moved to the new API yet either I believe. But checking doesn't hurt :).

cmthomps:
I have 2FA authorization enabled for Office 365.  Duplicati will work if you set-up an App Password.  
[https://support.office.com/en-us/article/create-an-app-password-for-office-365-3e7c860f-bda4-4441-a618-b53953ee1183](url).   Obviously bypassing the 2FA auth, but it works.

zappeee:
@cmthomps, I tried that but it still gives me errors unfortunately.

tygill:
FYI, the OAuth service used by Duplicati has been updated to support MS Graph authentication, so I believe you should be able to use OneDrive, OneDrive for Business, or SharePoint sites with 2FA / OAuth tokens if you use the `OneDrive v2` or `SharePoint v2` backends..

(This morning my OneDrive for Business and SharePoint sites aren't actually working, but it appears unrelated to Duplicati since the MS Graph API Explorer tool is also having errors.)

zappeee:
[...]

Issue #3024, opened 2018-02-09
Old OneDrive API being deprecated - needs to be rewritten to use the Graph API


tygill:
According to https://docs.microsoft.com/en-us/onedrive/developer/rest-api/concepts/migrating-from-live-sdk, the old Live Connect API (which is used by the current OneDrive backend) is being deprecated and disabled on November 1st, 2018. By the time that happens, we should either have the new API supported.

There is a [Microsoft.Graph NuGet package](https://www.nuget.org/packages/Microsoft.Graph/) which supports this API. But it requires bringing in the [System.Net.Http NuGet package](https://www.nuget.org/packages/System.Net.Http/) as well, instead of using the standard .NET version (which may or may not be a problem for non-Windows platforms). It might be easier to just use raw JSON instead of the library.

It looks like the new API will require different permissions, meaning users will need to re-authenticate. One idea to push people towards this would be to keep the existing backend, and use the new API in a new backend (OneDriveV2?). Once the new backend is stable, we could log a warning or error when the old backend is used, telling users that they will need to switch to the new backend by November 1st.

Moving to the new backend might also help prevent problems caused by the need for the workaround currently in use in the OneDrive backend described here: https://forum.duplicati.com/t/backup-failing-with-key-already-exists-exception-on-onedrive/1157.

tygill:
FYI, I've started taking a look at this (though if someone else magically gets it done before me I won't be offended :) )

tygill:
Since this has been merged, and the OAuth service has now been updated to support Graph API authentication, I'm closing this as resolved. If there are more minor issues along the way, they can be tracked separately.

Issue #2855, opened 2017-10-29
System.ArgumentException: An item with the same key has already been added.


fkappen:
- [x] I have searched open and closed issues for duplicates.
 - Found some old posts, tried solutions but nothing helped. 
 - with empty Backend error doesn't occur but a full new Backup would take 280 hours to be   uploaded once (6mbit/s upload 😞 )
----------------------------------------

## Environment info
- **Duplicati version**: 2.0.2.12
- **Operating system**: Mac OS High Sierra
- **Backend**: OneDrive

## Description

Backup fails after 120 successful Backups with error: System.ArgumentException: An item with the same key has already been added. Key: XXXX...

Tried to recreate Database, but the same error occurs.

Tried to make an Update of Duplicati - Upgraded from latest Beta to latest canary. 

Tried to reinstall Duplicati - same error. 

Think its depending on Files at the backend. 

System.ArgumentException: An item with the same key has already been added.

- **Actual result**:

29. Okt. 2017 17:46: Failed while executing "Repair" with id: 1
System.ArgumentException: An item with the same key has already been added. Key: duplicati-20170829T194128Z.dlist.zip.aes
  at Duplicati.Library.Main.BackendManager.List () [0x00049] in <82db7c3af7d944569906ea5ff7f18985>:0 
  at Duplicati.Library.Main.Operation.RecreateDatabaseHandler.DoRun (Duplicati.Library.Main.Database.LocalDatabase dbparent, System.Boolean updating, Duplicati.Library.Utility.IFilter filter, Duplicati.Library.Main.Operation.RecreateDatabaseHandler+NumberedFilterFilelistDelegate filelistfilter, Duplicati.Library.Main.Operation.RecreateDatabaseHandler+BlockVolumePostProcessor blockprocessor) [0x00057] in <82db7c3af7d944569906ea5ff7f18985>:0 
  at Duplicati.Library.Main.Operation.RecreateDatabaseHandler.Run (System.String path, Duplicati.Library.Utility.IFilter filter, Duplicati.Library.Main.Operation.RecreateDatabaseHandler+NumberedFilterFilelistDelegate filelistfilter, Duplicati.Library.Main.Operation.RecreateDatabaseHandler+BlockVolumePostProcessor blockprocessor) [0x00032] in <82db7c3af7d944569906ea5ff7f18985>:0 
[...]

PR #3118, opened 2018-03-20
Add new backends for OneDrive + OneDrive for Business, SharePoint, and Microsoft Groups based on the Microsoft Graph API

tygill:
This change adds new backends based on the Microsoft Graph API:

- `OneDriveV2` (protocol key: `onedrivev2`) for OneDrive and OneDrive for Business (the Graph API treats them as pretty much the same thing)
- `SharePointV2` (protocol key: `sharepoint`, since the old one was `mssp`) for SharePoint sites hosted in Office 365
- `MicrosoftGroup` (protocol key: `msgroup`) for using the drive associated with an Office 365 Group.

Users on the old `OneDrive`, `OneDriveForBusiness`, and `SharePoint` backends should be able to migrate backups to the new backends, though they will need to update authentication parameters and likely change backup paths.

Each of these backends is based on the new `MicrosoftGraphBackend` base class, which supports streaming, renaming, and quota features in addition to the core backend interface.

Note that instead of using the standard `OAuthHelper` class (which most other REST based backends seem to use), this backend uses the built-in `System.Net.Http.HttpClient` with a new `OAuthHttpMessageHandler` which adds authentication to each request. This was done in part because I'm more familiar with `HttpClient`, but also because the Graph APIs use colon characters in several URLs, which caused errors in the `OAuthHelper` when attempting to parse them as URIs. To help with this, I've also added an extension method to emulate the `await` keyword in non-`async` code.

I've tested the `OneDriveV2` backend on both OneDrive (personal) and OneDrive for Business, and to help with that, I've added several new tests to the `BackendTester` for renaming, quota reporting, and DNS name listing. I haven't tested either the `SharePointV2` or `MicrosoftGroup`, mostly due to not have a good site or group to test them against. However, as most of the logic is shared, I believe they should work.

Before this change is merged, the partner pull on the [Duplicati OAuth Helper](https://github.com/duplicati/oauth-handler/pull/3) should be merged and deployed - otherwise the OAuth links for these backends won't be valid. (My testing so far has been done by explicitly passing in access tokens, rather than generating my own Auth-ID.) I'd like to test this with the OAuth handler end-to-end to make sure things are all working.

This should resolve:
#2210: OneDrive for Business: OAuth support
#3101: Office 365 and 2FA
#3024: Old OneDrive API being deprecated (though I still want to do a followup to log a warning for users with backups configured using the old backend once these new backends have had a chance to prove their reliability for a while, so that users have a chance to switch over to the new backends before the old ones stop working)

It might also resolve / workaround:
#2855 (though it looks like people in the [forum thread](https://forum.duplicati.com/t/backup-failing-with-key-already-exists-exception-on-onedrive/1157) have reported it as having gone away..)
#2830 (might be the same as the above, but the user said it affected 'OneDrive for Business', but not clear if its still the same bug)
...and possibly others.


tygill:
I just added a pull request to @kees-z 's Duplicati manual adding details about how to use these backends:
https://github.com/kees-z/DuplicatiDocs/pull/2

PR #3129, opened 2018-03-25
GUI-Extension for Backup-Job

dbddhkpde:
Tried to extend the GUI with a small progress bar within the backup job "details":
- shows current action
- shows current file
- displays an icon for the active job (paused, running) accordingly
- show / hide the GUI-extension (progress bar, filename, action) only if page-event contains a current filename


kenkendk:
Great! Thanks for that!

If you feel like it, we also have a request that queued items get a small hour-glass and perhaps a number that indicates the queue position.

kees-z:
Awesome!

Also scheduled (but not currently queued) items could get a small calendar/(stop)watch/rotating arrow symbol to indicate that the backup runs automatically:
image

Another idea:
Backup jobs that last completed unsuccessfully, could get some indication (change of (background) color, question/exclamation mark or something similar).

dbddhkpde:
Good idea: next to come is the icon of scheduled jobs.

Issue #2179, opened 2016-12-17
No Keystone v3.x support in OpenStack/Swift configuration


kafeinnet:
### Version info
**Duplicati Version:** 2.0.1.35
**Operating System:** Ubuntu 16.10

### Bug description
Duplicati does not support OpenStack/Swift storage with Keystone 3.x identity. There is no Keystone version choice or domain name configuration possible.

But duplicity support it with those env vars :
SWIFT_USERNAME=username
SWIFT_PASSWORD=password
SWIFT_AUTHURL=https://identity.url.com/v3
SWIFT_AUTHVERSION=3
SWIFT_USER_DOMAIN_NAME=domain_name
SWIFT_PROJECT_DOMAIN_NAME=domain_name

Keystone v2.x will be deprecated soon, I think it's a good time to begin the 3.x support.

### Steps to reproduce
- Add a backup
- Chose OpenStack/Swift as destination

epol:
Pull request #3154 partially solves this issue, it supports user/password authentication but not token based authentication.

duplicatibot:
This issue has been mentioned on **Duplicati**. There might be relevant details there:

https://forum.duplicati.com/t/openstack-ovh-discontinuing-keystone-2-0-api/9097/3

kenkendk:
I think this is fixed with #3154
If token-based auth is needed, please create a new issue.

PR #3154, opened 2018-04-05
Openstack keystone v3 support

epol:
Added support for the new keystone v3 API, the old v2.0 APIs are still available, the v3 can be selected with a version option.
Only password login is implemented here for API v3, so token login is still missing.

Keystone API v2.0 has been deprecated years ago and many new openstack installations don't support them while API v3 is the current version of the APIs. See also issue #2179

With this patch API v3 support is added, but the default is still v2.0 so backward compatibility is preserved.


epol:
In my opinion most warnings in the quality review aren't correct, for example https://app.codacy.com/app/kenneth/duplicati/file/16136819852/issues/source?bid=3742986&fileBranchId=6972408#l240 says the 'url' is not used, but it's used here: https://github.com/epol/duplicati/blob/master/Duplicati/Library/Backend/OpenStack/OpenStackStorage.cs#L442

Duplicati/Library/Backend/OpenStack/OpenStackStorage.cs
```diff
@@ -122,6 +227,32 @@ public OpenStackAuthRequest(string tenantname, string username, string password,
...
         }
 
+        private class Keystone3AuthResponse
+        {
+            public TokenClass token { get; set; }
+
+            public class EndpointItem
+            {
+                // 'interface' is a reserved keyword, so we need this decorator to map it
+                [JsonProperty(PropertyName = "interface")]
```
kenkendk: You can also use `public string @interface { get; set; }`

kenkendk:
Great work! Full-stack fix :)

The Codacity setup is not yet refined enough. We can disable some of the rules.

epol:
Codacy has been helpful to find some forgotten code, so maybe it's better to have some extra warning and take them with a grain of salt ;)

Thank you for accepting the PR and for the great work on this software!

Issue #3106, opened 2018-03-17
KB/s missing off full screen in firefox


Roadrunneruk:
- [x] I have searched open and closed issues for duplicates.

----------------------------------------

## Environment info

- **Duplicati version**: Duplicati - 2.0.3.2_canary_2018-03-16

- **Operating system**: windows 10 with firefox
- **Backend**: Replace me

## Description

When running a backup and  firefox is at maximum the KB/s is cut off the display bar, but when the screen is made smaller the full display is shown, see pictures
![max size](https://user-images.githubusercontent.com/9529362/37554407-da4b054c-29cf-11e8-9579-1c978826037a.JPG)
![reduce size](https://user-images.githubusercontent.com/9529362/37554408-da6498c2-29cf-11e8-9228-90864a0d14cd.JPG)

## Steps to reproduce

1. run in firefox
2. 
3. ...

- **Actual result**:
look at pictures
- **Expected result**:
look at pictures

SanduRajapakse:
Alright, seems like an easy fix. Not sure why a fixed width was set, but I will correct it.

SanduRajapakse:
I'll take a look at this when I get a chance. The fix is straightforward, but it has a negative effect on the pause and throttle button. I will see what I can do to minimize the effect.

SanduRajapakse:
I made the necessary changes to fix this issue in pull request #3162. 

This will:
1) Truncate the length of the name (in the state bar) to keep the state fully visible
2) Show the full name (as a title) if the user hovers over the name

Once merged, this problem should be resolved.

warwickmm:
Resolved in pull request #3162.

PR #3162, opened 2018-04-08
GUI - State Display Bar Improvements

SanduRajapakse:
This is to address a few issues with the state display bar including an issue that makes it difficult to see the current status due to the length of the backup name as backup names lengths are not limited.

Note: The CSS file must be recompiled with updated LESS file

This will:
1) Truncate the length of the name (in the state bar) to keep the state fully visible
2) Show the full name (as a title) if the user hovers over the name

This addresses issue: #3106
and implementation was discussed here: 
https://forum.duplicati.com/t/proposal-for-a-modification-to-the-state-display-bar-green-info-bar-at-the-top/3133/6


kenkendk:
Thanks!

Issue #1287, opened 2015-02-10
Absurd warnings since updating to 2.0.0.81


David-Pa:
This happens with all my backups:

Warnings: [The supplied option ----auto-cleanup is not supported and will be ignored, The supplied option ----compression-module is not supported and will be ignored, The supplied option ----thread-priority is not supported and will be ignored, The supplied option ----all-versions is not supported and will be ignored, The supplied option ----asynchronous-upload-limit is not supported and will be ignored, The supplied option ----debug-output is not supported and will be ignored, The supplied option ----snapshot-policy is not supported and will be ignored, The supplied option ----usn-policy is not supported and will be ignored]

kenkendk:
Is this from the GUI ?

Can you edit the advanced options? Do they contain four ---- marks instead of the two -- they should have?

David-Pa:
Yes, it's from the GUI.

The advanced options contain two -- as they should.
I tried to change them to be without -- at all, but after saving I opened the settings again, and the -- were back.

HaraldOnline:
Here too: Warnings: [The supplied option ----snapshot-policy is not supported and will be ignored]

David-Pa:
I can confirm this issue is resolved.
Thanks!

HaraldOnline:
Me too! Thanks!

Issue #3126, opened 2018-03-24
Red tray icon


vers-one:
- [x] I have searched open and closed issues for duplicates.

----------------------------------------

## Environment info
- **Duplicati version**: 2.0.3.2_canary_2018-03-16
- **Operating system**: Windows 10

## Description
During the first setup I made a mistake in the backup task configuration, so it failed on the first run and tray icon became red. After that I fixed the mistake. There were a few successful runs within the last couple of days but the tray icon still stays red even there are no errors in UI. I tried everything: starting the task manually, starting it on schedule, restarting the service, clicking on the all log items, repairing and rebuilding the database, but the icon is still red (when the task is idle) no matter what.

vers-one:
Just installed it again from scratch on a totally different PC (also Windows 10). This time I didn't do any mistakes in the backup task configuration, but the tray icon turned red after the first successful run anyway.

warwickmm:
Does [this suggestion](https://forum.duplicati.com/t/windows-10-tray-icon-always-red-failed/835/3) help?  Nevertheless, we still need to figure out what is causing this.

vers-one:
Yep, interrupting the running task and dismissing the error notification did help. Thank you.

I still have this situation on my second PC, I'll leave it as is and will try to debug it to find out what is causing it.

vers-one:
Just installed duplicati-2.0.3.2_canary_2018-03-16-x64.msi on Windows 10 virtual machine and immediately got a red tray icon. I think I know what is the problem.

During the first start the database file doesn't exist so it gets created and initialized via
`DataConnection.FixInvalidBackupId()` (https://github.com/duplicati/duplicati/blob/master/Duplicati/Server/Program.cs#L282)
which eventually calls `ServerSettings.SaveSettings()` (https://github.com/duplicati/duplicati/blob/master/Duplicati/Server/Database/ServerSettings.cs#L105)

However the `ServerSettings` class constructor sets all the settings values to `null` (https://github.com/duplicati/duplicati/blob/master/Duplicati/Server/Database/ServerSettings.cs#L74) before that which means that `unacked-error` parameter (https://github.com/duplicati/duplicati/blob/master/Duplicati/Server/Database/ServerSettings.cs#L47) will be empty in the database.

[...]

Issue #1295, opened 2015-02-11
2.0.0.81- 2.0.0.81_preview_2015-02-06: adding backup job fails with "Failed to get server setup..."


huste:
as noted. Just "add backup" and be persented with the error message popup
"Failed to get server setup..." Restaring duplicaty tray does not cure anything.

This happens on Win7 and Win8.1

kenkendk:
Is this with Internet Explorer ?

since1868:
Having the same issue after updating today. On Chrome. Here's error log:

    2/11/2015, 3:13:33 PM Request for get-backup-defaults gave error
    System.ArgumentException: Could not determine JSON object type for type       Duplicati.Server.Database.Setting.
    at Newtonsoft.Json.Linq.JValue.GetValueType(Nullable`1 current, Object value)
    at Newtonsoft.Json.Linq.JContainer.AddInternal(Int32 index, Object content, Boolean skipParentCheck)
    at Newtonsoft.Json.Linq.JContainer.AddInternal(Int32 index, Object content, Boolean skipParentCheck)
    at Duplicati.Server.WebServer.ControlHandler.GetBackupDefaults(IHttpRequest request, IHttpResponse response, IHttpSession session, BodyWriter bw)
    at Duplicati.Server.WebServer.ControlHandler.Process(IHttpRequest request, IHttpResponse response, IHttpSession session)
    2/11/2015, 3:13:33 PM Reporting error gave error
    System.ObjectDisposedException: Cannot write to a closed TextWriter.
    at System.IO.StreamWriter.Flush(Boolean flushStream, Boolean flushEncoder)
    at Duplicati.Server.WebServer.ControlHandler.Process(IHttpRequest request, IHttpResponse response, IHttpSession session)

huste:
No, both computers have the lastest Firefox version installed. And Opera 12.17 (yes, I know, the old one) shows the same.

commerxchef:
I'm having the same problem.

I created a backup and it failed to save many options. I edited the backup and it then was able to save the options.

I ran the backup for the first time and noticed it complain that having no encryption was not recognized but it still continued without it.

The next day I tired to add a new backup and I encountered the same error. I've since deleted that backup in hopes of resolving the problem but nothing changed. My logs show the same errors as above.

I'll try the work around posted by since1868.

since1868:
[...]

PR #3166, opened 2018-04-09
Fix server setting property accessors

warwickmm:
This modifies several of the `ServerSettings` property accessors.  The `IsFirstRun` and `HasAskedForPasswordProtection` accessors were simplified by removing redundant handling of null or empty values.  By default, the `Utility.ParseBoolOption` method returns `true` for these cases.

The `UnackedError`, `UnackedWarning`, and `ServerPortChanged` accessors were modified so that null, empty or whitespace, and unparsable values return `false`, which seems more appropriate for these settings.  When we previously returned `true` for these cases, we would encounter issues such as the red tray icon for first installs (see issue #3126).

### DIFF
--- a/Duplicati/Server/Database/ServerSettings.cs
+++ b/Duplicati/Server/Database/ServerSettings.cs
@@ -188,10 +188,6 @@ namespace Duplicati.Server.Database
         {
             get
             {
-                var tp = m_values[CONST.IS_FIRST_RUN];
-                if (string.IsNullOrEmpty(tp))
-                    return true;
-
                 return Duplicati.Library.Utility.Utility.ParseBoolOption(m_values, CONST.IS_FIRST_RUN);
             }
             set
--- a/Duplicati/Server/Database/ServerSettings.cs
+++ b/Duplicati/Server/Database/ServerSettings.cs
@@ -198,23 +198,19 @@ namespace Duplicati.Server.Database
             }
         }
 
-		public bool HasAskedForPasswordProtection
-		{
-			get
-			{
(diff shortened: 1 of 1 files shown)


warwickmm:
The logic in the `ServerSettings.FixedInvalidBackupId` [accessor](https://github.com/duplicati/duplicati/blob/master/Duplicati/Server/Database/ServerSettings.cs#L534) seems a bit strange:
```
get
{
    if (m_values.ContainsKey(CONST.HAS_FIXED_INVALID_BACKUPID) && string.IsNullOrWhiteSpace(m_values[CONST.HAS_FIXED_INVALID_BACKUPID]))
        return false;
    else
        return Duplicati.Library.Utility.Utility.ParseBoolOption(m_values, CONST.HAS_FIXED_INVALID_BACKUPID);
}
```
I think `m_values.ContainsKey(CONST.HAS_FIXED_INVALID_BACKUPID)` will always return `true`, since `m_values` is initialized in the `ServerSettings` constructor to contain all the fields from the `CONST` class.  If that's the case, then the remaining logic feels inconsistent, since we return `false` for null or whitespace values, and `true` for unparsable values.

Is this logic intentional?  Or, can it be simplified to something like:
```
return Duplicati.Library.Utility.Utility.ParseBool(m_values[CONST.HAS_FIXED_INVALID_BACKUPID], false);
```
which would return `false` for null, whitespace, or unparsable values.

I refrained from making this change, as it appears this was introduced to resolve issues #1287 and #1295 and I didn't want to risk introducing a regression.

kenkendk:
I think we should change `ServerSettings.FixedInvalidBackupId` as well.
I think I just added the check as a "just in case", but just calling the parse method seems much better.

PR #3169, opened 2018-04-10
Add filter groups to replace default filters

tygill:
This change revives the disabled 'default filters' feature as revised 'filter groups', as discussed and described in #3030. This includes two main changes to how default filters worked:

- Instead of using a special command line parameter (`--default-filters`), filter groups are specified via the regular `--include` and `--exclude` parameters by wrapping the group name in curly brackets, e.g.: `--exclude={DefaultExcludes}`.
- Instead of grouping filters by operating system, filters are grouped by type (cache files, temporary files, OS files, etc.), and different folders are selected depending on the current operating system (meaning you can't use Windows filters on Linux, but that shouldn't be a problem). Supported groups today are:
  - `SystemFiles`
  - `OperationSystem`
  - `CacheFiles`
  - `TemporaryFiles`
  - `Applications`
  - `DefaultExcludes` (includes all of the above)
  - `DefaultIncludes` (currently includes nothing, and is hidden in the UI at the moment - left mostly as a placeholder for the future in case we want to specify some default groups to backup for different operating systems)

Because filter groups are specified in the same way as other filters, they can be interleaved as expected (e.g., you can `--include` something before `--exclude`-ing a group to override something in the group).

The filters UI has also been updated to have a basic dropdown for the available groups:
image

The one piece of feedback discussed in #3030 that has not yet been implemented is exposing the full list of paths affected by each filter group somewhere in the UI (which would require plumbing the list from the service into the UI via a custom request somehow, which is beyond my UI abilities). However, since I things as they are are functional, and most of the group names are generally self-descriptive, I'd like to see if people are OK merging this in it's present state (mostly so the code doesn't get stale and forgotten).


Duplicati/Library/Utility/FilterGroups.cs
```diff
@@ -0,0 +1,702 @@
...
+        /// </summary>
+        /// <param name="group">The groups to create the filters for</param>
+        /// <returns>Common filters</returns>
+        private static IEnumerable<string> CreateCommonFilters(FilterGroup group)
+        {
+            if (group.HasFlag(FilterGroup.CacheFiles))
+            {
+                // TODO: The control_dir_v2 might be under a different path for OEM branded instances.
+                // However, the AppName is loaded and controlled by the AutoUpdater assembly, which we can't reference here without an ugly circular dependency or dependency injection.
+                // What is the best way to solve this?
```
kenkendk: Lets leave it for now. We can do it by reflection, loading the assembly at runtime, and extracting the path but I don't want to delay merging it just to support that.

tygill: Opened #3175 for this.

Issue #3167, opened 2018-04-10
Attribute IsFSCaseSensitive is expensive to get


ltfish:
- [x] I have searched open and closed issues for duplicates.

----------------------------------------

## Environment info

- **Duplicati version**: 2.0.3.2_canary_2018-03-16
- **Operating system**: Windows 10
- **Backend**: Local hard drive

## Description

I did a profiling using Intel VTune on Duplicati 2 for a backup run. During the initial file counting phase, the profiler tells me that `System::Environment::GetEnvironmentVariable()` takes ~24% of the execution time. Further investigation shows that `FilterEntry::Matches` calls `Utility::get_IsFSCaseSensitive`, which calls `GetEnvironmentVariable()` for every file path, and this is not a fast operation.

I think a quick and hackish solution is simply caching the value of `GetEnvironmentVariable("FILESYSTEM_CASE_SENSITIVE")` because it should not change after the program starts. I'm posting this issue to see a) if this is a real performance issue, and b) if this proposed solution is acceptable. If answers to both questions are yes, I'll go ahead and prepare a PR to address this issue. Thanks.

## Steps to reproduce

a. Perform a profiling on Duplicati 2 using the "Basic Hotspots" profile in Intel VTune.
b. Start a backup job in Duplicati 2.
c. Wait for a while, and then terminate the backup task when Duplicati 2 is still counting files.
d. Stop profiling and take a look at the profiling result.

- **Actual result**:
`Utility::get_IsFSCaseSensitive` shows up as a hotspot taking ~24% of execution time in the profiling result.
- **Expected result**:
`Utility::get_IsFSCaseSensitive` should not show up as a hotspot in the profiling result.

## Screenshot

![get_isfscasesensitive_slow](https://user-images.githubusercontent.com/5071575/38543003-3973282a-3c58-11e8-82a3-5aa23809fe16.png)

Pectojin:
This seems pretty sane. I don't see any reason to not change it.

PR #3172, opened 2018-04-11
Cache IsFSCaseSensitive to speed up file counting.

ltfish:
`IsFSCaseSensitive` used to call `GetEnvironmentVariable()` every time it
was accessed, which unnecessarily slows down the file counting procedure.
Caching the value of this environment variable speeds up things quite a
bit.

### DIFF
--- a/Duplicati/Library/Utility/Utility.cs
+++ b/Duplicati/Library/Utility/Utility.cs
@@ -36,6 +36,11 @@ namespace Duplicati.Library.Utility
         public static long DEFAULT_BUFFER_SIZE => SystemContextSettings.Buffersize;
 
         /// <summary>
+        /// A cache of the FileSystemCaseSensitive property, which is computed upon the first access.
+        /// </summary>
+        private static bool? CachedIsFSCaseSensitive = null;
+
+        /// <summary>
         /// Gets the hash algorithm used for calculating a hash
         /// </summary>
         public static string HashAlgorithm { get { return "SHA256"; } }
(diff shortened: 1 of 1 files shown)


Duplicati/Library/Utility/Utility.cs
```diff
@@ -35,6 +35,14 @@ public static class Utility
         /// </summary>
         public static long DEFAULT_BUFFER_SIZE => SystemContextSettings.Buffersize;
 
+        /// <summary>
+        /// Boolean value parsed from the environment variable FILESYSTEM_CASE_SENSITIVE,
+        /// or null if it is not a valid boolean string.
+        /// </summary>
+        public static bool? EnvVar_FileSystemCaseSensitive = Utility.ParseBool(
+            Environment.GetEnvironmentVariable("FILESYSTEM_CASE_SENSITIVE"), null
```
warwickmm: This appears to have triggered many failures in Travis and AppVeyor...

If the `FILESYSTEM_CASE_SENSITIVE` variable is not set, `GetEnvironmentVariable("FILESYSTEM_CASE_SENSITIVE")` will return `null`.  In this case, the call to `Utility.ParseBool` will then return the default value `null`.  This is semantically different than before, where the default value provided to `Utility.ParseBool` is determined by `Utility.IsClientLinux && !Utility.IsClientOSX`.

ltfish: Thanks for looking into this. It triggered failures because the code wasn't correct (I was misusing `Utility.ParseBool()`). However, I don't quite get why my change would change the semantics.

Now I switched to a different way of implementation: Fully caching the whole thing. I hope it will work :)

ltfish:
This is to address issue #3167 .

kenkendk:
Great find. Simple and clean to fix.

PR #3171, opened 2018-04-11
IGNORE - Testing Travis failures

warwickmm:
Possibly relevant links:
* `https://github.com/travis-ci/travis-ci/issues/6861`
* `https://github.com/smackers/smack/issues/279`
* `https://github.com/mono/mono/issues/7416`
* `https://github.com/mono/mono/issues/6205`
* `https://github.com/Microsoft/msbuild/issues/2685`
* `https://github.com/Microsoft/msbuild/issues/3161`
* `https://stackoverflow.com/questions/13510465/the-mystery-of-stuck-inactive-msbuild-exe-processes-locked-stylecop-dll-nuget`
* `http://blog.danskingdom.com/tag/msbuild-msbuild-exe-slow-hangs-build-compile/`

Notes:
* It appears that a call to `csc.exe` hangs, which results in a stalled build.  The failure doesn't seem tied to a particular project, as it has stalled when compiling different projects.
* Killing a background process (mono) as suggested in one of the above links did not help.
* Skipping starting the server and running the GUI tests did not help.
* Building in debug mode did not help.
* Passing `/maxcpucount:1` to `msbuild` did not help.
* Passing `/nodeReuse:false` to `msbuild` did not help.

PR #3177, opened 2018-04-13
Fix Travis build issues by limiting concurrent processes

warwickmm:
It seems that `msbuild` has issues when building with multiple concurrent processes.  Once we restricted the build by providing the `/maxcpucount:1` option and using a single job in the Travis build, we:

1. Obtained 5 consecutive successful builds.
2. We then removed the `/maxcpucount:1` option and immediately encountered a failed build.
3. Adding back the `/maxcpucount:1` option, we then obtained another 10 consecutive successful builds.

Prior to these changes, 6 out of 9 builds failed (with `msbuild` hanging) while experimenting with various options, along with numerous failures in previous pull requests, etc.

After limiting Travis to a single job, the build times remain approximately the same as before (22-24 minutes).

Note that similar to the AppVeyor configuration, we skip the "BulkData" tests.  These changes also enable the "Utility" tests, which were not explicitly included before.

See pull request #3171 for experimental details.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -25,10 +25,6 @@ env:
   general:
     - USAGEREPORTER_Duplicati_LEVEL=none
     - AUTOUPDATER_Duplicati_SKIP_UPDATE=1
-  matrix:
-  - JOB=job1
-  - JOB=job2
-  - JOB=job3
 notifications:
   webhooks:
     urls:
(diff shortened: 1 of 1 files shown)


kenkendk:
Do you have a guess as to why Travis would have trouble running multiple jobs? My understanding is that it runs each job in a separate VM, so it should be 100% isolated environments.

The reason AppVeyour does not run the bulk data tests, is the time limit.
Travis has a similar time limit, but splitting the run into multiple jobs enables us to multiply the time limit and then get enough time to also run the bulk data tests.

I would prefer if we could keep the bulk data tests. Alternatively, have some mechanism to selectively run manually with the bulk data tests.

warwickmm:
Yes, that is strange.  I will have to experiment some more.  I'll close this pull request and re-open it if I have any new findings.

PR #3181, opened 2018-04-16
Fix `SpecialFolder` default filters

tygill:
The way Duplicati matches folders against filters, folder filters need to either end with a wildcard or a slash in order to match.

However, the 'control_dir_v2', 'Windows.old', and the various Environment.SpecialFolders filters did not have trailing slashes, leading to these filters not actually applying when they should.
This change fix the few specific filters that need this specification, and adds a new utiltiy method to produce filters for SpecialFolders in a way that makes sure they end with a trailing slash.

### DIFF
--- a/Duplicati/Library/Utility/FilterGroups.cs
+++ b/Duplicati/Library/Utility/FilterGroups.cs
@@ -346,7 +346,7 @@ namespace Duplicati.Library.Utility
                     yield return windir;
 
                     // Also exclude "C:\Windows.old\"
-                    yield return windir.TrimEnd('\\', '/') + ".old" + System.IO.Path.DirectorySeparatorChar;
+                    yield return Utility.AppendDirSeparator(windir.TrimEnd('\\', '/') + ".old");
                 }
             }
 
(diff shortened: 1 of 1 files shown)


Duplicati/Library/Utility/FilterGroups.cs
```diff
@@ -554,6 +554,35 @@ private static IEnumerable<string> CreateLinuxFilters(FilterGroup group)
...
+            string folderPath = Environment.GetFolderPath(specialFolder);
+            if (!string.IsNullOrEmpty(folderPath))
+            {
+                string filter = FilterGroups.CreateWildcardFilter(folderPath);
+
+                // Duplicati matches filters against folder paths exactly.
+                // Meaning a filter for 'C:\Windows' won't match 'C:\Windows\'.
+                // So this makes sure special folder's filter's have a trailing directory separator.
+                // (Alternatively, this could append '*' to all folder filters.)
+                if (filter[filter.Length - 1] != System.IO.Path.DirectorySeparatorChar)
```
warwickmm: Would the [Utility.AppendDirSeparator](https://github.com/duplicati/duplicati/blob/master/Duplicati/Library/Utility/Utility.cs#L425) method be useful here?

tygill: Yes, I believe it would. I didn't notice that utility method previously.

PR #3199, opened 2018-04-29
Refactoring of parsing for command line options.

verhoek:
I reordered some parts of the commandline options parsing part, splitting basically one large function in somewhat smaller functions. I reduced some isHelp checks by moving that up front. I don't think I violated any logic changes here, but please check.


warwickmm:
The diff is a bit difficult to look at, but I think this looks good.

kenkendk:
I agree with @warwickmm it is hard to follow the diff, but yes, looks good.

PR #3213, opened 2018-05-11
Include version information in Results

drwtsn32x:
I made this change because I wanted the Duplicati version information included in email results.
Not sure if this is the best way to access the version information, but it works.


kenkendk:
I like it. This means it will also be stored in the database with the log results, which makes sense from a debugging perspective.

PR #3221, opened 2018-05-15
Webapi/googledrive

verhoek:
As a proof of concept to remove some hard coded strings in rest calls, I refactored the googledrive backend. As a result, the usual benefits: less repetition, easier implementation for upcoming integration tests or unit tests against various backends, some decoupling of googledrive's api and duplicati logic.
To support this some helper functions in Utility Uri were added performing basic url/uri operations and in addition some elementary unit tests were added.

Code has been tested by uploading and restoring files on googledrive.

If approved, I would continue adjusting other backends in a similar fashion.

Codacy reports some of its pre-existing code issues to this pull request due to reformatting or editing the line in some other way.


Duplicati/UnitTest/UriUtilityTests.cs
```diff
@@ -0,0 +1,69 @@
...
+
+using NUnit.Framework;
+using System.Collections.Specialized;
+
+namespace Duplicati.UnitTest
+{
+    public class UriUtilityTests
+    {
+        [Test]
+        [Category("UrlUtility")]
```
warwickmm: The Travis configuration requires that we explicitly add each test category to the build.  Would you mind modifying the [travis.yml](https://github.com/duplicati/duplicati/blob/master/.travis.yml) file to include the `UrlUtility` category?

verhoek:
I added the additional unit test category to Travis yaml. Since I am not that familiar with Travis, I'm unsure whether the category is added in the correct place - possibly better would be within the stage Run unit tests where category Border is placed, to parallelize.

warwickmm:
Travis created a separate job for the `UriUtility` tests, so looks good to me.  Thanks.

kenkendk:
Looks good :)

The point of the categories is that Travis limits each test to 40 minutes (+ some slack), so we cannot run all tests in a single run. But since there is no limit on the total time, we can add as many categories as we like. 

We can keep it as you have it now, but some of the small tests should be grouped, as the boot+build takes > 3 minutes and the test < 5 seconds.

Issue #3210, opened 2018-05-10
Recreate investigation - Blocklist blocks not found by recreate?


tygill:
tl;dr: Block for a blocklist not present in database after recreate from remote - original database lists this block as living in an `Index` volume rather than a `Blocks` volume (but the indicated `Index` volume contains no mention of the block's hash).

Several weeks ago, I attempted to recreate my database for a backup after migrating to a new backend (from `OneDrive` to `OneDrive v2`). However, after the repair finished (which ended up processing all `dlist`, `dindex`, and `dblock` files on the remote), the new database was still not usable - when running the `LocalDatabase.VerifyConsistency` check, it would fail with an error stating: `Found inconsistency in the following files while validating database`, and then listed several files which had non-zero `actual` sizes, but a `dbsize` of zero. ([This](https://forum.duplicati.com/t/found-inconsistency-in-the-following-files-while-validating-database/525) forum post and bug #3085 seem to reference this same error case.)

_In that forum post and bug, it is mentioned that deleting the BlocksetIDs associated with the failing files can solve this error, but when I tried that, it just changed the error to one indicating that certain filesets were now missing files, which makes sense as removing anything referencing those Blocksets effectively removes those files from any filesets, so the only way I see to making that workaround actually work would be to also remove all now-broken filesets and start over with only the existing blocks._

Since it looked like a bug that was hitting several other people, I decided to try to dig in to see if I could track down exactly what was causing the bug - was something in the repair / recreate process broken, or was something in my remote backup corrupted / incorrect. In that process, I found a few things that seem quite suspicious, though they have to do with parts of Duplicati that I'm not currently very familiar with, so I'm hoping that by walking through this process I can get some pointers on what some of this might mean. I have a copy of my original SQLite database, as well as the recreated one, which I've used to try to understand what might have happened.

The first thing I noticed was that in the recreated database, there were no entries in either the `BlocklistHash` or `BlocksetEntry` tables for the BlocksetIDs of the inconsistent files, though the `Blockset` table did have the entry (which listed the correct `Length` and `FullHash`). This is ultimately the cause for the consistency error - the length of each file is computed by summing the sizes of each block that makes up the corresponding blockset, but since there are no blockset entries mapping to actual blocks, it computes a `dbsize` of zero, which doesn't match the value in the `Blockset` table.

[...]

Issue #3202, opened 2018-05-01
Internal consistency check failed, generated index block has wrong hash


SamiLehtinen:
- [ ] I have searched open and closed issues for duplicates.
- [x] Only "quick browser", not thorough scan.

----------------------------------------

## Environment info

- **Duplicati version**: v2.0.3.6-2.0.3.6_canary_2018-04-23
- **Operating system**: Ubuntu 16.04.4 Linux & Windows 10
- **Backend**: SFTP & FTPS & FILE

## Description

Backup sets basically get broken by just running backup. Three totally independent systems, with different backends and clients. As well as one "local / file" backup, where backup data is stored on second local disk. Something is seriously wrong if this happens?

"Internal consistency check failed, generated index block has wrong hash"

## Steps to reproduce

1. Just running normal backup.

## Debug log
`
Checking remote backup ...
  Listing remote folder ...
Fatal error => System.Exception: Internal consistency check failed, generated index block has wrong hash, 6yil2I1aMnlEYfq7luTpJh8RytKe5Gm7eeUuk0UAELU= vs I5/lRw4QR1CWI8t2ojmy/pgUFgAUcQqr8z+zpybt3rk=
  at Duplicati.Library.Main.Operation.Common.IndexVolumeCreator+<CreateIndexVolume>d__0.MoveNext () <0x409eb6e0 + 0x00b5b> in <filename unknown>:0 
--- End of stack trace from previous location where exception was thrown ---
  at System.Runtime.ExceptionServices.ExceptionDispatchInfo.Throw () <0x7f58718016d0 + 0x00029> in <filename unknown>:0 
  at System.Runtime.CompilerServices.TaskAwaiter.ThrowForNonSuccess (System.Threading.Tasks.Task task) <0x7f58717ff6b0 + 0x000a7> in <filename unknown>:0 
  at System.Runtime.CompilerServices.TaskAwaiter.HandleNonSuccessAndDebuggerNotification (System.Threading.Tasks.Task task) <0x7f58717ff630 + 0x0006b> in <filename unknown>:0 
  at System.Runtime.CompilerServices.TaskAwaiter.ValidateEnd (System.Threading.Tasks.Task task) <0x7f58717ff5e0 + 0x0003a> in <filename unknown>:0 
  at System.Runtime.CompilerServices.TaskAwaiter 1[TResult].GetResult () <0x7f58717ff8d0 + 0x00017> in <filename unknown>:0 
[...]

PR #3225, opened 2018-05-17
Adds a check for BlocksetIDs that don't have any corresponding BlocksetEntrys to the broken fileset check.

tygill:
Bug #3210 (and several other possibly related ones) had cases where a recreated database would have these invalid blockset IDs. This should allow list-broken-files to find these corrupt filesets.

It seems like they were caused by filesets referring to a blockset by hash, but not having any index or block file actually containing the block for that hash.

Also fixes countMissingInformation query - it should join with missingBlocklistVolumes, rather than missingBlocklistEntries, as it is selecting the VolumeID.

### DIFF
--- a/Duplicati/Library/Main/Database/LocalListBrokenFilesDatabase.cs
+++ b/Duplicati/Library/Main/Database/LocalListBrokenFilesDatabase.cs
@@ -30,7 +30,11 @@ UNION
 )
 WHERE ""BlocksetID"" IS NULL OR ""BlocksetID"" IN 
   (
-    SELECT ""BlocksetID"" FROM ""BlocksetEntry"" WHERE ""BlockID"" NOT IN (SELECT ""ID"" FROM ""Block"")
+    SELECT DISTINCT ""BlocksetID"" FROM (
+      SELECT ""BlocksetID"" FROM ""BlocksetEntry"" WHERE ""BlockID"" NOT IN (SELECT ""ID"" FROM ""Block"")
+    UNION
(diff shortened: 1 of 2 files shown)


warwickmm:
It looks like several tests have failed with the message:
```
Recreated database has missing blocks and 3 broken filelists. Consider using "list-broken-files" and "purge-broken-files" to purge broken data from the remote store and the database.
```
Are the tests now misguided, or have these modifications changed some expected behavior?

tygill:
That is a good question - I'll have to take a look.

kenkendk:
Awesome detective work. I think I have fixed the same issue in another branch, but I ended up with the exact same problem that you have now fixed:
https://travis-ci.org/duplicati/duplicati/builds/377652799

tygill:
Glad I wasn't the only one to find and fix this - I've been trying to find out where the invalid blocksets are coming from so I could get my backup restarted on 2.0.3.6, but last night finally downgraded to 2.0.3.5 so my backup could go ahead and finish.

After purging invalid blocks using this logic, and restarting my backup with the official 2.0.3.6 build and a recreated database, I've started hitting the error in #3202, and had some notes at the end from trying to dig into what could be causing that. But there's enough in the core workings of Duplicati that I'm not familiar enough with to get to the bottom yet - I keep finding something that looks off, only to later realize that my assumption about it was incorrect and the code there should actually be working correctly :/

kenkendk:
Now that I figure out the problem I was battling, I think they just appeared similar, but are in fact two different problems.

PR #3226, opened 2018-05-18
Minor refactoring and modifications in travis build script.

verhoek:
Corrected a spelling typo in comparision (trivial change, unfortunately many files affected). 
Reduced duplication in FilterCollector.cs. 

Most significantly I moved all travis build steps previously defined in the travis yaml to its own script with a trap that stops on any error. Reason for this is unit tests kept failing due to directories not being there, until I found out that the zip files the script is supposed to download were broken and not unextractable. Previously, any error in the script steps would not cause the build to stop and "silently" continue. All steps are critical though. if anything has a non-zero exit code, the build should in fact stop. 

In addition, I extended the new build script to be a bit more failsafe on zip files -  in my case broken zip files were previously downloaded in my cache, preventing a re-download but also causing to break any future build due to missing directories and failing unit tests. This has been fixed by testing the zip files if there, if broken or missing, it will redownload.

Another upshot of this is that the travis build script can now be executed and hence tested locally.


kenkendk:
Great work!

PR #3229, opened 2018-05-20
Bugfixes commandline

verhoek:
Commandline previously crashed on "help find", "help --include" and "help idontexist".


Duplicati/CommandLine/Program.cs
```diff
@@ -163,31 +163,16 @@ public static int RealMain(string[] args)
...
-            bool isHelp = cargs.Count == 0 || (cargs.Count >= 1 && string.Equals(cargs[0], "help", StringComparison.OrdinalIgnoreCase));
-
-            if (isHelp) 
+            if (cargs.Count == 0 || (string.Equals(cargs[0], "help", StringComparison.OrdinalIgnoreCase))) 
             {
                 return Commands.Help(outwriter, setup, cargs, options, filter);
             }
 
-            if ((options.ContainsKey("parameters-file") && !string.IsNullOrEmpty("parameters-file")) || (options.ContainsKey("parameter-file") && !string.IsNullOrEmpty("parameter-file")) || (options.ContainsKey("parameterfile") && !string.IsNullOrEmpty("parameterfile")))
+            if (options.ContainsKey("parameters-file") && !string.IsNullOrEmpty(options["parameters-file"]))
```
kenkendk: This changes the logic a bit. Previously, it was possible to use `parameter-file`, `parameterfile` and `parameters-file`. With this change, only the last version is accepted.

This could potentially break existing backups, if one of the other names are used. What is the motivation to remove the alias names?

Duplicati/CommandLine/Commands.cs
```diff
@@ -133,11 +133,7 @@ public static int Examples(TextWriter outwriter, Action<Duplicati.Library.Main.C
     
         public static int Help(TextWriter outwriter, Action<Duplicati.Library.Main.Controller> setup, List<string> args, Dictionary<string, string> options, Library.Utility.IFilter filter)
         {
-            if (args.Count < 1)
-                Duplicati.CommandLine.Help.PrintUsage(outwriter, "help", options);
-            else
-                Duplicati.CommandLine.Help.PrintUsage(outwriter, args[0], options);
-                
+            Duplicati.CommandLine.Help.PrintUsage(outwriter, args.Count == 0? "help" : args[args.Count - 1], options);
```
kenkendk: This changes the logic a bit, as it would previously use the first argurment, but now it uses the last argument.
I don't have a preference for either, but was wondering if there was an idea behind the change?

Duplicati/CommandLine/Program.cs
verhoek: My bad, I didn't recognize these were aliases at all :) I reverted this to be backwards compatible, even though it strikes me to have these aliases for an option (other options don't have this). In the new commit I rewrote the code to avoid duplication. Notice that in the old code, it should have checked on the value option[parameterOption] to be not null or empty, instead it check a constant string to be non-empty which would hence always be false.

Duplicati/CommandLine/Commands.cs
verhoek: Thanks for pointing this out. I did realize this but at the time didn't come up with a better solution. I added a commit that takes the first argument after help if present, based on the knowledge that the first argument, if any, is always help (otherwise we wouldn't have arrived in this function).

Duplicati/CommandLine/Program.cs
kenkendk: There are some options that have aliases, but it is not common. Maybe it would be better if I had not introduced aliases here, but I figured that users would have a hard time remembering what the option was called, and now it is in there :).

PR #3234, opened 2018-05-25
Use Equals to compare CultureInfo instances

warwickmm:
Since the `Equals` method is overridden in the `CultureInfo` class, we should use it instead of the `==` operator to compare two instances.


kenkendk:
Yes, that distinction has bitten me more than once :/

PR #3237, opened 2018-05-25
(MS Graph) When the delete operation fails due to a 'file not found' error, we don't need to bubble up an exception

tygill:
(MS Graph) When the delete operation fails due to a 'file not found' error, we don't need to bubble up an exception


kenkendk:
There is some logic for handling this, but it requires that the backend throws a `FileMissingException`.
Your fix works too, but we do not get any log information that this happened.
I think we should either write a log message about ignoring the error, or throw `FileMissingException`.

tygill:
I would rather this do the standard thing - I didn't know about FileMissingException. I've changed Delete to throw it, and also updated Rename and Get to throw it if there is an underlying DriveItemNotFoundException.

PR #3242, opened 2018-05-26
Remove ineffective statements

warwickmm:
The removed code was introduced in revision ba58651e5a ("Fix for CreateFolder with SSH"), presumably to fix an issue where some SSH backends did not like the trailing slash when creating new folders.  However, since we are not using the output argument of the call to `Substring`, these statements have no effect.


kenkendk:
Maybe we should change it to:
```
p = p.Substring(0, p.Length - 1);
```

Or do you think it works better if we ignore the trailing slash?

warwickmm:
Since we've been effectively ignoring the trailing slash without any problems, I think it's fine (and cleaner) to continue ignoring it.

PR #3244, opened 2018-05-26
feature tempdir for backend + fix preliminary erasing of callcontextsettings

verhoek:
Added a **tempdir** option to set a default tempdir for the backend as an alternative to use the current system's path. Previously this happened through setting environment variables which did not work reliably: while debugging on Linux, **TMPDIR** was set yet System.IO.Path.GetTempPath did not respect its value.

Instead, a more direct approach simply sets a default value that if present overrides System.IO.Path.GetTempPath. Note that tempdir configured on a job level, as for example configured in the gui, will take precedence over this default tempdir, just as it did over  
System.IO.Path.GetTempPath.

See also: https://forum.duplicati.com/t/linux-tempdir-not-affective/3536

While implementing this, I observed a rather subtle and most likely unwanted behavior in that different realizations of the parametrized CallContextSettings were destroying each other's settings due to a common shared key "duplicati-settings". In particular, a httpcontextsettings.settings would erase an independent systemcontextsettings.settings (or oauth) because a new contextid would be generated. 
This behavior has been changed in that different realizations of CallContextSettings have also distinct contextid's by default (this can be overridden by explicitly giving a common key in startcontext though).


Duplicati/Server/Runner.cs
```diff
@@ -438,7 +439,7 @@ public static Duplicati.Library.Interface.IBasicResults Run(IRunnerData data, bo
 
 
             var backup = data.Backup;
-            Duplicati.Library.Utility.TempFolder tempfolder = null;
+            Duplicati.Library.Utility.TempFolder tempfolder = new Duplicati.Library.Utility.TempFolder();
```
kenkendk: This will cause it to create a temporary folder, even when it is not needed. The temp folder is only required when including the job setups inside the backup.

kenkendk:
Great work. I did not realize that there were actually places where the tempdir was call directly.
If you change the extra tempfolder creation line back, I will merge it.

verhoek:
Thanks for the review. I adjusted your commented part after understanding better what it should do based on your comment. 

- Behavior before: if a task config is to be stored, then store it using a new tempfolder instance, and then keep using the tempfolder, which is either null or the new instance from the task config storing part. 

- Behavior now: If a task config is to be stored, store it with a tempfolder instance. Do not reuse the tempfolder instance for further execution of code.

I am not sure if the tempfolder instance is used in further execution of Runner.Run. It doesn't look like it and I guess all information is available in the options that are used to initialize the controller. I could be wrong.

kenkendk:
The `store-task-config` option is used to signal that a copy of the server backup settings should be included in the backup. It is not exposed to the UI, because we never figured out the best way to show it.

It works by using the `control-files` option, which allows you to inject files into the `dlist` files. The code just creates a single `.json` file with the backup settings, and then appends the name of this file to the user-supplied `control-files` value (if any).

I don't think your new version works, because once the temporary file (and folder) goes out of scope, they are deleted. The previous setup kept the reference around until the operation was finished, such that the file path was valid for the duration of the backup.

verhoek:
Thanks for the explanation, that sounds reasonable. The logic is again as it was before.

kenkendk:
Nice work 👍

PR #3254, opened 2018-06-02
Fix line ending issues

warwickmm:
This fixes several cases where carriage returns were not followed by line feeds, which caused many editors (and GitHub) to render multiple lines as a single line.

It appears that these carriage-return-only lines prevented Codacy from performing its analysis, as new issues are now being reported in these files.


kenkendk:
I think it was a template file in Monodevelop or similar, that had some weird line endings. I fixed most of them some time ago, but it seems they keep popping up.

PR #3256, opened 2018-06-05
Use helper method to read from Streams and add unit tests

warwickmm:
A call to `Stream.Read` can theoretically read just a single byte.  While local file and memory streams will likely read as many bytes as requested, it is not guaranteed.  As such, we should use the `Utility.ForceStreamRead` method in cases where a `Stream` should be read until it is exhausted, or a maximum number of bytes has been read.

We also added some unit tests to verify the behavior of the `Utility.ForceStreamRead` method.

### DIFF
--- a/Duplicati/Library/Main/Operation/RestoreHandler.cs
+++ b/Duplicati/Library/Main/Operation/RestoreHandler.cs
@@ -610,8 +610,8 @@ namespace Duplicati.Library.Main.Operation
                                             if (sourcestream.Length > block.Offset)
                                             {
                                                 sourcestream.Position = block.Offset;
-                                                
-                                                var size = sourcestream.Read(blockbuffer, 0, blockbuffer.Length);
+
+                                                int size = Library.Utility.Utility.ForceStreamRead(sourcestream, blockbuffer, blockbuffer.Length);
                                                 if (size == block.Size)
                                                 {
                                                     var key = Convert.ToBase64String(hasher.ComputeHash(blockbuffer, 0, size));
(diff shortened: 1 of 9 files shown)


Duplicati/Library/Main/Operation/RestoreHandler.cs
```diff
@@ -610,8 +610,8 @@ private static void ScanForExistingSourceBlocksFast(LocalRestoreDatabase databas
                                             if (sourcestream.Length > block.Offset)
                                             {
                                                 sourcestream.Position = block.Offset;
-                                                
-                                                var size = sourcestream.Read(blockbuffer, 0, blockbuffer.Length);
+
+                                                int size = Library.Utility.Utility.ForceStreamRead(sourcestream, blockbuffer, blockbuffer.Length);
                                                 if (size == block.Size)
```
warwickmm: This condition feels odd to me.  Is `sourcestream` guaranteed to be `block.Size` bytes in length?

Duplicati/Library/Main/Operation/RestoreHandler.cs
```diff
@@ -720,7 +720,7 @@ private static void ScanForExistingSourceBlocks(LocalRestoreDatabase database, O
                                                 using (var sourcefile = m_systemIO.FileOpenRead(source.Path))
                                                 {
                                                     sourcefile.Position = source.Offset;
-                                                    var size = sourcefile.Read(blockbuffer, 0, blockbuffer.Length);
+                                                    int size = Library.Utility.Utility.ForceStreamRead(sourcefile, blockbuffer, blockbuffer.Length);
                                                     if (size == targetblock.Size)
```
warwickmm: This condition feels odd to me.  Is `sourcefile` guaranteed to be `block.Size` bytes in length?

Duplicati/Library/Main/Operation/RestoreHandler.cs
kenkendk: As I understand it, it checks if the data we just read into the buffer has the same length as the block we are looking for. In theory, we could use matching data where we just read `block.size` bytes each time, but that produces a number of non-aligned blocks which is generally not supported as it increases the processing time dramatically.
As an example, if `block.size == 1` we would read a single byte and hash it each time, where now we read full blocks until EOF.

Duplicati/Library/Main/Operation/RestoreHandler.cs
kenkendk: I think this is the same logic as before. We look for a specific block, and ignore those that do not match.

Issue #2640, opened 2017-08-13
Directory junctions are restored as empty folders on Windows


mkatsevVR:
I have:
- [x] searched open and closed issues for duplicates
This seems identical to #1475 except on Windows this time

----------------------------------------

### Version info
**Duplicati Version:** 2.0.1.73_experimental_2017-07-15
**Operating System:** Windows 10 x64
**Backend:** B2

### Bug description
When directory junctions are restored, an empty folder is created in their place. 
### Steps to reproduce
- Make a backup for a Windows user folder (just as an example, since it already has some junctions, e.g. 'My Documents' is a junction that points to 'Documents'...)
- Restore the backup

**Actual result:** 
'Documents' is restored correctly, but 'My Documents' is an empty folder
```
08/09/2017  10:49 AM    <DIR>          Documents
08/13/2017  02:47 PM    <DIR>          My Documents
```
**Expected result:** 
'My Documents' should point to 'Documents'
```
08/13/2017  03:00 PM    <DIR>          Documents
04/27/2017  01:15 AM    <JUNCTION>     My Documents [C:\Users\Max\Documents]
```

### debug log
```
Aug 13, 2017 2:47 PM: Failed to apply metadata to file: "t:\restore_test\duplicati_main\Users\Max\My Documents\", message: File already exists: t:\restore_test\duplicati_main\Users\Max\My Documents
Aug 13, 2017 2:47 PM: Creating missing folder t:\restore_test\duplicati_main\Users\Max\My Documents for target t:\restore_test\duplicati_main\Users\Max\My Documents\ 
```

verglor:
Both directory junctions and directory symlinks are restored as empty folder.
File hardlinks are restored as regular copy of file.
File symlinks are not restored at all.

verglor:
According to last comment in https://github.com/duplicati/duplicati/issues/1047#issuecomment-51188887:

> The symlinks are now stored, but they are not restored.
> This  will be fixed once the metadata restore is implemented.

Is it planned to implement metadata restore ?

donestefan:
Still present in 2020 on Duplicati 2.0.5.1_beta_2020-01-18 
Content of Junctions seem to be backed up correctly (judging be the backup size). But the restore dialog in the GUI shows the junction to be empty. Restore of a junction results in a warning "Restore completed without errors but no files were restored"
I tested with Backend "Webdav" and "OnedriveV2".

PR #3262, opened 2018-06-07
Windows symlink restore problems

snamds:
- Fix. Alphaleonis.Win32.Filesystem.File.CreateSymbolicLink wrong parameters order. !?!?!?!?
- Fix. Relative target links didn't work
- Fix. Folder was created previous symlink creation


snamds:
I was evaluating this software to use it and found that symlink restore doesn't work in windows.

· The Alphaleonis.Win32.Filesystem.File.CreateSymbolicLink parameters order was incorrect.
· The target passed to CreateSymbolicLink doesn't support relative paths. If you use relative paths, they are created using the execution path as reference. I've used the Path.Combine function to make it absolute.
· The code created an empty folder before evaluating the metadata info. When the metadada was evaluated, the symlink could not be created because the folder already existed.

snamds:
There was a conflict with last changes in RestoreHandler.cs. Now it's ok.

kenkendk:
Suppose you have this structure:
```
C:\a\
C:\a\b\
C:\a\d\
C:\a\b\c.txt
C:\a\d\f.txt
```

If you restore `*.txt`, you need to ensure that `C:\a\b\` and `C:\a\d\` are created, otherwise you will not be able to create the files `C:\a\b\c.txt` and `C:\a\d\f.txt`. The code you removed is making sure this happens.

Even if `C:\a\b\c.txt` is a symlink, you still need to create `C:\a\b\` and that will not interfere with creating the symlink. Are you seeing something else?

snamds:
Ok @kenkendk , now I understand what you meant with "Is it another problem that it recreates the symlink folder and not the parent because the path ends with a backslash?"
Yes that was the problem. I've changed the "m_systemIO.PathGetDirectoryName" with "Duplicati.Library.Utility.Utility.GetParent"

snamds:
I think https://github.com/duplicati/duplicati/issues/2640 will be fixed with this pr

PR #3266, opened 2018-06-11
Clean up tempdir environment vars

verhoek:
I came across some TMPDIR environment stuff that is now obsolete. I cleaned this up plus simplified SQLITE_TMPDIR logic.

### DIFF
--- a/Duplicati/Library/Main/Controller.cs
+++ b/Duplicati/Library/Main/Controller.cs
@@ -22,7 +22,6 @@ using System.Linq;
 #endregion
 using System;
 using System.Collections.Generic;
-using System.Text;
 using Duplicati.Library.Utility;
 
 namespace Duplicati.Library.Main
@@ -55,11 +54,6 @@ namespace Duplicati.Library.Main
         /// The thread running the current task
         /// </summary>
         private System.Threading.Thread m_currentTaskThread = null;
-
-        /// <summary>
-        /// Holds various keys that need to be reset after running the task
-        /// </summary>
-        private readonly Dictionary<string, string> m_resetKeys = new Dictionary<string, string>();
 
         /// <summary>
         /// The thread priority to reset to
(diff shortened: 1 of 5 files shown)


kenkendk:
The reason for the TMPDIR environment variable is that SQLite relies only on this parameter to set the temporary folder. If we remove this code, the `--tempdir` setting will not affect the SQLite temporary folder?

verhoek:
It seems you're right, though the sqlite documentation is confusing here.
The --tempdir is always stored in Library.Utility.TempFolder.SystemTempPath . And this property is used to set SQLITE_TMPDIR, which is used to set the SQLite temporary folder.
I thought this was it, **but**, as I just found out, SQLITE_TMPDIR is apparently only used on unix/linux/cygwin, and on windows it's TMP/TEMP/...
See also

http://sqlite.1065341.n5.nabble.com/pragma-temp-store-directory-is-deprecated-what-is-the-alternative-td83280.html

and 

https://sqlite.org/tempfiles.html#tempdir (section 5). 

To be on the safe side, I will add again setting the TMP or TEMP environment variables for SQLite plus a comment explaining why we do this.

kenkendk:
Ok, so the new logic is to just set the SQLite env variable during the loading.
I have not tested it, but is this sufficient to make sure all future temp-folder usage by SQLite is using that folder? If so, this is a cleaner approach than what I have previously implemented.

I am considering spawning a new process for running the backup. This will make it simpler to have temporary environment variables and make it possible to kill the process, but it requires some bridging to communicate between the two.

kenkendk:
Reading through the sqlite source, it looks like Linux reads it on the first attempt to get a temporary folder, but Windows appears to repeatedly read the system temporary folder via the API call.

verhoek:
I didn't think of that - I didn't test the environment setting of TMP/TEMP, I tested the tempdir code only on linux. And there the other variable SQLITE_TMPDIR is used which works fine. I will remove the TEMP/TMP variable setting from the SQLiteHelper and move it back to around where SystemTempPath is set, so it has a more global scope.

verhoek:
After some reconsideration, I did not move the TEMP/TMP setting to a more global scope (like tempfolder/callcontext settings) because there we don't have much control if these environment variables are changed or not while SQLite is being used.

The environment variables are now set once SQLite is to start, and not changed/restored. It is not ideal yet in the sense that if at some future time, another external program is used that also requires adjusting TMP/TEMP environment variables, this could cause conflicts.

kenkendk:
Yeah, it is not ideal at all. We really need a setting in SQLite that allows us to control the temp folder, but it would be years before such a change propagated to all the distros and platforms, so lets use your version for now, and if I get the spawned process working, we can simply set the tempdir for the child process in a single place and forget about it.

Issue #3245, opened 2018-05-28
Prompting for encryption passphrase uses 100% of CPU


mhorowitz:
- [x] I have searched open and closed issues for duplicates.

----------------------------------------

## Environment info
- **Duplicati version**: 2.0.3.3_beta_2018-04-02
- **Operating system**: macOS 10.12.6
- **Backend**: b2, ssh

## Description
Running cli commands which prompt for encryption key spin using 100% of the CPU.  I get the prompt `Enter encryption passphrase:`, then before I even type the password, the process starts using 100% of the CPU.  If I type a password here, it echos.  At this point, nothing ever seems to happen again.

This did not happen with 2.0.2.1

## Steps to reproduce
1. duplicati-cli test b2://... 1

- **Actual result**:
100% CPU in Activity Montor
- **Expected result**:
whatever the command is supposed to do.  Not echoing my encryption key.

Also repros with `ssh://` url, or with `list-broken-files` command.  `help` works fine.

If I use the `--passphrase` cli flag, instead of entering the key interactively, everything appears to work fine.

## Debug log
I tried using the --logfile option, but it wasn't created.

verhoek:
I also encountered this behavior and implemented a fix. The reason this happens: a child process is started that uses Console.ReadKey within a while(true) loop, but the standard input is never correctly redirected to the child process. I am still checking if there are other places in the code where similar things happen, and will then open a pull request.

warwickmm:
Fixed in pull request #3267.

PR #3267, opened 2018-06-12
Fix - Console.ReadKey for encryption passphrase

verhoek:
See https://github.com/duplicati/duplicati/issues/3245 .

Before, the update manager didn't share the console with the actual duplicati process properly, preventing Console.ReadKey for the encryption passphrase to be processed.

Made other places with standard stream redirecting more explicit. I didn't see immediate problems here but please double check if no similar issues can arise here.

### DIFF
--- a/Duplicati/Library/AutoUpdater/UpdaterManager.cs
+++ b/Duplicati/Library/AutoUpdater/UpdaterManager.cs
@@ -1106,10 +1106,7 @@ namespace Duplicati.Library.AutoUpdater
                 {
                     CreateNoWindow = true,
                     UseShellExecute = false,
-                    RedirectStandardError = true,
-                    RedirectStandardInput = true,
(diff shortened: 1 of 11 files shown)


kenkendk:
The idea with the redirects is that the spawned process can send data to the parent process stdout (and parent stdin -> child stdin).

If we remove the code as suggested, that functionality goes away?

Maybe we just need to remove the buffers on the streams to make readkey work?

verhoek:
That's what I understood from the code as well, but the "UseShellExecute = false" implies they share the same console, whence stdout/err/in. From what I saw, if in addition streams are shared, this seems to break it, stdin doesn't work properly anymore, thought stdout/err still get redirected. Also, I tested this just on linux, I don't know how it behaves on windows.

kenkendk:
Thanks for the link, I learned something new :)

Issue #3276, opened 2018-06-16
Concurrency issues with Box.com (null reference exception)


warwickmm:
## Environment info

- **Duplicati version**: 2.0.3.6
- **Operating system**: Arch Linux with 4.16.13 kernel
- **Mono version**: Mono JIT compiler version 5.12.0
- **Backend**: Box.com

## Description
With version 2.0.3.6, I am unable to upload files to a Box.com backend.  When the backup runs, no errors are explicitly reported, and the UI always says "Backup: 0 bytes / 0 Versions".  The same backup configuration works fine in version 2.0.3.5.

Digging through the server logs, the following entry appears multiple times:
```
"ClassName":"System.NullReferenceException"
"Message":"Object reference not set to an instance of an object"
"Data":null
"InnerException":null
"HelpURL":null
"StackTraceString":"  at Duplicati.Library.Backend.Box.BoxBackend.Put (System.String remotename, System.IO.Stream stream) [0x00049] in /home/warwickmm/documents/code/duplicati/Duplicati/Library/Backend/Box/BoxBackend.cs:216
```
On line 216 of `BoxBackend.cs`, we reference the `m_filecache` field, which is set to `null` in several places in the code.  I'm wondering if there's a race condition in how this field is being assigned and referenced?

Although the exception and backend are the same as in #3062, I believe the cause of the issues are different.

Issue #3062, opened 2018-02-21
Box.com: "Object reference not set to an instance of an object" error


weslowsk:
- [X ] I have searched open and closed issues for duplicates.

----------------------------------------

## Environment info

- **Duplicati version**: 2.0.2.19_canary_2018-02-12
- **Operating system**: Linux Mint 18
- **Backend**: Box.com

## Description

Issue #2349 was marked as fixed. However, the issue persists.

## Steps to reproduce
Same as previously referenced issue.

- **Actual result**:
The backup runs for several minutes (sometimes over an hour) and then shows an error and doesn't show "successful run".
- **Expected result**:
Backup completes successfully

Bruceforce:
## Environment info


- **Duplicati version**: 2.0.2.19_canary_2018-02-12
- **Operating system**: Linux 3.16.0-5-amd64 SMP Debian 3.16.51-3+deb8u1 (2018-01-08) x86_64 GNU/Linux
- **Mono version**: 5.8.0.108
- **Backend**: Box.com

I can confirm this one. For me the backup ran for ~2 hours and then I got the error message. So I had to schedule the backup to start every two hours. After it once completed there were only small changes and the backup job was successful on these runs. Only the first "bigger" backup which took longer time caused problems for me.

weslowsk:
The frequent backup (i.e. once every x hours) may be a workaround for me too. I'll try it. Thanks for the tip.

weslowsk:
The workaround worked for me. But, it did take 10-20 scheduled backups (every x hours) and at least 3 database delete/rebuilds to finally get about 13GB backed up into Box...it's a pain. Root cause determination of why it fails and a fix for the failure is still needed.

taferris:
Bummer!    I was hopeful that 2.0.3.2_canary_2018-03-16 which had the note of "Fixed a problem with the box.com backend" was going to solve this, but I am still getting the dreaded "Object reference not set to an instance of an object" when backing up to Box.com.

I will say the new logging gives much more information!

I am going to try upgrading mono and then try again and if it fails will clean and post the logs.

dgileadi:
I'm getting this error too on v2.0.3.3_beta_2018-04-02, macOS High Sierra 10.13.5. I've been manually restarting the backup instead of setting the shorter interval and that seems to work too.

It's a bit frustrating that the log doesn't show the exception, so I don't have the stack trace.

warwickmm:
[...]

PR #3277, opened 2018-06-16
Avoid null reference exceptions with Box.com backend

warwickmm:
This fixes an issue in the `BoxBackend` class, where a reference to a `Dictionary` was being assigned to `null` in several places.  By clearing the `Dictionary` instead, we avoid encountering null reference exceptions when performing concurrent operations.

This addresses issue #3276, and possibly issue #3062.


kenkendk:
We should not use the backend concurrently. There should be a single instance, and it should only be accessed one request at a time (enforced by the `BackendManager`).

Issue #3149, opened 2018-04-04
Handling of "files that are missing from the remote storage" at end of backup


u1735067:
- [x] I have searched open and closed issues for duplicates.

----------------------------------------

## Environment info
- **Duplicati version**: 2.0.3.1_experimental_2018-03-16
- **Operating system**: Win10 1709 x64
- **Backend**: ACD

## Description
At the end of my backup, I got the following messages : 
* `Found 1 files that are missing from the remote storage, please run repair`
* `Missing file: duplicati-20180404T012548Z.dlist.zip.aes`

And the UI indicated `Last successful run:01/24/2018`, the same as before the backup.

However, running repair showed that the file was actually present, and repair happily reports:
* `Destination and database are synchronized, not making any changes`
* `LastBackupDate: 04/04/2018 03:25:48`

But that doesn't changes the state of the UI, still indicating `Last successful run:01/24/2018`

I note 3 things here: 
* In the logs, I can't find a report about the backup, even if it failed. Only a `Fatal error`. The stats about it seems to be lost (and it might be related to the fact the UI isn't updated?).
* While repair thinks things are rights, it seems to be missing to verify some UI informations
* Maybe that's already the case, but maybe the backend should wait a bit before listing the files after the last upload, and/or retry a few times to be sure that's not temporary, maybe with a little delay between each tries.

Also, I tried to use "Create bug report" (mainly to see what it produces), but it took forever (I guess I hit 2298) and I had to kill Duplicati to stop it, "Stop now" didn't work.

## Steps to reproduce
1. Run a backup, for which some files are not returned in the last listing
2. Also, run repair

- **Actual result**:
  * Backup informations are not logged
  * Duplicati didn't retry the listing
  * Repair failed to update the UI informations, while it shows everything is OK (last backup is OK)
- **Expected result**:
  * Maybe log informations / stats about the backup, even if it failed
  * Duplicati retry the listing before throwing a fatal error
  * Repair should also update the UI informations
- **Workaround**:
  * Run another backup after the "failed" one, backup count is correctly updated (+2)

## The fatal error
```
Fatal error
Duplicati.Library.Interface.UserInformationException: Found 1 files that are missing from the remote storage, please run repair
   à Duplicati.Library.Main.Operation.FilelistProcessor.VerifyRemoteList(BackendManager backend, Options options, LocalDatabase database, IBackendWriter log, String protectedfile)
   à Duplicati.Library.Main.Operation.BackupHandler.PostBackupVerification()
[...]

PR #3285, opened 2018-06-21
Amazon cloud drive delay problems (amzcd-consistency-delay)

snamds:
- time delay between instances of AmzCD class was not respected.


snamds:
I have created two static dictionaries. One to store per file delays and the other to store the delay per authid.
Operations like put, get, etc use the file delay. Operations like list, createDirectory, etc use the authid delay.
This solves this problem: https://github.com/duplicati/duplicati/issues/3149

verhoek:
Does it make sense to convert them to concurrent dictionaries? Since they're static with regards to multithreading (I don't know if multi threading occurs here).

snamds:
I've already locked access to dictionaries so concurrent access is not a problem. Anyway, the problem was that delay logic was made using local variables of the AmzCD class. As multiples instances are used, one instance didn't know the last access to ACD of other instances.

Issue #3278, opened 2018-06-18
Duplicati file permissions on Linux reveals sftp passwords to all


sylerner:
- [x ] I have searched open and closed issues for duplicates.

----------------------------------------

## Environment info

- **Duplicati version**: 2.0.3.3_beta_2018-04-02
- **Operating system**: Linux
- **Backend**: sftp

## Description

By default, on most Linux systems, file permissions include everyone having read access to everyone's files.  I.g., the default umask on many distributions (including openSUSE) is 0022, allow group and others all permissions except write.

This is a particular problem with .config/Duplicati/Duplicati-server.sqlite, since it contains - in clear text - my sftp password - which is also my login password on the target system.

The permissions for .config/Duplicati/ and for all duplicati files should explicitly be set by installation and/or all duplicati programs execution to deny read, write and execute/browse permissions to group and other, only allowing the owner access to the directories and files.

## Steps to reproduce

1. Run DB Browser for SQLite (sqlitebrowser)
2. Open /home/someone_else/.config/Duplicati/Duplicati-server.sqlite
3. Click Browse Data
4. Select Option table
5. View passphrase record

- **Actual result**:
Anyone can open anyone's Duplicati-server.sqlite file and view stored passwords - with passphrase record in clear text.  They and also modify any records desired, changing options or corrupting the database.
- **Expected result**:
Unable to open anyone's duplicati files except their own.  Duplicati installation and/or program execution should explicitly set directly and file permissions to protect files from other users.

## Screenshots

![screenshot_20180618_013803](https://user-images.githubusercontent.com/20540525/41519898-daed4e6c-7298-11e8-8933-87d813ce4c7b.png)

sylerner:
The clear text password in the above screenshot has been whited out.

piegamesde:
I'm using sftp and this bothered me too. When using an ssh keyfile for login, you can easily avoid this problem.

sylerner:
First issue: If a user is not aware of the issue, they wouldn't know to move to using a ssh keyfile.

Second issue: My target is a Synology NAS.  There is no way to set up the use of ssh keyfiles via their GUI.  While I can log in to the NAS to set up a keyfile, Synology's security mechanisms will interfere with this when rebooted.

[...]

PR #3287, opened 2018-06-21
Fix/sqlite file permissions

verhoek:
See https://github.com/duplicati/duplicati/issues/3278 . Just for Linux (Windows I could add later). By default all sqlite files are created with file permissions 600 (I guess this is enough, if not let me know, maybe group rights have to be elevated in case of technical users duplicati runs with?). If a file already exists, permissions are unchanged. Reason for this is some users might want to alter permissions manually after creation.

Fix is not ideal, nicer would be to do it through SQLite's open. It looks like there is no functionality for this (http://sqlite.1065341.n5.nabble.com/about-default-file-permission-of-SQLite-database-file-td19582.html).

### DIFF
--- a/Duplicati/Library/SQLiteHelper/DatabaseUpgrader.cs
+++ b/Duplicati/Library/SQLiteHelper/DatabaseUpgrader.cs
@@ -182,7 +182,7 @@ namespace Duplicati.Library.SQLiteHelper
         /// </summary>
         /// <param name="connection">The database connection to use</param>
         /// <param name="sourcefile">The file the database is placed in</param>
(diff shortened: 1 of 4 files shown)


kenkendk:
Looks good.

However, I do not like the compile flags, a they would prevent a user from compiling on Windows and running on Linux. The problem is that `Mono.Posix` is not available on Windows, so I usually put stuff that depends on `Mono.Posix` into the [UnixSupport](https://github.com/duplicati/duplicati/tree/master/thirdparty/UnixSupport) library.

I will merge your changes, and then move the calls into the `UnixSupport` class.

verhoek:
Great! I also did not like doing this, though failed to come up with something better. This unix project is a nice idea, I didn't know it existed.

PR #3301, opened 2018-06-28
fix for special case

mnaiman:


mnaiman:
@kenkendk can you please merge this change soon, previous broke hyper-v backup in some cases, thx

Issue #3236, opened 2018-05-25
VSS Error ID 8194 in Windows Event Viewer


Germs2004:
- [x] I have searched open and closed issues for duplicates.

----------------------------------------

## Environment info
- **Duplicati version**: 2.0.3.3_beta_2018-04-02
- **Operating system**: Windows Server 2016
- **Backend**: A different local Windows Server 2016, shared folder

## Description
I run Duplicati as a Windows Service running as LocalSystem, and have snapshot-policy On and usn-policy On.  It seems to do successful backups and restores.  The Duplicati logs show no errors about VSS but there are errors in the Windows Event Viewer every time a Duplicati backup job runs.

Timeline:
```
Backup job started: 10:52:41
First error in Event Viewer: 10:52:46 AM
Second error in Event Viewer (duplicate of first): 10:59:15 AM
Begin time of Delete phase: 11:00:27 AM
Backup job End Time: 11:01:31 AM
```

Event Viewer error (2 of these):
```
Level = Error
Source = VSS
Event ID = 8194

Volume Shadow Copy Service error: Unexpected error querying for the IVssWriterCallback interface.  hr = 0x80070005, Access is denied.
. This is often caused by incorrect security settings in either the writer or requestor process. 

Operation:
   Gathering Writer Data

Context:
   Writer Class Id: {e8132975-6f93-4464-a53e-1050253ae220}
   Writer Name: System Writer
   Writer Instance ID: {2d2433ca-ba80-4d45-867b-aeebfb418c1e}
```

Duplicati backup job Result log:
```
DeletedFiles: 0
DeletedFolders: 0
ModifiedFiles: 2
ExaminedFiles: 280941
OpenedFiles: 2
AddedFiles: 0
SizeOfModifiedFiles: 73363456
SizeOfAddedFiles: 0
SizeOfExaminedFiles: 376658058583
SizeOfOpenedFiles: 73363456
NotProcessedFiles: 0
AddedFolders: 0
TooLargeFiles: 0
FilesWithError: 0
ModifiedFolders: 0
ModifiedSymlinks: 0
AddedSymlinks: 0
DeletedSymlinks: 0
PartialBackup: False
Dryrun: False
MainOperation: Backup
CompactResults:
    DeletedFileCount: 0
    DownloadedFileCount: 0
    UploadedFileCount: 0
    DeletedFileSize: 0
    DownloadedFileSize: 0
    UploadedFileSize: 0
    Dryrun: False
    MainOperation: Compact
[...]

PR #3302, opened 2018-06-29
default vss excluded writer

mnaiman:
Solves https://github.com/duplicati/duplicati/issues/3236


kenkendk:
The change only affects the documentation, it does not set the default option.
I will merge and update the code to also exclude this writer by default.

PR #3309, opened 2018-07-04
Remove async modifier from methods that lack await

warwickmm:
A method marked `async` that does not use `await` will run synchronously. By removing the `async` modifiers, we make it more explicit that these are synchronous methods.  This also addresses some compiler warnings, and avoids the creation of the state machine used to handle asynchronous operations.

In doing so, we also renamed these methods by removing `Async` from their names.  Furthermore, since they no longer have the `async` modifier, the do not need to return a `Task`, and do not need to be awaited.


Duplicati/Library/Main/Operation/Common/BackendHandler.cs
```diff
@@ -435,7 +434,7 @@ public Task<Library.Utility.TempFile> GetFileForTestingAsync(string remotename,
         private async Task<bool> DoPut(FileEntryItem item, bool updatedHash = false)
         {
             // If this is not already encrypted, do it now
-            await item.Encrypt(m_options).ConfigureAwait(false);
+            item.Encrypt(m_options);
```
warwickmm: The [other invocation](https://github.com/duplicati/duplicati/pull/3309/files#diff-59e4ce9a1a5775dcf7c9d1fa42821830R224) of `Encrypt` is performed on the thread pool using `Task.Run`.  I'm not sure if it makes sense to do so here as well?

Pectojin:
This looks sane. I've resolved a merge conflict introduced after your PR was submitted but I'm having a little trouble getting them to build in appveyor due to 
```
Operation\Backup\FilePreFilterProcess.cs(127,118): error CS0117: 'MetadataGenerator' does not contain a definition for 'GenerateMetadataAsync' [C:\projects\duplicati\Duplicati\Library\Main\Duplicati.Library.Main.csproj]
```
The "fault" is here: https://github.com/duplicati/duplicati/blob/master/Duplicati/Library/Main/Operation/Backup/FilePreFilterProcess.cs#L127
It was added after your PR, so it references the missing method.

warwickmm:
Thanks for your help, @Pectojin.  I believe revision 51cfb3d should fix the build.

PR #3314, opened 2018-07-06
Log sql variables

mnaiman:
All SQL statements are logged to file when Profiling is set.
Profiling is so verbose in other parts, that in SQL worth to have it.
For chasing bugs or understanding whats happening is valuable.


JonMikelV:
Am I reading that correctly that it will include the SQL parameters in the logs as well?  I've kind of been wanting that for a while, but am a bit worried about potential security / auth. leakage in some calls...

kenkendk:
Yes, the logging will include the SQL parameters. However, the local database (which is where the logging code is active) only contains file+block hashes and filenames. Since the filenames are present on the machine, I do not see any security issues with adding this logging.

But there is a performance issue here, as the string building is done for each block lookup, even if the user has not activated logging. Due to the way the logging system works, the handler does not know if the the message is going to be used or not.

If we include this, we should perhaps add an extra commandline option that enables logging of even the performance critical parts. This would also improve performance overall, becase we would not be calling the log system for messages that we do not want anyway.

kenkendk:
I have added a switch to enable detailed database logging (`--profile-all-database-queries`), and will merge this PR later today.

PR #3317, opened 2018-07-09
Refactoring/ftp

verhoek:
Some refactoring and fixes such that the backend tester for ftp works again.

### DIFF
--- a/Duplicati/CommandLine/BackendTester/Program.cs
+++ b/Duplicati/CommandLine/BackendTester/Program.cs
@@ -137,9 +137,9 @@ namespace Duplicati.CommandLine.BackendTester
         {
             string allowedChars = ValidFilenameChars;
             if (options.ContainsKey("extended-chars"))
-                allowedChars += options["extended-chars"];
-            else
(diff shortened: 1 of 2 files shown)


verhoek:
If the dictionary contains the extended-chars, it could still be possible that no value is supplied afaik. We both check the key existence and value. 

The old behavior did change because it breaks on some ftp backends like pro-ftpd: one cannot just simply create filenames with the given set of extended chars, it will break since such filenames are not allowed. Alternatively we can try to escape or quote such filenames, but in general that depends on the backend. Since this code is used for all backends, I think it is good to be more conservative. The user always has the option to define a more extended set of characters.

kenkendk:
When I started writing Duplicati, I had the idea that the backends could be used in other projects, so I prepared for a number of other ways to use it. The unittest was then made to test for various encoding issues that would occur for non-standard filenames. Duplicati will not (by default anyway) use special characters for filenames, so the change is "safe". In my unittests, I always set the characterlist to something without special characters when testing FTP.

PR #3324, opened 2018-07-18
Jon mikel v cert update mac os

Pectojin:
As discussed in https://forum.duplicati.com/t/2-0-3-7-on-mac-missing-certificates/3858


Pectojin:
Builds fine and fits in with the previous messages.
screen shot 2018-07-18 at 22 48 36

Issue #3342, opened 2018-08-18
Culture-aware file path comparisons


warwickmm:
It's generally [recommended](https://docs.microsoft.com/en-us/dotnet/standard/base-types/best-practices-strings) that file paths be compared using ordinal rules.  However, the `Utility.ClientFilenameStringComparer` and `Utility.ClientFilenameStringComparison` properties return values that result in culture-aware comparisons.

Furthermore, I am seeing inconsistent behavior across platforms.  Consider the following code:
```C#
Console.WriteLine($"Default culture: {System.Globalization.CultureInfo.CurrentCulture}");
System.Threading.Thread.CurrentThread.CurrentCulture = new System.Globalization.CultureInfo("en-US");
Console.WriteLine($"Switched culture to: {System.Globalization.CultureInfo.CurrentCulture}");

string withSoftHyphen = "ani\u00ADmal";
string noHyphen = "animal";

Console.WriteLine($"\"{withSoftHyphen}\" compares to \"{noHyphen}\": {withSoftHyphen.CompareTo(noHyphen)}");
Console.WriteLine($"\"{withSoftHyphen}\" ordinal compares to \"{noHyphen}\": {String.CompareOrdinal(withSoftHyphen, noHyphen)}");

string ddzs = "ddzs";
string dzsdzs = "dzsdzs";

Console.WriteLine($"With respect to {System.Globalization.CultureInfo.CurrentCulture}, is \"{ddzs}\" equal to \"{dzsdzs}\": {ddzs.Equals(dzsdzs, StringComparison.CurrentCulture)}");

System.Threading.Thread.CurrentThread.CurrentCulture = new System.Globalization.CultureInfo("hu-HU");
Console.WriteLine($"Switched culture to: {System.Globalization.CultureInfo.CurrentCulture}");
Console.WriteLine($"With respect to {System.Globalization.CultureInfo.CurrentCulture}, is \"{ddzs}\" equal to \"{dzsdzs}\": {ddzs.Equals(dzsdzs, StringComparison.CurrentCulture)}");
```

On Windows 7 with the "Region and Language Format" set to "English (United States)" using Visual Studio Professional 2017 Version 15.3.3 targeting the .NET 4.6.1 framework, the above code prints out:
```
Default culture: en-US
Switched culture to: en-US
"ani­mal" compares to "animal": 0
"ani­mal" ordinal compares to "animal": 64
With respect to en-US, is "ddzs" equal to "dzsdzs": False
[...]

PR #3354, opened 2018-08-31
Use ordinal sort rules to compare path names

warwickmm:
It's [recommended](https://docs.microsoft.com/en-us/dotnet/standard/base-types/best-practices-strings) that file paths be compared using ordinal rules.  However, the `Utility.ClientFilenameStringComparer` and `Utility.ClientFilenameStringComparison` properties returned values that resulted in culture-aware comparisons.  By using ordinal sort rules, we avoid issues where two different files on the filesystem can have names that are equivalent in certain cultures (e.g., `"ddzs"` and `"dzsdzs"` are equivalent with respect to the `hu-HU` culture), which could lead to strange "collisions" in the code.

This addresses issue #3342.

### DIFF
--- a/Duplicati/Library/Utility/Utility.cs
+++ b/Duplicati/Library/Utility/Utility.cs
@@ -1054,12 +1054,12 @@ namespace Duplicati.Library.Utility
         /// <summary>
         /// Gets a string comparer that matches the client filesystems case sensitivity
         /// </summary>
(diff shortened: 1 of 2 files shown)


warwickmm:
Could this change adversely affect existing backups, where some paths stored in the local database (or in the remote filelist) might now compare differently to those on the filesystem?

Pectojin:
Thanks 👍

PR #3355, opened 2018-08-31
Added some missing image formats

ravisorg:
NEF, ARW, and HEIC

### DIFF
--- a/Duplicati/Library/Main/default_compressed_extensions.txt
+++ b/Duplicati/Library/Main/default_compressed_extensions.txt
@@ -99,6 +99,9 @@
 .dng #Adobe Digital Negative
 .cr2 #Canon RAW Image Format
 .webp #WebP Image
+.nef #Nikon RAW format
+.arw #Sony RAW format
+.heic #High Efficiency Image File Format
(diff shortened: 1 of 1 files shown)


Pectojin:
Wouldn't NEF and ARW compress really well since it's raw image data, i.e no compression applied yet?

ravisorg:
That's actually a really good question. I just assumed they were fairly uncompressable, since memory card space on cameras is fairly limited and it makes sense for camera manufacturers to store their data as efficiently as possible. 

But that's an assumption not a fact, so I just ran a couple tests. Both done using some random NEF and ARW files on my computer (eg: not selected with any specific criteria) and using WinRAR in ZIP mode and highest (slowest) compression level.

arw-compression
nef-compression

Results as I read them are that there is very negligible compression possible on these file types. 7z or some other format may have better luck I'm not sure.

Hopefully that helps!

Pectojin:
Thanks for the quick tests! 

After a little googling it definitely seems both support compression so I think you're right in your assumption that these would very often be compressed due to memory card constraints.

PR #3365, opened 2018-09-06
[FIX] Localizations that have a sub-code are not loaded

mikaelmello:
The loader tries to load files in the format `code-subCode` while the files are saved as `code_subCode`.


kenkendk:
Thanks!

Issue #3372, opened 2018-09-09
FailedToUseChangeJournal: Failed to compare two elements in the array.


mnaiman:
- **Duplicati version**: latest .11
- **Operating system**: Win

## Description
When USN policy is on on SOME machines it throw this error. There is missing null reference check.
@dgehri Can you please take a look on that?

2018-09-06 23:00:17 +02 - [Error-Duplicati.Library.Main.Operation.BackupHandler-FailedToUseChangeJournal]: Failed to use change journal for volume "C:\": Failed to compare two elements in the array.
System.InvalidOperationException: Failed to compare two elements in the array. ---> System.NullReferenceException: Object reference not set to an instance of an object.
   at Duplicati.Library.Snapshots.USNJournal.SortedRecords.<>c.<GetParentOf>b__5_0(Record left, Record right)
   at System.Collections.Generic.ArraySortHelper`1.InternalBinarySearch(T[] array, Int32 index, Int32 length, T value, IComparer`1 comparer)
   at System.Collections.Generic.ArraySortHelper`1.BinarySearch(T[] array, Int32 index, Int32 length, T value, IComparer`1 comparer)
   --- End of inner exception stack trace ---
   at System.Collections.Generic.ArraySortHelper`1.BinarySearch(T[] array, Int32 index, Int32 length, T value, IComparer`1 comparer)
   at System.Collections.Generic.List`1.BinarySearch(Int32 index, Int32 count, T item, IComparer`1 comparer)
   at Duplicati.Library.Snapshots.USNJournal.SortedRecords.GetParentOf(Record usnRecord)
   at Duplicati.Library.Snapshots.USNJournal.ResolveFullPaths(ICollection`1 records)
   at Duplicati.Library.Snapshots.USNJournal.GetRecords(Int64 minUsn)
   at Duplicati.Library.Snapshots.USNJournal.<GetChangedFileSystemEntries>d__14.MoveNext()
   at Duplicati.Library.Snapshots.UsnJournalService.Initialize(IFilter emitFilter, IEnumerable`1 prevJournalData)

PR #3377, opened 2018-09-12
refs #3372: tentative fix for FailedToUseChangeJournal

dgehri:
@kenkendk : Possible fix for #3372


kenkendk:
Very plausible fix for that error 👍

PR #3378, opened 2018-09-12
refs ##3311 : fix for PathTooLongException

dgehri:


kenkendk:
In other places I rely on AlphaFS to work on paths that are rejected by .Net and the old Windows file API. If there is still a problem with `PathTooLongException`, we should change to using AlphaFS to manipulate the paths  in the USN module.

PR #3391, opened 2018-09-20
Add Description field to backups

### LINKED ISSUE #2214: Add backup description field for notes
Hello, please consider adding new fields to write a description to the backup or notes. The advantage would be that the name of backups could be short and concise. Thanks

<bountysource-plugin>
---
Want to back this issue? **[Post a bounty on it!](https://www.bountysource.com/issues/40469909-add-backup-description-field-for-notes?utm_campaign=plugin&utm_content=tracker%2F4870652&utm_medium=issues&utm_source=github)** We accept bounties via [Bountysource](https://www.bountysource.com/?utm_campaign=plugin&utm_content=tracker%2F4870652&utm_medium=issues&utm_source=github).
</bountysource-plugin>

sffetlio:
Looking through the code I wanted to start with a small PR, so here it is. 
Fixes #2214

### DIFF
--- a/Duplicati/Server/Database/Backup.cs
+++ b/Duplicati/Server/Database/Backup.cs
@@ -57,6 +57,10 @@ namespace Duplicati.Server.Database
         /// </summary>
         public string Name { get; set; }
         /// <summary>
+        /// The backup description
+        /// </summary>
+        public string Description { get; set; }
+        /// <summary>
         /// The backup tags
         /// </summary>
         public string[] Tags { get; set; }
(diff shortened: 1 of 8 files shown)


kenkendk:
Wow, you found all the spots that needs to change!

Makes me wish I had used an ORM instead of all that manual SQL fiddling.

There is only one thing that is needed, which is to change the version to 6:
https://github.com/duplicati/duplicati/blob/master/Duplicati/Server/Database/Database%20schema/Schema.sql#L157

Otherwise new users will have a v5 database, with the `Description` field, but still execute the `6` update script giving an error.

sffetlio:
Fixed :)

PR #3392, opened 2018-09-21
Simplify method resolution by removing default parameters

warwickmm:
This simplifies some method overloads by removing unnecessary default parameters.  The previous implementations resulted in an overload being hidden by one without a default parameter, making it unclear which method was being called.  For example, consider the following signatures:
```c#
public MultipartItem(string content, string contenttype = null, string name = null, string filename = null)
public MultipartItem(string contenttype, string name = null, string filename = null)
```
From the [documentation](https://docs.microsoft.com/en-us/dotnet/csharp/programming-guide/classes-and-structs/named-and-optional-arguments),

> If two candidates are judged to be equally good, preference goes to a candidate that does not have optional parameters for which arguments were omitted in the call. This is a consequence of a general preference in overload resolution for candidates that have fewer parameters.


kenkendk:
I tend to overdo the default parameters :)

PR #3399, opened 2018-09-29
Remove unused variables, fields, and unnecessary assignments

warwickmm:
This removes some unused local variables, unused fields, and unnecessary variable assignments.  This cleans up the code slightly while also reducing the number of compiler warnings.


Pectojin:
I love the commits that remove more code than they add 👍

warwickmm:
No problem at all.  Thanks for noticing this.  I think the intent was to just remove the assigned variable `logscope`, since it appeared to be unused.

PR #3404, opened 2018-10-02
Fix server status strings grammar and consistency

lucascosti:


lucascosti:
Some minor fixes for Hacktoberfest :)

If there are any other strings/docs that could use some attention, let me know!

mikaelmello:
I'd recommend to open the template file under `Localization/webroot` and `Localization/duplicati`, look for inconsistencies and then fix them on their original files.

The downside for that is that it will make a lot of translations outdated

PR #3414, opened 2018-10-07
Avoid potential comparison of value type with null

warwickmm:
While there aren't currently any cases where the generic type is a value type, this will make the code slightly more robust to future implementations.


warwickmm:
Oops, I was a bit careless there.  Thanks for noticing this.  Pull request #3417 should address this issue more appropriately.

PR #3422, opened 2018-10-11
Fix error in POST requests with empty form

mikaelmello:
Sometimes `info.Request.Form`, which is of type `HttpForm`, is empty and has a value of `HttpInput.Empty` instead of `HttpForm.EmptyForm`. Because of that, the check to see if the form was empty before trying to add items did not return true.


mikaelmello:
Update: this did not fix it. The error happens deeper inside HttpServer.

PR #3423, opened 2018-10-11
Revert hunk from 9ce1e80. Log redirect needs to be setup

mikaelmello:
If the log redirect is not setup, the live logs do not work at all because the scope used is the root one, which has IsolatingScope set to true.

Reverting hunk from 9ce1e80

### DIFF
--- a/Duplicati/Server/Program.cs
+++ b/Duplicati/Server/Program.cs
@@ -250,6 +250,9 @@ namespace Duplicati.Server
 
             try
             {
+                // Setup the log redirect
+                var logscope = Library.Logging.Log.StartScope(Program.LogHandler, null);
+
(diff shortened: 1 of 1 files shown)


warwickmm:
Thanks for fixing this @mikaelmello, I was a bit careless earlier.  Does the assigned variable `logscope` appear to be unused?  If so, can we remove it?

mikaelmello:
No problem, sorry again for my previous comment.

My guess is that when you removed the expression intelligence did not
accuse anything, so not? Not sure though, I won't have a computer close for
the next few hours

Em qui, 11 de out de 2018 12:02, warwickmm <notifications@github.com>
escreveu:

Issue #3433, opened 2018-10-16
Backups with an empty description are not possible


jeroenmeulenaar:
- [x ] I have searched open and closed issues for duplicates.

----------------------------------------

## Environment info
- **Duplicati version**: debug, 2018.10.14
- **Operating system**: Win7/32bits
- **Backend**: n/a

## Description
The "description" field should be non-empty, but the GUI doesn't check on this.

## Steps to reproduce
Add a new backup, but leave "Description" empty. The GUI doesn't check this, but the final step 5 will reply with "Unable to save schedule or backup object: constraint failed".

- **Actual result**:
An error is thrown
- **Expected result**:
Backups with an empty 'description' should be allowed.

Pectojin:
Yup. This one is related too #3428

There are some issues with default values.

jeroenmeulenaar:
> Yup. This one is related too #3428
Hmm, I thought I searched the open issues :).

To me the simplest option would be to remove the "NOT NULL" constraint from the database; looks better than tricking it with an empty string.

mikaelmello:
@jeroenmeulenaar While I agree with you, @Pectojin said:

> That way we don't need to deal with null exceptions other places in the application since it should always be a string.

And in my opinion having an empty description makes more sense than a NULL value.

Regarding the error displayed when the description is not filled, I made a quick hack here: https://github.com/LacunaSoftware/duplicati/commit/8634c85c6fdc45f86ad88066b17712f1fbf17547

jeroenmeulenaar:
> @jeroenmeulenaar While I agree with you, @Pectojin said:
> 
> > That way we don't need to deal with null exceptions other places in the application since it should always be a string.
> 
> And in my opinion having an empty description makes more sense than a NULL value.

Okay, I agree with the annoyance of having a NULL value, so let's make it indeed into an empty string.
 
> Regarding the error displayed when the description is not filled, I made a quick hack here: [LacunaSoftware@8634c85](https://github.com/LacunaSoftware/duplicati/commit/8634c85c6fdc45f86ad88066b17712f1fbf17547)

Maybe it's then good to update the UI as well, and display "Description (optional)" instead of "Description"?

mikaelmello:
> Maybe it's then good to update the UI as well, and display "Description (optional)" instead of "Description"?

Great suggestion, thanks. I'll make all those changes in a separated branch and create a PR.

PR #3436, opened 2018-10-17
Make description an optional field for backups

### LINKED ISSUE #3428: Description nullable on old installations but not nullable on fresh ones
## Description

Since the recent addition of a description field for backups, if you upgrade from an older version, by migrating the `Backup` schema is altered and the `Description` column is added with `NULL` as the default value.

However, the schema itself (used on new installations) declares `Description` as `NOT NULL`, hence two problems arrive:

1. There is an inconsistency that can only be fixed by resetting the database. Should we make the `Column` column nullable? If not, the migration should probably add the new column as nullable, update all entries to an empty string and make the column not nullable. Is that a good approach?

2. There is nothing handling a missing description error. An error while validating the insertion into the db is propagated all the way to the UI, displaying this:

image

I'm not sure how you guys would want it fixed, so I don't have a PR ready.

mikaelmello:
Checks whether it has been set and acts accordingly to avoid exceptions

Update migration to initialize all existent rows to Description=''.

Note that after some research, default values for TEXT columns were not
supported by MySQL a few years ago, however Duplicati uses SQLite
which seems to fully support this feature. If not, a workaround is to use:

```
ALTER TABLE "Backup" ADD COLUMN "Description" TEXT NULL;
UPDATE "Backup" SET "Description"='' WHERE "Description" IS NULL;
ALTER TABLE "Backup" ALTER COLUMN "Description" TEXT NOT NULL;
```

Closes #3428 and #3433

### DIFF
--- a/Duplicati/Server/webroot/ngax/scripts/controllers/EditBackupController.js
+++ b/Duplicati/Server/webroot/ngax/scripts/controllers/EditBackupController.js
@@ -221,7 +221,6 @@ backupApp.controller('EditBackupController', function ($rootScope, $scope, $rout
             return;
         }
 
-
         if (encryptionEnabled) {
             if ($scope.PassphraseScore === '') {
                 DialogService.dialog(gettextCatalog.getString('Missing passphrase'), gettextCatalog.getString('You must enter a passphrase or disable encryption'));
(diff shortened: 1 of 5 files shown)


Duplicati/Server/webroot/ngax/scripts/controllers/EditBackupController.js
```diff
@@ -487,6 +486,13 @@ backupApp.controller('EditBackupController', function ($rootScope, $scope, $rout
         var filters = $scope.Backup.Filters;
         $scope.Backup.Filters = [];
 
+        // If Description is anything other than a string, we are either creating a new
```
mikaelmello: anything other than a string includes 'undefined'

Pectojin:
I think you got it all covered. Just waiting for it to finish appveyor and codacy checks.

Issue #3311, opened 2018-07-05
PathTooLongException when using USNJournal


jojje:
- [X ] I have searched open and closed issues for duplicates.

----------------------------------------

## Environment info

- **Duplicati version**: 2.0.3.9_canary_2018-06-30
- **Operating system**: Windows 7 Pro (64 bit)
- **Backend**: "Local folder or drive"

## Description
When re-running a backup set which contains long file/folder names, and using the Journal USN feature introduced in #3184, the backup fails with a `System.IO.PathTooLongException`.

_Full scan_ backups of the set works however, such as when making the initial backup, or when changing the include / exclude filters for the set, so the problem only seems to occur when the USN Journal code branch is triggered.

## Steps to reproduce
1. Create a directory or filename with an absolute path longer tan 240 or 255 characters respectively.
2. Create a backup set that includes the dir / folder and set the advanced option `usn-policy` to `Required`.
3. Run the backup set two times.

- **Actual result**:
The second time the backup is executed it fails with the exception `System.IO.PathTooLongException`.

- **Expected result**:
Successful backup without the mentioned exception.

## Debug log
```
2018-07-05 18:39:42 +02 - [Error-Duplicati.Library.Main.Operation.BackupHandler-FatalError]: Fatal error
System.IO.PathTooLongException: The specified path, file name, or both are too long. The fully qualified file name must be less than 260 characters, and the directory name must be less than 248 characters.
   at System.IO.PathHelper.GetFullPathName()
   at System.IO.Path.LegacyNormalizePath(String path, Boolean fullCheck, Int32 maxPathLength, Boolean expandShortPaths)
   at System.IO.Path.GetFullPathInternal(String path)
   at System.IO.DirectoryInfo.Init(String path, Boolean checkHost)
   at Duplicati.Library.Snapshots.USNJournal.GetVolumeRootFromPath(String path)
   at Duplicati.Library.Snapshots.UsnJournalService.IsPathEnumerated(String path)
   at Duplicati.Library.Main.Operation.BackupHandler.<>c__DisplayClass12_0.<RunMainOperation>b__2(String path, Int64 fileSize)
   at Duplicati.Library.Main.Database.LocalBackupDatabase.AppendFilesFromPreviousSetWithPredicate(IDbTransaction transaction, Func`3 exclusionPredicate, Int64 fileSetId, Int64 prevFileSetId, DateTime
timestamp)
[...]

PR #3456, opened 2018-10-24
Issue 3311 PathTooLongException when using USNJournal

verhoek:
Yet another try for https://github.com/duplicati/duplicati/issues/3311#issuecomment-432563388 .


Duplicati/Library/AutoUpdater/UpdaterManager.cs
```diff
@@ -1075,7 +1075,7 @@ public static int RunFromMostRecent(System.Reflection.MethodInfo method, string[
         public static int RunFromMostRecentSpawn(System.Reflection.MethodInfo method, string[] cmdargs, AutoUpdateStrategy defaultstrategy = AutoUpdateStrategy.CheckDuring)
         {
             // If the update is disabled, go straight in
-            if (DISABLE_UPDATE_DOMAIN)
+            if (!DISABLE_UPDATE_DOMAIN)
```
kenkendk: I have a setting disabling the updates via an environment variable in the `.csproj`: 

`AUTOUPDATER_Duplicati_SKIP_UPDATE=1`

Duplicati/Library/Snapshots/USNJournal.cs
```diff
@@ -1,650 +1,652 @@
...
+                }
+            }
+        }
+
+        public static string GetVolumeRootFromPath(string path)
+        {
+            if (path == null)
+                throw new Exception(Strings.USNHelper.UnexpectedPathFormat);
+
+            return AlphaFS.Path.GetPathRoot(path);
```
kenkendk: In other places, I use a custom version of `System.IO`, like this:

```csharp
private static SystemIOWindows IO_WIN = new SystemIOWindows();
```

It does not yet have `GetPathRoot()` but it has many of the other methods, and we can add `GetPathRoot()` to the `ISystemIO` interface.

Duplicati/Library/AutoUpdater/UpdaterManager.cs
verhoek: :) My hacky way for debugging. Thanks for the tip.

Duplicati/Library/Snapshots/USNJournal.cs
verhoek: Agreed, better to decouple this.

verhoek:
Don't merge yet, this afternoon I will try to make the entire Snapshots part more consistent by moving alphaFS functionality to SystemIOWindows. Also outside of snapshots GetPathRoot is used through the native system io, which we might need to care about.

kenkendk:
> Also outside of snapshots GetPathRoot is used through the native system io, which we might need to care about.

The design is such that the all file/folder/path access should go through the snapshots (or `NoSnapshotWindows` if not snapshot is active).
This will make sure that AlphaFS is used when required.

I see two places where `GetPathRoot()` is used directly. 
One is in the restore method and is only affected if the restore target is a too long path.
The other is in the file backend and is only affected if the backup destination path is too long.

Ideally, both should be changed to go through the custom `ISystemIO` interface.

verhoek:
Intermediate status update:

I created a separate sub project for the io wrappers to prevent circular dependencies.

I'm not satisfied with the static property IO_OS in SystemIO (a similar switch was before inside SnapshotUtility) - it has a boolean argument depending on windows/linux. It would be nice to use Utility.IsClientWindows here directly, but then I (might) create again create circular dependencies.

I'm also not satisfied with the naming of the methods / vars / props yet.

At least the code is at this stage more consistent than it was, we don't cast implementations to interfaces anymore, and should be more failsafe on the windows side.

kenkendk:
> It would be nice to use Utility.IsClientWindows here directly, but then I (might) create again create circular dependencies.

It would be nice to have `IO_OS` in the `Utility` project, but since it depends on AlphaFS, we would carry that dependency around.

If we keep it in `Snapshot`, we limit the dependency on AlphaFS to that project which really needs it. The only place we need access to `IO_OS` outside `Main` and `Snapshot` are in the file backend?

But I guess if `Snapshot` just pulls in the `IO` project, then we end up with more or less the same as we have now. You should be able to reference the `Utility` project from `IO` because `Utility` does not depend on `IO` ?

verhoek:
Thanks for thinking along :)

> > It would be nice to use Utility.IsClientWindows here directly, but then I (might) create again create circular dependencies.
> 
> It would be nice to have `IO_OS` in the `Utility` project, but since it depends on AlphaFS, we would carry that dependency around.

I thought of only having the IO_OS in Utility, and the Linux/System.IO and Windows IO in Library.IO. But then I start to split functionality a bit across packages - at least only the entry point for IO functionality.

> If we keep it in `Snapshot`, we limit the dependency on AlphaFS to that project which really needs it. The only place we need access to `IO_OS` outside `Main` and `Snapshot` are in the file backend?

I don't know if Snapshot really needs to know AlphaFS or better IVssBackupComponents exists. I'm looking into the possibility to separate IVssBackupComponents functionality and the snapshot logic. Right now they're pretty much interwoven.

> But I guess if `Snapshot` just pulls in the `IO` project, then we end up with more or less the same as we have now. You should be able to reference the `Utility` project from `IO` because `Utility` does not depend on `IO` ?

Utility does use IO calls. Yesterday I spotted LinuxPowerSupplyState the Path Combine, though Linux is 'safe' since it uses the standard System.IO.  On the other hand, FileBackedList also contains System.IO calls.

I'd prefer if code uses Library.IO uniformly instead of using sometimes standard System.IO and sometimes Library.IO, since then it's likely one forgets to consider possible issues with Windows. It could work but I have no idea which other System.IO calls might throw a pathtoolong exception.

My reasoning to create a separate package is that IO is a low level functionality that can possible be used by all sub projects, and would like to prevent cross references etc.

kenkendk:
> Utility does use IO calls. Yesterday I spotted LinuxPowerSupplyState the Path Combine, though Linux is 'safe' since it uses the standard System.IO. On the other hand, FileBackedList also contains System.IO calls.

Those should be safe as they only use system defined paths. Many things on the system will break if `TMPDIR` is longer than 260 chars.

> I'd prefer if code uses Library.IO uniformly instead of using sometimes standard System.IO and sometimes Library.IO, since then it's likely one forgets to consider possible issues with Windows. 

I agree, even though it sucks having to not use the otherwise nice built-in interface in `System.IO`.

> It could work but I have no idea which other System.IO calls might throw a pathtoolong exception.

AFAIK, all `System.IO` calls (except `IsValidPath` and similar) call the validation routine and throws an exception for any non-plain path.

> My  reasoning to create a separate package is that IO is a low level  functionality that can possible be used by all sub projects, and would  like to prevent cross references etc.

Yes, I guess it has to be that way.

verhoek:
So I changed a bunch, I hope it is somehow understandable what happened.

What about WinNativeMethods in Utility? It looks like it isn't used at all.
Utility.Utility became a monster class, I'd like to take more file IO functionality out of there to Library.IO, but most of it is quite entangled.
I'd rather do that at a later point in time.

I guess I should add some IO unit tests too.

kenkendk:
> So I changed a bunch, I hope it is somehow understandable what happened.
Yeah, massive change, but all looks good.

> What about WinNativeMethods in Utility?
We should just remove it. I think it is a left-over from when I tried using the `OpenForBackup` methods to access locked files, but the Windows implementation is a real mess.

> Utility.Utility became a monster class
Yeah. The `Library.Utility` module is a dumping ground for much of the shared functionality. Originally the idea was to create a re-usable library of common functions, but I don't think it serves that purpose anymore. I also hate the name with `Utility.Utility.abc()` being used all over, but I have not found a good way to fix it.

I only have two minor issues with the changes.

1) There are still at least one place calling `System.IO.DirectorySeparatorChar.ToString()` and we should make it consistent and only use `Library.IO.Util.DirectorySeparatorString` instead.

2) The `Library.IO.IO_OS` should be a property and not need the flag. I know this is caused by the circular depency on `Library.Utility`.

A suggestion is to move `IsClientLinux` and friends to `Library.IO` and then just let the "old" properties in `Library.Utility.Utility` call `Library.IO.Util`. This way the properties are in both modules, but share the implementation.

Maybe we should take the opportunity to make a new `Platform` class that has just these properties. And perhaps also rename `IsClientLinux` to `IsClientPosix` as that is more accurate (MacOS is BSD derived, but Posix compliant).

As this class is performance critical, we should also make it with a static initializer instead of trying to determine properties dynamically.

verhoek:
> > So I changed a bunch, I hope it is somehow understandable what happened.
> 
> Yeah, massive change, but all looks good.
> 
> > What about WinNativeMethods in Utility?
> 

Got removed.

> I only have two minor issues with the changes.
> 
>     1. There are still at least one place calling `System.IO.DirectorySeparatorChar.ToString()` and we should make it consistent and only use `Library.IO.Util.DirectorySeparatorString` instead.
> 

Well it felt like there were at least a 1000 places.. :)

>     2. The `Library.IO.IO_OS` should be a property and not need the flag. I know this is caused by the circular depency on `Library.Utility`.
> 
> 
> A suggestion is to move `IsClientLinux` and friends to `Library.IO` and then just let the "old" properties in `Library.Utility.Utility` call `Library.IO.Util`. This way the properties are in both modules, but share the implementation.
> 

I'm fine with that. To do this, I renamed Library.IO to Libary.Common.IO. I did this to pull in the platform class, but platform has not necessarily to do with IO. 
To create a package just for platform seemed at the moment overkill.
I hope this doesn't turn into a second utility class, though I guess utility should not necessarily be used by all other packages and can contain also more complex stuff.


> Maybe we should take the opportunity to make a new `Platform` class that has just these properties. And perhaps also rename `IsClientLinux` to `IsClientPosix` as that is more accurate (MacOS is BSD derived, but Posix compliant).
> 

Ok.

> As this class is performance critical, we should also make it with a static initializer instead of trying to determine properties dynamically.

Ok, I still have to look into this.

verhoek:
Unless you have further comments I think it can be merged.

PR #3458, opened 2018-10-24
Fix useless notification "Warning: Got 0 warning(s)"

mikaelmello:
When a restore was made in which 0 files were restored, the Result has Success set to "Warning" and then a notification displaying "Warning: Got 0 warning(s)" appears

Another solution to avoid this was to remove the condition to return "Warning" when 0 files were restored but this could affect things that I don't know about and the problem actually belongs to the notification, so I've decided to simply show warning notifications if and only if we actually have notifications or warnings.

I don't know if this is the behavior wanted by you, just thought it would be a nice PR.

### DIFF
--- a/Duplicati/Library/Main/Operation/RestoreHandler.cs
+++ b/Duplicati/Library/Main/Operation/RestoreHandler.cs
@@ -462,6 +462,8 @@ namespace Duplicati.Library.Main.Operation
                     Logging.Log.WriteInformationMessage(LOGTAG, "RestoreFailures", "Failed to restore {0} files, additionally the following files failed to download, which may be the cause:{1}{2}", fileErrors, Environment.NewLine, string.Join(Environment.NewLine, brokenFiles));
                 else if (fileErrors > 0)
                     Logging.Log.WriteInformationMessage(LOGTAG, "RestoreFailures", "Failed to restore {0} files", fileErrors);
+                else if (result.FilesRestored == 0)
+                    Logging.Log.WriteWarningMessage(LOGTAG, "NoFilesRestored", null, "Restore completed without errors but no files were restored");
 
                 // Drop the temp tables
                 database.DropRestoreTable();
--- a/Duplicati/Library/Main/ResultClasses.cs
+++ b/Duplicati/Library/Main/ResultClasses.cs
@@ -610,7 +610,7 @@ namespace Duplicati.Library.Main
             {
                 if (Errors != null && Errors.Any())
                     return ParsedResultType.Error;
-                else if ((Warnings != null && Warnings.Any()) || FilesRestored == 0)
+                else if (Warnings != null && Warnings.Any())
                     return ParsedResultType.Warning;
                 else
                     return ParsedResultType.Success;
--- a/Duplicati/Server/Runner.cs
+++ b/Duplicati/Server/Runner.cs
@@ -780,7 +780,7 @@ namespace Duplicati.Server
                     );
                 }
             }
-            else if (result.ParsedResult != Library.Interface.ParsedResultType.Success && (result.Warnings.Any() || result.Errors.Any()))
+            else if (result.ParsedResult != Library.Interface.ParsedResultType.Success)
             {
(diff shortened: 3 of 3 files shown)


kenkendk:
I would like to know why the `ParsedResult` is set to `warning` if there are zero warnings.
Do you know why?

Other than that I agree, showing "Warning: 0 warnings" is pretty useless :).

kenkendk:
Ok, I see the problem now. The restore operation has a check to see if zero files are restored, but this does not create any warning messages (as I would expect).

I suggest we change it, so the controller emits a message if zero files are changed (around line 470 in `RestoreHandler.cs`):
```csharp
if (m_result.FilesRestored == 0 && m_result.ParsedResult == ParsedResultType.Success)
    Logging.Log.WriteWarningMessage(LOGTAG, "NoFilesRestored", null, "Restore completed without errors, but no files were restored");
```

Then we can remove the extra logic in the Restore result class, so it becomes:
```csharp
public override ParsedResultType ParsedResult
{
    get
    {
        if (Errors != null && Errors.Any())
            return ParsedResultType.Error;
        else if (Warnings != null && Warnings.Any())
            return ParsedResultType.Warning;
        else
            return ParsedResultType.Success;
    }
}
```
I prefer this approach as just not popping up the warning message hides the fact that something unexpected has happened.

mikaelmello:
I think this is better too. Will do it tomorrow.

A suggestion: the current UI to see errors or warnings directly related to backups is so lacking. You have to click on show, then click the item of the last operation to open the _entire_ log and manually search for the Errors and Warnings fields.

Over the next week I'll take a look into that too

kenkendk:
> A suggestion: the current UI to see errors or warnings directly related to backups is so lacking.

I agree. When I made it, there were many other things (there still is), so it was a matter of just getting something that works. The current solution is just dumping the log table into the browser tab, and it could be soooo much better.

For starters, we should tag each backup run with a unique ID, and write that into the log table, so we can group messages into what run it belongs to. This will make it much more user friendly, when it displays a list of operations.

We should also record the log ID for each message, such that the notification can point to the actual message, insted of just opening the log page.

There is a mention of this idea here: https://forum.duplicati.com/t/suggestion-to-improve-queue-and-notifications/3067

I recall there was a more detailed suggestion somewhere that included a design with status icons next to each entry, but I cannot find it.

mikaelmello:
Which log `id` and `message` do you recommend me to use?

mikaelmello:
Done. I'll also map a reasonable amount of strings to `Strings.cs` so they can be i18n'd.

kenkendk:
Arh, I just started a canary build, so this will go in the next build.

> Edit: LacunaSoftware@e97310b Can I follow with these changes?

Sure.

mikaelmello:
> A suggestion: the current UI to see errors or warnings directly related to backups is so lacking. You have to click on show, then click the item of the last operation to open the entire log and manually search for the Errors and Warnings fields.

Following up on this, what do you think of `PrintSerializeObject()` writing JSON? With that change I guess it would be way easier to parse the LogData messages and display them nicely on the Web UI.

kenkendk:
> Following up on this, what do you think of PrintSerializeObject() writing JSON? With that change I guess it would be way easier to parse the LogData messages and display them nicely on the Web UI.

Fine by me. We can also introduce something that has a more UI specific layout. I think the raw `PrintSerializeObject()` includes the log lines, which can be quite a lot of data to transfer, even locally.

mikaelmello:
I did a quick test with the following changes:
image

In a backup with 2000 files it does not seem to get big. I guess only Logs above Warning are stored.

https://pastebin.com/rUVeE1dM

kenkendk:
There is the "collectionlimit" which sets how many of each list are included before an ellipsis is added, so maybe we can just use that (set it to `0`), and then get the error/warning messages via a direct call if the user expands the UI element where they are displayed?

mikaelmello:
I'll move the posts to a proper issue and continue from there

Issue #3460, opened 2018-10-26
Path on server needs / at beginning when configurations are edited


aSpeX2009:
- [x] I have searched open and closed issues for duplicates.

----------------------------------------

## Environment info

- **Duplicati version**: 2.0.3.12_canary_2018-10-23
- **Operating system**: Ubuntu 18.04
- **Backend**: OneDrive V2

## Description

Since the last canary I have to add a "/" to the beginning of "Path on server" when I edit backup-configurations. Otherwise the path is not found, a new database is generated and i get an error-message.
When I add the slash and save the configuration it gets removed the next time I open the config of the backup-job.

## Steps to reproduce

1. Edit backup-configuration without adding a / to the path on server (change time etc.)
2. Start a backup
3. Get an error-message
4. Cry
5. Add a / to path on server
6. Repair database
7. Start backup
8. Get no error-message 

- **Actual result**:
Error-message below
- **Expected result**:
Do a backup without errors

## Debug log

[...]

PR #3462, opened 2018-10-27
Fix URLs in Microsoft Graph backends

warwickmm:
This fixes an issue introduced in revision f9022fec66cf33cb28e039b3046c31a111008493 ("Avoid referencing virtual members in constructor"), where the `MicrosoftGraphBackend` constructor was provided URLs with unexpected `/` characters.

This addresses issue #3460.


warwickmm:
My apologies for introducing this error.  My local test suite was still using the "Live" API, which allowed this to slip through.  :disappointed:

PR #3468, opened 2018-11-04
Feature/fix path storage2

wolfganggg:
Fixes #


kenkendk:
I am waiting a bit to inject this into the master branch so we can squeeze out a new beta before applying this, as it makes it difficult to go back.

Do you see significant space savings when using this?

Pectojin:
@kenkendk can we merge this now that the beta is out?

Pectojin:
I'm gonna merge this after the tests finish. Just for the sake of completeness, what's the id of the issue this fixes so we can close it?

warwickmm:
If it helps, I ran a variety of tests on this branch and didn't detect any issues.  I also have a backup configuration (~30,000 files, 3.6 GB source) and the resulting size of the local database after the initial backup was 20,791,296 bytes.  When running the same configuration using the current `master` (c1b0611e908d2db67a1fde171c0765d7d9cddea2), the size of the local database after the initial backup was 27,131,904 bytes.

Pectojin:
I also had a positive experience in my tests. My backups were too small to detect much of a space saving. Inspecting the database everything also looked good, so I'm comfortable with this change. 

I also think it will potentially allow us to do some cool stuff with restore since we can find the directory first before looking at the actual files.

warwickmm:
I suspect that this addresses issue #1283?

PR #3475, opened 2018-11-07
removed unused info icon

Pectojin:
The info icon doesn't bring any info and that style of info icons aren't really used anywhere else.
screenshot 2018-11-07 at 21 38 54


kenkendk:
The intention was that the `i` would open a help page, but that part was never implemented.

PR #3489, opened 2018-11-12
Actual calendar cates in "Last successful run" / "Next scheduled run"

### LINKED ISSUE #2124: Actual Calendar Dates in "Last successful run" / "Next scheduled run" [$20]
Requests adding the ability to display actual calendar dates (November 10th) versus "Last Thursday" as a configurable option.

"Tomorrow" and "Last Tuesday" is fine when the user has a few backup jobs, but it's more difficult to visualize the big picture when there are many scheduled jobs. 

This is purely a cosmetic request. It isn't intended to change functionality in any way.

<bountysource-plugin>
---
Want to back this issue? **[Post a bounty on it!](https://www.bountysource.com/issues/39221815-actual-calendar-dates-in-last-successful-run-next-scheduled-run?utm_campaign=plugin&utm_content=tracker%2F4870652&utm_medium=issues&utm_source=github)** We accept bounties via [Bountysource](https://www.bountysource.com/?utm_campaign=plugin&utm_content=tracker%2F4870652&utm_medium=issues&utm_source=github).
</bountysource-plugin>

mikaelmello:
Closes #2124

This PR adds an option called 'force-actual-date' that lets the user decides if they want to see the full date or relative dates

I'm not a native English speaker and I am pretty sure the strings I've used to describe the option are far from good, so any suggestions are welcome

I also made this PR without knowing if this feature is welcome, up for you to review


Duplicati/Server/webroot/ngax/scripts/controllers/SystemSettingsController.js
```diff
@@ -108,7 +108,7 @@ backupApp.controller('SystemSettingsController', function($rootScope, $scope, $l
                 if ($scope.updateChannel != $scope.originalUpdateChannel)
                     AppService.post('/updates/check');
 
-                $location.path('/');
+                $route.reload();
```
Pectojin: I get a javascript error `ReferenceError: $route is not defined` in SystemSettingsController.js when running the code and trying to save the settings. It still saves the settings but nothing happens in the UI.

Duplicati/Server/webroot/ngax/scripts/filters/parsetimestamp.js
```diff
@@ -14,3 +14,13 @@ backupApp.filter('parsetimestamp', function(AppUtils) {
       return AppUtils.toDisplayDateAndTime(value);
   }
 });
+
+backupApp.filter('parseForceActualDate', function(parsetimestampFilter, momentFilter) {
```
Pectojin: Aesthetically, I'd like the `parseForceActualDate` filter to be named something more generic since, without the setting enabled, `parseForceActualDate` won't parse the actual date.

Pectojin:
The flag works, but I think there's some inconsistency between the parameter description and the actual result:
`Nov 12, 2018, 8:01 AM`
`4 Nov 2018 15:54`
I don't know which way is best, but they should be the same.

A little out of scope, but I actually like the date display `Oct 7, 2018, 20:02` better than `07/10/2018` -  probably damage from dealing with both European and American date stamps - Perhaps it would make sense to use that across both regular and forced date display and then have the only difference be the relative display of "Today" or "Last Tuesday"?

kenkendk:
> The flag works, but I think there's some inconsistency between the parameter description and the actual result:
> Nov 12, 2018, 8:01 AM
> 4 Nov 2018 15:54

I think this depends on the browser locale settings. I personally favor the ISO standard (`yyyy-mm-dd`), but I think we should display the dates in whatever their browser says is the "right way".

Pectojin:
Hmm, if it's locale based that makes the description a little more tougher. But I suppose respecting locale is the superior choice.

Duplicati/Server/webroot/ngax/scripts/controllers/SystemSettingsController.js
mikaelmello: I accidentally discarded the hunk with that addition when removing something else, will update

Duplicati/Server/webroot/ngax/scripts/filters/parsetimestamp.js
mikaelmello: Any suggestions?

Pectojin: `parseDate`? Unless it overlaps with some existing filter

mikaelmello:
I'm struggling to get the dates already in the UI to be updated right after I change `--force-actual-date`. I have to do a complete refresh by pressing F5 for it to reload the new setting. Any suggestions?

kenkendk:
> I'm struggling to get the dates already in the UI to be updated right after I change --force-actual-date. I have to do a complete refresh by pressing F5 for it to reload the new setting. Any suggestions?

I think you need to load the data in `SystemSettingsController.js` and not in `AppController.js` IIRC, the other scopes listen to changes in the system settings scope and will trigger a refresh when that scope changes.

kenkendk:
@mikaelmello Are you still stuck with the refresh problem?

mikaelmello:
> @mikaelmello Are you still stuck with the refresh problem?

Sorry, I've just been focusing on other things lately, will get some time in the next days for that

Pectojin:
Was anything missing from this PR besides the refresh problem? The functionality seemed to be there when I tested so it's a shame to have this PR left unmerged.

BlueBlock:
This looks good but I'm not sure how to get it to refresh. This is angular1 I think? it might need to be a hack of a solution to get it to refresh.

Pectojin:
I had completely forgotten about this 😅

Yep, it's angular 1, but honestly I think it might be worth merging without the auto refresh. 

We can always add it later.

BlueBlock:
I would agree. A TODO comment line should be added in the code for the refresh once it is merged.

Pectojin:
Ugh, the indentation has been fixed in upstream so the diffs on Options.cs and Strings.cs is 100%.

I'm gonna have to edit that locally to resolve conflicts. It'll land in a new PR...

Pectojin:
Okay I ended up fixing it here anyway because somehow my on branch squashed everything into one commit... Not too happy about wiping ownership on the commits.

PR #3490, opened 2018-11-12
Require confirmation of UI password when updating it

### LINKED ISSUE #3480: Require password confirmation on Settings
It would be really good to require a password confirmation before allowing any changes. This is a really simple feature that avoids setting a wrong password (because of a typo, for example).

I've set this feature on our own version of Duplicati, so a PR can be easily done if the feature is welcome.

mikaelmello:
Closes #3480 

Discussion also on the issue

image

image


Pectojin:
There appears to be an issue if I set a password, save, go back to settings, save again without changing the password.

It's overwriting the server password with an empty string, so it's impossible to log in. I couldn't replicate that issue on the master branch.

Other than that edge case every other case I tested worked, so we just need to deal with it not submitting the password if it's unchanged (e.g. empty).

kenkendk:
I can't remember the server-side handling, but I would prefer if the server rejects attempts to set the password to empty.

mikaelmello:
> There appears to be an issue if I set a password, save, go back to settings, save again without changing the password.
> It's overwriting the server password with an empty string, so it's impossible to log in. I couldn't replicate that issue on the master branch.

I had the passwords being set to null when I reproduced these steps. Just fixed

> I can't remember the server-side handling, but I would prefer if the server rejects attempts to set the password to empty.

It just blindly accepts any string and sets it as a normal setting. It is up to the client to salt and hash.

Pectojin:
I got home too late to test last night... I'll take some time out today.

Pectojin:
Everything looks good.

Only funny thing is that it'll ignore anything you put in the confirm password box if you haven't changed the password. No big deal.

Issue #3498, opened 2018-11-17
Password mandatory for asymmetric GPG encryption in the web interface


b3yond:
- [x] I have searched open and closed issues for duplicates.

----------------------------------------

## Environment info

- **Duplicati version**: 2.0.4.4
- **Operating system**: Debian 9
- **Backend**: webDAV

## Description

GPG only seems to work with a workaround on the command line. This workaround is artificially blocked in the web interface.

According to #2270 you can use GPG as encryption module at backup when you specify `--passphrase=unused`. The web interface artificially forbids you to not use a passphrase and requires you to set a passphrase in the General configuration of the backup (first page). 

Setting `--passphrase=unused` manually at Advanced Options doesn't change anything in the web interface. I guess it's overridden by the encryption passphrase from the first page.

If you set the encryption passphrase from the first page to "unused", it accepts the command. But when you execute the backup, it backups zero files.

The backup logs say everything is fine though.

In short: GPG encryption is unusable in the web interface right now.

## Steps to reproduce

1. Set a backup in the web interface.
2. Set passphrase to unused in the first page
3. Set the encrpytion module to GPG in the first page.
4. Set the rest of the GPG encryption flags in the 5th page:
  * `--gpg-encryption-command=--encrypt`
  * `--gpg-encryption-switches=--recipient\ "mail@example.org"`
5. Save the settings and start backup
6. See that the backup says: "Backup: 0 bytes / 0 Versions"
7. Check that the backend stays empty, no files saved at all

- **Actual result**:
It does not backup anything

- **Expected result**:
When I set the passphrase to "unused", it backs everything up, and I can restore it later using only the GPG private key, but not the passphrase.

## Screenshots
![screenshot_2018-11-17_10-53-14](https://user-images.githubusercontent.com/19255313/48659948-a2da3b00-ea59-11e8-8109-724f31d535e4.png)
![screenshot_2018-11-17_10-55-27](https://user-images.githubusercontent.com/19255313/48659953-a968b280-ea59-11e8-8aaa-8de82d954d21.png)
![screenshot_2018-11-17_10-55-15](https://user-images.githubusercontent.com/19255313/48659952-a7065880-ea59-11e8-9b38-f0a931f78c9d.png)
![screenshot_2018-11-17_11-18-05](https://user-images.githubusercontent.com/19255313/48659984-7c68cf80-ea5a-11e8-9184-d85904f4db77.png)
[...]

PR #3506, opened 2018-11-20
various GPG changes

verhoek:
I made some changes motivated by:
https://github.com/duplicati/duplicati/issues/3498

In particular:

- When using asymmetric GPG encryption, a passphrase is not required. 
In fact, the passphrase box disappears (this is a design decision, I'm open for discussion here :)
My argument is, Duplicati does not need to know the passphrase just to encrypt. In fact, Duplicati might not need to know the passphrase in case of GPG at all - more below)
- The UI does not complain anymore for having an empty passphrase and we don't have to use random strings such as "unused"
- Improved logging, i.e., if GPG crashes, we see it (the issue https://github.com/duplicati/duplicati/issues/3498 was afaik related to a faulty recipient option, but no error ever appeared)
- Restoring a GPG asymmetric encrypted backup set has the possibility to specify a passphrase in the second step.
- If no passphrase is supplied, then (at least on Linux), the system's keychain is invoked. Upshot: Duplicati doesn't need to store GPG's passphrase at all.
- Also restoring any other backup set gives the possibility to supply a passphrase, when Duplicati detects that no passphrase is stored. I could narrow this I guess on checking if encryption was ever used for the backup set.

There is some minimal logic in the UI checking if asymmetric GPG is active. In theory, all the above works just as well for symmetric GPG - also there the system's keychain will be invoked (on Linux). We could add a suggestion that it's not necessarily necessary to input a passphrase there, as long as the system's key chain kicks in.

Any feedback regarding corner cases I might have overlooked are more than welcome.

### DIFF
--- a/Duplicati/Server/Runner.cs
+++ b/Duplicati/Server/Runner.cs
@@ -183,18 +183,21 @@ namespace Duplicati.Server
                 filters);
         }
 
-        public static IRunnerData CreateRestoreTask(Duplicati.Server.Serialization.Interface.IBackup backup, string[] filters, DateTime time, string restoreTarget, bool overwrite, bool restore_permissions, bool skip_metadata)
+        public static IRunnerData CreateRestoreTask(Duplicati.Server.Serialization.Interface.IBackup backup, string[] filters, 
+                                                    DateTime time, string restoreTarget, bool overwrite, bool restore_permissions, 
+                                                    bool skip_metadata, string passphrase = null)
         {
-            var dict = new Dictionary<string, string>();
-            dict["time"] = Duplicati.Library.Utility.Utility.SerializeDateTime(time.ToUniversalTime());
+            var dict = new Dictionary<string, string>
+            {
+                ["time"] = Library.Utility.Utility.SerializeDateTime(time.ToUniversalTime()),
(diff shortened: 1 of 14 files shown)


Duplicati/Server/Runner.cs
```diff
@@ -183,18 +183,21 @@ public static IRunnerData CreateListTask(Duplicati.Server.Serialization.Interfac
                 filters);
         }
 
-        public static IRunnerData CreateRestoreTask(Duplicati.Server.Serialization.Interface.IBackup backup, string[] filters, DateTime time, string restoreTarget, bool overwrite, bool restore_permissions, bool skip_metadata)
+        public static IRunnerData CreateRestoreTask(Duplicati.Server.Serialization.Interface.IBackup backup, string[] filters, 
+                                                    DateTime time, string restoreTarget, bool overwrite, bool restore_permissions, 
+                                                    bool skip_metadata, string passphrase = null)
```
warwickmm: From what I can tell, it appears there is only one usage of this method and it's been modified to provide the `passphrase`.  Should we just make this parameter required, instead of making it optional?

verhoek: Thanks, that's true. In that case, it's better to make it required.

Issue #2601, opened 2017-07-17
Flag logs that contain a warning


dimejo:
I have:
- [x] searched open and closed issues for duplicates

----------------------------------------

### Version info

**Duplicati Version:** 2.0.1.61_experimental_2017-05-30
**Operating System:** Windows 10 Pro
**Backend:** I've no clue what that means ...

### Bug description
I'm running daily backups of my Windows profile. Some of these backups generate warning beacuse a file was locked and could not be read. When clicking on the red warning indicator to the bottom I'm redirected to the list of logs for this backup job, but there is no indication which log contains a warning, therefor I have to open every single log.

### Screenshots
The log for '10. Jul. 2017' contains a warning but it is not visible until I open it.

![screenshot_2017-07-17_18-10-28_duplicati](https://user-images.githubusercontent.com/25005220/28277828-739be410-6b1b-11e7-9031-58ebef3f04df.png)

shiek:
The log WebUI view is really a pain and needs work.
As a workaround open the local sqlite db of your backup with http://sqlitebrowser.org
Then select the "LogData" table to easily browse all the entries.

PR #3510, opened 2018-11-23
[WIP] Improve UI for backup logs

mikaelmello:
The idea is to stop showing a dump of serialized logs and have a decent and simple dashboard so the user can quickly view relevant info.

1. Save logs in DB serialized in JSON as opposed to a custom Duplicati serialization format
2. Parse the JSON in the front-end to allow easy access to data
3. Display the data
---

Need help:
- [ ] Limit number of elements in the `IEnumerable` fields of `BasicResults` in order to save space when storing the serialized version on db. The properties, which are `Warnings`, `Errors` and `Messages` can grow quite large on big backups.
- [ ] Discuss about usage of colors
- [ ] Which other information on logs are also relevant to be displayed?

Paging @kenkendk, @Pectojin, @warwickmm, @kees-z and @renestach


kees-z:
Nice!

Colors look good to me, the scheme matches the other pages in the Web UI.
Maybe it looks more clean if the values are aligned a bit, something like this:

image

I also would like to see a colored symbol for every operation. The top level lines ("Nov 23, 2018 8.25 AM - Operation: Backup") should be preceeded with a green, yellow or red symbol, indicating if there are warnings or errors for that job.
In that case, all logs can be collapsed by default, browsing the list makes it easy to locate operations with errors and warnings. This feature is discussed [here](https://forum.duplicati.com/t/feature-request-better-reporting/46?u=kees-z).

Regarding extra information, could summaries of sub-operations be listed in the main view? Like result of compacting, verifying etc, including a green, red or yellow symbol for each sub-operation.

mikaelmello:
> Maybe it looks more clean if the values are aligned a bit, something like this

I thought about it but I don't know if I really agree and this initial design was simpler, but anyway, do you have hints regarding the better way to do that? CSS is not my strong suit

> I also would like to see a colored symbol for every operation. The top level lines ("Nov 23, 2018 8.25 AM - Operation: Backup") should be preceeded with a green, yellow or red symbol, indicating if there are warnings or errors for that job.

+1

> In that case, all logs can be collapsed by default, browsing the list makes it easy to locate operations with errors and warnings. This feature is discussed here.

They currently are collapsed by default, I let both of them expanded because only one had Warnings.length > 0

> Regarding extra information, could summaries of sub-operations be listed in the main view? Like result of compacting, verifying etc, including a green, red or yellow symbol for each sub-operation.

Sure. However there are two caveats:

1. Sub-operations don't usually have their own `Warnings/Errors/Messages` properties, getting from their parents instead. By "usually" I mean all that I've seen but I'm not sure if it is consistent. Therefore if there is an error/warning in any of them, it will show up in the rest. The first solution that comes to my mind is to have their `getters` always returning an instance of their own arrays or if they don't have them, null instead of getting their parents'
2. Can you explain or point me to somewhere explaining the roles or meaning of each sub-result?

kees-z:
> do you have hints regarding the better way to do that? CSS is not my strong suit

No! Sorry, I'm not a coder at all...

> They currently are collapsed by default, I let both of them expanded because only one had Warnings.

I thought so, just mentioned it because preceeding each line with a result symbol will make it a lot easier to browse the logs and track down errors/warnings.

> However there are two caveats:

Just looked at [kenkendk's proposal](https://forum.duplicati.com/t/feature-request-better-reporting/46/6?u=kees-z). I have no idea how it could be technically accomplished, but I like the idea. In his example there is a yellow symbol at the root level, below a few green and one yellow symbol, indicating what part of the operation generated the warning.
But if something like this is hard to implement, your design looks ok to me.

mikaelmello:
I've added the icons and updated the preview on the OP.

Regarding the sub-results, I'll wait for @kenkendk 's opinion on why sub-results get their parents errors/warnings and if it is possible to remove it. That would also make the process of limiting arrays' sizes easier.

kees-z:
In the new screenshot, "Warnings" and "Errors" are expandable if they are <> 0. IMO, this is a very nice alternative for listing the status with colored symbols like @kenkendk 's example. If the description of warnings and errors are grouped there and easy to understand, you can quick and easy see which tasks did nog complete successfully and what went wrong.

You removed the small triangles, which makes it less clear that the lines van be expanded. For me, I wouldn't mind if the triangle symbols were added back to each log line.

mikaelmello:
> You removed the small triangles, which makes it less clear that the lines van be expanded. For me, I wouldn't mind if the triangle symbols were added back to each log line.

I removed them because they didn't look good neither along with the new icons or at the end of the line. Perhaps a different icon would work, will try monday

warwickmm:
Nice work!

> I removed them because they didn't look good neither along with the new icons or at the end of the line. Perhaps a different icon would work, will try monday

I think providing a visual cue that the logs can be expanded is important.  This is often a source of confusion for users, who often claim that they cannot see any logs because it's not clear that clicking on the entry expands it to show more details.

warwickmm:
I think this will take care of issue #2601, and perhaps many others.  :+1:

Duplicati/Server/webroot/ngax/index.html
```diff
@@ -232,7 +232,7 @@
                 </ul>
             </div>
 
-            <div ng-view class="content">
+            <div ng-view class="content" style="width: 100%">
```
mikaelmello: Setting this attribute in the CSS file did not work by the way.

I could really use some help with css...

kenkendk:
> Regarding the sub-results, I'll wait for @kenkendk 's opinion on why sub-results get their parents errors/warnings and if it is possible to remove it. That would also make the process of limiting arrays' sizes easier.

The reason is that the result classes "forwards" some properties to the base. 

Internally, the `backup` operation runs, `backup`, `delete`, `compact`, `verify`. The result classes for `delete`, `compact` and `verify` are created with the `backup` results as the "`parent`" instance. This sharing means that log messages are stored in the parent, and this causes the `ParsedResult` property to evaluate to something unexpected.

Looking back at this, I think the benefits from this convoluted sharing are less than the advantages.
If we remove the sharing, and allow each class to just maintain the same properties, we just need to manually combine the results.

We could maintain the backendstatistics to be shared, but otherwise we simply need to sum up the transfer sizes and counts.

mikaelmello:
> The reason is that the result classes "forwards" some properties to the base.
> 
> Internally, the `backup` operation runs, `backup`, `delete`, `compact`, `verify`. The result classes for `delete`, `compact` and `verify` are created with the `backup` results as the "`parent`" instance. This sharing means that log messages are stored in the parent, and this causes the `ParsedResult` property to evaluate to something unexpected.
> 
> Looking back at this, I think the benefits from this convoluted sharing are less than the advantages.
> If we remove the sharing, and allow each class to just maintain the same properties, we just need to manually combine the results.
> 
> We could maintain the backendstatistics to be shared, but otherwise we simply need to sum up the transfer sizes and counts.

I think I'm in the right path with the changes I`ve made in the latest commit? https://github.com/duplicati/duplicati/pull/3510/commits/57bd9d1e416b3536615e8bc2effd694de4378689

Here are the logs from a backup run after these changes:

```
{
    "DeletedFiles": 0,
    "DeletedFolders": 0,
    "ModifiedFiles": 0,
    "ExaminedFiles": 200,
    "OpenedFiles": 200,
    "AddedFiles": 200,
    "SizeOfModifiedFiles": 0,
    "SizeOfAddedFiles": 2457600000,
    "SizeOfExaminedFiles": 2457600000,
    "SizeOfOpenedFiles": 2457600000,
    "NotProcessedFiles": 0,
    "AddedFolders": 1,
    "TooLargeFiles": 0,
    "FilesWithError": 0,
    "ModifiedFolders": 0,
    "ModifiedSymlinks": 0,
    "AddedSymlinks": 0,
    "DeletedSymlinks": 0,
    "PartialBackup": false,
    "Dryrun": false,
    "MainOperation": "Backup",
    "CompactResults": {
      "DeletedFileCount": 0,
      "DownloadedFileCount": 0,
      "UploadedFileCount": 0,
      "DeletedFileSize": 0,
      "DownloadedFileSize": 0,
      "UploadedFileSize": 0,
      "Dryrun": false,
      "MainOperation": "Compact",
      "ParsedResult": "Success",
      "Version": "2.0.0.7 (Current)",
      "EndTime": "2018-11-26T13:52:12.8534765Z",
      "BeginTime": "2018-11-26T13:52:12.8093973Z",
      "Duration": "00:00:00.0440792",
      "Messages": null,
      "Warnings": null,
      "Errors": null,
      "BackendStatistics": null
    },
    "DeleteResults": null,
    "RepairResults": null,
    "TestResults": {
      "MainOperation": "Test",
      "Verifications": [
        {
          "Key": "duplicati-20181126T135156Z.dlist.zip",
          "Value": []
        },
        {
          "Key": "duplicati-if790ee648e514eec9bb88571027c26be.dindex.zip",
          "Value": []
        },
        {
          "Key": "duplicati-b350a2f32f75e455dbe3ccc8e08b633fb.dblock.zip",
          "Value": []
        }
      ],
      "ParsedResult": "Success",
      "Version": "2.0.0.7 (Current)",
      "EndTime": "2018-11-26T13:52:13.347553Z",
      "BeginTime": "2018-11-26T13:52:13.0362838Z",
      "Duration": "00:00:00.3112692",
      "Messages": null,
      "Warnings": null,
      "Errors": null,
      "BackendStatistics": null
    },
    "ParsedResult": "Success",
    "Version": "2.0.0.7 (Current)",
    "EndTime": "2018-11-26T13:52:13.3996782Z",
    "BeginTime": "2018-11-26T13:51:55.6682278Z",
    "Duration": "00:00:17.7314504",
    "Messages": [
      "2018-11-26 11:51:55 -02 - [Information-Duplicati.Library.Main.Controller-StartingOperation]: A operação Backup foi iniciada",
      "2018-11-26 11:51:56 -02 - [Information-Duplicati.Library.Main.BasicResults-BackendEvent]: Backend event: List - Started:  ()",
      "2018-11-26 11:51:56 -02 - [Information-Duplicati.Library.Main.BasicResults-BackendEvent]: Backend event: List - Completed:  ()",
      "2018-11-26 11:52:12 -02 - [Information-Duplicati.Library.Main.BasicResults-BackendEvent]: Backend event: Put - Started: duplicati-b350a2f32f75e455dbe3ccc8e08b633fb.dblock.zip (52,54 KB)",
      "2018-11-26 11:52:12 -02 - [Information-Duplicati.Library.Main.BasicResults-BackendEvent]: Backend event: Put - Completed: duplicati-b350a2f32f75e455dbe3ccc8e08b633fb.dblock.zip (52,54 KB)",
      "2018-11-26 11:52:12 -02 - [Information-Duplicati.Library.Main.BasicResults-BackendEvent]: Backend event: Put - Started: duplicati-if790ee648e514eec9bb88571027c26be.dindex.zip (11,26 KB)",
      "2018-11-26 11:52:12 -02 - [Information-Duplicati.Library.Main.BasicResults-BackendEvent]: Backend event: Put - Completed: duplicati-if790ee648e514eec9bb88571027c26be.dindex.zip (11,26 KB)",
      "2018-11-26 11:52:12 -02 - [Information-Duplicati.Library.Main.BasicResults-BackendEvent]: Backend event: Put - Started: duplicati-20181126T135156Z.dlist.zip (8,20 KB)",
      "2018-11-26 11:52:12 -02 - [Information-Duplicati.Library.Main.BasicResults-BackendEvent]: Backend event: Put - Completed: duplicati-20181126T135156Z.dlist.zip (8,20 KB)",
      "2018-11-26 11:52:12 -02 - [Information-Duplicati.Library.Main.Database.LocalDeleteDatabase-CompactReason]: Compacting not required",
      "2018-11-26 11:52:12 -02 - [Information-Duplicati.Library.Main.BasicResults-BackendEvent]: Backend event: List - Started:  ()",
      "2018-11-26 11:52:12 -02 - [Information-Duplicati.Library.Main.BasicResults-BackendEvent]: Backend event: List - Completed:  (3 bytes)",
      "2018-11-26 11:52:13 -02 - [Information-Duplicati.Library.Main.BasicResults-BackendEvent]: Backend event: Get - Started: duplicati-20181126T135156Z.dlist.zip (8,20 KB)",
      "2018-11-26 11:52:13 -02 - [Information-Duplicati.Library.Main.BasicResults-BackendEvent]: Backend event: Get - Completed: duplicati-20181126T135156Z.dlist.zip (8,20 KB)",
      "2018-11-26 11:52:13 -02 - [Information-Duplicati.Library.Main.BasicResults-BackendEvent]: Backend event: Get - Started: duplicati-if790ee648e514eec9bb88571027c26be.dindex.zip (11,26 KB)",
      "2018-11-26 11:52:13 -02 - [Information-Duplicati.Library.Main.BasicResults-BackendEvent]: Backend event: Get - Completed: duplicati-if790ee648e514eec9bb88571027c26be.dindex.zip (11,26 KB)",
      "2018-11-26 11:52:13 -02 - [Information-Duplicati.Library.Main.BasicResults-BackendEvent]: Backend event: Get - Started: duplicati-b350a2f32f75e455dbe3ccc8e08b633fb.dblock.zip (52,54 KB)",
      "2018-11-26 11:52:13 -02 - [Information-Duplicati.Library.Main.BasicResults-BackendEvent]: Backend event: Get - Completed: duplicati-b350a2f32f75e455dbe3ccc8e08b633fb.dblock.zip (52,54 KB)"
    ],
    "Warnings": [],
    "Errors": [],
    "BackendStatistics": {
      "RemoteCalls": 8,
      "BytesUploaded": 73725,
      "BytesDownloaded": 73725,
      "FilesUploaded": 3,
      "FilesDownloaded": 3,
      "FilesDeleted": 0,
      "FoldersCreated": 0,
      "RetryAttempts": 0,
      "UnknownFileSize": 0,
      "UnknownFileCount": 0,
      "KnownFileCount": 3,
      "KnownFileSize": 73725,
      "LastBackupDate": "2018-11-26T11:51:56-02:00",
      "BackupListCount": 1,
      "TotalQuotaSpace": 0,
      "FreeQuotaSpace": 0,
      "AssignedQuotaSpace": -1,
      "ReportedQuotaError": false,
      "ReportedQuotaWarning": false,
      "MainOperation": "Backup",
      "ParsedResult": "Success",
      "Version": "2.0.0.7 (Current)",
      "EndTime": "0001-01-01T00:00:00",
      "BeginTime": "2018-11-26T13:51:55.6682278Z",
      "Duration": "00:00:00",
      "Messages": null,
      "Warnings": null,
      "Errors": null,
      "BackendStatistics": null
    }
  }
```

mikaelmello:
@kenkendk , how can I get a BackupResult where DeleteResults is not null? I've tried a lot of things but none seem to work.

The reason I want to do that is to know what exactly `DeletedSets` is and how it is serialized.

mikaelmello:
I've updated with the latest screenshots.

Now the only things missing are data about `DeleteResults` and `ListChangesResults`.

I absolutely have no idea of when `ListChangesResults` is logged nor their precise meaning.

Could you guys provide more feedback?

warwickmm:
> how can I get a BackupResult where DeleteResults is not null?

I think [this](https://github.com/duplicati/duplicati/blob/master/Duplicati/Library/Main/Operation/BackupHandler.cs#L250) is the only place where the `BackupResults.DeleteResults` setter is called.  Looks related to compacting or retention policy?

warwickmm:
> I absolutely have no idea of when ListChangesResults is logged nor their precise meaning.

From what I can tell, `ListChangesResults` is populated when calling `compare` from the command line?

mikaelmello:
> > how can I get a BackupResult where DeleteResults is not null?
> 
> I think [this](https://github.com/duplicati/duplicati/blob/master/Duplicati/Library/Main/Operation/BackupHandler.cs#L250) is the only place where the `BackupResults.DeleteResults` setter is called. Looks related to compacting or retention policy?

That's right. I thought DeleteResults was related to deleting files, not backup versions

warwickmm:
This is looking great.  If I can make a small suggestion, what are your thoughts about  making the indentation under an expanded item slightly larger?

image

mikaelmello:
> This is looking great. If I can make a small suggestion, what are your thoughts about making the indentation under an expanded item slightly larger?

I agree, it does look better now.

image

mikaelmello:
The last commit (https://github.com/duplicati/duplicati/pull/3510/commits/ec5fee9f9a51b9bfedf201e76ef7a9050117b62e) has the best way I could find to limit the number of elements that are serialized.

I could neither reassign the properties nor configure a setting in the serializer saying "On arrays, serialize up to X elements".

There are two more things I'd like to implement:

1. Make the limit of serialization a global setting that can be changed from the UI.
2. With this change, when the log view displays the quantity of warnings/errors/etc it will only display up to the size we limited the arrays to be, so something is lost. I think it would be a good idea to have something like "TotalWarningsCount" property separated from the `Warnings` property, in which we get the total size of `Warnings`. This way we can show the correct number of errors/warnings but display only a limited quantity.

@kenkendk , any opinions?

kenkendk:
> Make the limit of serialization a global setting that can be changed from the UI.

The `PrintSerializeObject` method has a `collectionlimit` parameter that is used for this purpose.

> With this change, when the log view displays the quantity of warnings/errors/etc it will only display up to the size we limited the arrays to be, so something is lost. I think it would be a good idea to have something like "TotalWarningsCount" property separated from the Warnings property, in which we get the total size of Warnings. This way we can show the correct number of errors/warnings but display only a limited quantity.

That will make it very obvious that data is missing. The current (previous?) approach was to insert a log line with `...`, indicating that some stuff was cut off.

mikaelmello:
@kenkendk I've create an option that sets this limit, I'll just post a screenshot to show the current work:

image

There's a problem though, I can't find a way to properly access options from the `ResultClasses.cs` file, I've thought of the following:

1. Breaking encapsulation by letting `m_options` (or an `Options` getter) from `Main/Controller.cs` be public
2. Create a single method on `Main/Controller.cs` that exposes the exact option I want. The problem is that this class does not have a single method performing a similar function and this would do something like "going off standard", I don't know a correct English expression to explain it.
3. I can't access from `Server` because `Server` references `Library.Main`.
4. I don't feel like passing `m_options` through multiple layers of constructors and function calls until it gets to `ResultClasses.cs`.

The easiest path was to not allow this as an option and just be a defined constant.

kees-z:
Great work! I hope to see this soon in an upcoming Canary release.

Another idea (don't know if it's possible/easy to implement): could the JSON data in the _Complete log_ section be reformatted somehow? This would make it more readable for the end user.

For example, these rules could be applied to the unformatted text:

* Remove all surrounding quotes and trailing commas.
* Make text before colon bold and green color, remove first colon and make remaining text grey.
* Replace every capital with a space and convert to lowercase, except the first character (example: replace _DeletedFileSize_ with _Deleted file size_) 
* Indent everything between curly braces.
* Collapse everything between square braces (and make it axpandable).
* Replace dates and times with something more readable, for example replace _2018-11-26T13:52:13.0362838Z_ with _2018-11-26 13:52:13_, optionally converted to local time zone.
* Replace value _null_ with something like _-_.

This will give a similar look to the JSON data as the summary above it.

mikaelmello:
> Another idea (don't know if it's possible/easy to implement): could the JSON data in the Complete log section be reformatted somehow? This would make it more readable for the end user.

That is a great idea that I hadn`t thought about. 

> Another suggestion for the _Complete log_ view:
> Can the grey text box and scroll bar be replaced with an expand/collapse symbol?
> Clicking an icon to expand the log could then display the log completely, making it easier to view the contents of the log.

Also good, would be best if the user could change the size of the gray box too.

kees-z:
> Also good, would be best if the user could change the size of the gray box too.

Personally, I would _remove_ the grey box completely and display everything without using scroll bars. If everything between `[` and `]` is collapsed by default (and the _Complete log_ section itself too), there will be about 75 lines under _Complete log_ when you expand it.
For me, this is acceptable, logs are already indented and separated using horizontal separation lines.

mikaelmello:
So after messing around with these suggestion I've come to the conclusion that formatting and syntax highlighting the complete log is not worth the trouble.

Anything relevant is already supposed to be displayed in a nicely manner so:
> - Replace every capital with a space and convert to lowercase, except the first character (example: replace DeletedFileSize with Deleted file size)
> - Replace dates and times with something more readable, for example replace 2018-11-26T13:52:13.0362838Z with 2018-11-26 13:52:13, optionally converted to local time zone.
> - Replace value null with something like -.
> - EDIT: Convert number of bytes to KB/MB/GB/TB and add unit (replace 16903871149 with 15,74 GB)

Shouldn't be necessary, if they are we've kind of failed in summarizing the important information above.

Getting the complete log formatted will take more time than I'd like to. The only thing I think it is important for the complete log is that it should be easily viewed and copied (possibly exported). That wasn't possible with a small gray box and no way to copy everything.

mikaelmello:
This is the final result, I feel that this is enough

https://imgur.com/a/RWFnjNR

kees-z:
Agreed. Fair point about the ability to copy/export the complete log.
If the full log is collapsed by default after expanding a line from the list (which seems to be the case), I guess browsing the logs is convenient and has a clean look.
Thanks for your efforts, this seems to be a great UX improvement.

BTW what's behind the _Result_ lines with grey question mark symbols? Does the content have an unknown status, or is it information without any status at all? If it's just information, maybe a minus symbol is more appropriate.  A question mark suggests a status that could not be retrieved.

warwickmm:
Great work!

mikaelmello:
> BTW what's behind the Result lines with grey question mark symbols? Does the content have an unknown status, or is it information without any status at all? If it's just information, maybe a minus symbol is more appropriate. A question mark suggests a status that could not be retrieved.

They happened on old results that were serialized not in JSON. So the status could not be retrieved.

mikaelmello:
> They happened on old results that were serialized not in JSON. So the status could not be retrieved.

Adding to that, they were in the middle of functioning results because I ran backups in another branch that weren't serializing the results in JSON.


@kenkendk, anything else you'd want before possibly merging?

Issue #2184, opened 2016-12-20
Some Advanced Options are listed twice


kees-z:
I have:
- [X] searched open and closed issues for duplicates

----------------------------------------

### Version info

**Duplicati Version:** 2.0.1.35
**Operating System:** Windows 10 Professional / Windows 10 Enterprise

### Bug description
Under Settings, the advanced options in the section "Configure http requests" are listed twice.

### Steps to reproduce
- Open the web interface
- Click Settings
- Open the  drop-down menu Add advanced option

**Actual result:** All entries in this section are listed twice.
**Expected result:** All entries should be listed once.

### Screenshots

![duplicati - configure http requests](https://cloud.githubusercontent.com/assets/24669568/21345590/229a21cc-c6a1-11e6-8c6b-2e9f3328d930.png)

warwickmm:
Resolved in pull request #3511.

PR #3511, opened 2018-11-23
Remove duplicate connection entries from advanced options list

warwickmm:
This removes duplicate connection entries from the advanced options list.  Since the `IConnectionModule` [interface](https://github.com/duplicati/duplicati/blob/master/Duplicati/Library/Interface/IConnectionModule.cs#L27) implements the `IGenericModule` interface, we do not need to add the connection modules separately.

This addresses issue #2184.


Pectojin:
Looks good to me 👍

PR #3512, opened 2018-11-24
Fix commandline GUI regression

Pectojin:
This fixes a regression identified ts678 https://forum.duplicati.com/t/release-2-0-4-3-canary-2018-11-13/5279/11

Since the export defaults to not exporting passwords (e.g. `export-passwords=false`) the commandline GUI no longer works until you manually add the password.


mikaelmello:
Looks good to me

Issue #1451, opened 2015-09-02
Duplicati 2.0 is not capable of reading files with path name larger than 256 characters


mickmm:
I'm using duplicati 2.0 preview and i'm having an error like this with files larger than 256 characters:

2015-07-03 02:06:38Z - Warning: Failed to process path: h:\some_large _path.JPG
System.ArgumentException: Nombre no válido.
Nombre del parámetro: name
   en System.Security.AccessControl.NativeObjectSecurity.CreateInternal(ResourceType resourceType, Boolean isContainer, String name, SafeHandle handle, AccessControlSections includeSections, Boolean createByName, ExceptionFromErrorCode exceptionFromErrorCode, Object exceptionContext)
   en System.Security.AccessControl.NativeObjectSecurity..ctor(Boolean isContainer, ResourceType resourceType, String name, AccessControlSections includeSections, ExceptionFromErrorCode exceptionFromErrorCode, Object exceptionContext)
   en System.Security.AccessControl.DirectorySecurity..ctor(String name, AccessControlSections includeSections)
   en Duplicati.Library.Snapshots.SystemIOWindows.GetAccessControlFile(String path)
   en Duplicati.Library.Snapshots.SystemIOWindows.GetMetadata(String path)
   en Duplicati.Library.Snapshots.WindowsSnapshot.GetMetadata(String file)
   en Duplicati.Library.Main.Operation.BackupHandler.GenerateMetadata(String path, FileAttributes attributes)
   en Duplicati.Library.Main.Operation.BackupHandler.HandleFilesystemEntry(String path, FileAttributes attributes)

2015-07-03 02:06:38Z - Warning: Failed to process path: h:\some_large _path.JPG
System.ArgumentException: Nombre no válido.
Nombre del parámetro: name
   en System.Security.AccessControl.NativeObjectSecurity.CreateInternal(ResourceType resourceType, Boolean isContainer, String name, SafeHandle handle, AccessControlSections includeSections, Boolean createByName, ExceptionFromErrorCode exceptionFromErrorCode, Object exceptionContext)
   en System.Security.AccessControl.NativeObjectSecurity..ctor(Boolean isContainer, ResourceType resourceType, String name, AccessControlSections includeSections, ExceptionFromErrorCode exceptionFromErrorCode, Object exceptionContext)
   en System.Security.AccessControl.DirectorySecurity..ctor(String name, AccessControlSections includeSections)
[...]

PR #3517, opened 2018-11-26
Fix/enumerate files

verhoek:
See https://forum.duplicati.com/t/problem-with-backup-since-update/5385/2 .
The second commit contains a proposal to remove repeating many times the same code block.
I don't know how this would affect performance since there are quite a few IO calls - feedback would be good.

### DIFF
--- a/Duplicati/Library/Common/IO/SystemIOWindows.cs
+++ b/Duplicati/Library/Common/IO/SystemIOWindows.cs
@@ -288,18 +288,28 @@ namespace Duplicati.Library.Common.IO
             return null;
         }
 
-        public IEnumerable<string> EnumerateFileSystemEntries(string path)
+        private static T TryNativeFirst<T>(Func<string, T> nativeIOFunc, Func<string, T> alternativeIOFunc, string path)
         {
             if (!IsPathTooLong(path))
-                try { return System.IO.Directory.EnumerateFileSystemEntries(path); }
+                try { return nativeIOFunc(path); }
                 catch (System.IO.PathTooLongException) { }
                 catch (System.ArgumentException) { }
 
-            var r = Alphaleonis.Win32.Filesystem.Directory.GetFileSystemEntries(PrefixWithUNC(path));
-            for (var i = 0; i < r.Length; i++)
-                r[i] = StripUNCPrefix(r[i]);
+            return alternativeIOFunc(path);
+        }
 
-            return r;
+        public IEnumerable<string> EnumerateFileSystemEntries(string path)
+        {
+            IEnumerable<string> alphaFS(string _path)
+            {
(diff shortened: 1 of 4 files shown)


Duplicati/Library/Common/IO/SystemIOWindows.cs
```diff
@@ -288,20 +288,47 @@ public string GetSymlinkTarget(string file)
             return null;
         }
 
+        private static T TryNativeFirst<T>(Func<string, T> nativeIOFunc, Func<string, T> alternativeIOFunc, string path)
+        {
+            if (!IsPathTooLong(path))
+                try { return nativeIOFunc(path); }
+                catch (System.IO.PathTooLongException) { }
+                catch (System.ArgumentException) { }
```
warwickmm: Can I ask why we're ignoring the `ArgumentException` here?  Are there cases where `Directory.EnumerateFiles` will throw this, but the AlphaFS version is able to handle the provided path?

warwickmm: Thanks for the explanation.  That makes sense.  If we do decide to drop the extra `catch`, we can do so in a separate pull request.  This way, we won't pollute this one with too many unrelated changes.

warwickmm:
This looks good to me.  I can't say much about how this affects performance though.  The local function is small enough that maybe it gets inlined?  Perhaps we should opt for the cleaner implementation and worry about optimization after it is demonstrated that this is an issue.

Was your plan to refactor all the methods to use `TryNativeFirst` in this branch, or in a separate pull request?

Duplicati/Library/Common/IO/SystemIOWindows.cs
warwickmm: Actually, it looks like there was a case where the .NET library threw an `ArgumentException` when it should have thrown a `PathTooLongException`.  Not sure if this is still the case, but this might be worth documenting in the code to clarify the extra `catch`.

See https://github.com/duplicati/duplicati/issues/1451#issuecomment-137708490

verhoek:
You're right, I should have kept the fix with the refactoring separate. I will create another PR with the refactoring because I was indeed planning to refactor the entire class. You can merge the first commit in if you can and ignore the second, or merge them both in and I'll resolve the merge conflict in the second PR. I can't force push over this PR since the discussion in this PR doesn't make much sense anymore.

warwickmm:
I'm not sure what happens to this pull request if only some of the commits are merged.  Would it show as merged, or closed?  Perhaps one option is to push a commit that reverts the changes in 94f34ce5a19b542b6b2e63c0f8e1aea9baf2ebe5?

Also, I don't know if I should be merging in changes.  I'll leave that for @kenkendk and @Pectojin.

kenkendk:
Looks good.

I am sure the runtime will inline the calls; it is usually quite aggressive with inlining.

PR #3520, opened 2018-11-28
Pr/remove repeated code

verhoek:
As discussed in the other PR. I still need to optimize a few things - optional boolean prefixUnc is not ideal, but convenient.
Since almost everything to alphafs is apparently prefixed with UNC, at least I will switch the 
default value or set it always to true without an associated argument.
I also will take a look if I can reduce the number of helper functions.

### DIFF
--- a/Duplicati/Library/Backend/File/FileBackend.cs
+++ b/Duplicati/Library/Backend/File/FileBackend.cs
@@ -173,7 +173,7 @@ namespace Duplicati.Library.Backend
             if (!systemIO.DirectoryExists(m_path))
                 throw new FolderMissingException(Strings.FileBackend.FolderMissingError(m_path));
 
-            foreach (string s in systemIO.EnumerateFileSystemEntries(m_path))
+            foreach (string s in systemIO.EnumerateFiles(m_path))
             {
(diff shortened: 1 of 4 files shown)


verhoek:
There's a problem with the unit tests on app veyor. I probably switched some calls or alike, I will test on Windows myself tomorrow.

verhoek:
It seems to work now. We should review this carefully, it's quite possible I mixed calls up. I didn't optimize the boolean prefixUNC  yet, since I believe we can remove this altogether:

https://github.com/alphaleonis/AlphaFS/wiki/Using-AlphaFS-from-code

saying "This means you never have to fiddle again with long path prefixes such as : \\?\ or \\?\UNC\" , which is precisely what's happening here. And even in our case, we don't fiddle consistently, so I suspect the prefixing is unnecessary.

I prefer to remove the prefixing in another PR to avoid making too many possible breaking changes in one go.

kenkendk:
This seems to be tangled with the previous PR, but nonetheless it looks sound.

I like the comment about AlphaFS automatically adding the UNC prefix, and I agree we should switch to that.

Issue #3072, opened 2018-03-01
Backup job database not deleted when deleting job


Pectojin:
- [X] I have searched open and closed issues for duplicates.

----------------------------------------

## Environment info

- **Duplicati version**: 2.0.2.20
- **Operating system**: Ubuntu 16.04 & macOS High Sierra
- **Backend**: Any

## Description

The `Delete the local database` checkbox on the `Delete backup` page does not actually cause the database to be deleted.

I can manually delete the database on the `Database` page, so it's definitely not a permission issue.

## Steps to reproduce

1. Go to the delete backup config page
2. Check the `Delete the local database` checkbox
3. Delete the database config

- **Actual result**:
The backup is deleted as if I had not checked the delete db checkbox
- **Expected result**:
DB deleted

## Debug log
no logs seem to be generated during the deletion of the backup relating to the database

sylerner:
I am having the same issue with version 2.0.3.3_beta_2018-04-02 running on openSUSE Linux Leap 15.0.

PR #3526, opened 2018-12-02
Fix incorrect reference to database path

warwickmm:
This fixes an issue where the wrong key was used to access the options `Dictionary`, which caused the local database to not be deleted when deleting a backup configuration.

This addresses issue #3072.


Pectojin:
Nice! That almost feels silly 👍

Issue #2857, opened 2017-10-30
List of backup timestamps contains duplicates


joelluijmes:
- [x] I have searched open and closed issues for duplicates.

----------------------------------------

## Environment info

- **Duplicati version**: 2.0.2.12_canary_2017-10-20 (headless installation)
- **Operating system**: Debian 8 x64
- **Backend**: STACK (WebDAV)

## Description

When running the backup I get the following exception:
```Failed: List of backup timestamps contains duplicates: 30-10-2017 13:00:00, 30-10-2017 12:00:00, 30-10-2017 11:00:00, 30-10-2017 10:00:00, 30-10-2017 09:00:00, 30-10-2017 08:00:00, 30-10-2017 07:00:00, 30-10-2017 06:00:00, 30-10-2017 05:00:00, 30-10-2017 04:00:00, 30-10-2017 03:00:00, 30-10-2017 02:00:00, 30-10-2017 01:00:00, 30-10-2017 00:00:00, 29-10-2017 23:00:00, 29-10-2017 22:00:00, 29-10-2017 21:00:00, 29-10-2017 20:00:00, 29-10-2017 19:00:00, 29-10-2017 18:00:00, 29-10-2017 17:00:00, 29-10-2017 16:00:00, 29-10-2017 15:00:00, 29-10-2017 14:00:00, 29-10-2017 13:00:00, 29-10-2017 12:00:00, 29-10-2017 11:00:00, 29-10-2017 10:00:00, 29-10-2017 09:00:00, 29-10-2017 08:00:00, 29-10-2017 07:00:00, 29-10-2017 06:00:00, 29-10-2017 05:00:00, 29-10-2017 04:00:00, 29-10-2017 03:00:00, 29-10-2017 02:00:00, 29-10-2017 02:00:00, 29-10-2017 01:00:00, 29-10-2017 00:00:00, 28-10-2017 23:00:00, 28-10-2017 22:00:00, 28-10-2017 21:00:00, 28-10-2017 20:00:00, 28-10-2017 19:00:00, 28-10-2017 18:00:00, 28-10-2017 17:00:00, 28-10-2017 16:00:00, 28-10-2017 15:00:00, 28-10-2017 14:36:33, 28-10-2017 13:09:12
[...]

Issue #1364, opened 2015-04-24
DST changes scheduled backup time


go-vegan:
I've set my backups to run at 11:00 each day, but since the switch to DST, they run on hour later (12:00).
##

marceldev89:
This is still the case, this weekend the scheduled time moved back an hour.

Luto73:
I can confirm this bug. A scheduled backup which ran at 23:00 UTC+2 (MEST) is now running at 22:00 UTC+1 (MET).

EraYaN:
It is a shame due to schedules "free-data" hours. So I really want the backup to start at 0:00 and not 23:00 (which it changed to last week). Not sure how to solve this issue. Maybe an option to store the time in the local computer time?

antonpiatek-imanage:
Is there any technical reason not to store the schedule in local timezone? 
I also schedule my backup to run over lunch time to minimise the impact on me (full homedir is an expensive job).

If there's no good reason not to store it this way I might even try and learn c#

Pectojin:
Storing time in local computer timezone seems to be what @kenkendk intended when he opened #1168.

Just hasn't been implemented yet. Although, usually it's better to store everything in UTC and then convert it. Is it too difficult to convert it everywhere in the application?

kenkendk:
Everything is actually stored in UTC (and I prefer that it stays that way), but for this particular issue, the problem is that UTC is DST-free, and this causes it to "differ" from the local time.

It should be a matter of changing the writer to store it in local-time:
https://github.com/duplicati/duplicati/blob/master/Duplicati/Library/Main/BackendManager.cs#L507

Then changing the reader to not read it like UTC:
https://github.com/duplicati/duplicati/blob/master/Duplicati/Server/Database/Connection.cs#L318

The hard part is then changing the logic to use the local time here:
https://github.com/duplicati/duplicati/blob/master/Duplicati/Server/Scheduler.cs#L234

We should keep the "last" timestamp in UTC, and then make sure the function work correctly when using UTC/local compares.

uok:
This is still happening, not a big problem but causes some confusion. Hope there is a fix for this possible? Thanks!

janmadera:
Using Duplicati - 2.0.4.5_beta_2018-11-28 issue still the same. Is there anyone on it? Looks like something basic, that is know since many months / years now but still not solved.

auanasgheps:
[...]

PR #3528, opened 2018-12-04
Use local time when checking if day is allowed

mikaelmello:
https://forum.duplicati.com/t/incorrect-start-time/1391/21

When using UTC time to check if a day is allowed, this causes unexpected behavior for the user because they expect to allow/disallow days on their local time, not UTC.

### DIFF
--- a/Duplicati/Server/Scheduler.cs
+++ b/Duplicati/Server/Scheduler.cs
@@ -374,10 +374,11 @@ namespace Duplicati.Server
         /// <returns>True if the backup is allowed to run, false otherwise</returns>
         private static bool IsDateAllowed(DateTime time, DayOfWeek[] allowedDays)
         {
+            var localTime = time.ToLocalTime();
             if (allowedDays == null || allowedDays.Length == 0)
                 return true;
             else
-                return Array.IndexOf<DayOfWeek>(allowedDays, time.DayOfWeek) >= 0; 
+                return Array.IndexOf<DayOfWeek>(allowedDays, localTime.DayOfWeek) >= 0; 
         }
 
     }


kenkendk:
I am fine with merging this.

But a larger change that we should do (later) is to store the scheduled time in local offset, not in UTC.

Pectojin:
How big a change would storing that offset be? Implementing that change will resolve at least a couple issues on the tracker.

mikaelmello:
Are there any other reasons for storing them in local time aside from DST changes?

What would happen in the following scenario?

- Backup runs at 00:00
- When the clock gets to 01:00, the end of DST makes it go back to 00:00
- A new backup is run with the same DateTime as before.

Pectojin:
DST essentially breaks two things:
- Hourly backups will try to create two dlist with same timestamp causing failure #2857
- DST changes backup schedule #1364

If Duplicati tries to create two backups with same timestamp it fails. In a sense it soft bricks the backup because manual intervention is required to resolve the issue.

If you don't have hourly backups you're only affected by the schedule change, though.

mikaelmello:
Changing backup schedule really is a big deal. However aren't the hourly conflicts bound to happen anyway?

Pectojin:
Not if we keep using the same timezone for the dlist timestamps and convert them internally in Duplicati for the GUI. But alternatively we can allow both to exist as Kenneth originally suggested.

kenkendk:
> Hourly backups will try to create two dlist with same timestamp causing failure #2857

That problem is "easy" to fix, as it is only the listing/parsing code.

> DST changes backup schedule #1364

That problem is the one I was aiming for.

Anyway, anyone opposed to at least merging this to get rid of the wrong day-of-week issue?

Pectojin:
Yes. Let's battle the scope creep.

Issue #3395, opened 2018-09-23
Filter groups code is extremely slow


gabrielambuehl:
----------------------------------------

## Environment info

- **Duplicati version**: Latest canary (but seemingly true ever since filter groups was added)
- **Operating system**: win10
- **Backend**: onedrive, google, b2 (seems irrelevant)

## Description
Doing a backup over my roughly 30GB, 30’000 profile (all living on a fast SSD) with filter group default excludes takes roughly 12 minutes, at least 80% of which is spent counting files (counter goes up by about 50-100 files per second.
If I kick out filter group default excludes and just a have a bunch of excluded directories otherwise, it goes down to 2-3 minutes, most of which is spent checking backend files (which I guess cannot be avoided given duplicati’s architecture).
It definitely feels like there is something seriously wrong in the filter group code.

## Steps to reproduce
Add default filter group to a backup.

tshcherban:
Can we use other approach for large groups? One large regex (for default excludes) is very slow to execute (~95k steps per each ~50-char file path). Maybe we can use list of directories to exclude and ommit directory listing if it not fits into filter.

mr-flibble:
Did you try **--disable-file-scanner=true** ? I can only recommend it for large backups. 

_When this flag is enabled, the scanner that computes the size of source files is disabled, and instead the reported size is read from the database. Using this flag can speed up the backup by reducing disk access, but will give a less accurate progress indicator._

warwickmm:
I've been benchmarking parts of the code, and can demonstrate the behavior when simply enumerating the file system entries (by calling `Utility.EnumerateFileSystemEntries`) with and without the default filter.  For a directory with around 250,000 items, I'm seeing the following results:
```
         Method |      Mean |    Error |   StdDev |
--------------- |----------:|---------:|---------:|
 DefaultFilters | 111.731 s | 2.0954 s | 1.8575 s |
      NoFilters |   5.443 s | 0.1080 s | 0.2956 s |
```
Next step would be to take a closer look at the relevant code and perhaps run a profiler to find the hot spots.

Dobatymo:
The regex could probably converted to a finite-state machine.

tshcherban:
Seems that we are doing some waste job scanning all files from directory, which should be excluded entirely. Will check later.

warwickmm:
[...]

PR #3545, opened 2018-12-18
Convert regex pattern to wildcard in default excludes group

warwickmm:
This provides a workaround for the poor performance of the default excludes group exhibited by regex filters.  While this wildcard pattern will potentially exclude more files than the previous regex pattern, in
practice this should not often be the case.

This does not resolve the performance issue of the regex filters discussed in issue #3395.  However, I think in general we should try to avoid expensive application-specific filters since these will affect the performance for all users of the default excludes group, whether the application is present or not.

Below are some benchmarks for enumerating (using `Utility.EnumerateFileSystemEntries`) the entries in a directory that contains ~155,000 files and ~33,000 directories.  The `DefaultFilters` method represents the default filters as of revision a0af8d00c, and the `DefaultFiltersWildcardOnly` method represents the default filters as modified in this branch.  The `SingleRegexFilter` method consists of a single regex filter, and the `DualRegexFilter` consists of two regex filters.  None of the filters matched any entry in the directory being enumerated, so the number of entries being tested should be pretty consistent across the tests.

```
                     Method |     Mean |    Error |   StdDev | Scaled | ScaledSD |
--------------------------- |---------:|---------:|---------:|-------:|---------:|
                  NoFilters |  3.420 s | 0.0282 s | 0.0264 s |   0.05 |     0.00 |
             DefaultFilters | 63.868 s | 0.3148 s | 0.2790 s |   1.00 |     0.00 |
 DefaultFiltersWildcardOnly | 24.730 s | 0.3938 s | 0.3683 s |   0.39 |     0.01 |
          SingleRegexFilter | 46.562 s | 0.7440 s | 0.6959 s |   0.73 |     0.01 |
            DualRegexFilter | 86.367 s | 1.0413 s | 0.9740 s |   1.35 |     0.02 |

// * Legends *
  Mean     : Arithmetic mean of all measurements
  Error    : Half of 99.9% confidence interval
  StdDev   : Standard deviation of all measurements
  Scaled   : Mean(CurrentBenchmark) / Mean(BaselineBenchmark)
  ScaledSD : Standard deviation of ratio of distribution of [CurrentBenchmark] and [BaselineBenchmark]
  1 s      : 1 Second (1 sec)
```
The above results clearly illustrate the prohibitive cost of a regex filter.  I think [there are ways](https://github.com/duplicati/duplicati/issues/3395#issuecomment-437704755) that we can improve this, but I would need to gain more insight into why the current implementation is the way it is.

### DIFF
--- a/Duplicati/Library/Utility/FilterGroups.cs
+++ b/Duplicati/Library/Utility/FilterGroups.cs
@@ -606,18 +606,6 @@ namespace Duplicati.Library.Utility
         }
 
         /// <summary>
-        /// Creates a Regex filter
-        /// </summary>
-        /// <param name="filter">Filter text</param>
-        /// <returns>Regex filter</returns>
-        private static string CreateRegexFilter(string filter)
-        {
-            // Create a filter with the given name.
-            // However, in order to match paths correctly, the directory separators need to be normalized to match the system default.
-            return "[" + filter.Replace(FilterGroups.RegexEscapedAltDirectorySeparatorChar, FilterGroups.RegexEscapedDirectorySeparatorChar) + "]";
-        }
-
-        /// <summary>
         /// Gets a list of exclude paths from the MacOS system
         /// </summary>
         /// <returns>The list of paths to exclude on OSX backups.</returns>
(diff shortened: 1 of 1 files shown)


warwickmm:
> Would it be reasonable to also remove the CreateRegexFilter method since it is not used anymore and we should avoid it in the future?

Good idea.  I'll remove it when I get a chance.

warwickmm:
Revision faf2b0df474e7f4155dd81990fca30d2b5b94b99 removes the unused `CreateRegexFilter` method to discourage the creation of default regex filters.

Issue #3550, opened 2018-12-19
RUMPS issue on High Sierra introduced between 2.0.4.5_beta_2018-11-28 and  2.0.4.6_canary_2018-12-11 


BrettDikeman:
- [x ] I have searched open and closed issues for duplicates.

----------------------------------------

## Environment info

- **Duplicati version**: 2.0.4.6_canary_2018-12-11 or higher
- **Operating system**: OS X High Sierra
- **Backend**: N/A

## Description
Installing a release newer than 2.0.4.5_beta_2018-11-28 (possibly starting with 2.0.4.6_canary_2018-12-11, I don't have the time to exhaustively nail down which release) results in a Duplicati installation that cannot run on an up-to-date, largely-unmodified OS X High Sierra system.

## Steps to reproduce

1. Install High Sierra
2. Install 2.0.4.6_canary_2018-12-11 or newer (including as-of-writing current canary release v2.0.4.8-2.0.4.8_canary_2018-12-13

## Result

**Actual result**:
- As installer finishes, Dock shimmies as the Duplicati process repeatedly dies & is relaunched by launchctl
- mono consumes a large amount of CPU time

## Debug log

```
$ /Applications/Duplicati.app/Contents/MacOS/duplicati 
Unexpected message: None
Unexpected message: Exception in thread Thread-1:
Unexpected message: Traceback (most recent call last):
Unexpected message: File "/System/Library/Frameworks/Python.framework/Versions/2.7/lib/python2.7/threading.py", line 810, in __bootstrap_inner
Unexpected message: self.run()
Unexpected message: File "/System/Library/Frameworks/Python.framework/Versions/2.7/lib/python2.7/threading.py", line 763, in run
Unexpected message: self.__target(*self.__args, **self.__kwargs)
Unexpected message: File "/Applications/Duplicati.app/Contents/Resources/OSXTrayHost/osx-trayicon-rumps.py", line 108, in get_input
Unexpected message: app.menu = parsemenus(cfg)
Unexpected message: File "/Applications/Duplicati.app/Contents/Resources/OSXTrayHost/osx-trayicon-rumps.py", line 51, in parsemenus
Unexpected message: mn = rumps.MenuItem(n["Text"])
Unexpected message: KeyError: 'Text'
```

[Forum post](https://forum.duplicati.com/t/high-sierra-client-wont-run-duplicati-repeatedly-crashes-with-keyerror-text/5713)

[...]

PR #3552, opened 2018-12-20
Add back properties required by RUMPS (Mac status bar)

warwickmm:
These properties were removed in pull request #3534 when replacing write-only properties with methods.  While the properties were not being read in C#, there is Python code that depends on their availability.

This addresses issue #3550.

### DIFF
--- a/Duplicati/GUI/Duplicati.GUI.TrayIcon/RumpsRunner.cs
+++ b/Duplicati/GUI/Duplicati.GUI.TrayIcon/RumpsRunner.cs
@@ -32,24 +32,19 @@ namespace Duplicati.GUI.TrayIcon
 
             private readonly RumpsRunner m_parent;
 
-            private string m_text;
-            private MenuIcons m_icon;
-            private bool m_enabled;
-            private bool m_default;
-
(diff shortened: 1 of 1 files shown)


Pectojin:
Just for ease of test/confirmation:

1. checkout Duplicati master branch on MacOS
2. build and run TrayIcon in Visual Studio
3. Keep an eye on Activity Monitor and console output, after a few moments a mono child process `Python` will spike in memory and CPU usage and it will generate huge amounts of logs.
4. shut it down. You may have to force quit the Python process as it seems to go zombie.
5. checkout the debug/rumps_runner branch
6. build and run
7. repeat. The console output will be as usual and the python child process will use virtually no resources.

warwickmm:
Sorry about this.  I seem to be breaking a few things recently.

Pectojin:
That's what we have canaries for :)
I don't think there was any way to know this was going to break from the changes made, it wasn't even obvious after it broke.

kenkendk:
> Sorry about this. I seem to be breaking a few things recently.

No worries. Stuff breaks when you change things. I prefer the changes over stand-still :)

PR #3557, opened 2018-12-22
added passphrase confirmation

### LINKED ISSUE #3553: In "Export backup configuration" we enter encryption key without verification

## Environment info

- **Duplicati version**: 2.0.3.3_beta_2018-04-02
- **Backend**: Backblaze B2

## Description

The "Export backup configuration" screen offers the (necessary) option to encrypt the config file that will be exported. If this option is used, the user is asked to input the encryption key. 

This key is inputted only once, and is done in a hidden password field. At this point a typo would result in a config file which cannot be decrypted. (OK, so a cautious user might test restoring from the file, but we shouldn't rely on this)

RFE: Add a duplicate field here, compare them, and complain if they do not match, as per convention when inputting passwords etc elsewhere.

Pectojin:
This fixes #3553

screenshot 2018-12-22 at 18 40 47


verhoek:
Looks OK and tested works fine.

PR #3559, opened 2018-12-23
Autoupdater : Allow generation of alternative priv/pub key pair

verhoek:
This enables the possibility to build a zip or release process with a parallel priv/pub key pair not enforced by using the one implied from the manifest used currently in the main release process.

Update: also allow for non-standard gpg keyring location.

Update: Removed again alternate keyring location -  it's more elegant to import/export keys before launching the autoupdater.


verhoek:
Had to add another cmdline option for stdin passphrase functionality to work at all with gpg2. I guess the previous cmdline worked only with a much older gpg2 version.

verhoek:
Actually, I'm going to remove the entire gpg stuff from autoupdatebuilder since it's completely the gpg part to sign is fully independent from the rest. It will go in a separate gpg wrapper program that uses @kenkendk 's sharpaes wrapper to load keys. Using that, we can use this tool to also sign installers in a secure way avoiding echo pipes with passphrases.

verhoek:
Created GnupgSigningTool as said and updated the legacy building logic. @kenkendk : I cannot test this myself, I hope you can check that the gpg signatures are still being generated as expected.

kenkendk:
The GnuPG signing is not a critical part for the updater. The signatures are published, such that they can be verified by end users, but they are not checked as part of the built-in updater.

I will merge, and then test the signatures after the next build.

PR #3562, opened 2018-12-25
Improve performance of wildcard filters

warwickmm:
This improves the performance of wildcard filters by avoiding unnecessary array allocations.  The mono profiler indicated that a large percentage of time was spent allocating arrays that were used as stacks.  By replacing these with `Stack` objects, we avoid this unnecessary overhead.

This also adds some tests to verify the behavior of the wildcard filter.

Note that this does not improve the performance of the default excludes filters.  To address that issue, we need to avoid converting them to a regex filter and preserve their wildcard nature.

### DIFF
--- a/Duplicati/Library/Utility/FilterExpression.cs
+++ b/Duplicati/Library/Utility/FilterExpression.cs
@@ -186,9 +186,9 @@ namespace Duplicati.Library.Utility
             /// <note>From: http://www.c-sharpcorner.com/uploadfile/b81385/efficient-string-matching-algorithm-with-use-of-wildcard-characters/</note>
             private static bool IsWildcardMatch(string input, string pattern)
             {
-                int[] inputPosStack = new int[(input.Length + 1) * (pattern.Length + 1)];   // Stack containing input positions that should be tested for further matching
-                int[] patternPosStack = new int[inputPosStack.Length];                      // Stack containing pattern positions that should be tested for further matching
-                int stackPos = -1;                                                          // Points to last occupied entry in stack; -1 indicates that stack is empty
+                Stack<int> inputPosStack = new Stack<int>(); // Stack containing input positions that should be tested for further matching
+                Stack<int> patternPosStack = new Stack<int>(); // Stack containing pattern positions that should be tested for further matching
+
                 bool[,] pointTested = new bool[input.Length + 1, pattern.Length + 1];       // Each true value indicates that input position vs. pattern position has been tested
                 int inputPos = 0;   // Position in input matched up to the first multiple wildcard in pattern
                 int patternPos = 0; // Position in pattern matched up to the first multiple wildcard in pattern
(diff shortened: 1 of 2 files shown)


Duplicati/Library/Utility/FilterExpression.cs
```diff
@@ -186,9 +186,9 @@ private static Regex GetFilterGroupRegex(string filterGroupName)
...
             private static bool IsWildcardMatch(string input, string pattern)
             {
-                int[] inputPosStack = new int[(input.Length + 1) * (pattern.Length + 1)];   // Stack containing input positions that should be tested for further matching
-                int[] patternPosStack = new int[inputPosStack.Length];                      // Stack containing pattern positions that should be tested for further matching
-                int stackPos = -1;                                                          // Points to last occupied entry in stack; -1 indicates that stack is empty
+                Stack<int> inputPosStack = new Stack<int>(); // Stack containing input positions that should be tested for further matching
+                Stack<int> patternPosStack = new Stack<int>(); // Stack containing pattern positions that should be tested for further matching
+
                 bool[,] pointTested = new bool[input.Length + 1, pattern.Length + 1];       // Each true value indicates that input position vs. pattern position has been tested
```
warwickmm: I made several attempts to replace this multidimensional array with various data structures (`HashSet<Tuple<int, int>>`, `bool[][]`), but all my attempts proved unsuccessful in improving the performance.  For some reason, this allocation does not seem to have as large of an effect on performance as the removed array allocations above.

warwickmm:
Prior to version v2.0.4.8-2.0.4.8_canary_2018-12-13, enumerating files with the default filters would take 18-20x the time it took with no filters.  These changes, together with those from pull request #3545, should hopefully reduce the impact to around 1.5-2x.

mikaelmello:
Wonderful work!

warwickmm:
Unfortunately, I'm having difficulty realizing the performance gains in an actual backup configuration.  My targeted benchmarks definitely show some improvement (as described above), but attempts to observe this in an actual backup has proved more difficult.

I have some more testing I need to do before we consider merging this and will report back with what I find.

mikaelmello:
Which tools do you use for benchmarking and how do you analyze performance on actual backups?

warwickmm:
For benchmarking `Utility.EnumerateFileSystemEntries` with various filters, I was using [BenchmarkDotNet](https://github.com/dotnet/BenchmarkDotNet).  For the actual backup, I was just clicking Run in the UI and looking at the reported timings.

warwickmm:
I think these changes are ok.  However, I would like to reword the commit message before we merge, as this does not address issue #3395.  Once I fix the message, I'll edit the pull request title accordingly.

It looks like the reason why this doesn't improve the performance of the default excludes is because the code converts all of those wildcard filters into one giant regex (possibly because the default excludes used to contain a regex filter).  I think the code already has support for treating these as a collection of wildcard filters.  I will try and look into that next.

Issue #3542, opened 2018-12-15
Restoring fails with TargetInvocationException


Phaiax:
- [x] I have searched open and closed issues for duplicates.

----------------------------------------

## Environment info

- **Duplicati version**: 2.0.4.8_canary_2018-12-13
- **Operating system**: Windows 10 (Version 10.0.17134 Build 17134)
- **Backend**: SSH@Synology

## Description

I tried restoring one file, to a special destination folder, "Write Multiple versions with timestamp" was selected, then I clicked "restore".
The Website stated "Dateien und Ordner erfolgreich wiederhergestellt." (="Files and folders were successfully restored"), at the same time a Windows notification and a Website notification appeared, stating "Error while running Foo-Bar-Backup" and "Ein Aufrufziel hat einen Ausnahmefehler verursacht." (="A target invocation has produced an exceptional error"). Clicking "Show" on the error forwarded to the backup log which did not contain any information about the error. Instead in the system log under "About", I found the backtrace below.
In the restore destination folder, no files appeared (still empty).

## Steps to reproduce

1. Install Duplicaty, the current newest canary
2. Add backup, no special options, to SSH, encrypted
3. Start backup (on the remote location, a few `duplicati-...-*.zip.aes` files appeared.
4. Try restoring a file

5. I tried restarting the dupicati server, but the problem persists.

- **Actual result**: Restoring did not work, even if the website stated it was successful (but there were error notifications)
- **Expected result**: Restoring should work or the website should also state an error. The "Show" link in the notification opens the correct log.

If I can help you any further, please ask.
Thank you for your time.
Daniel

## Debug log

The text in the stacktrace can be translated to:
"Empty passphrase not allowed; Parameter name: passphrase at Duplicati.Library.Encryption.AESEncryption..ctor("

```
System.Reflection.TargetInvocationException: Ein Aufrufziel hat einen Ausnahmefehler verursacht. 
---> System.ArgumentException: Leere Passphrase nicht erlaubt Parametername: passphrase bei String passphrase, Dictionary`2 options) 
--- Ende der internen Ausnahmestapelüberwachung --- 
bei System.RuntimeMethodHandle.InvokeMethod(Object target, Object[] arguments, Signature sig, Boolean constructor) 
bei System.Reflection.RuntimeConstructorInfo.Invoke(BindingFlags invokeAttr, Binder binder, Object[] parameters, CultureInfo culture) 
[...]

PR #3563, opened 2018-12-25
Only create passphrase option if passphrase is provided

warwickmm:
This fixes an issue introduced in revision f7800e8a2bd where passphrase-less restore operations from the UI would result in the following exception:
```
System.Reflection.TargetInvocationException: Exception has been thrown by the target of an invocation. ---> System.ArgumentException: Empty passphrase not allowed
Parameter name: passphrase
```
This addresses issue #3542.


verhoek:
Thanks a lot for fixing!

PR #3567, opened 2018-12-27
Fix Duplicati shutdown issue on Synology

drwtsn32x:
The current Synology start-stop-status does not properly shut down Duplicati.  It terminates only the main process but not the child processes.

Example processes before Duplicati is stopped via Synology Package Manager:

```
bash-4.3# ps ax | grep Dup
15236 ?        Sl     0:00 mono /volume1/@appstore/Duplicati/Duplicati.Server.exe
15242 ?        Sl     0:01 /volume1/@appstore/Mono/usr/local/bin/mono-sgen /volume1/@appstore/Duplicati/Duplicati.Server.exe
```

After telling Duplicati to stop, only PID 15236 is terminated.  PID 15242 remains, and this causes problems if you try to start Duplicati again.  This issue also causes problems when people try to upgrade as a stop/start cycle is needed in such a case.

With this change the script will look up the process group ID (PGID) of the main PID and issue a kill command to that whole group.  This will ensure the child processes are also shut down.

### DIFF
--- a/Installer/Synology/scripts/start-stop-status
+++ b/Installer/Synology/scripts/start-stop-status
@@ -68,7 +68,7 @@ DaemonStop() {
 	DaemonStatus
 	if [ $? == 1 ]; then
 		echo "Stopping ${PACKAGE_NAME_SIMPLE}."
-		kill -- -$(ps -o pgid= $(cat "$PID_FILE") | grep -o '[0-9]*')
+		pkill -P $(cat "$PID_FILE")
 		rm -f "$PID_FILE"
 		
 		rm -f /usr/local/etc/nginx/conf.d/dsm.duplicati.conf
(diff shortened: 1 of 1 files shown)


verhoek:
Hi, thanks for your PR. Can you change it to use pkill -P $PID ?

drwtsn32x:
Doesn't "pkill -P" only match the child processes for a given PID?  What about the main process itself?  Or maybe we should use "pkill -g $PGID" instead as that also works on PGIDs like "kill -- -$PGID"

drwtsn32x:
Disregard - I just tested pkill -P $PID and it *did* properly kill the PID and its child processes.  I agree it is cleaner. I'll update the pull request. Thanks!

drwtsn32x:
I've never updated a pull request before. Hopefully I did it correctly.

drwtsn32x:
> Does anyone else feel like this is actually a workaround

Yes, I do. I'm not sure what is responsible for the root problem of not terminating the child process. Is it Mono's responsibility?  Duplicati's?

If someone can identify the root cause and fix it, then this workaround should be un-done.  In the meantime the workaround is better than nothing, and there doesn't seem to be any negative consequences.

PR #3568, opened 2018-12-29
Improve performance of default excludes filters

warwickmm:
This improves the performance of the default excludes filters by not converting them to a large regex filter.  Previously, the `FilterExpression.Expand` method would forward the filter group name to the `FilterEntry` constructor, which would combine the group's patterns into one large regex filter.  This would result in a significant decrease in performance, as the regex filters are much slower than the wildcard filters.

Instead, we can first expand the filter group into its constituent patterns before presenting them to the `FilterEntry` constructor.  This preserves their wildcard nature and allows us to take advantage of faster filter matching algorithms.

This addresses issue #3395.

### DIFF
--- a/Duplicati/Library/Utility/FilterExpression.cs
+++ b/Duplicati/Library/Utility/FilterExpression.cs
@@ -406,16 +406,19 @@ namespace Duplicati.Library.Utility
             if (string.IsNullOrWhiteSpace(filter))
                 return null;
 
-            if (filter.Length < 2 ||
-                (filter.StartsWith("[", StringComparison.Ordinal) && filter.EndsWith("]", StringComparison.Ordinal)) ||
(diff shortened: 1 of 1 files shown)


warwickmm:
I built this in release mode and tested with a backup configuration of ~100,000 files (8 GB) to a local backend, where none of the files matched the default exclude filters.  The average performance numbers  are below:

* No filters (v2.0.4.9): 2 minutes 55 seconds
* Default excludes (v2.0.4.9): 8 minutes 33 seconds
* Default excludes (this branch): 3 minutes 51 seconds
* Default excludes (this branch with changes from pull request #3562): 3 minutes 8 seconds

I also tested this against sources with folders that would match the default excludes filters, and they were properly excluded.

PR #3581, opened 2019-01-02
Avoid redundant calls to Util.AppendDirSeparator

warwickmm:
This removes some redundant calls to `Util.AppendDirSeparator`.  Since the `Utility.EnumerateFileSystemEntries` method always calls `Util.AppendDirSeparator` before pushing a folder name onto the stack, we do not need to do so when providing the collection of sources, or when popping an entry from the stack.

### DIFF
--- a/Duplicati/Library/Utility/Utility.cs
+++ b/Duplicati/Library/Utility/Utility.cs
@@ -215,7 +215,7 @@ namespace Duplicati.Library.Utility
 
                 while (lst.Count > 0)
                 {
-                    var f = Util.AppendDirSeparator(lst.Pop());
+                    var f = lst.Pop();
 
                     yield return f;
 
--- a/Duplicati/Library/Utility/Utility.cs
+++ b/Duplicati/Library/Utility/Utility.cs
@@ -198,7 +198,6 @@ namespace Duplicati.Library.Utility
                 rootpath = Util.AppendDirSeparator(rootpath);
                 try
                 {
-
                     var attr = attributeReader?.Invoke(rootpath) ?? FileAttributes.Directory;
                     if (callback(rootpath, rootpath, attr))
                         lst.Push(rootpath);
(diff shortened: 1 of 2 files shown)


Duplicati/Library/Utility/Utility.cs
```diff
@@ -215,7 +214,7 @@ public static IEnumerable<string> EnumerateFileSystemEntries(string rootpath, En
 
                 while (lst.Count > 0)
                 {
-                    var f = Util.AppendDirSeparator(lst.Pop());
+                    var f = lst.Pop();
```
verhoek: A few lines below apparently also other folders are pushed on the stack, not having an append dirseparator appended anymore by this change. If you understand when to add '/' or '\' at the end of a folder, and when, it would be good to comment this or make it sensible by reading the code (for me it looks puzzling). This entire function should be also  be rewritten, it's too long and has too deep indentations. Is it relevant we use a stack as data type here and if so, why?

warwickmm: It looks to me that every argument to `lst.Push` is the result of a call to `Util.AppendDirSeparator`.  Can you point me to a specific line that is lacking this?  It's very possible that I missed something.

I agree that this method is quite unwieldy and should be refactored.  My understanding is that the stack is used to keep track of folders that have not been excluded by a filter.  This way, we do not enumerate (and attempt to match against the filters) entries underneath an excluded directory.  As to why we use a `Stack` vs. some other data structure, I am not sure.  For this tree-traversal case with ignored nodes, a `Stack` does seem pretty efficient.

I'm not sure when a trailing `/` is really needed or not.  There are lots of places in the code that use a trailing `/` to differentiate between a directory and a file, so my hope was to not change any behavior here.  I also don't quite understand the exception handling here...will have to study that some more.

verhoek: I think you're right, I was thinking about line 230 but in 225 it appends one too. When I refactored some IO calls I also left it intact with the append dir separator, since I am also unsure what the exact idea behind it is. Maybe regarding the filter logic - though I can think of better ways to solve that.

Issue #3296, opened 2018-06-26
Add  backup-test-percentage in addition to backup-test-samples [$25]


sylerner:
- [x] I have searched open and closed issues for duplicates.

----------------------------------------

## Environment info

- **Duplicati version**: 2.0.3.3_beta_2018-04-02
- **Operating system**: Linux
- **Backend**: sftp

## Description

Checking only 1 sample per backup when there are thousands of files doesn’t make much of a dent, especially if only backing up once per day.

While backup-test-samples provides the option to test more, a user often will have no idea how many files they have in their backup.

From a user point of view, I would think the ability to specify what percentage of your backup you want checked after each backup would be a friendlier method than having to specify an absolute number of samples.

On a similar topic, should the default check be something like 1% of the backup set?

Assuming that duplicati keeps track of who’s already been checked (I believe I’ve read that it does), a rolling 1% would - assuming there isn’t too much change each day - in a reasonable amount of time have checked 100% of the backup.

## Steps to reproduce

1. Perform backup

- **Actual result**:
Only checks one sample, no way to guess how many samples to make a dent w/o overdoing it.
- **Expected result**:
Ability to check a specific percentage of a backup set per backup, with a reasonable default value so that the entire backup set gets checked in a reasonable period of time (e.g., less than a year of daily backups).

JonMikelV:
There is a bounty for this currently at $25.
https://www.bountysource.com/issues/60122728-add-backup-test-percentage-in-addition-to-backup-test-samples

Pectojin:
Implemented in #3582 👍

Pectojin:
@warwickmm you should claim the bounty https://www.bountysource.com/issues/60122728-add-backup-test-percentage-in-addition-to-backup-test-samples

warwickmm:
Actually, I'd prefer not to claim the bounty.  Also, my employer is quite strict regarding paid outside business activities.  Is there a way for the bounty to be claimed by the more general team somehow?

Pectojin:
Bountysource can do payouts to paypal, so maybe it can be funneled directly at the Duplicati paypal? 

Otherwise @kenkendk might be able to handle it.

duplicatibot:
This issue has been mentioned on **Duplicati**. There might be relevant details there:

https://forum.duplicati.com/t/local-archive-and-backup-verification-scheme-help-new-user/10450/2

duplicatibot:
[...]

PR #3582, opened 2019-01-03
Add option to specify percentage of samples for verification

warwickmm:
This adds an option to specify the percentage of samples selected for verification.  While the `backup-test-samples` option allows the user to specify the number of samples for verification, this may be small relative to the (unknown) number of remote files.  By specifying a percentage, the user has greater confidence that the entirety of the remote files will be tested in a reasonable amount of time.

Note that unlike the `backup-test-samples` option, the default value for this new option is `0`.  This was done with consideration for bandwidth caps, etc.

This addresses issue #3296.


Duplicati/Library/Main/Options.cs
```diff
@@ -1734,6 +1735,18 @@ public bool NoBackendverification
...
                 
+        /// <summary>
+        /// Gets the percentage of samples to test during a backup operation
+        /// </summary>
+        public long BackupTestPercentage
+        {
+            get
+            {
+                m_options.TryGetValue("backup-test-percentage", out string s);
+                return string.IsNullOrEmpty(s) ? 0 : Math.Min(Math.Max(long.Parse(s), 0), 100);
```
verhoek: Can you validate the percentage? If parse throws exception or the number is negative, tell user has to be a number between 0 and 100. I know, negative gets mapped to 0, which should be ok.

Duplicati/Library/Main/Operation/BackupHandler.cs
```diff
@@ -271,15 +271,17 @@ private void PostBackupVerification()
                 backend.WaitForComplete(m_database, null);
             }
 
-            if (m_options.BackupTestSampleCount > 0 && m_database.GetRemoteVolumes().Any())
+            long remoteVolumeCount = m_database.GetRemoteVolumes().LongCount(x => x.State == RemoteVolumeState.Verified);
+            long samplesToTest = Math.Max(m_options.BackupTestSampleCount, (long)Math.Round(remoteVolumeCount * (m_options.BackupTestPercentage / 100D), MidpointRounding.AwayFromZero));
```
verhoek: Why do you multiple with the number of remote volumes?

Duplicati/Library/Main/Options.cs
warwickmm: Good idea.  Revision 52790cc7a9759e5824466e2f870201e3175bbc54 validates the value and throws an exception if it is out of range.

Duplicati/Library/Main/Operation/BackupHandler.cs
warwickmm: I believe this should yield the number of files to test, which is what needs to be provided to `TestHandler.DoRun`.  The `GetRemoteVolumes` method returns entries for all the `dindex`, `dlist`, and `dblock` files on the backend.

Duplicati/Library/Main/Options.cs
verhoek: Ok, but Long.parse will throw a parsing exception if it's not a long. Can you catch it and output the message you added already?

warwickmm: No problem.  Revision b4232ce1bfc1084eaa58704fb90ec21511044a47 catches any parse exception and provides a more meaningful message to the user.

Duplicati/Library/Main/Operation/BackupHandler.cs
verhoek: I'm not familiar with this part of the code yet, but I understood from the issue you referred to that this percentage is with respect to the number of files in a backup set that should be tested. What is the meaning of a remote volume ? If it corresponds to a backed up file on a remote backend, then I understand.

warwickmm: My understanding is that both the `backup-test-samples` and `backup-test-percentage` options refer to the files on the backend.  For example, if the number of samples to test is 3, then it will attempt to verify up to 3 `dindex` files, 3 `dlist` files, and 3 `dblock` files, for a total of up to 9 files.  The total number of files tested is reflected in `TestResults: VerificationsActualLength` in the backup log.

The method that selects the files to be tested is [here](https://github.com/duplicati/duplicati/blob/master/Duplicati/Library/Main/Database/LocalTestDatabase.cs#L112).

verhoek: Ok, but how many actual files from the backup set are contained in dindex? If there's one dblock containing the entire backup set, and I specify 1% of my files should be verified, then nothing gets verified due to the rounding. This is a border case but still. So this percentage is with respect to blocks. Is it possible to find out, how many files are in a block and guarantee at least this percentage of files is checked? I'm also not sure what the best solution is, because I guess you can only verify blockwise.

warwickmm: The example you mention is the reason why we take the maximum of the samples implied by the `backup-test-samples` and `backup-test-percentage` options.  If there's a better approach, please let me know.

I don't think it's possible to guarantee how many user files are verified, since a file's blocks may be spread out across many `dblock` files.

Pectojin: I like the balance between always validating at least 3, but up to a percentage. 

I don't think worrying about individual user files makes sense. It's bad for performance, bad for user understanding, and it doesn't actually matter for validation what files you validate as long as you validate all volumes eventually.

verhoek: > The example you mention is the reason why we take the maximum of the samples implied by the `backup-test-samples` and `backup-test-percentage` options. If there's a better approach, please let me know.
> 
> I don't think it's possible to guarantee how many user files are verified, since a file's blocks may be spread out across many `dblock` files.

That's true. But can you then refer to blocks  instead of files in the outputs of the help command, to make this more specific?

warwickmm: One difficulty is that the option actually does pertain to files (`dindex`, `dlist`, and `dblock` files), and not blocks.  I tried to be consistent with the help for the `backup-test-samples` and `full-remote-verification` options so that users would recognize that the options are related.  

If the help text is unclear, perhaps we can reword these from
> After a backup is completed, some files are selected for verification on the remote backend.

to

> After a backup is completed, some files from the remote backend are selected for verification.

If that makes sense, I'd prefer that we make the change in a separate pull request as it feels a bit out of scope here.  The text in the user's manual should then be changed as well to be consistent.

Pectojin:
Codacy is complaining because it's a property getter capable of throwing errors. This is fine since it has to be caught somewhere server side and that's first time we load the options.

Duplicati/Library/Main/Operation/BackupHandler.cs
Pectojin: Seems fair

verhoek: I agree that in the already existing option it is also not precise with the phrasing 'some files' and I think this should be fixed. I know that they are indeed files, but we can say precisely what kind of files these are: the dlist, dblock and dindex files. One could mistake them for user files, which I would have had first in mind with a backup program and without in depth knowledge of the internals of it. And this is not true.

I hope you see my point with that :) I will merge it in and hope you or otherwise I will create a merge request with the help.

@Pectojin : I agree with that.

warwickmm:
> Codacy is complaining because it's a property getter capable of throwing errors. This is fine since it has to be caught somewhere server side and that's first time we load the options.

I saw that too, and there are other getters in the `Options` class where exceptions are thrown as well.  Perhaps we can later modify the `Options` class to perform more validation in the constructor.  I also noticed that there are several places where we should use an instance of `Options`, instead of just a `Dictionary<string, string>` to avoid issues such as in #3526.

Duplicati/Library/Main/Operation/BackupHandler.cs
warwickmm: That makes sense to me.  I'll create a pull request with the above mentioned modification to the help and we can continue the discussion there.

Pectojin:
I was originally thinking that we should just validate this input before saving it to the database. However, it's not unlikely that options will change (down/upgrades), so just blindly trusting what's saved in the DB isn't an option.

Validating all parameters in the constructor sounds like the right way to go.

Duplicati/Library/Main/Operation/BackupHandler.cs
warwickmm: Pull request #3585 was created to clarify the help text.

warwickmm:
> ... why create an additional parameter instead of changing backend-test-samples to a string and treating it as a percentage if it starts or egg ends with %?

Mainly because this approach just never crossed my mind.  :grin:

verhoek:
From above, using both parameters give additional functionality with taking the max of the too (i.e. the 1% example above).

Issue #3579, opened 2019-01-02
Remove the LocalBackupDatabase.BuildLookupTable method?


warwickmm:
I was browsing the code and came across the `LocalBackupDatabase.BuildLookupTable` method.  The only usage is in the [BackupHandler class](https://github.com/duplicati/duplicati/blob/master/Duplicati/Library/Main/Operation/BackupHandler.cs#L462), where it is accompanied by the following comment:
```
// This should be removed as the lookups are no longer used
m_database.BuildLookupTable(m_options);
```
If the method is no longer being used, we should remove it.  Otherwise, we should clarify the comment to avoid any confusion.

Pectojin:
Well, let's try! :) #3584

PR #3584, opened 2019-01-04
Remove unsed method and call

Pectojin:
As discussed in #3579

Seems to still compile and run fine. No visible issue from this change in my testing.


warwickmm:
I tested various operations (backup, compact, repair, restore, test) and everything appeared to work normally.  Can anyone (@kenkendk?) comment on what the `BuildLookupTable` does and why it isn't needed anymore?

kenkendk:
> Can anyone (@kenkendk?) comment on what the BuildLookupTable does and why it isn't needed anymore?

The code will build a lookup table for paths, such that queries checking if the file has changed can _potentially_ be done faster. Unfortunately, building the table also takes a bit of time, so the overall savings were never really great.

The code is only activated if `--disable-filepath-cache=false` is set. I don't think anyone has reported significant speedups from this, so we can remove it but we should set the option to be deprecated such that users having it toggled on gets a warning (add an extra string argument with a message saying is not longer doing anything):
https://github.com/duplicati/duplicati/blob/master/Duplicati/Library/Main/Options.cs#L514

And this should be removed as well:
https://github.com/duplicati/duplicati/blob/master/Duplicati/Library/Main/Options.cs#L1578-L1589

Pectojin:
I've removed the toggle method and, I think, set the flag as deprecated.

Pectojin:
Missed one reference to the deleted method.

kenkendk:
Great work!

PR #3596, opened 2019-01-11
Add User-Agent to AzureBlob backend requests

### LINKED ISSUE #3521: Azure Blob useragent enhancement
## Environment info
- **Duplicati version**: Latest
- **Operating system**: All
- **Backend**: Azure Blob

## Description
Please add custom user-agent string to AzureBlob connector. 

Please try hooking the user agent into the operation context sending request event as below to prepend to the existing value of the library:

```
OperationContext.GlobalSendingRequest += (sender, args) =>
    {
        string userAgent = <your user agent> + " " + Microsoft.WindowsAzure.Storage.Shared.Protocol.Constants.HeaderConstants.UserAgent;                
                args.Request.UserAgent = userAgent;
    };
```

Or if using latest SDK, hook into CustomUserAgent property: [OperationContext.CustomUserAgent Property ](https://docs.microsoft.com/dotnet/api/microsoft.windowsazure.storage.operationcontext.customuseragent?view=azure-dotnet#Microsoft_WindowsAzure_Storage_OperationContext_CustomUserAgent)

If possible, please align to the format: APN/1.0 {Product Name}/{Product Version} {Feature Name}/{Feature Version}
e.g. APN/1.0 Duplicati/2.0 AzureBlob/2.0

mikaelmello:
Closes #3521 

@xyh1 is that it? I haven't tested to see if it works, just wanted to get a PR created.

Also, we didn't use the OperationContext class before, should we? Even if all our operations are on top of CloudStorageAccount, will the change in OperationContext have the intended effect?

### DIFF
--- a/Duplicati/Library/Backend/AzureBlob/AzureBlobWrapper.cs
+++ b/Duplicati/Library/Backend/AzureBlob/AzureBlobWrapper.cs
@@ -62,11 +62,18 @@ namespace Duplicati.Library.Backend.AzureBlob
 
         public AzureBlobWrapper(string accountName, string accessKey, string containerName)
         {
-            _containerName = containerName;
+            OperationContext.GlobalSendingRequest += (sender, args) =>
+            {
+                string userAgent = "APN/1.0 Duplicati/2.0 AzureBlob/2.0 " + Microsoft.WindowsAzure.Storage.Shared.Protocol.Constants.HeaderConstants.UserAgent;
+                args.Request.UserAgent = userAgent;
+            };
+
             var connectionString = string.Format("DefaultEndpointsProtocol=https;AccountName={0};AccountKey={1}",
                 accountName, accessKey);
             var storageAccount = CloudStorageAccount.Parse(connectionString);
             var blobClient = storageAccount.CreateCloudBlobClient();
+
+            _containerName = containerName;
             _container = blobClient.GetContainerReference(containerName);
         }
 


Duplicati/Library/Backend/AzureBlob/AzureBlobWrapper.cs
```diff
@@ -62,11 +62,18 @@ public string[] DnsNames
 
         public AzureBlobWrapper(string accountName, string accessKey, string containerName)
         {
-            _containerName = containerName;
+            OperationContext.GlobalSendingRequest += (sender, args) =>
+            {
+                string userAgent = "APN/1.0 Duplicati/2.0 AzureBlob/2.0 " + Microsoft.WindowsAzure.Storage.Shared.Protocol.Constants.HeaderConstants.UserAgent;
```
verhoek: I wouldn't put in a hard coded string like this, rather a constant. You could add a property to the class for setting this. Also if we want to supply a user agent to services (is this the only place we do it), so that services can see what program is being used (whether this is a good thing I will leave out of the discussion), it could be a global setting for consistency.

kenkendk: In other places I have used the library's version:
https://github.com/duplicati/duplicati/blob/master/Duplicati/Library/Backend/WEBDAV/WEBDAV.cs#L371

kenkendk: Sounds good to me. 👍

kenkendk: > Cons: The service could target our users? I don't know, either way, with a opt-out option this would be mitigated.

Should that happen, we can discuss the appropriate action. Generally, a service acting on the User-Agent header is asking for trouble, but it is a nice option for service debugging, stats and customer support (i.e. "the software you are using has a know issue, please upgrade").

If a provider wants to block Duplicati requests, I don't think we should play cat-and-mouse with them. Their users should complain or change providers.

kenkendk: @mikaelmello  I can make the change if you prefer?

mikaelmello: I guess that's better

kenkendk:
Closing this and merging from #3630

Issue #3592, opened 2019-01-08
dindex.zip files placed outside of FTP backup path


balchen:
- [x] I have searched open and closed issues for duplicates.

----------------------------------------

## Environment info

- **Duplicati version**: 2.0.4.5_beta_2018-11-28
- **Operating system**: Windows Server 2008 R2 64 bit
- **Backend**: IIS FTP

## Description

Duplicati is set up to backup to FTP at /duplicati_path. Files are mainly placed in /duplicati_path, but some files are placed in /. These are all dindex.zip files. Duplicati seems to assume it can upload files to /. This would not be possible if the FTP user didn't also have write access to /, which it might not have moving forward.

I have three backup jobs from two different Duplicati server installations, and all three place files in /, in addition to their designated backup paths.

## Steps to reproduce

1. setup Duplicati to backup to FTP
2. run a backup
3. inspect the parent folder of the FTP backup folder

- **Actual result**: 
Files placed outside of configured backup path
- **Expected result**: 
No files placed outside of configured backup path

balchen:
From looking at the code, the reason appears to be a bug in the AlternativeFTPBackend. FluentFTP.FtpClient will silently reconnect if there's a connection failure. AlternativeFTPBackup will only call SetWorkingDirectory once, when it instantiates FtpClient.

If the connection fails during upload, FtpClient will silently reconnect, and the retry will execute in the root path (whatever that is for the FTP user), not the Duplicati configured path.

I've created a pull request with a suggested fix. #3600

PR #3600, opened 2019-01-12
Call SetWorkingDirectory every time an open FTP connection is request…

balchen:
…ed to prevent files from ending up in the root folder if FtpClient fails and silently reconnects.

Fix for issue #3592


kenkendk:
Good work on finding a hard-to-replicate issue. I guess the FTP library considers "silent reconnect" to be a nice feature, but it ends up breaking things.

Issue #3366, opened 2018-09-06
Backblaze B2 Application Keys fail to authorize


sshaikh:
- [ x] I have searched open and closed issues for duplicates.

----------------------------------------

## Environment info
- **Duplicati version**: 2.0.3.3
- **Operating system**: Debian
- **Backend**: Backblaze B2

## Description
Connection test fails when trying to use application keys - using the master key works.

## Steps to reproduce
Create new B2 bucket
Create and note new master key
Create and note new application key that is restricted to use the above bucket
Create a new backup job in Duplicati that backs up to this bucket
Use master key - connection test works
Use application key - connection test fails

- **Actual result**:
Connection test fails with "Failed to connect: 401 - unauthorized:"
- **Expected result**:
Connection test passes

## Notes
The use of B2 application keys differs from the master key, so needs explicit support (I suspect this support isn't present)

ts678:
There are reports of success, e.g. [here](https://forum.duplicati.com/t/linux-mint-19-and-backblaze-b2/4465/3?u=ts678) and [here](https://forum.duplicati.com/t/new-user-with-duplicati-feedback/4565/6?u=ts678).   It also works for me, although I haven't tested extensively.

Do you have specific information on what sort of support is needed? I was hoping it's just docs and labels.

sshaikh:
I can confirm that using the key id as the account id makes application keys work - but I had to ask on the forum as it wasn't clear.

So yes, perhaps the change is for docs/labels or a hint on the setting screen?

bph84:
FWIW, Googling and the above comment was how I got past the issue.

poleguy:
This should be re-opened. It's still an issue for any new users who haven't stumbled here. On the backup destination page, the box reads "B2 Account ID" and "B2 Cloud Storage Account ID" and they should be changed to read: "B2 Application Key ID" and "B2 Cloud Storage Application Key ID" as that's what's actually required.

In addition the error messages if you get it wrong are not helpful. Change those to add 'Be sure to enter the application key ID, not the account ID in the "B2 Application Key ID" field.'

ts678:
This is a [new](https://www.backblaze.com/blog/b2-application-keys/) B2 scheme, the old one still works, and wording should probably offer a choice as [B2 does](https://www.backblaze.com/b2/docs/b2_authorize_account.html):

>  You can use either the master application key or a normal application key.
>
[...]

PR #3647, opened 2019-02-05
Removed account id to reduce confusion

Pectojin:
Cosmetic change for https://github.com/duplicati/duplicati/issues/3366#issuecomment-460830882

Nothing needs to change in the backend code, since it's always worked with Application ID, but since Account ID does not work anymore this will reduce confusion.

### DIFF
--- a/Duplicati/Server/webroot/ngax/templates/backends/b2.html
+++ b/Duplicati/Server/webroot/ngax/templates/backends/b2.html
@@ -9,8 +9,8 @@
 </div>
 
 <div class="input text">
-    <label for="b2_username" translate>B2 Account ID or Application ID</label>
-    <input type="text" name="b2_username" id="b2_username" ng-model="$parent.Username" placeholder="{{'B2 Cloud Storage Account ID' | translate}}" />
+    <label for="b2_username" translate>B2 Application ID</label>
+    <input type="text" name="b2_username" id="b2_username" ng-model="$parent.Username" placeholder="{{'B2 Cloud Storage Application ID' | translate}}" />
 </div>
 <div class="input password">
     <label for="b2_password" translate>B2 Application Key</label>


Pectojin:
Yes, all languages will have to retranslate the phrase. Which is ideal since they've all been saying the wrong thing presumably.

ts678:
[b2_authorize_account](https://www.backblaze.com/b2/docs/b2_authorize_account.html) says:

You can use either the master application key or a normal application key.

You'll find the master key on the B2 Cloud Storage Buckets page on the web site. When using the master key, the "key id" is your account ID, and the "application key" is the one you got from the web site.

and both ways still work here. I'm worried by [rclone](https://rclone.org/b2/) which sounds like yours, so I'm not sure which is correct...

ikjadoon:
Thank you for the follow-up, ts678, and thank you for creating this commit, pectojin.

[My previous comment simplified](https://github.com/duplicati/duplicati/issues/3366#issuecomment-460830882) it a bit too much and ts678 is correct that they _both can work_, but the wording in Duplicati is just a tad off.

In B2, there are two types of access. To give access to **all buckets**, the API will only accept:

1. `Account ID` (automatically generated for each B2 account) plus...
2. `Master Application Key` (must be generated in the B2 Buckets menu)

To_ give access to **one bucket**, the API will only accept:

1. `Key ID` (generated upon creating a new application key) plus...
2. `Application Key` (generated upon creating a new application key)

Thus, _**four**_ unique IDs and only the "matching set" will authorize properly in the B2 API. Account ID & Master Application ID must always go together; Key ID and Application Key must always go together. I could see Duplicati users needing either or other, so perhaps removing "account ID" might end up adding more confusion. 

I apologize for my previous inaccurate comment.

For what it's worth, B2's documentation explains this distinction....not easily and I needed to read it thrice to understand the two authentication sets. For further validation, [Transmit's explanation matches the above](https://library.panic.com/transmit/transmit5/bb2-faq/), as does [this comment by a Backblaze employee](https://www.backblaze.com/blog/b2-application-keys/#comment-4015356187):

image

### TL;DR both can be used, but Duplicati should add "**Master** Application Key" to the text

Perhaps like so:

**Account ID** (all buckets) or **Key ID** (one bucket)
**Master Application Key** (all buckets) or **Application Key** (one bucket)

If that's too much text, then perhaps just adding "Master Application Key" to the second box's prompt and users can try to mentally match them.

Pectojin:
That makes sense, although a bit confusing. 

I think "application id" does a decent job of conveying "[bucket or master] application id", since both are called application id's.

ikjadoon:
Absolutely confusing, I agree. B2's vague GUI doesn't help. 

// 

Valid point, though B2 doesn't ever actually use "Application ID" and you can't find that on the interface. It's either "ApplicationKeyID" (API's name) or "KeyID" (interface's name).

image

Maybe `Key/Account ID` and `Application/Master Key` ?

Issue #3653, opened 2019-02-08
Allow to use custom domain for b2 DownloadUrl


aahung:
- [x] I have searched open and closed issues for duplicates.

----------------------------------------

## Environment info

- **Duplicati version**: 2.0.4.5
- **Operating system**: macOS 10.14.3 (18D42)
- **Backend**: b2

## Description

B2 allow we utilize Cloudflare to serve as a frontend and therefore we can use our custom domains for b2. Allowing setting DownloadUrl will save bandwidth cost when Duplicati downloads files from b2 since traffic between Cloudflare and b2 does not cost money.

aahung:
See PR #3654

PR #3654, opened 2019-02-09
Allow to use custom domain for b2 DownloadUrl #3653

aahung:
- how custom download url can be set with option "b2-download-url"

And the option will look like this:

image

Network-wise, the download url has been replaced as well:
image

### DIFF
--- a/Duplicati/Library/Backend/Backblaze/B2.cs
+++ b/Duplicati/Library/Backend/Backblaze/B2.cs
@@ -29,6 +29,7 @@ namespace Duplicati.Library.Backend.Backblaze
         private const string B2_ID_OPTION = "b2-accountid";
         private const string B2_KEY_OPTION = "b2-applicationkey";
 		private const string B2_PAGESIZE_OPTION = "b2-page-size";
+        private const string B2_DOWNLOAD_URL_OPTION = "b2-download-url";
 
 		private const string B2_CREATE_BUCKET_TYPE_OPTION = "b2-create-bucket-type";
         private const string DEFAULT_BUCKET_TYPE = "allPrivate";
@@ -40,6 +41,7 @@ namespace Duplicati.Library.Backend.Backblaze
         private readonly string m_urlencodedprefix;
         private readonly string m_bucketType;
         private readonly int m_pagesize;
+        private readonly string m_downloadUrl;
         private readonly B2AuthHelper m_helper;
         private UploadUrlResponse m_uploadUrl;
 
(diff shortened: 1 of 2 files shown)


aahung:
#3653

Duplicati/Library/Backend/Backblaze/B2.cs
```diff
@@ -163,6 +181,7 @@ public IList<ICommandLineArgument> SupportedCommands
                     new CommandLineArgument("auth-username", CommandLineArgument.ArgumentType.String, Strings.B2.AuthUsernameDescriptionShort, Strings.B2.AuthUsernameDescriptionLong),
                     new CommandLineArgument(B2_CREATE_BUCKET_TYPE_OPTION, CommandLineArgument.ArgumentType.String, Strings.B2.B2createbuckettypeDescriptionShort, Strings.B2.B2createbuckettypeDescriptionLong, DEFAULT_BUCKET_TYPE),
                     new CommandLineArgument(B2_PAGESIZE_OPTION, CommandLineArgument.ArgumentType.Integer, Strings.B2.B2pagesizeDescriptionShort, Strings.B2.B2pagesizeDescriptionLong, DEFAULT_PAGE_SIZE.ToString()),
+                    new CommandLineArgument(B2_DOWNLOAD_URL_OPTION, CommandLineArgument.ArgumentType.String, Strings.B2.B2downloadurlDescriptionShort, Strings.B2.B2downloadurlDescriptionLong),
```
Pectojin: I would recommend using the overloaded `CommandLineArgument` method that allows defining default value.

Then you can delete the part of the long description that states the default value.

`public CommandLineArgument(string name, ArgumentType type, string shortDescription, string longDescription, string defaultValue, string[] aliases)
            : this(name, type, shortDescription, longDescription, defaultValue)`

aahung: Actually there is no default hardcoded value for `downloadUrl`. I put it in the description as `https://f00X.backblazeb2.com` where `X` depends on the account. Maybe I need to clarify it in the description a little bit. Thanks

aahung:
@Pectojin I revised the description in [49c7050](
https://github.com/duplicati/duplicati/pull/3654/commits/49c70503b8765eb7cc0df786617a49c3fd6927ae) to avoid the confusion.

Duplicati/Library/Backend/Backblaze/B2.cs
Pectojin: Oh, that makes more sense

Pectojin:
Adding this parameter to the destination form when editing/creating a B2 backup may be a nice addition to this, but lets merge the base functionality as is.

PR #3655, opened 2019-02-09
Fix Sia API authentication not sending credentials

### LINKED ISSUE #3002: Can't use Sia backend with password authentification

## Environment info
- **Duplicati version**: 2.0.2.17_canary_2018-01-23
- **Operating system**: Debian 8
- **Backend**: Sia Decentralize Cloud

## Description
Duplicati fails to authenticate with the Sia API when providing the correct password. Duplicati works perfectly when no password is set for both duplicati and Sia.

## Steps to reproduce
1. Setup Sia with a password for the API: `siad --authenticate-api`.
2. Add a backup on duplicati and provide the `server password`.
3. Run the backup.

- **Actual result**:
Duplicati throw an error when trying to communicate with the Sia API.
- **Expected result**:
Working fine.

## Screenshots
selection_273
selection_275

## Debug log
```
System.Exception: /renter/upload/backup/duplicati-20180202T182644Z%2Edlist%2Ezip?source=%2Ftmp%2Fdup-b9d2175f-95f3-45d9-b8ff-39b3f3dc4ab9 failed, response: {"message":"API authentication failed."}

  at Duplicati.Library.Main.Operation.BackupHandler.HandleFilesystemEntry (Duplicati.Library.Snapshots.ISnapshotService snapshot, Duplicati.Library.Main.BackendManager backend, System.String path, System.IO.FileAttributes attributes) [0x0000e] in <4ba6789697e24783ae57b0acf97a6486>:0 
  at Duplicati.Library.Main.Operation.BackupHandler.RunMainOperation (Duplicati.Library.Snapshots.ISnapshotService snapshot, Duplicati.Library.Main.BackendManager backend) [0x0018a] in <4ba6789697e24783ae57b0acf97a6486>:0 
  at Duplicati.Library.Main.Operation.BackupHandler.Run (System.String[] sources, Duplicati.Library.Utility.IFilter filter) [0x008a5] in <4ba6789697e24783ae57b0acf97a6486>:0 
  at Duplicati.Library.Main.Controller+<>c__DisplayClass17_0.<Backup>b__0 (Duplicati.Library.Main.BackupResults result) [0x00036] in <4ba6789697e24783ae57b0acf97a6486>:0 
  at Duplicati.Library.Main.Controller.RunAction[T] (T result, System.String[]& paths, Duplicati.Library.Utility.IFilter& filter, System.Action`1[T] method) [0x0014b] in <4ba6789697e24783ae57b0acf97a6486>:0 
  at Duplicati.Library.Main.Controller.Backup (System.String[] inputsources, Duplicati.Library.Utility.IFilter filter) [0x00068] in <4ba6789697e24783ae57b0acf97a6486>:0 
  at Duplicati.Server.Runner.Run (Duplicati.Server.Runner+IRunnerData data, System.Boolean fromQueue) [0x00458] in <66004e16a6074efdaebb0aaaf058863a>:0 
```

Ajedi32:
.NET ignores credentials with blank usernames. We can work around this by sending a bogus username, which Sia simply ignores. I'm pretty new to .NET so I'm not sure if this is the best way to solve the problem, but I tested it and it _does_ seem to work.

Fixes #3002

### DIFF
--- a/Duplicati/Library/Backend/Sia/Sia.cs
+++ b/Duplicati/Library/Backend/Sia/Sia.cs
@@ -20,7 +20,7 @@ namespace Duplicati.Library.Backend.Sia
         private readonly int m_apiport;
         private readonly string m_targetpath;
         private readonly float m_redundancy;
-        private readonly System.Net.NetworkCredential m_user;
+        private readonly string m_authorization;
 
         public Sia() {
 
@@ -55,11 +55,9 @@ namespace Duplicati.Library.Backend.Sia
             if (m_targetpath.Length == 0)
                 m_targetpath = "backup";
 
-            m_user = new System.Net.NetworkCredential();
-            if (options.ContainsKey(SIA_PASSWORD))
-            {             
-                m_user.Password = options[SIA_PASSWORD];
-            }
+            m_authorization = options.ContainsKey(SIA_PASSWORD) && !string.IsNullOrEmpty(options[SIA_PASSWORD])
+                ? "Basic " + System.Convert.ToBase64String(System.Text.Encoding.GetEncoding("ISO-8859-1").GetBytes(":" + options[SIA_PASSWORD]))
+                : null;
         }
 
         private System.Net.HttpWebRequest CreateRequest(string endpoint)
@@ -67,10 +65,10 @@ namespace Duplicati.Library.Backend.Sia
             string baseurl = "http://" + m_apihost + ":" + m_apiport;
             System.Net.HttpWebRequest req = (System.Net.HttpWebRequest)System.Net.HttpWebRequest.Create(baseurl + endpoint);
 
-            if (!string.IsNullOrEmpty(m_user.Password))
+            if (m_authorization != null)
             {
-                req.Credentials = m_user;
-                req.PreAuthenticate = true;
+                // Manually set Authorization header, since System.Net.NetworkCredential ignores credentials with empty usernames
+                req.Headers.Add("Authorization", m_authorization);
             }
 
             req.KeepAlive = false;


verhoek:
Does this work with an empty string for the username as well (which is different than the not being present of the username)? According to SIA, an empty string is required (https://sia.tech/docs/#authentication). This commit gives a username and probably works since SIA ignores the username. Relying implicitly on this fact can break if SIA starts to check non-empty user names.

Ajedi32:
No, the username is an empty string by default, which results in .NET not sending any authentication header at all.

I suppose I could try setting the header directly rather than relying on .NET's credentials system. I'll give that a shot...

verhoek:
Ok, that could be, I didn't see it in the api of httpwebrequest. The authentication header you can set directly yes, but see how much hassle that is with basic auth and encoding. Otherwise, I think this is fine since it's simpler and we'd need to fix it once SIA changes behaviour.

Ajedi32:
Yeah, I didn't see any information about this in the .NET documentation either. I was only able to determine what was happening by observing the API requests with ngrep.

I don't think manually setting the header would be _too_ difficult; the encoding is just Base64. Your call though. I'm fine with either approach.

verhoek:
Indeed: https://stackoverflow.com/questions/25852551/how-to-add-basic-authentication-header-to-webrequest/25852562 . I guess it's better to do it like that explicitly and add a comment why we are doing it in this way.

Ajedi32:
Alright, will do.

Ajedi32:
^ That should do it.

verhoek:
Thanks! Independent of the fixed issue but regarding basic auth, I just noticed base url is fixed to be http not https. I never looked into the SIA backend nor am familiar with SIA but this looks questionable.

Ajedi32:
As things currently stand I believe the Sia server has to be running on the same machine as Duplicati anyway, so https shouldn't be necessary.

I can imagine a few scenarios where it could be useful to have Sia running on a different machine, but it doesn't seem like those use-cases would be very well supported anyway by the current design of Sia's API.

The file upload API, for example, [takes a path to a file on the local machine as one of its arguments](https://sia.tech/docs/#renter-upload-siapath-post). That wouldn't work very well if the server was on a different machine from Duplicati.

Issue #3652, opened 2019-02-08
Disk full in /tmp, dup-xxxx files not getting cleaned up


instantlinux:
- [x ] I have searched open and closed issues for duplicates.

----------------------------------------

## Environment info

- **Duplicati version**: 2.0.4.2_experimental_2018-11-12
- **Operating system**: Debian Stretch (docker)
- **Backend**: Backblaze B2

## Description

```
Error while running <backup name>
Disk full. Path /tmp/dup-30a88439-df78-4a2e-8840-5763ee2b27df
```
## Steps to reproduce

1. Set up a backup
2. Run for a couple of weeks

- **Actual result**:
Backup didn't complete
- **Expected result**:
The /tmp volume should be cleaned up after each run

## Screenshots

10GB of tmp files created in past 2 weeks:
```
# ls -l
total 9229696
drwxr-xr-x 2 root backup      4096 Jan 26 12:55 HttpServer
-rw-r--r-- 1 root backup 524304366 Feb  4 07:24 dup-0d5f37bd-af00-4135-98ca-e0a3b2459967
-rw-r--r-- 1 root backup       156 Jan 28 02:06 dup-0ee35bc2-01eb-4a5f-9d49-6324b44f4d56
-rw-r--r-- 1 root backup 524319268 Feb  4 06:34 dup-137e3a87-fdfa-4473-b443-adf27fd46e95
-rw-r--r-- 1 root backup    555615 Feb  4 06:24 dup-1726a950-3ad2-453e-9940-526655ca74c4
-rw-r--r-- 1 root backup 524351678 Jan 30 09:49 dup-1e897042-35ca-4850-8050-656d950d7b10
-rw-r--r-- 1 root backup       338 Feb  7 03:53 dup-20bb8b9f-0df9-4162-bf58-5d9bfd51f6f0
-rw-r--r-- 1 root backup 524259161 Feb  7 03:53 dup-20c1038c-952f-4cdd-a510-e95c0b695ff1
-rw-r--r-- 1 root backup    299676 Feb  6 09:15 dup-2186b888-72ce-41d6-9ce5-bb107a78eaa9
-rw-r--r-- 1 root backup       156 Jan 29 03:01 dup-21ca1928-d0e2-4888-9d75-ad1e0fcad742
-rw-r--r-- 1 root backup       156 Feb  4 09:35 dup-2293a43b-0768-40d2-9c9f-12bf4f54b4ad
[...]

PR #3660, opened 2019-02-11
Ensure removal of VolumeWriterBase temporary files

warwickmm:
This ensures that instances of `VolumeWriterBase` are disposed of after their use.  Each `VolumeWriterBase` contains a `TempFile`, which represents a temporary file that is deleted upon disposal.  If we do not dispose of these instances, temporary files may be left behind.

Some of these instances could not be put inside a `using` block, as the variables are being redefined inside a loop (the object in a `using` block must be `readonly`).  In these cases, we manually call `Dispose` before the variable reassignment.

This concerns issue #3652.

### DIFF
--- a/Duplicati/Library/Main/Operation/RepairHandler.cs
+++ b/Duplicati/Library/Main/Operation/RepairHandler.cs
@@ -304,134 +304,138 @@ namespace Duplicati.Library.Main.Operation
                             }
                             else if (n.Type == RemoteVolumeType.Index)
                             {
-                                var w = new IndexVolumeWriter(m_options);
-                                newEntry = w;
-                                w.SetRemoteFilename(n.Name);
-
-                                var h = Library.Utility.HashAlgorithmHelper.Create(m_options.BlockHashAlgorithm);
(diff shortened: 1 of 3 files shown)


warwickmm:
I was unable to reproduce the issue described in issue #3652 (I could not induce my test cases to require a compact operation), but disposing of these instances seems like the right thing to do and should reduce the number of temporary files left behind.

Duplicati/Library/Main/Operation/RepairHandler.cs
```diff
@@ -304,137 +304,141 @@ public void RunRepairRemote()
                             }
                             else if (n.Type == RemoteVolumeType.Index)
                             {
-                                var w = new IndexVolumeWriter(m_options);
```
warwickmm: The following hunks were due to having MonoDevelop enclose the variables in a `using` block.  Unfortunately, this resulted in a messy diff.

warwickmm:
Looks like some tests are failing.  I'll have to investigate some more.

Issue #3634, opened 2019-01-28
asynchronous-upload-limit setting ignored


jheyneman:
- [x] I have searched open and closed issues for duplicates.

----------------------------------------

## Environment info

- **Duplicati version**:  2.0.4.12_canary_2019-01-16
- **Operating system**: Debian 
- **Backend**: Google Drive

## Description

asynchronous-upload-limit = 2 is ignored

## Steps to reproduce

1. Set asynchronous-upload-limit to anything other than default (4)
2. Run Backup and monitor temp folder.

- **Actual result**:
More than 2 backup archives are in temp directory specified. In my case, reaching 5 complete archives and 1 partial before my 12GB ramdisk runs out of space. My backup then errors out with a disk space error.
- **Expected result**:
I expect 2 backup archives to be pending for upload in the temp directory, up to 3 if you include the current file being uploaded.

## Screenshots

![image](https://user-images.githubusercontent.com/21336105/51858465-c98eda80-2302-11e9-9f59-2dea2cdee9ad.png)
![image](https://user-images.githubusercontent.com/21336105/51858541-f8a54c00-2302-11e9-9914-ac9bd6a3ef78.png)

warwickmm:
I believe this is fixed with pull request #3662.  If you find that the `asynchronous-upload-limit` option is still begin ignored, please let us know be reopening this issue.

huste:
I just installed Duplicati 2.0.4.18 on my father's PC (Win10, running from the Tray, not as a service).
As he has a really lame internet connection, I decided to use 
  asynchronous-upload-limit=20
  asynchronous-concurrent-upload-limit=1
both, on the backup job an din the general settings.
While the latter works and only uploads one ZIP archive at a time, the former doesn't. I still get 4 pre-zipped packages in the temporary directory.
The size for each package is set to 50MB in the job, which gets respected.

The backup job uses a plain FTP connection.

@warwickmm: I don't seem to be able to reopen this one, so could someone do this, please?

jheyneman:
@huste I had forgotten that I logged this and never tested it. I'll give it a shot on my end and post back.

warwickmm:
Pull request #3662 cleaned up temporary files for _successful_ backup and compact operations.  Temporary files from failed (or aborted) operations will still be left behind and is a separate issue that needs to be addressed.

huste:
[...]

PR #3662, opened 2019-02-17
Remove temporary files from backup and compact operations

warwickmm:
This ensures removal of temporary files that were left behind in the following situations:
1. After a backup operation, a temporary file containing the `filelist.json` was left in the temporary folder.  This file remained whether or not files needed to be uploaded to the backend.
2. If the compact operation is run but [ShouldCompact](https://github.com/duplicati/duplicati/blob/master/Duplicati/Library/Main/Operation/CompactHandler.cs#L138) is `false`, a few small temporary files would be left behind.
3. If the compact operation is run but [ShouldCompact](https://github.com/duplicati/duplicati/blob/master/Duplicati/Library/Main/Operation/CompactHandler.cs#L138) is `true`, many temporary files (I believe containing the filelists and unencrypted blocks) would be left behind.  These files could be large in size, and would result in a user's temporary directory getting filled up.

Note that the temporary files might not be removed in cases where exceptions are thrown.  This is due to the `IDisposable` objects not being contained in `using` statements, as the variables are being reassigned inside a loop.

This addresses issue #3652, and possibly issue #3634.

### DIFF
--- a/Duplicati/Library/Main/Operation/CompactHandler.cs
+++ b/Duplicati/Library/Main/Operation/CompactHandler.cs
@@ -110,20 +110,19 @@ namespace Duplicati.Library.Main.Operation
                         newvolindex = new IndexVolumeWriter(m_options);
                         newvolindex.VolumeID = db.RegisterRemoteVolume(newvolindex.RemoteFilename, RemoteVolumeType.Index, RemoteVolumeState.Temporary, transaction);
                         db.AddIndexBlockLink(newvolindex.VolumeID, newvol.VolumeID, transaction);
-                        newvolindex.StartVolume(newvol.RemoteFilename);
                     }
-                    
+
                     long blocksInVolume = 0;
                     long discardedBlocks = 0;
                     long discardedSize = 0;
                     byte[] buffer = new byte[m_options.Blocksize];
                     var remoteList = db.GetRemoteVolumes().Where(n => n.State == RemoteVolumeState.Uploaded || n.State == RemoteVolumeState.Verified).ToArray();
-                    
+
                     //These are for bookkeeping
                     var uploadedVolumes = new List<KeyValuePair<string, long>>();
                     var deletedVolumes = new List<KeyValuePair<string, long>>();
                     var downloadedVolumes = new List<KeyValuePair<string, long>>();
-                    
+
                     //We start by deleting unused volumes to save space before uploading new stuff
                     var fullyDeleteable = (from v in remoteList
                                            where report.DeleteableVolumes.Contains(v.Name)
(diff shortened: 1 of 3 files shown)


Duplicati/Library/Main/Operation/CompactHandler.cs
```diff
@@ -236,13 +236,18 @@ where report.CompactableVolumes.Contains(v.Name)
                             }
                         }
                     }
-                    
+                    else
+                    {
+                        newvolindex.Dispose();
+                        newvol.Dispose();
```
warwickmm: This removes files that remained after a compact operation was run, but `report.ShouldCompact` is `false`.

Duplicati/Library/Main/Volumes/FilesetVolumeWriter.cs
```diff
@@ -150,6 +150,7 @@ public void AddControlFile(string localfile, CompressionHint hint, string filena
         public override void Dispose()
         {
             this.Close();
+            base.Dispose();
```
warwickmm: This removes the temporary file containing the `filelist.json` that remained after each backup operation.  The base class contains a temporary file that is removed upon calling `Dispose`, so we need to call the `base` implementation here.

Duplicati/Library/Main/Volumes/VolumeWriterBase.cs
```diff
@@ -52,9 +52,6 @@ protected VolumeWriterBase(Options options, DateTime timestamp)
             else
                 m_localfile = new Library.Utility.TempFile();
 
-            // TODO(danstahr): This is a hack! Figure out why the file is being disposed of prematurely.
-            m_localfile.Protected = true;
-
```
warwickmm: This removes a workaround that was introduced in revision 31ac40083ab85926b656e87f937b90dd02e69a5f in pull request #2637.  From the comments in that pull request, it appears that there were issues when running tests in Travis with temporary files being removed prematurely.  It's unclear if that is still an issue (it appears that the underlying cause was unknown), but hopefully it has been resolved.

This prevented the temporary files created during a compact operation from being removed.

kenkendk:
@warwickmm Have you tested on Windows? IIRC there were some issues with the temporary files being locked on Windows, but no problems on Linux/MacOS.

warwickmm:
Sorry, I should have mentioned that I was only able to replicate/test on Linux.  I did not do any testing on Windows or MacOS.

sttepan:
Как это работает?

kenkendk:
@warwickmm Since the AppVeyor unittest passes, I think this is safe for Windows use also.

PR #3663, opened 2019-02-18
Remove unused Options properties

warwickmm:
These collections of options appear to be unused, and are not very well maintained.  Options are often introduced and these properties were not being updated.  Most of these properties have not been referenced in a commit patch since revision a0521da54d1614eae426e12f5b76da600376ef66.


kenkendk:
The intention was that the options should be used to warn the user if they apply options that do not work for the given operation. But it was not maintained and the UI is always sending all options which would trigger warnings in the UI.

PR #3681, opened 2019-02-28
Epic/keyring/1 refactor server program

verhoek:
Some refactoring of server/program.cs


Duplicati/Server/Program.cs
```diff
@@ -2,18 +2,22 @@
...
+using Duplicati.Library.Common.IO;
+
 namespace Duplicati.Server
 {
     public class Program
     {
+
+        private static readonly List<string> AlternativeHelpStrings = new List<string> { "help", "/help", "usage", "/usage", "--help" };
+
+        private static readonly List<string> ParameterFileOptionStrings = new List<string> { "parameters-file", "parameters-file", "parameterfile" };
```
warwickmm: The first two entries here look identical.  Is this a typo?

warwickmm: Actually, it looks like the previous implementation had this as well.  Perhaps we can use this opportunity to clean this up?

verhoek: I removed it, thanks for noticing.

Duplicati/Server/KeyManager/PasswordFromDatabase.cs
```diff
@@ -0,0 +1,44 @@
...
+        }
+
+        public string GetPassphrase(long backupId)
+        {
+            ISetting[] settings = con.GetSettings(backupId);
+            return settings.Where(s => s.Name == "passphrase")
+                           .Select(s => s.Value).FirstOrDefault();
+        }
+
+        public string PasswordSalt => throw new NotImplementedException();
```
warwickmm: I can't seem to find any usages of `PasswordSalt`.  Is this property needed for something in the future?

Duplicati/Server/KeyManager/PasswordFromDatabase.cs
```diff
@@ -0,0 +1,44 @@
...
+//  Foundation, Inc., 59 Temple Place, Suite 330, Boston, MA 02111-1307 USA
+using System;
+using Duplicati.Server.Database;
+using Duplicati.Server.Serialization.Interface;
+using System.Linq;
+
+
+namespace Duplicati.Security.KeyManager
+{
+    public class PasswordFromDatabase : IPasswordService
```
warwickmm: I don't see any usages of the `PasswordFromDatabase` class or the `IPasswordService` interface in the diff.  Did you intend to reference these somewhere, or are these for future use?

verhoek: This accidentally slipped in due to IDE eagerness of adding unstaged files. I tried to smoothen it out with rebasing but did not manage to do so (due to merge commits or similar), so removed these files in an extra commit.

Duplicati/Server/Program.cs
kenkendk: I _think_ it was supposed to be `parameter-file`, `parameters-file`, `parameterfile` to help users guessing their way with the commandline.

But if it did not work previously with the singular form, we can just remove it.

Issue #3673, opened 2019-02-24
Failed put operations do not result in error


warwickmm:
If all the `Put` operations for a backup fail, no error is reported to the user.  Even worse, the UI updates the "Last successful backup" time, which can lead the user to believe that the backup was successful.  The backup log shows a green check mark, and indicates 0 errors and 0 warnings.  

![image](https://user-images.githubusercontent.com/17345532/53294374-6d509680-379a-11e9-8f43-81a856c282d6.png)

![image](https://user-images.githubusercontent.com/17345532/53294430-670eea00-379b-11e9-9116-adb291bb8276.png)

If this occurs on the first backup, the user might sense something is wrong if they happen to see "Backup:0 bytes / 0 Versions".  If this occurs on a subsequent backup, it's unclear if there's any way for the user to know that the backup was unsuccessful.

Sample logs here:
```
  "WarningsActualLength": 0,
  "ErrorsActualLength": 0,
  "Messages": [
    "2019-02-23 18:33:36 -08 - [Information-Duplicati.Library.Main.Controller-StartingOperation]: The operation Backup has started",
    "2019-02-23 18:33:36 -08 - [Information-Duplicati.Library.Main.BasicResults-BackendEvent]: Backend event: List - Started:  ()",
    "2019-02-23 18:33:36 -08 - [Information-Duplicati.Library.Main.BasicResults-BackendEvent]: Backend event: List - Completed:  ()",
    "2019-02-23 18:33:36 -08 - [Information-Duplicati.Library.Main.Operation.FilelistProcessor-SchedulingMissingFileForDelete]: scheduling missing file for deletion, currently listed as Uploading: duplicati-20190224T023001Z.dlist.zip.aes",
    "2019-02-23 18:33:36 -08 - [Information-Duplicati.Library.Main.Operation.FilelistProcessor-RemoteUnwantedMissingFile]: removing file listed as Temporary: duplicati-bc4d8df45af10459f8587af1a0f257f8f.dblock.zip.aes",
    "2019-02-23 18:33:36 -08 - [Information-Duplicati.Library.Main.Operation.FilelistProcessor-RemoteUnwantedMissingFile]: removing file listed as Deleting: duplicati-be48686392a7c4b0ca436bbc410529767.dblock.zip.aes",
[...]

PR #3684, opened 2019-03-06
Implement parallel uploads to a single backend

### LINKED ISSUE #2026: Support parallel uploads for a backend [$381]
Some of the cloud backends (e.g. Dropbox) seem bandwidth limited by the cloud provider for individual connections (upload/download). One way to achieve higher throughput would be to upload backup volumes in parallel. Obviously this only works if there is more than 1 backup volume to be uploaded.
The same approach would also be possible for achieving higher throughput on downloads.

However, an important implication for the backend implementations is that they should be **thread-safe** , i.e. calling PUT should be possible in parallel.  Perhaps, an additional interface could be defined, IThreadSafeBackend or something, that IBackends can implement for indicating that they support parallel upload/download operations. 
## 

<bountysource-plugin>
---
There is a **[$50 open bounty](https://www.bountysource.com/issues/38460805-support-parallel-uploads-for-a-backend?utm_campaign=plugin&utm_content=tracker%2F4870652&utm_medium=issues&utm_source=github)** on this issue. Add to the bounty at [Bountysource](https://www.bountysource.com/?utm_campaign=plugin&utm_content=tracker%2F4870652&utm_medium=issues&utm_source=github).
</bountysource-plugin>

seantempleton:
The BackendUploader was heavily refactored to support concurrent uploads to a single backend. The functionality of encrypting, hashing, and creating index volumes were moved to the DataBlockProcessor and SpillCollectorProcess to allow the upload threads to continuously upload data rather than having to stop and encrypt data for example.

- The IBackend and IStreamingBackend interfaces were updated to change the Put() function to return a Task. All backend implementations were updated to support the new signature.
- Added a new option, asynchronous-concurrent-upload-limit to control how many uploads are allowed to happen concurrently. Defaults to 4.
- Added a new class to track the progress of all concurrent uploads and report it to the BackendProgressUpdater as well as throttle the uploads if necessary.

Resolves #2026


Duplicati/Library/Main/Operation/Backup/DataBlockProcessor.cs
```diff
@@ -95,12 +93,24 @@ public static Task Run(BackupDatabase database, Options options, ITaskReader tas
...
-                            
+
                                 blockvolume.Close();
 
                                 await database.CommitTransactionAsync("CommitAddBlockToOutputFlush");
 
-                                await self.Output.WriteAsync(new VolumeUploadRequest(blockvolume, indexvolume));
+                                FileEntryItem blockEntry = CreateFileEntryForUpload(blockvolume, options);
+
+                                IndexVolumeWriter indexVolume = null;
```
warwickmm: Minor point, but given that there is already a local variable `indexvolume`, can we rename this (perhaps to `indexVolumeWriter`)?  Having two local variables that differ only in case can make this difficult to read.

Duplicati/Library/Main/Operation/Backup/BackendUploader.cs
```diff
@@ -118,80 +115,55 @@ public Task Run()
...
+                        else if (req is IndexVolumeUploadRequest indexUpload)
+                            inProgress.Add(UploadVolumeWriter(indexUpload.IndexVolume));
+                        else if (req is FlushRequest flush)
                         {
                             try
                             {
-                                while(inProgress.Count > 0)
-                                    await inProgress.Dequeue().Value;
-                                active = 0;
+                                await Task.WhenAll(inProgress);
```
warwickmm: I'm not sure if I'm reading this correctly, but it seems that previously if a queued `Task` threw an exception, the remaining `Tasks` would not be awaited.  Now, all the `Tasks` are run until they complete.  If that's true, what are the side effects (if any) or differences in behavior (good or bad)?

Duplicati/Library/Main/Operation/Backup/BackendUploader.cs
```diff
@@ -85,98 +122,323 @@ public static Task Run(Common.BackendHandler backend, Options options, Common.Da
...
-                            task = new KeyValuePair<int, Task>(1, backend.UploadFileAsync(((IndexVolumeUploadRequest)req).IndexVolume));
-                        else if (req is FlushRequest)
+                        else if (req is FlushRequest flush)
                         {
                             try
                             {
-                                while(inProgress.Count > 0)
-                                    await inProgress.Dequeue().Value;
-                                active = 0;
+                                await Task.WhenAll(workers.Select(w => w.Task));
```
warwickmm: I'm not sure if I'm reading this correctly, but it seems that previously if a queued `Task` threw an exception, the remaining `Tasks` would not be awaited.  Now, all the `Tasks` are run until they complete.  If that's true, what are the side effects (if any) or differences in behavior (good or bad)?

seantempleton: I think you're reading that right. I was in the mindset of completing the uploads so it seemed to make sense to use `WhenAll` but you bring up a good point. In normal operation if an exception occurs we want to cancel all of the other currently in progress uploads so we should have the same behavior here where we are waiting for the last few uploads.

I think the downside to using `WhenAll` here would be that if an exception occurs we would wait for the other files to finish or throw their own exceptions, which could potentially take a long time. And then later on those files would just get deleted from the backend anyway.

In normal operation where everything works properly this is obviously the nicest way to go about waiting for the uploads. Unless you have other thoughts I'll go ahead with changing this to cancel the other uploads when an exception occurs.

Duplicati/Library/Main/Operation/Backup/DataBlockProcessor.cs
seantempleton: I agree, it is confusing. It slipped past me.. good catch.

Duplicati/UnitTest/RandomErrorBackend.cs
```diff
@@ -46,12 +48,12 @@ private void ThrowErrorRandom()
                 throw new Exception("Random upload failure");
         }
         #region IStreamingBackend implementation
-        public void Put(string remotename, Stream stream)
+        public async Task Put(string remotename, Stream stream, CancellationToken cancelToken)
```
warwickmm: Is there a reason why some implementations of `Put` use `async/await`, while others do not?

Duplicati/Library/Main/Operation/Backup/BackendUploader.cs
```diff
@@ -86,18 +115,18 @@ public static Task Run(Common.BackendHandler backend, Options options, Common.Da
...
                 
-                while(!await self.Input.IsRetiredAsync && await taskreader.ProgressAsync)
+                while(!await self.Input.IsRetiredAsync && await m_taskreader.ProgressAsync)
                 {
                     try
                     {
                         var req = await self.Input.ReadAsync();
 
-                        if (!await taskreader.ProgressAsync)
+                        if (!await m_taskreader.ProgressAsync)
```
verhoek: I am not a fan of the m_* member variable "convention" and would like to phase them out. I know the code is written mostly in such a way that it's not clear what local vars and member vars are -mainly due to large functions. It's quite possible I am the only one disliking the m_ style notation in duplicati team.

seantempleton: I dislike the m_* style as well. Being my first PR for the project I didn't want to step on any toes and kept with the current style. If there is an agreed upon standard I would be happy to switch to using that. I'm partial to m* or to no prefix at all myself.

warwickmm: If it were up to me, I would prefer to always use the `this` keyword when referring to instance members.  At least for me, it is an immediate and clear indication that the variable refers to an instance member.

I have also tried to refrain from disturbing the local style too much, as in the end, consistency should perhaps be the ultimate goal.

seantempleton: warwickmm that's actually the other one I dislike. :) But I can get used to anything as long as it's consistent. If I really need to know the scope of a variable Visual Studio will show that in the Intellisense tooltip. I realize not everyone has that information available to them though. And I generally like to keep functions short enough that it is easy to tell anyway.

warwickmm: For me, moving the cursor and hovering is too much effort (especially on a laptop).  But yes, as long as we're consistent I am fine with any style.

Duplicati/Library/Main/Operation/Backup/BackendUploader.cs
```diff
@@ -369,28 +380,29 @@ private async Task DoPut(FileEntryItem item, IBackend backend, CancellationToken
...
+        private void HandleProgress(ThrottledStream ts, long progress, string path)
         {
             if (!m_taskreader.TransferProgressAsync.WaitForTask().Result)
                 throw new OperationCanceledException();
 
-            // Update the throttle speeds if they have changed
-            string tmp;
-            m_options.RawOptions.TryGetValue("throttle-upload", out tmp);
+            var updateThrottleSpeeds = false;
+            m_options.RawOptions.TryGetValue("throttle-upload", out var tmp);
```
warwickmm: Is it possible to use `m_options.MaxUploadPrSecond` here?  I would like to reduce the number of string references to options and use the named properties when feasible.

seantempleton: That makes sense. I'll get that done in my next commit.

Duplicati/Library/Main/Operation/Backup/BackendUploader.cs
seantempleton: I looked through a lot of files and most of them are using the m_ style. I'll plan on leaving this as-is for now. I think having a coding standard should be a separate conversation where all interested parties can help make the final decision.
If the community agrees on a different standard before this PR is merged I can always update it then.

warwickmm: Sounds good.  Thanks.

verhoek: That's fine.  I just noticed that you changed it to the m_* style actively, whence the remark. We should discuss coding styles in another place and you shouldn't change it here. I hope I can find some time soon to read the actual code of your implemented interesting feature.

warwickmm:
I finally made it through all the changes (it was easier looking at them one commit at a time) and it looks pretty reasonable to me.  Nice work.  Hopefully others will find time to look at this as well, since the number of changes isn't trivial.

It seems that users occasionally experience issues where (presumably after an interrupted or cancelled operation) their local database and remote files become out of sync.  Were you able to perform any kind of testing to see how interruptions are handled and reported?

verhoek:
> I finally made it through all the changes (it was easier looking at them one commit at a time) and it looks pretty reasonable to me. Nice work. Hopefully others will find time to look at this as well, since the number of changes isn't trivial.
> 

I started looking into it today a bit, so far it looks OK but as you said a lot of changes + I'm unfamiliar with this part of the code and well, it isn't easy for me to dissect the way it's written (the old code). I do like that some parts got moved to places where they apparently make more sense.

Duplicati/Library/Main/Operation/Backup/BackendUploader.cs
```diff
@@ -85,98 +120,331 @@ public static Task Run(Common.BackendHandler backend, Options options, Common.Da
...
+            await m_database.LogRemoteOperationAsync("put", item.RemoteFilename, JsonConvert.SerializeObject(new { Size = item.Size, Hash = item.Hash }));
+            await m_stats.SendEventAsync(BackendActionType.Put, BackendEventType.Started, item.RemoteFilename, item.Size, updateProgress: false);
+            m_progressUpdater.StartFileProgress(item.RemoteFilename, item.Size);
+
+            var begin = DateTime.Now;
+
+            if (!m_options.DisableStreamingTransfers && backend is IStreamingBackend streamingBackend)
+            {
+                using (var fs = File.OpenRead(item.LocalFilename))
+                using (var ts = new ThrottledStream(fs, m_initialUploadThrottleSpeed, 0))
```
verhoek: Did behaviour change here? Before a download limit was set, which is removed by setting the third argument to 0?
Suppose I have no upload limit, do not want my download bandwith to exceed 100kB/s, and I am uploading with 20MB/s and in order to do so, I download 500kB/s. Does this imply that I will violate the download 100kB/s limit?
Was this respected previously?

I am not saying either is correct or wrong, but if this is the case, behaviour changed. If I understand correctly, to maintain the old behaviour, you'd have to also start doing transfer speed bookkeeping in the other "download" direction in FileProgressThrottler?

seantempleton: You are correct, that would be a change in behavior.
When I was playing with the code on my own I set that to 0 and thought, why do I need to limit the download speed when I'm uploading! I kind of glossed over it after that, but I agree that needs to be put back.

seantempleton: I looked into this and I don't think this matters after all. When uploading data we of course upload it, but we don't download anything until the verify step. And that step uses different code to do the downloading.

Additionally the `ProgressReportingStream` only has one value that it tracks for progress so it cannot currently differentiate between downloads and uploads.

All _that_ said I did add in the code to handle throttling downloads but it will not currently be used so I don't know if it is worth it to push that code. What do you think?

warwickmm: I'm not very familiar with this part of the code, but it sounds like it at least deserves of a comment explaining why the download throttle value is what it is.

seantempleton: That makes sense, I'll add a comment as to why there is no download throttle value.

seantempleton:
How much weight is placed on Codacy? Is it required that every issue raised is fixed? For instance it says there are several member variables that are not used, but they definitely are. And it wants `BackendUploader` to implement `IDisposable` because of the `CancellationTokenSource`. Reading about that the recommendation is you really only need to dispose it if you are using a linked `CancellationTokenSource`.

warwickmm:
I'd take the Codacy issues as just suggestions.  There are times where they are clearly false flags, and some things are very subjective.  Sometimes, a simple refactoring raises existing issues that I would rather not touch in a particular pull request.  However, there are times where the issues raised are quite meaningful, so I would take a look at those.  Hopefully reviewers are looking at the Codacy results as well, and will point out the ones that they think are important.

Duplicati/Library/Main/Operation/Backup/BackendUploader.cs
verhoek: I meant that even if you don't actively download files (like for verification), still packets are sent to be able to upload. It's two ways, you can't have 0 bytes in one direction without affecting the other side. My guess was the current code might take care of that, but who knows. I would not push unused code.

seantempleton: Some data is downloaded when we upload in terms of TCP ACK packets for instance. Or if an uploaded packet is lost a retry will occur but that happens below the level of Duplicati so we aren't even aware of it, let alone able to measure it.

kenkendk:
Wow, large refactoring and update. You even fixed some of the backends and made them really `async` instead of just returning `Task.FromResult(true)`. I do not fully understand the updated upload process, but I do not see any reason it should not work.

For Codacy, I think @warwickmm has the right attitude. It suggests improvements, but they do not always make things better, just stricter.

The `Put` method should be called `PutAsync` to follow the .Net conventions ? Are there reasons not to use the `Async` suffix here? I do not see a need for a non-async `Put`, as the caller can just use `PutAsync(...).Wait();` to synchronize immediately.

Other than taht, I am good with merging this PR.

For future PRs we could consider:
- Should we change `IBackend` to use `async` for all methods, like `GetAsync`, `ListAsync` etc?
- Should we remove the `IStreamingBackend` and just have a `SupportsStreaming` property ?
- The `Utility.AsyncHttpRequest` should be removed. It was added to support proper timeout and cancellation to the HTTP requests, but since we have `async` methods, this is not needed anymore.

seantempleton:
> I do not fully understand the updated upload process, but I do not see any reason it should not work.

If you have any specific questions I'd be happy to answer them. This is a bit simplified but the `BackendUploader` is mostly the same functionality wise but instead of running each upload through the `RunOnMain` function, it starts a task. Then some code was shuffled in from `BackendHandler` and other code shuffled out to the `DataBlockProcessor` and `SpillCollectorProcess`.

> The `Put` method should be called `PutAsync` to follow the .Net conventions ? Are there reasons not to use the `Async` suffix here? I do not see a need for a non-async `Put`, as the caller can just use `PutAsync(...).Wait();` to synchronize immediately.

I agree, I will change it to `PutAsync`.

I like those future changes to consider. I was wondering if I could swap out using async/await where `AsyncHttpRequest` was used, but decided I was changing enough things already.

Thanks for the comments kenkendk!

warwickmm:
I ran my tests again on a variety of backends and most of them (Amazon Cloud Drive, B2, Box, Dropbox, Google Cloud Storage, Google Drive, Jottacloud, Local, Microsoft One Drive, OpenDrive WebDAV, SFTP) ran without issues.  

However, my tests fail with the Mega backend.  The initial backup returns an exit code of `1` instead of `0`.  When running the initial backup from the UI, it appears that it ran fine when in fact only a `dlist` file appears in the backend folder (perhaps we're running into issue #3673 here?).  The logs show that an "Object reference not set to an instance of an object" error was encountered at some point.  A subsequent backup run yields the "Found 1 remote files that are not recorded in local storage, please run repair" error.

These tests pass when using the most recent `master` (38dfdffb903a13a65169006b0d9444c25f92ef43), so I'm thinking that perhaps there's a small issue introduced in this branch.

seantempleton:
I have no way of testing it but the changes for the Mega backend were trivial and I think I tracked the issue down. I was passing null to their upload function instead of an instance of their `IProgress` class.

warwickmm:
My test suite is now passing.  :smile:

mapsouza:
This was great! Could it be implemented for downloads as well?

Issue #3248, opened 2018-05-30
Slow Download speed with Jottacloud in case of recovery


theincogtion:
- **Duplicati version**: 2.0.3.5
- **Operating system**: QNAP QTS (Linux)
- **Backend**: RJottacloud

## Description
Download from one file only is incredibly slow
Download one single file from Jottacloud gives you maximum 2MB/s
With Jottacloud Software it is possible to Download 6x faster because it allows 6 parellel downloads

## Steps to reproduce
Restore files from Jottacloud

## Screenshots
Duplicati performance
![grafik](https://user-images.githubusercontent.com/10852327/40731157-c0e62690-6430-11e8-9bb4-fb6700dfd180.png)

Jottacloud Software Download 6x streams

![grafik](https://user-images.githubusercontent.com/10852327/40731293-043abe92-6431-11e8-9a03-e6f85d57642d.png)

Jottacloud Software single Download
![grafik](https://user-images.githubusercontent.com/10852327/40731418-4a2f2f1e-6431-11e8-8ecf-ac9a2ba16504.png)

Looks for me if parallelity would solve the issue.

albertony:
This seems to be similar to #2026 (that one is more focused on upload, but it does also mention download), and then it is not something specific to this backend implementation but more a general issue -  and not a trivial one: 

> My conclusion: it is not an easy extension, I think it requires a complete rewrite of the BackendManager and that possibly has repercussions on the rest of Duplicati. Not sure if anyone is up for such a daunting task.

That being said, Jottacloud seem to be moving to a newer API in their own clients. Upgrading our backend to that could perhaps speed things up, although I think they use it only for uploads. Unfortunately I haven't figured out all the necessary details about how it works yet..

PR #3691, opened 2019-03-19
Implement Jottacloud multithreaded downloading support

nescafe2002:
Hello,

Please consider implementing multithreaded downloading for Jottacloud by merging this commit.

Jottacloud limits single tcp streams to 18.5 Mbps which can be dealt with by downloading larger files in parallel chunks. I've implemented parallelized downloading using a queue of tasks of a certain max size (4 by default).

Increasing threads count to 8 can result in a download speed of 80~155Mbps depending on volume size.

explorer_2019-03-19_20-53-40
explorer_2019-03-19_20-53-09


Duplicati/Library/Backend/Jottacloud/Jottacloud.cs
```diff
@@ -222,6 +254,64 @@ public IEnumerable<IFileEntry> List()
...
+        /// <returns></returns>
+        public IFileEntry Info(string remotename)
+        {
+            var doc = new System.Xml.XmlDocument();
+            try
+            {
+                // Send request and load XML response.
+                var req = CreateRequest(System.Net.WebRequestMethods.Http.Get, remotename, "", false);
+                var areq = new Utility.AsyncHttpRequest(req);
+                using (var resp = (System.Net.HttpWebResponse)areq.GetResponse())
```
warwickmm: Is this statement needed?  The return value isn't used, and the following line appears to read the response.

nescafe2002: Nice catch! It seems to work fine without the extra GetResponse() call.

nescafe2002:
Also should fix issue #3248.

Duplicati/Library/Backend/Jottacloud/Jottacloud.cs
```diff
@@ -146,6 +162,22 @@ public Jottacloud(string url, Dictionary<string, string> options)
...
+            m_threads = options.ContainsKey(JFS_THREADS) ? int.Parse(options[JFS_THREADS]) : 4;
+
+            if (!options.TryGetValue(JFS_CHUNKSIZE, out var tmp))
+            {
+                tmp = JFS_DEFAULT_CHUNKSIZE;
+            }
+
+            var chunksize = Utility.Sizeparser.ParseSize(tmp, "mb");
+
+            if (chunksize > int.MaxValue || chunksize < 1024)
```
warwickmm: I think `chunksize` is a `long` here.  What's the reasoning for comparing it to `int.MaxValue`?

Duplicati/Library/Backend/Jottacloud/Jottacloud.cs
```diff
@@ -146,6 +162,22 @@ public Jottacloud(string url, Dictionary<string, string> options)
...
+            {
+                throw new ArgumentOutOfRangeException(nameof(chunksize), string.Format("The chunk size cannot be less than {0}, nor larger than {1}", 1024, int.MaxValue));
```
warwickmm: It might be helpful to include the units (B?) in the exception message.

warwickmm: When I specify an invalid chunk size, the UI shows an error but with the message "Unwrapped TargetInvocationException".  Are you able to replicate this?

Duplicati/Library/Backend/Jottacloud/Jottacloud.cs
```diff
@@ -222,6 +253,63 @@ public IEnumerable<IFileEntry> List()
...
+                throw;
+            }
+            // Handle XML response. Since we in the constructor demand a folder below the mount point we know the root
+            // element must be a "folder", else it could also have been a "mountPoint" (which has a very similar structure).
+            // We must check for "deleted" attribute, because files/folders which has it is deleted (attribute contains the timestamp of deletion)
+            // so we treat them as non-existant here.
+            var xFile = doc.DocumentElement;
+            if (xFile.Attributes["deleted"] != null)
+            {
+                throw new FileMissingException();
```
warwickmm: It might be helpful to provide an exception message here to help in debugging.

Duplicati/Library/Backend/Jottacloud/Jottacloud.cs
```diff
@@ -49,6 +50,21 @@ public class Jottacloud : IBackend, IStreamingBackend
         private readonly System.Net.NetworkCredential m_userInfo;
         private readonly byte[] m_copybuffer = new byte[Duplicati.Library.Utility.Utility.DEFAULT_BUFFER_SIZE];
 
+        private static readonly string JFS_DEFAULT_CHUNKSIZE = "5mb";
+        private static readonly string JFS_DEFAULT_THREADS = "4";
+        private readonly int m_threads;
+        private readonly int m_chunksize;
```
warwickmm: It looks like this value is always being compared to `longs`.  Is there a reason why we declare it to be an `int` here?

Duplicati/Library/Backend/Jottacloud/Jottacloud.cs
nescafe2002: I'm not sure what is considered helpful in this case. File name, url, xml body? I see DropBox returns some json here.

nescafe2002: I've added remotename to the exception. Also de-duplicated some xml handling code.

Duplicati/Library/Backend/Jottacloud/Jottacloud.cs
nescafe2002: Agreed, but I tried to confirm to existing logic/ux (https://github.com/duplicati/duplicati/blob/master/Duplicati/Library/Main/Options.cs#L1343)

Duplicati/Library/Backend/Jottacloud/Jottacloud.cs
nescafe2002: System.IO.BinaryReader.ReadBytes(int count) only seems to accept ints, hence the limitation of max chunk size.

Duplicati/Library/Backend/Jottacloud/Jottacloud.cs
```diff
@@ -146,6 +162,22 @@ public Jottacloud(string url, Dictionary<string, string> options)
             m_url_device = JFS_ROOT + "/" + m_userInfo.UserName + "/" + m_device;
             m_url        = m_url_device + "/" + m_mountPoint + "/" + m_path;
             m_url_upload = JFS_ROOT_UPLOAD + "/" + m_userInfo.UserName + "/" + m_device + "/" + m_mountPoint + "/" + m_path; // Different hostname, else identical to m_url.
+
+            m_threads = options.ContainsKey(JFS_THREADS) ? int.Parse(options[JFS_THREADS]) : 4;
```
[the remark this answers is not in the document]
nescafe2002: Agreed

Duplicati/Library/Backend/Jottacloud/Jottacloud.cs
nescafe2002: Agreed

Duplicati/Library/Backend/Jottacloud/Jottacloud.cs
nescafe2002: Yes, it's thrown here: https://github.com/duplicati/duplicati/blob/master/Duplicati/Library/DynamicLoader/BackendLoader.cs#L101 (with inner exception {"The chunk size cannot be less than 1024, nor larger than 2147483647"}). I think this issue falls out of scope of this pull request.

nescafe2002: Using FormatSizeString now resulting in: {"The chunk size cannot be less than 1.00 KB, nor larger than 2.00 GB\r\nParameter name: chunksize"}

warwickmm: You're correct.  I have a proposed solution in pull request #3695.

Duplicati/Library/Backend/Jottacloud/Jottacloud.cs
warwickmm: I'm not sure.  Hopefully someone that's more familiar with the backend implementations can help out here.

warwickmm:
Hmmm...using revision c93e0eed4da8b1c89519e60594d0d9eb3063d2e8 I cannot perform a successful initial backup.  Was there an errant suggestion or some other change that was made?  Are you able to reproduce the following failures?
```
Backup started at 3/21/2019 9:35:02 PM
Checking remote backup ...
  Listing remote folder ...
Scanning local files ...
  9 files need to be examined (12.20 MB)
  Uploading file (3.73 KB) ...
  Uploading file (1.25 KB) ...
  Uploading file (3.56 KB) ...
Compacting remote backup ...
Checking remote backup ...
  Listing remote folder ...
Verifying remote backup ...
Remote backup verification completed
  Downloading file (1.25 KB) ...
  0 files need to be examined (0 bytes)
  Downloading file (1.25 KB) ...
  Downloading file (1.25 KB) ...
  Downloading file (1.25 KB) ...
  Downloading file (1.25 KB) ...
Failed to process file duplicati-20190322T043502Z.dlist.zip.aes => The file duplicati-20190322T043502Z.dlist.zip.aes was downloaded and had size 0 but the size was expected to be 1277
  Downloading file (3.56 KB) ...
  Downloading file (3.56 KB) ...
  Downloading file (3.56 KB) ...
  Downloading file (3.56 KB) ...
  Downloading file (3.56 KB) ...
Failed to process file duplicati-i45ec002f95c2479c8f6225bc0a9ec533.dindex.zip.aes => The file duplicati-i45ec002f95c2479c8f6225bc0a9ec533.dindex.zip.aes was downloaded and had size 0 but the size was expected to be 3645
  Downloading file (3.73 KB) ...
  Downloading file (3.73 KB) ...
  Downloading file (3.73 KB) ...
  Downloading file (3.73 KB) ...
  Downloading file (3.73 KB) ...
Failed to process file duplicati-b899c0b53e60e4267aaf1440502b1b3a3.dblock.zip.aes => The file duplicati-b899c0b53e60e4267aaf1440502b1b3a3.dblock.zip.aes was downloaded and had size 0 but the size was expected to be 3821
  Duration of backup: 00:02:25
  Remote files: 3
  Remote size: 0 bytes
  Total remote quota: 0 bytes
  Available remote quota: 0 bytes
  Files added: 6
  Files deleted: 0
  Files changed: 0
  Data uploaded: 8.54 KB
  Data downloaded: 0 bytes
```

nescafe2002:
Good catch (again)! I have deleted one line of code in deduplication in last commit. This has been restored.

Duplicati/Library/Backend/Jottacloud/Jottacloud.cs
kenkendk: The way I have handled this in other backends, like GCloud etc, is to simply keep a cache of items in the backend. This is required as each item has a GUID of sorts, which is not related to the filename, so even though the API requests something by name, we need to translate it into a GUID. To avoid repeated (recursive) lookups, I opted to cache them.

If you need something similar, you can do the same, but we can also consider it there is a better way. Not immediately sure what that way would be though, as there are case where the filename is all we know (i.e. we did not call `List()` on before calling `Get()`).

kenkendk:
@warwickmm I do not have a setup for testing JottaCloud, if this PR works for you, please merge it.

warwickmm:
Looks good to me.  :+1:

PR #3697, opened 2019-03-22
Ensure that failed put operations throw exception

warwickmm:
This fixes an issue where failed Put operations would not result in an exception.  The `FlushBackend` method [uses Task.WhenAny](https://github.com/duplicati/duplicati/blob/57b28182cb714c46226645b47fa80cb3d23f1f79/Duplicati/Library/Main/Operation/BackupHandler.cs#L360) to await the first completion of the uploader task or the flush request.  Since the uploader was setting the status of the flush request in the [finally block](https://github.com/duplicati/duplicati/blob/57b28182cb714c46226645b47fa80cb3d23f1f79/Duplicati/Library/Main/Operation/Backup/BackendUploader.cs#L180), the flush request would complete first when an exception was thrown, causing the uploader's exception to be unobserved.  To resolve this,  we check that the uploader's status is not faulted before setting the flush request status.

In order for the database transactions to be rolled back, we instantiate the `IDbTransaction` instance earlier so that the `Rollback` method is invoked in the [finally](https://github.com/duplicati/duplicati/blob/57b28182cb714c46226645b47fa80cb3d23f1f79/Duplicati/Library/Main/Operation/BackupHandler.cs#L558) block.  Otherwise, a backup version would be recorded in the database even if all the Puts failed.

This addresses issue #3673.

### DIFF
--- a/Duplicati/Library/Main/Operation/Backup/BackendUploader.cs
+++ b/Duplicati/Library/Main/Operation/Backup/BackendUploader.cs
@@ -39,6 +39,12 @@ namespace Duplicati.Library.Main.Operation.Backup
     {
         public Task<long> LastWriteSizeAsync { get { return m_tcs.Task; } }
         private readonly TaskCompletionSource<long> m_tcs = new TaskCompletionSource<long>();
+        
+        public void TrySetCanceled()
+        {
+            m_tcs.TrySetCanceled();
+        }
+        
         public void SetFlushed(long size)
         {
             m_tcs.TrySetResult(size);
@@ -171,6 +177,7 @@ namespace Duplicati.Library.Main.Operation.Backup
                                 Task finishedTask = await Task.WhenAny(workers.Select(w => w.Task)).ConfigureAwait(false);
                                 if (finishedTask.IsFaulted)
                                 {
+                                    flush.TrySetCanceled();
                                     ExceptionDispatchInfo.Capture(finishedTask.Exception).Throw();
                                 }
 
--- a/Duplicati/Library/Main/Operation/Backup/BackendUploader.cs
+++ b/Duplicati/Library/Main/Operation/Backup/BackendUploader.cs
@@ -166,11 +166,12 @@ namespace Duplicati.Library.Main.Operation.Backup
                         }
                         else if (req is FlushRequest flush)
                         {
+                            Task finishedTask = null;
                             try
                             {
                                 while (workers.Any())
                                 {
-                                    var finishedTask = await Task.WhenAny(workers.Select(w => w.Task)).ConfigureAwait(false);
+                                    finishedTask = await Task.WhenAny(workers.Select(w => w.Task)).ConfigureAwait(false);
                                     if (finishedTask.IsFaulted)
                                         ExceptionDispatchInfo.Capture(finishedTask.Exception).Throw();
                                     workers.RemoveAll(w => w.Task == finishedTask);
@@ -179,7 +180,10 @@ namespace Duplicati.Library.Main.Operation.Backup
                             }
                             finally
                             {
-                                flush.SetFlushed(lastSize);
+                                if (finishedTask != null && !finishedTask.IsFaulted)
+                                {
+                                    flush.SetFlushed(lastSize);
+                                }
                             }
                             break;
                         }
--- a/Duplicati/Library/Main/Operation/Backup/BackendUploader.cs
+++ b/Duplicati/Library/Main/Operation/Backup/BackendUploader.cs
@@ -166,25 +166,19 @@ namespace Duplicati.Library.Main.Operation.Backup
                         }
                         else if (req is FlushRequest flush)
                         {
-                            Task finishedTask = null;
-                            try
+                            while (workers.Any())
                             {
-                                while (workers.Any())
+                                Task finishedTask = await Task.WhenAny(workers.Select(w => w.Task)).ConfigureAwait(false);
+                                if (finishedTask.IsFaulted)
                                 {
-                                    finishedTask = await Task.WhenAny(workers.Select(w => w.Task)).ConfigureAwait(false);
-                                    if (finishedTask.IsFaulted)
-                                        ExceptionDispatchInfo.Capture(finishedTask.Exception).Throw();
-                                    workers.RemoveAll(w => w.Task == finishedTask);
+                                    ExceptionDispatchInfo.Capture(finishedTask.Exception).Throw();
                                 }
-                                uploadsInProgress = 0;
+
+                                workers.RemoveAll(w => w.Task == finishedTask);
+                                flush.SetFlushed(lastSize);
                             }
-                            finally
-                            {
-                                if (finishedTask != null && !finishedTask.IsFaulted)
-                                {
-                                    flush.SetFlushed(lastSize);
-                                }
-                            }
+                            
+                            uploadsInProgress = 0;
                             break;
                         }
 
--- a/Duplicati/Library/Main/Operation/Backup/BackendUploader.cs
+++ b/Duplicati/Library/Main/Operation/Backup/BackendUploader.cs
@@ -182,9 +182,9 @@ namespace Duplicati.Library.Main.Operation.Backup
                                 }
 
                                 workers.RemoveAll(w => w.Task == finishedTask);
-                                flush.SetFlushed(lastSize);
                             }
                             
+                            flush.SetFlushed(lastSize);
                             uploadsInProgress = 0;
                             break;
                         }
--- a/Duplicati/Library/Main/Operation/BackupHandler.cs
+++ b/Duplicati/Library/Main/Operation/BackupHandler.cs
@@ -393,10 +393,10 @@ namespace Duplicati.Library.Main.Operation
 
                 Task parallelScanner = null;
                 Task uploaderTask = null;
+                using (var db = new Backup.BackupDatabase(m_database, m_options))
                 try
                 {
                     // Setup runners and instances here
-                    using(var db = new Backup.BackupDatabase(m_database, m_options))
                     using(var backendManager = new BackendManager(m_backendurl, m_options, m_result.BackendWriter, m_database))
                     using(var filesetvolume = new FilesetVolumeWriter(m_options, m_database.OperationTimestamp))
                     using(var stats = new Backup.BackupStatsCollector(m_result))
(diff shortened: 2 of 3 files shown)


warwickmm:
I'm not that familiar with this part of the code, but these changes do seem to resolve the issue without many side effects.  To verify this, I did the following:
  
1. Remove write permission to local destination folder.
2. Observe error when running initial backup.
3. Observe in the restore list that no versions are available.
4. Add write permission to destination folder.
5. Run backup, which completes successfully without warnings or errors.
6. Modify source file and run another backup, which completes successfully.
7. Verify that files from both versions can be restored.

Duplicati/Library/Main/Operation/BackupHandler.cs
```diff
@@ -503,9 +506,6 @@ private async Task RunAsync(string[] sources, Library.Utility.IFilter filter)
                         // Make sure we have the database up-to-date
                         await db.CommitTransactionAsync("CommitAfterUpload", false);
 
-                        // TODO: Remove this later
-                        m_transaction = m_database.BeginTransaction();
```
warwickmm: It's not clear to me why the `TODO` comment about removing this exists, so I left it in above.

warwickmm:
Looks like I missed some test failures.  I'll have to investigate some more.

Duplicati/Library/Main/Operation/Backup/BackendUploader.cs
```diff
@@ -179,7 +180,10 @@ public Task Run()
                             }
                             finally
                             {
-                                flush.SetFlushed(lastSize);
+                                if (finishedTask != null && !finishedTask.IsFaulted)
+                                {
+                                    flush.SetFlushed(lastSize);
```
seantempleton: I know you're still working on some changes but I wrote up some thoughts.

In my opinion an even easier way of doing this is to change the `FlushBackend` method in the BackupHandler and just remove the `WhenAny` call on the uploader and flush request. await the uploader, that will catch any exceptions and wait for all uploads to be finished, and then await the flush request to get the value.

In some ways I could see both removing the `WhenAny` call and adding this code here. If we encounter an exception uploading a file then we truly haven't flushed the uploads successfully so it makes sense not to set the last volume size. Also in the exception scenario the uploader task is guaranteed to finish first in the `WhenAny` and then we'll just await the uploader anyway, throwing the exception due to the await call.

In the happy path where everything uploads fine the flush task will be guaranteed to finish first in the `WhenAny` and we'll never await the uploader task. It seems safer to me to always await the uploader task to be guaranteed to see any exceptions.

Also, this line here makes me slightly nervous in that if it's never called and someone only awaits the flush request, then they will hang indefinitely. As things are written now I don't think that is possible though. It's kind of a weird situation. There are two tasks for the caller to await, the first one can throw exceptions which they should know about and the second task has the value they want, but the second task can actually be completed before the first task.

Sorry, I can ramble sometimes as I let my thoughts roll out. I hope I was at least partially helpful! :)

warwickmm: Thanks @seantempleton.  I like your idea of simplifying the `FlushBackend` method.  I'll look into this some more.

The strange this is, a lot of the failed unit tests are passing for me locally.

warwickmm:
I simplified the `FlushBackend` method, and got rid of the `try-finally` in the `BackendUploader`, which didn't seem necessary.  To avoid potential hanging due to awaiting the `FlushRequest.LastWriteSizeAync`, I exposed a `FlushRequest.SetCanceled` method to cancel the task when the uploader task faults.

The unit tests are passing for me locally, so we'll have to see if they continue to fail in Travis.

warwickmm:
OK, I am not sure what's going on here.  The tests pass for me locally, but are failing in both Travis and AppVeyor.  I suspect it has to do with the change to when the database transaction is started, but without that an erroneous version is recorded in the database when all the Put operations fail.

I admit that I am not very familiar with how we handle the database transactions, so any help would be greatly appreciated!

warwickmm:
I'm still baffled by the inconsistent test results.  My NUnit console runner, CLR version, and .NET Framework versions are the same as in Travis.  However, I am using a newer kernel (5.0.2 vs. 4.4.0), while Travis is using a new version of mono (5.18.1 vs. 5.18.0).  I would hope that a different version of mono wouldn't make that big of a difference, but I'll attempt to upgrade to see if that explains it.

Mine:
```
NUnit Console Runner 3.5.0 
Copyright (C) 2016 Charlie Poole

Runtime Environment
   OS Version: Linux 5.0.2.1 
  CLR Version: 4.0.30319.42000

Test Files
    ./Duplicati/UnitTest/bin/Release/Duplicati.UnitTest.dll

Test Filters
    Where: cat==Border

...
<test output here>
...

Run Settings
    DisposeRunners: True
    WorkDirectory: /home/warwickmm/documents/code/duplicati
    NumberOfTestWorkers: 1
    ImageRuntimeVersion: 4.0.30319
    ImageTargetFrameworkName: .NETFramework,Version=v4.5
    ImageRequiresX86: False
    ImageRequiresDefaultAppDomainAssemblyResolver: False

Test Run Summary
  Overall result: Passed
  Test Count: 16, Passed: 16, Failed: 0, Inconclusive: 0, Skipped: 0
  Start time: 2019-03-24 16:39:18Z
    End time: 2019-03-24 16:53:16Z
    Duration: 837.790 seconds
```

Travis:
```
NUnit Console Runner 3.5.0 
Copyright (C) 2016 Charlie Poole
Runtime Environment
   OS Version: Linux 4.4.0.101 
  CLR Version: 4.0.30319.42000
Test Files
    ./Duplicati/UnitTest/bin/Release/Duplicati.UnitTest.dll
Test Filters
    Where: cat==Border

...
<test output here>
...

Run Settings
    DisposeRunners: True
    WorkDirectory: /home/travis/build/warwickmm/duplicati
    NumberOfTestWorkers: 1
    ImageRuntimeVersion: 4.0.30319
    ImageTargetFrameworkName: .NETFramework,Version=v4.5
    ImageRequiresX86: False
    ImageRequiresDefaultAppDomainAssemblyResolver: False
Test Run Summary
  Overall result: Failed
  Test Count: 16, Passed: 7, Failed: 9, Inconclusive: 0, Skipped: 0
    Failed Tests - Failures: 0, Errors: 9, Invalid: 0
  Start time: 2019-03-23 18:43:28Z
    End time: 2019-03-23 18:54:06Z
    Duration: 637.705 seconds
```

Duplicati/Library/Main/Operation/Backup/BackendUploader.cs
```diff
@@ -166,21 +172,20 @@ public Task Run()
...
+                                    flush.SetCanceled();
+                                    ExceptionDispatchInfo.Capture(finishedTask.Exception).Throw();
                                 }
-                                uploadsInProgress = 0;
-                            }
-                            finally
-                            {
+
+                                workers.RemoveAll(w => w.Task == finishedTask);
                                 flush.SetFlushed(lastSize);
```
seantempleton: I think this should be outside of the while loop. Without tracking the task that corresponds to the `lastSize` variable we can't know when it is done.

seantempleton:
The unit tests are passing on my Windows machine as well. (Except for a few. They are not finding the test data for some reason, but I'm pretty sure it's not related.)

I'm trying to figure out where the transaction might be that the unit tests are saying we are trying to commit. I haven't found it yet.

Duplicati/Library/Main/Operation/Backup/BackendUploader.cs
warwickmm: Nice catch, thanks.  Fixed in f52217e992c4738e93d10cc46a2aee6babba1c6f.

warwickmm:
> The unit tests are passing on my Windows machine as well. (Except for a few. They are not finding the test data for some reason, but I'm pretty sure it's not related.)

Yes, there are a few tests that depend on some external data.  You can see the `wget` calls in the [build.sh](https://github.com/duplicati/duplicati/blob/master/build.sh) script.

My tests pass whether I use my IDE's test runner, or the above [build.sh] script, which is what Travis is using.  It's strange that it is failing on both Travis and AppVeyor, while it appears to be passing locally for us on Windows and Linux.

According to the logs, it looks like the commit is occurring when an object is being disposed, but I still can't figure why I am unable to reproduce the failure.

seantempleton:
Well, I guess I'll have to take my statement back. I just hit it in the debugger. I ran all the BorderTests in the debugger at once and it stopped in the Run10mb test.

Duplicati/Library/Main/Operation/BackupHandler.cs
kenkendk: I think the reason for the `TODO` was that I working on rewriting all the operations to use channels and a single (shared) database transaction. But for time constraint reasons, I had to stop that and just implement the backup/upload part with channels and leave the other operations as they were. They require a transaction to be active to avoid writing partial results, which is why the transaction is started.

seantempleton:
I don't know the absolute fine details but I think I've narrowed down the problem. The transaction for `m_database` is started before the BackendUploader is finished. (The `uploaderTask` is passed to `FlushBackend()` several lines later.) In `DoPut()` in the BackendUploader the last line commits the transaction to the database. `m_database` from the BackupHandler is ultimately passed to the BackendUploader through the `db` variable, an instance of BackupDatabase.

So previously the BackendUploader would be completely finished before starting a new transaction in `m_database`, but now the BackupHandler is trying to begin a transaction while the BackendUploader is trying to commit transactions using the same variable / connection.

warwickmm:
Thanks @seantempleton.  Here's what I've gathered so far:

If I revert the change in this pull request to when `BeginTransaction` is called, the failed Puts do result in an error, but an errant backup version appears in the database.  It shows up in the restore list, but if you actually try to restore something you will encounter an error.  If I put a `try` around the [await FlushBackend](https://github.com/duplicati/duplicati/blob/master/Duplicati/Library/Main/Operation/BackupHandler.cs#L501) and call `await db.RollbackTransactionAsync()` in the `catch`, it does not help.  The errant backup version still exists.

However, with the (likely incorrect) call to `BeginTransaction` in this pull request, the above does not occur.  No errant backup version exists (_something_ appears to be rolled back) and subsequent backups (after enabling write permissions) complete without issue.  The big problem here is the failing tests that I am unable to reproduce locally.  This really concerns me, as this is making me losing confidence in my local unit test results.  :disappointed:

warwickmm:
I'm going to close this as something is clearly not right here and this should not be merged.  This issue will require some more investigation.

PR #3707, opened 2019-04-02
Mask password field for local file backends

drwtsn32x:


kenkendk:
I guess we should do that for all backends.

drwtsn32x:
This is the only backend I've used where I noticed it wasn't masked. But yes I agree - it should be this way everywhere.

drwtsn32x:
I checked other html files and found s3 and openstack also could be updated. I'll submit another pull request....

Issue #3633, opened 2019-01-27
Cache Files filter group doesn't exclude Chrome files on Mac/Linux


warwickmm:
The "Cache Files" filter group currently excludes the following pattern for Chrome cache files
```
*/Google/Chrome/*cache*
```
However, this pattern only matches for Chrome on Windows.  See the [documentation](https://chromium.googlesource.com/chromium/src/+/HEAD/docs/user_data_dir.md#User-Cache-Directory) for the cache location on Mac OS X and Linux.

Forum discussion [here](https://forum.duplicati.com/t/cache-folders-not-being-ignored/6141).

rohan013:
Hello,

I would like to pick up this issue.

flafleur:
Chrome cache files should already filtered on Mac OS X by Cache filter group.  Current filter is "*/Library/Caches/" and documentation refers to "[user cache dir] ~/Library/Caches/Google/Chrome/Default".

I have added ".cache" for Linux to exclude ".cache" when Cache filter group is selected.

warwickmm:
Thanks @flafleur.  Looking at the documentation, it seems that the cache path on Mac OS X can sometimes be the same as the profile dir `~/Library/Application Support/Google/Chrome/Default`, which is not matched by the filter?
> 1. If Library/Application Support is an ancestor of the profile dir, the user cache dir is Library/Caches plus the relative path from Application Support to the profile dir.
> 2. Otherwise, the user cache dir is the same as the profile dir.

I did forget to look at the system-specific filters.  It appears that on Linux, the [existing filters](https://github.com/duplicati/duplicati/blob/master/Duplicati/Library/Utility/FilterGroups.cs#L557) are simply inadequte.

warwickmm:
#3713 adds the `.cache` folder to the Linux excludes.  We should consider first checking the `$XDG_CACHE_HOME` environment variable.  If it is not set or empty, we can then resort to excluding `.cache`.

> $XDG_CACHE_HOME defines the base directory relative to which user specific non-essential data files should be stored. If $XDG_CACHE_HOME is either not set or empty, a default equal to $HOME/.cache should be used.

https://specifications.freedesktop.org/basedir-spec/basedir-spec-latest.html

PR #3708, opened 2019-04-03
Cache Files filter group doesn't exclude Chrome files on Mac/Linux #3633

flafleur:
Chrome cache files should already filtered on Mac OS X by Cache filter group. Current filter is "*/Library/Caches/" and documentation refers to "[user cache dir] ~/Library/Caches/Google/Chrome/Default".

I have added ".cache" for Linux to exclude ".cache" when Cache filter group is selected.

### DIFF
--- a/Duplicati/Library/Utility/FilterGroups.cs
+++ b/Duplicati/Library/Utility/FilterGroups.cs
@@ -556,8 +556,10 @@ namespace Duplicati.Library.Utility
             }
             if (group.HasFlag(FilterGroup.CacheFiles))
             {
+                yield return FilterGroups.CreateWildcardFilter(@"*/.cache/");                
                 yield return FilterGroups.CreateWildcardFilter(@"*/.config/google-chrome/Default/Cookies");
-                yield return FilterGroups.CreateWildcardFilter(@"*/.config/google-chrome/Default/Cookies-journal");
+                yield return FilterGroups.CreateWildcardFilter(@"*/.config/google-chrome/Default/Cookies-journal");                
+
             }
             if (group.HasFlag(FilterGroup.Applications))
             {


warwickmm:
Fixes #3633.

warwickmm:
Looking at the [documentation](https://chromium.googlesource.com/chromium/src/+/HEAD/docs/user_data_dir.md#User-Cache-Directory), is there a case on Mac OS X where the cache dir is the same as the profile dir (`~/Library/Application Support/Google/Chrome/Default`)?


> 2. Otherwise, the user cache dir is the same as the profile dir.

Duplicati/Library/Utility/FilterGroups.cs
```diff
@@ -556,8 +556,10 @@ private static IEnumerable<string> CreateLinuxFilters(FilterGroup group)
             }
             if (group.HasFlag(FilterGroup.CacheFiles))
             {
+                yield return FilterGroups.CreateWildcardFilter(@"*/.cache/");                
                 yield return FilterGroups.CreateWildcardFilter(@"*/.config/google-chrome/Default/Cookies");
-                yield return FilterGroups.CreateWildcardFilter(@"*/.config/google-chrome/Default/Cookies-journal");
+                yield return FilterGroups.CreateWildcardFilter(@"*/.config/google-chrome/Default/Cookies-journal");                
+
```
warwickmm: Can we remove this blank line?

flafleur: Yes you can remove it, or I have to?

flafleur:
> Looking at the [documentation](https://chromium.googlesource.com/chromium/src/+/HEAD/docs/user_data_dir.md#User-Cache-Directory), is there a case on Mac OS X where the cache dir is the same as the profile dir (`~/Library/Application Support/Google/Chrome/Default`)?
> 
> > 1. Otherwise, the user cache dir is the same as the profile dir.

You're right, but this folder doesn't only include cache.  How to know if we are only exclude cache from this path "~/Library/Application Support/Google/Chrome/Default"?

Duplicati/Library/Utility/FilterGroups.cs
warwickmm: It would be nice if you can remove it.  When possible, we would prefer to clean up the changes before merging them so that we don't have to remember to do so later.

warwickmm:
I don't have a Mac, so I don't know what the exact folder layout looks like.  In this case, does the cache appear in a subdirectory of the profile directory?

Duplicati/Library/Utility/FilterGroups.cs
flafleur: Sorry create a new Pull request.  I'm new with git.  New PR #3713

flafleur:
I don't have Mac too, but the issue [here](https://forum.duplicati.com/t/cache-folders-not-being-ignored/6141)  seems to be for Linux.

warwickmm:
I believe the original poster mentions that the filter didn't work on either their Mac or Linux machine, which led me to think that the default filters weren't sufficient for both environments.

PR #3709, opened 2019-04-03
Added french canadian localization 

flafleur:
Added french canadian localization

Pull all localizations from transifex and recompile all languages.


kenkendk:
Looks good to me.

Issue #3644, opened 2019-02-04
Error / DB corruption: Unexpected difference in fileset version X: found Y entries, but expected Z


SamSirry:
- [x] I have searched open and closed issues for duplicates.

----------------------------------------

## Environment info

- **Duplicati version**: 2.0.4.12 (Canary)
- **Operating system**: Windows 7 Pro x64 (8 CPU cores, 8 GB RAM)
- **Backend**: SFTP (SSH)

_General Settings:_

    --allow-sleep=true
    --usn-policy=Auto
    --tempdir=E:\Temp\SYSTEM\Duplicati
    --retry-delay=20s
    --number-of-retries=10
    --dont-compress-restore-paths=true
    --concurrency-block-hashers=6
    --concurrency-max-threads=8
    --concurrency-compressors=3

_Job-specific settings:_

    --ssh-fingerprint=ssh-rsa 2048 {xxx}
    Remote volume size: 10MB
    Retention: Keep all backups

## Description

The error "Unexpected difference in fileset version X: found Y entries, but expected Z" gets generated in certain cases when the backup operation gets interrupted before completion.
It seems to be a corruption in the local job database, and fixing it requires either deleting the backup set by running a manual Duplicati command line (tried it only once and didn't manage to delete successfully), or a full rebuild of the database.

## Steps to reproduce

1. Create a backup job with the destination set to SSH/SFTP.
2. Start the job, and let it upload a couple of volumes.
3. Attempt to stop it by requesting to "Stop after upload". It is a known bug that this will not actually stop the job, and it will go on.
4. Now stop the job by selecting the option "Stop now".
5. Attempt to start the job again. The error `Unexpected difference in fileset version X: found Y entries, but expected Z` will be generated.

- **Actual result**:
The backup job terminates apparently normally in step 4, but in step 5 an error message is displayed the the backup job terminates before processing any data (to be backed up) files.
- **Expected result**:
In step 5, starting the backup job should proceed normally, just like any job that gets started manually, first checking the remote data, scanning the local files for changes, uploading new blocks etc...

## Screenshots

_none_

## Debug log

_Error from the internal log:_

    Duplicati.Library.Interface.UserInformationException: Unexpected difference in fileset version 1: 2019-01-30 16:19:11 (database id: 1), found 76 entries, but expected 417 at 
[...]

Issue #1088, opened 2014-08-13
"pause" doesn't stop the upload


agrajaghh:
When I click the pause button in the GUI or at the tray icon, duplicati still continues to upload. 

I'm running Duplicati 2.0.0.57- 2.0_preview_2014-08-12 on Windows 7 and Windows 8.1.

I guess it just doesn't start new jobs, but it would be good to be able to pause running backups as well. Sometimes you need the bandwith for other things (sending a big email, gaming, ...)

rocifier:
I like that solution personally agrajaghh. I'm going to take a stab at doing that by adding a new event which if queried returns the percentage the current file is uploaded. That way the gui can make a choice about what logic to prompt the user with and finally commit to pause() if the user desires, which will always cut off the current upload. Doesn't seem too trivial though given the inheritance hierarchy :/

kenkendk:
The progress event contains the current upload progress.
The fields BackendFileSize and BackendFileProgress should contain the relevant information.

This is reported in the event 'server-progress-updated' and in the function 'update_progress_and_schedules'.

Patches are always welcome :)

rocifier:
Cool. That's kind of the easy part of that done then. The hard part, I'm finding, is how to make pause() stop the current upload.

kenkendk:
That needs to be handled in the the callback for the upload:
https://github.com/duplicati/duplicati/blob/master/Duplicati/Library/Main/BackendManager.cs#L591

rocifier:
I'm confused.. what does TaskControlRendevouz() do? The code inside it seems to indicate it pauses?

kenkendk:
The task control is the method I use to signal changes to the running task.
In the current setup, it appears that pause() will actually also pause an active upload.
It has been a while since I looked at that part, so there might be something else going on.

The idea is that the active thread occasionally calls TaskControlRendevouz(). If a request for stop, pause or similar is active, the call will block or throw an exception.

A dry-run is a run where no files are updated (local and remote stays the same), like a simulation run.

rocifier:
I think I see the issue that is happening here. The project code everywhere is pausing correctly. However, (at least for the S3 backend), we have the line:
    PutObjectResponse objectAddResponse = m_client.PutObject(objectAddRequest);
[...]

PR #3712, opened 2019-04-03
Fix pause and resume, and check for cancel while uploading

### LINKED ISSUE #3565: job pausing/stopping doesn't work in 2.0.4.5_beta_2018-11-28
- **Duplicati version**: 2.0.4.5_beta_2018-11-28
- **Operating system**: OS X High Sierra 10.13.6

## Description
Aborting a backup job doesn't work; this may be specific to the scanning stage, or perhaps due to server and tray/web GUI running from separate plists due to High Sierra permissions problems?

The job in question was run with the following options (passphrase and folder exclusion statements removed):

--full-remote-verification=true
--use-block-cache=true
--thread-priority=idle
--full-block-verification=true
--backup-name=homedir
--dbpath=/var/root/.config/Duplicati/86847676818883747767.sqlite
--encryption-module=aes
--compression-module=zip
--dblock-size=100MB
--retention-policy=1W:1D,4W:1W,12M:1M
--aes-set-threadlevel=4
--exclude-files-attributes=temporary
--disable-module=console-password-input
--dry-run=true

## Steps to reproduce

1. Use separate plists for server and tray icon
2. Start a backup
3. During the scanning stage, click on (X) in web UI, select immediate
4. Click pause icon in web UI

Stopping the job requires force-quitting the mono process.

On a related note (I will probably open a separate bug for this), my whole reason for stopping the job was due to use-block-cache not working; I'm seeing a massive amount of write activity. **In dry run mode with use-block-cache, duplicati essentially re-writes the entire contents of my SSD. This is incredibly bad for an SSD!)**

```
## Debug log
Backup started at 12/26/2018 3:32:23 PM
Checking remote backup ...
  Listing remote folder ...
Scanning local files ...
  74690 files need to be examined (5.53 GB) (still counting)
  80923 files need to be examined (5.85 GB) (still counting)
  86682 files need to be examined (5.98 GB) (still counting)
  96166 files need to be examined (8.52 GB) (still counting)
  107703 files need to be examined (11.26 GB) (still counting)
  120194 files need to be examined (36.15 GB) (still counting)
  130064 files need to be examined (45.56 GB) (still counting)
  140289 files need to be examined (56.31 GB) (still counting)
  150345 files need to be examined (57.19 GB) (still counting)
  161269 files need to be examined (58.02 GB) (still counting)
  168300 files need to be examined (57.83 GB) (still counting)
  176195 files need to be examined (58.15 GB) (still counting)
  184556 files need to be examined (134.90 GB) (still counting)
  194669 files need to be examined (134.83 GB) (still counting)
  202100 files need to be examined (134.65 GB) (still counting)
  212829 files need to be examined (134.63 GB) (still counting)
  221705 files need to be examined (134.48 GB) (still counting)

System.Threading.ThreadAbortException: Thread was being aborted.
  at Duplicati.Library.Main.Controller.RunAction[T] (T result, System.String[]& paths, Duplicati.Library.Utility.IFilter& filter, System.Action`1[T] method) [0x00239] in <c6c6871f516b48f59d88f9d731c3ea4d>:0 
  at Duplicati.Library.Main.Controller.Backup (System.String[] inputsources, Duplicati.Library.Utility.IFilter filter) [0x00068] in <c6c6871f516b48f59d88f9d731c3ea4d>:0 
  at Duplicati.CommandLine.Commands.Backup (System.IO.TextWriter outwriter, System.Action`1[T] setup, System.Collections.Generic.List`1[T] args, System.Collections.Generic.Dictionary`2[TKey,TValue] options, Duplicati.Library.Utility.IFilter filter) [0x00119] in <04206d56f2084515b87543fcb90a7e00>:0 
  at (wrapper delegate-invoke) System.Func`6[System.IO.TextWriter,System.Action`1[Duplicati.Library.Main.Controller],System.Collections.Generic.List`1[System.String],System.Collections.Generic.Dictionary`2[System.String,System.String],Duplicati.Library.Utility.IFilter,System.Int32].invoke_TResult_T1_T2_T3_T4_T5(System.IO.TextWriter,System.Action`1<Duplicati.Library.Main.Controller>,System.Collections.Generic.List`1<string>,System.Collections.Generic.Dictionary`2<string, string>,Duplicati.Library.Utility.IFilter)
  at Duplicati.CommandLine.Program.ParseCommandLine (System.IO.TextWriter outwriter, System.Action`1[T] setup, System.Boolean& verboseErrors, System.String[] args) [0x00313] in <04206d56f2084515b87543fcb90a7e00>:0 
  at Duplicati.CommandLine.Program.RunCommandLine (System.IO.TextWriter outwriter, System.IO.TextWriter errwriter, System.Action`1[T] setup, System.String[] args) [0x00002] in <04206d56f2084515b87543fcb90a7e00>:0 
System.Threading.ThreadAbortException: Thread was being aborted.
  at Duplicati.CommandLine.Program.RunCommandLine (System.IO.TextWriter outwriter, System.IO.TextWriter errwriter, System.Action`1[T] setup, System.String[] args) [0x000bb] in <04206d56f2084515b87543fcb90a7e00>:0 
  at Duplicati.Server.WebServer.RESTMethods.CommandLine+<>c__DisplayClass4_0.<POST>b__0 (Duplicati.Library.Main.IMessageSink sink) [0x00030] in <fe28905ee30b422e8d475f1cfdb85515>:0 

```

seantempleton:
The BasicResults class was not calling Pause or Resume on
m_taskController. m_taskController is the instance of ITaskReader that
is passed to all of the worker classes that use it to check if they
should pause or stop.

Changed the BackendUploader to check TransferProgressAsync instead of
ProgressAsync, as the transfer function is meant for the worker that is
uploading files.

Added two new phases, Paused and Paused_WaitForUpload. These are used
to inform the user about the state of the operation when a pause has been requested.
The current transfer rate will display while waiting for uploads to finish.

The BackendUploader will now periodically timeout waiting for input (files ready
to upload) and will check if a pause is requested. It will then wait for all
current uploads to finish and signal to the IOperationProgressUpdater
that the phase should be changed to Paused.

This also helps when using "Stop Now" as the BackendUploader will check if a cancel is requested while uploads are currently in progress, and then cancel the uploads rather than waiting for them to finish.

Fixes #3565

### DIFF
--- a/Duplicati/Library/Main/Operation/Backup/BackendUploader.cs
+++ b/Duplicati/Library/Main/Operation/Backup/BackendUploader.cs
@@ -147,11 +147,21 @@ namespace Duplicati.Library.Main.Operation.Backup
                         if (m_operationUpdater.CurrentPhase == OperationPhase.Paused_WaitForUpload)
                         {
                             await WaitForUploadsToFinish(flush: null).ConfigureAwait(false);
+                            uploadsInProgress = 0;
                             m_operationUpdater.UpdatePhase(OperationPhase.Paused);
                             m_progressUpdater.Pause();
                             if (!await m_taskReader.TransferProgressAsync.ConfigureAwait(false))
+                            {
+                                (request as FlushRequest)?.SetFlushed(-1);
                                 break;
+                            }
                             m_progressUpdater.Resume();
+                        }
+                        else if (!await m_taskReader.TransferProgressAsync.ConfigureAwait(false))
+                        {
+                            m_cancelTokenSource.Cancel();
+                            (request as FlushRequest)?.SetFlushed(-1);
+                            break;
                         }
 
                         if (request == null)
@@ -160,14 +170,13 @@ namespace Duplicati.Library.Main.Operation.Backup
                         await HandleRequest(request).ConfigureAwait(false);
                         uploadsInProgress++;
                         if (request is FlushRequest)
-                        {
-                            uploadsInProgress = 0;
                             break;
-                        }
 
                         if (uploadsInProgress >= m_maxConcurrentUploads)
                         {
-                            await Task.WhenAny(m_workers.Select(w => w.Task)).ConfigureAwait(false);
+                            if (!await AnyUploadCompleted().ConfigureAwait(false))
+                                continue;
+
                             uploadsInProgress--;
 
                             var failedUploads = m_workers.Where(w => w.Task.IsFaulted).Select(w => GetInnerMostException(w.Task.Exception)).ToList();
@@ -210,6 +219,21 @@ namespace Duplicati.Library.Main.Operation.Backup
                     m_workers.ForEach(w => w.Dispose());
                 }
             });
+        }
+
+        /// <returns><c>true</c> if an upload task completed or <c>false</c> if the operation is requested to pause or stop.</returns>
+        private async Task<bool> AnyUploadCompleted()
+        {
+            var twoSeconds = TimeSpan.FromSeconds(2);
+            while (true)
+            {
+                await Task.Delay(twoSeconds).ConfigureAwait(false);
+                if (m_operationUpdater.CurrentPhase == OperationPhase.Paused_WaitForUpload || !await m_taskReader.TransferProgressAsync.ConfigureAwait(false))
+                    return false;
+
+                if (m_workers.Any(w => w.Task.IsCompleted))
+                    return true;
+            }
         }
 
         private static Exception GetInnerMostException(Exception ex)
@@ -263,7 +287,7 @@ namespace Duplicati.Library.Main.Operation.Backup
             {
                 try
                 {
-                    WaitForUploadsToFinish(flush);
+                    await WaitForUploadsToFinish(flush).ConfigureAwait(false);
                 }
                 finally
                 {
@@ -306,12 +330,12 @@ namespace Duplicati.Library.Main.Operation.Backup
             {
                 if (item.IsRetry)
                     await RenameFileAfterErrorAsync(item).ConfigureAwait(false);
-                await DoPut(item, worker.Backend, cancelToken).ConfigureAwait(false);
+                return await DoPut(item, worker.Backend, cancelToken).ConfigureAwait(false);
             },
             item, worker, cancelToken).ConfigureAwait(false);
         }
 
-        private async Task<bool> DoWithRetry(Func<Task> method, FileEntryItem item, Worker worker, CancellationToken cancelToken)
+        private async Task<bool> DoWithRetry(Func<Task<bool>> method, FileEntryItem item, Worker worker, CancellationToken cancelToken)
         {
             item.IsRetry = false;
             var retryCount = 0;
@@ -328,8 +352,7 @@ namespace Duplicati.Library.Main.Operation.Backup
                 {
                     if (worker.Backend == null)
                         worker.Backend = m_backendFactory();
-                    await method().ConfigureAwait(false);
-                    return true;
+                    return await method().ConfigureAwait(false);
                 }
                 catch (Exception ex)
                 {
@@ -402,10 +425,10 @@ namespace Duplicati.Library.Main.Operation.Backup
             item.RemoteFilename = newname;
         }
 
-        private async Task DoPut(FileEntryItem item, IBackend backend, CancellationToken cancelToken)
+        private async Task<bool> DoPut(FileEntryItem item, IBackend backend, CancellationToken cancelToken)
         {
             if (cancelToken.IsCancellationRequested)
-                return;
+                return false;
 
             if (item.TrackedInDb)
                 await m_database.UpdateRemoteVolumeAsync(item.RemoteFilename, RemoteVolumeState.Uploading, item.Size, item.Hash);
@@ -414,7 +437,7 @@ namespace Duplicati.Library.Main.Operation.Backup
             {
                 Logging.Log.WriteDryrunMessage(LOGTAG, "WouldUploadVolume", "Would upload volume: {0}, size: {1}", item.RemoteFilename, Library.Utility.Utility.FormatSizeString(new FileInfo(item.LocalFilename).Length));
                 item.DeleteLocalFile();
-                return;
+                return true;
             }
 
             await m_database.LogRemoteOperationAsync("put", item.RemoteFilename, JsonConvert.SerializeObject(new { Size = item.Size, Hash = item.Hash }));
@@ -434,26 +457,33 @@ namespace Duplicati.Library.Main.Operation.Backup
             else
                 await backend.PutAsync(item.RemoteFilename, item.LocalFilename, cancelToken).ConfigureAwait(false);
 
+            var success = !cancelToken.IsCancellationRequested;
+
             var duration = DateTime.Now - begin;
             m_progressUpdater.EndFileProgress(item.RemoteFilename);
-            Logging.Log.WriteProfilingMessage(LOGTAG, "UploadSpeed", "Uploaded {0} in {1}, {2}/s", Library.Utility.Utility.FormatSizeString(item.Size), duration, Library.Utility.Utility.FormatSizeString((long)(item.Size / duration.TotalSeconds)));
 
-            if (item.TrackedInDb)
-                await m_database.UpdateRemoteVolumeAsync(item.RemoteFilename, RemoteVolumeState.Uploaded, item.Size, item.Hash);
-
-            await m_stats.SendEventAsync(BackendActionType.Put, BackendEventType.Completed, item.RemoteFilename, item.Size);
-
-            if (m_options.ListVerifyUploads)
+            if (success)
             {
-                var f = backend.List().FirstOrDefault(n => n.Name.Equals(item.RemoteFilename, StringComparison.OrdinalIgnoreCase));
-                if (f == null)
-                    throw new Exception(string.Format("List verify failed, file was not found after upload: {0}", item.RemoteFilename));
-                else if (f.Size != item.Size && f.Size >= 0)
-                    throw new Exception(string.Format("List verify failed for file: {0}, size was {1} but expected to be {2}", f.Name, f.Size, item.Size));
+                Logging.Log.WriteProfilingMessage(LOGTAG, "UploadSpeed", "Uploaded {0} in {1}, {2}/s", Library.Utility.Utility.FormatSizeString(item.Size), duration, Library.Utility.Utility.FormatSizeString((long)(item.Size / duration.TotalSeconds)));
+
+                if (item.TrackedInDb)
+                    await m_database.UpdateRemoteVolumeAsync(item.RemoteFilename, RemoteVolumeState.Uploaded, item.Size, item.Hash);
+
+                await m_stats.SendEventAsync(BackendActionType.Put, BackendEventType.Completed, item.RemoteFilename, item.Size);
+
+                if (m_options.ListVerifyUploads)
+                {
+                    var f = backend.List().FirstOrDefault(n => n.Name.Equals(item.RemoteFilename, StringComparison.OrdinalIgnoreCase));
+                    if (f == null)
+                        throw new Exception(string.Format("List verify failed, file was not found after upload: {0}", item.RemoteFilename));
+                    else if (f.Size != item.Size && f.Size >= 0)
+                        throw new Exception(string.Format("List verify failed for file: {0}, size was {1} but expected to be {2}", f.Name, f.Size, item.Size));
+                }
             }
 
             item.DeleteLocalFile();
             await m_database.CommitTransactionAsync("CommitAfterUpload");
+            return success;
         }
 
         private void HandleProgress(ThrottledStream stream, long progress, string path)
--- a/Duplicati/Library/Main/Operation/Backup/BackendUploader.cs
+++ b/Duplicati/Library/Main/Operation/Backup/BackendUploader.cs
@@ -253,6 +253,10 @@ namespace Duplicati.Library.Main.Operation.Backup
             {
                 // Timeout exception is expected and handled
             }
+            catch (TaskCanceledException)
+            {
+                // Either the read will be tried again or uploads have been canceled
+            }
 
             return null;
         }
--- a/Duplicati/Library/Main/Operation/Backup/BackendUploader.cs
+++ b/Duplicati/Library/Main/Operation/Backup/BackendUploader.cs
@@ -39,12 +39,12 @@ namespace Duplicati.Library.Main.Operation.Backup
     {
         public Task<long> LastWriteSizeAsync { get { return m_tcs.Task; } }
         private readonly TaskCompletionSource<long> m_tcs = new TaskCompletionSource<long>();
-        
+
         public void SetFlushed(long size)
         {
             m_tcs.TrySetResult(size);
         }
-        
+
         public void TrySetCanceled()
         {
             m_tcs.TrySetCanceled();
@@ -100,22 +100,25 @@ namespace Duplicati.Library.Main.Operation.Backup
         private CancellationTokenSource m_cancelTokenSource;
         private readonly DatabaseCommon m_database;
         private long m_initialUploadThrottleSpeed;
+        private long m_lastSize;
         private int m_lastThrottleCheckTime;
         private long m_lastUploadThrottleSpeed;
         private int m_maxConcurrentUploads;
         private readonly Options m_options;
         private readonly FileProgressThrottler m_progressUpdater;
         private readonly StatsCollector m_stats;
+        private readonly IOperationProgressUpdaterAndReporter m_operationUpdater;
         private readonly ITaskReader m_taskReader;
+        private List<Worker> m_workers = new List<Worker>();
 
-
-        public BackendUploader(Func<IBackend> backendFactory, Options options, DatabaseCommon database, ITaskReader taskReader, StatsCollector stats)
+        public BackendUploader(Func<IBackend> backendFactory, Options options, DatabaseCommon database, ITaskReader taskReader, StatsCollector stats, IOperationProgressUpdaterAndReporter operationUpdater)
         {
             m_backendFactory = backendFactory;
             m_options = options;
             m_taskReader = taskReader;
             m_stats = stats;
             m_database = database;
+            m_operationUpdater = operationUpdater;
             m_progressUpdater = new FileProgressThrottler(stats, options.MaxUploadPrSecond);
         }
 
@@ -128,77 +131,46 @@ namespace Duplicati.Library.Main.Operation.Backup
 
             async self =>
             {
-                var workers = new List<Worker>();
                 m_maxConcurrentUploads = m_options.AsynchronousConcurrentUploadLimit <= 0 ? int.MaxValue : m_options.AsynchronousConcurrentUploadLimit;
                 m_initialUploadThrottleSpeed = m_options.AsynchronousConcurrentUploadLimit <= 0 ? int.MaxValue : m_options.MaxUploadPrSecond / m_maxConcurrentUploads;
-                var lastSize = -1L;
                 var uploadsInProgress = 0;
+                var waitForInputTimeout = TimeSpan.FromSeconds(2);
                 m_cancelTokenSource = new CancellationTokenSource();
                 m_progressUpdater.Run(m_cancelTokenSource.Token);
 
                 try
                 {
-                    while (!await self.Input.IsRetiredAsync && await m_taskReader.ProgressAsync)
+                    while (!await self.Input.IsRetiredAsync)
                     {
-                        var req = await self.Input.ReadAsync();
+                        var request = await GetRequest(self.Input, waitForInputTimeout);
 
-                        if (!await m_taskReader.ProgressAsync)
-                            break;
-
-                        var worker = workers.FirstOrDefault(w => w.Task.IsCompleted && !w.Task.IsFaulted);
-                        if (worker == null)
+                        if (m_operationUpdater.CurrentPhase == OperationPhase.Paused_WaitForUpload)
                         {
-                            worker = new Worker(m_backendFactory());
-                            workers.Add(worker);
+                            await WaitForUploadsToFinish(flush: null).ConfigureAwait(false);
+                            m_operationUpdater.UpdatePhase(OperationPhase.Paused);
+                            m_progressUpdater.Pause();
+                            if (!await m_taskReader.TransferProgressAsync.ConfigureAwait(false))
+                                break;
+                            m_progressUpdater.Resume();
                         }
 
-                        if (req is VolumeUploadRequest volumeUpload)
-                        {
-                            if (volumeUpload.IndexVolume == null)
-                                worker.Task = Task.Run(() => UploadFileAsync(volumeUpload.BlockEntry, worker, m_cancelTokenSource.Token));
-                            else
-                                worker.Task = Task.Run(() => UploadBlockAndIndexAsync(volumeUpload, worker, m_cancelTokenSource.Token));
+                        if (request == null)
+                            continue;
 
-                            lastSize = volumeUpload.BlockVolume.SourceSize;
-                            uploadsInProgress++;
-                        }
-                        else if (req is FilesetUploadRequest filesetUpload)
+                        await HandleRequest(request).ConfigureAwait(false);
+                        uploadsInProgress++;
+                        if (request is FlushRequest)
                         {
-                            worker.Task = Task.Run(() => UploadVolumeWriter(filesetUpload.Fileset, worker, m_cancelTokenSource.Token));
-                            uploadsInProgress++;
-                        }
-                        else if (req is IndexVolumeUploadRequest indexUpload)
-                        {
-                            worker.Task = Task.Run(() => UploadVolumeWriter(indexUpload.IndexVolume, worker, m_cancelTokenSource.Token));
-                            uploadsInProgress++;
-                        }
-                        else if (req is FlushRequest flush)
-                        {
-                            while (workers.Any())
-                            {
-                                Task finishedTask = await Task.WhenAny(workers.Select(w => w.Task)).ConfigureAwait(false);
-                                if (finishedTask.IsFaulted)
-                                {
-                                    flush.TrySetCanceled();
-                                    ExceptionDispatchInfo.Capture(finishedTask.Exception).Throw();
-                                }
-
-                                Worker finishedWorker = workers.Single(w => w.Task == finishedTask);
-                                workers.Remove(finishedWorker);
-                                finishedWorker.Dispose();
-                            }
-                            
-                            flush.SetFlushed(lastSize);
                             uploadsInProgress = 0;
                             break;
                         }
 
                         if (uploadsInProgress >= m_maxConcurrentUploads)
                         {
-                            await Task.WhenAny(workers.Select(w => w.Task)).ConfigureAwait(false);
+                            await Task.WhenAny(m_workers.Select(w => w.Task)).ConfigureAwait(false);
                             uploadsInProgress--;
 
-                            var failedUploads = workers.Where(w => w.Task.IsFaulted).Select(w => GetInnerMostException(w.Task.Exception)).ToList();
+                            var failedUploads = m_workers.Where(w => w.Task.IsFaulted).Select(w => GetInnerMostException(w.Task.Exception)).ToList();
                             if (failedUploads.Any())
                             {
                                 if (failedUploads.Count == 1)
@@ -209,12 +181,12 @@ namespace Duplicati.Library.Main.Operation.Backup
                         }
                     }
                 }
-                catch (Exception ex) when (!ex.IsRetiredException())
+                catch (Exception)
                 {
                     m_cancelTokenSource.Cancel();
                     try
                     {
-                        await Task.WhenAll(workers.Select(w => w.Task));
+                        await Task.WhenAll(m_workers.Select(w => w.Task));
                     }
                     catch
                     {
@@ -222,7 +194,7 @@ namespace Duplicati.Library.Main.Operation.Backup
                     }
                     finally
                     {
-                        workers.ForEach(w => w.Dispose());
+                        m_workers.ForEach(w => w.Dispose());
                     }
                     throw;
                 }
@@ -230,12 +202,12 @@ namespace Duplicati.Library.Main.Operation.Backup
                 try
                 {
                     m_stats.SetBlocking(true);
-                    await Task.WhenAll(workers.Select(w => w.Task));
+                    await Task.WhenAll(m_workers.Select(w => w.Task));
                 }
                 finally
                 {
                     m_stats.SetBlocking(false);
-                    workers.ForEach(w => w.Dispose());
+                    m_workers.ForEach(w => w.Dispose());
                 }
             });
         }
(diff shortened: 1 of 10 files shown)


warwickmm:
Nice!  I believe this might address issue #1088 as well?

warwickmm:
When testing this, I did encounter a strange case when pausing did appear to pause the backup, but the progress bar kept updating with "X files (Y bytes) to go at Z MB/s", where X would slowly decrease.  This could lead the user to think that the backup was not actually paused.  Are you able to reproduce this?
1. Start a backup.
2. Wait until the progress bar says "X files (Y bytes) to go at Z MB/s".
3. Pause the backup.
4. Watch the progress bar and observe the "Z MB/s" slowly decreasing.

seantempleton:
My understanding was the intention of #1088 is to pause or cancel the actual uploads, rather than pausing the backup job and letting the uploads that are currently in progress finish.

I can indeed reproduce the slowly decreasing transfer rate. The thread that is doing the transfer rate computing continues to run when everything is paused. It could be paused when the user presses pause, but since the uploads actually continue until they finish then it wouldn't really be telling the truth.

I can see a way to make it reflect the transfer rate and then show nothing after all uploads are done by checking the state of `m_transfer` in `TaskControl`. (It's not a public value right now and I'm not sure the intention is that it should be.) If paused wait for active uploads to finish and then show nothing.
The `TransferProgressAsync` function causes the calling thread to wait so that can't be used to stop updating the rate since it would need to keep calculating while in the paused state.

Although I kind of want to tackle the todo comment in `BackendProgressUpdater` about using a dynamic window for calculating the transfer rate. Maybe have a 10 second rolling window. Once paused it would still continue to show the rate while the uploads are going, then once they are finished it would slowly wind down to 0 over 10 seconds. That is probably outside of the scope of this change though. Thoughts?

warwickmm:
I'm not very familiar with how the progress is updated, but what if we add some elements to `OperationPhase`, say `Pause_WaitForUpload` and `Paused`, and invoke `this.m_operationProgressUpdater.UpdatePhase` in the `ResultClasses`?  We can then inform the user that we are about to pause (but an in progress upload needs to be finished), and then later indicate that we are paused (after the upload is finished).

I think this would require modifying `Duplicati/Server/webroot/ngax/scripts/services/ServerStatus.js` to handle these new states as well.

Would that even work?  Or a dumb idea?

seantempleton:
I like your proposed idea in general. The problem is that calling `ProgressAsync` or `TransferProgressAsync` causes the calling thread to wait if we are in a paused state. So we have no way of knowing ahead of time if we would need to update the `OperationPhase`. In addition the BackendUploader or someone would have to know when all current uploads are finished and then update the UI again, but that can't happen if a thread is waiting for the `ManualResetEvent` in `BasicResults` to be signaled.

I went ahead and tried out adding a rolling window to `BackendProgressUpdater` for calculating the transfer rate per second. It was actually a lot easier than I thought it was going to be. When the uploads finish while paused the displayed transfer rate slowly drops over 20 seconds and then disappears.

If the thinking is that we ought to solve the issue in a more direct way, similar to your proposal, I'll save that for a future PR.

warwickmm:
Thanks for looking into this.  I think as long as the UI makes it clear the the backup is paused, that's an improvement.

Perhaps we should focus on the issue with pausing and stopping here, and handle the notification separately.

warwickmm:
It looks like ignoring the CoCoL retired exception causes the failed Put exceptions in issue #3673 to be unobserved.  Are we sure that this is a safe thing to do?  Are we possibly ignoring other important exceptions by doing this?

kenkendk:
> I like your proposed idea in general. The problem is that calling ProgressAsync or TransferProgressAsync causes the calling thread to wait if we are in a paused state. So we have no way of knowing ahead of time if we would need to update the OperationPhase

Is a solution that the call itself updates the state? Otherwise there could be an `AreWePaused` property, although that should be used carefully, as there could be a race between reading the property and calling the method.

> When testing this, I did encounter a strange case when pausing did appear to pause the backup, but the progress bar kept updating with "X files (Y bytes) to go at Z MB/s", where X would slowly decrease. This could lead the user to think that the backup was not actually paused.

The intention was to avoid pausing the transfers, as that would cause timeouts and retries. So a pause would stop processing files (effectively pause the CPU usage) but let the active uploads continue. Looking at the code, it seems that invoking Pause will now also pause transfers?

> It looks like ignoring the CoCoL retired exception causes some of the failed Put exceptions in issue #3673 to be unobserved. Are we sure that this is a safe thing to do? Are we possibly ignoring other important exceptions by doing this?

@warwickmm : The `RetiredException` is thrown whenever the channel is closed (or disposed). Once this happens, the channel is considered "retired" and always throws the exceptions.

As a design in CoCoL, this looks nice, but when using it in a real scenario, like Duplicati, it has some drawbacks. Sometimes the closing of a channel is intentional and part of a normal flow, which is when we want to ignore it. Other times, the channel is closed because something else breaks, and in this case we do still want to ignore it, but there is no mechanism to report the original error.

I think I want to change CoCoL to allow it to report an error close and a normal close, like `NormalRetiredException` and `ErrorRetiredException`, where the latter has a `ReasonException` property (names are not final, but the general idea is the same).

seantempleton:
@warwickmm I think I misread your idea with a new operation phase at first. Rereading it, I like it for getting us to the `Paused_WaitForUpload` state. Then the problem is getting to the `Paused` state. We could have the `BackendUploader` take `IOperationProgressUpdaterAndReporter` in its constructor. That has an event for when the phase changes as well as the ability to update the phase. The `BackendUploader` could keep note of phase changes and when we hit the `Paused_WaitForUpload` state then after each upload finishes, that thread can check if it was the last one, and if so then update to `Paused`.

It will probably be a little clunky due to needing to handle multi-threading issues in the `BackendUploader` and it also might be a little awkward because `Paused_WaitForUpload` would be triggered in `BasicResults` and `Paused` would be triggered in the `BackendUploader`. But I think that should work.

> Looking at the code, it seems that invoking Pause will now also pause transfers?

@kenkendk This will not pause the uploads that are currently in progress but will prevent any new ones. The `BackendUploader` calls `TransferProgressAsync` at the beginning of it's while loop, but all of the uploads on other threads will continue on as normal. I could rename that boolean from `alsoTransfers` to `alsoNewTransfers` to make that clearer.

In regards to the `RetiredException` I'll put that back to how it was. I agree it wouldn't be a good idea to swallow those exceptions in cases where it legitimately needs to be reported. I like your idea @kenkendk of having unique exceptions for each scenario.

seantempleton:
I changed this up a little bit from the initial plan. I added the CurrentPhase property to `IOperationProgress` so the `BackendUploader` could check what phase we are in. It also now uses a timeout when waiting for input, and catches the exception from the timeout since it is expected, and no longer calls `TransferProgressAsync` regularly so the thread is not blocked when a pause happens.

If the phase changes to Pause_WaitForUpload it will wait for current uploads, change the phase to paused, and then call `TransferProgressAsync` to wait for when it should resume.

The StateController on the frontend will show the current transfer rate while waiting for uploads to finish and then disappear when the phase changes to Paused.

warwickmm:
This is great.  I think the updated progress is very useful.

I did come across a case where stopping (via Stop Now) the backup did not actually stop the uploads.  The UI indicated the usual error (the thread was aborted), but the uploads continued.  Are you able to replicate this?

Duplicati/Library/Main/Operation/Backup/BackendUploader.cs
```diff
@@ -235,6 +209,59 @@ private static Exception GetInnerMostException(Exception ex)
...
+            }
+            else if (request is IndexVolumeUploadRequest indexUpload)
+            {
+                worker.Task = Task.Run(() => UploadVolumeWriter(indexUpload.IndexVolume, worker, m_cancelTokenSource.Token));
+            }
+            else if (request is FlushRequest flush)
+            {
+                try
+                {
+                    WaitForUploadsToFinish(flush);
```
warwickmm: Does this need to be awaited?

seantempleton: Good catch. Fixed.

seantempleton:
I was able to reproduce not stopping when pressing Stop Now. I need to start doing a better job of testing after my initial changes. That worked for me on my initial commit but I broke it in a later commit. Thanks for testing and catching problems warwickmm.

I was also having problems with the transfer rate displaying when waiting for uploads to finish during a pause. I am confident that was working earlier so I think that is another case of not testing after making changes. I moved the if statement where that work was being done. Would you mind testing that part so I know I'm not crazy?

warwickmm:
I can see the transfer rate when waiting for uploads to finish before pausing.

seantempleton:
Not sure if it is clear or not but this is ready for review. Since the last push all the testing I have thrown at it has worked for me.

warwickmm:
Thanks @seantempleton.  I tested this again and unfortunately I still see uploads continuing after requesting "Stop Now".  The UI appears as if the operation has stopped (after a "The thread was aborted" error), but the uploads (to an SFTP backend) continue.  The backup source is a little over 200 MB, the stop was requested when about 15 MB of each of the 4 `dblock` files had been uploaded, but the uploads continue until each `dblock` file is 50 MB.  Are you able to replicate this?

seantempleton:
Arg, seems I'm not done yet after all. I'll see if I can reproduce that.

seantempleton:
I was testing the FTP backend and was getting discouraged because it seemed to be working properly and then I reread your message and realized you said SFTP! Turns out I somehow didn't actually implement cancellation in the SFTP backend so I'll fix that.

warwickmm:
Ah, that makes sense.  I should have clarified that I was using SFTP earlier.

Duplicati/Library/Main/Operation/Backup/BackendUploader.cs
```diff
@@ -126,75 +129,51 @@ public Task Run()
...
-                                {
-                                    flush.TrySetCanceled();
-                                    ExceptionDispatchInfo.Capture(finishedTask.Exception).Throw();
-                                }
-
-                                workers.RemoveAll(w => w.Task == finishedTask);
-                            }
-                            
-                            flush.SetFlushed(lastSize);
                             uploadsInProgress = 0;
```
warwickmm: Is this assignment necessary?  The `uploadsInProgress` variable looks to be unused in this code path.

warwickmm:
I switched to a Google Drive backend and pausing/stopping appears to work without any issues.  Is the plan to fix the SFTP backend here, or in a separate pull request?

seantempleton:
I'll fix that in this pull request.

warwickmm:
Thanks @seantempleton.  I was able to stop the backup successfully with the SFTP backend.

However, after stopping/restarting a backup several times, I did encounter a few cases of
```
Unexpected difference in fileset version 1: 4/29/2019 7:18:16 PM (database id: 1), found 5 entries, but expected 11
```
```
2019-04-29 19:43:04 -07 - [Warning-Duplicati.Library.Main.Operation.FilelistProcessor-MissingRemoteHash]: remote file duplicati-b0a447be5bf4d43898a98ee7531d10eb6.dblock.zip.aes is listed as Uploaded with size 13030122 but should be 52428925, please verify the sha256 hash
"07ZE5Q9/gRI0i6N/xlcYXTSdxvspmApTlW/nHAzb5LQ="
```
```
2019-04-29 19:45:12 -07 - [Error-Duplicati.Library.Main.Operation.TestHandler-FailedToProcessFile]: Failed to process file duplicati-bc57dfc389e9d477a9ea9e485f5339950.dblock.zip.aes
```
```
2019-04-29 19:46:55 -07 - [Error-Duplicati.Library.Main.Operation.RepairHandler-RemoteFileVerificationError]: Failed to perform verification for file: duplicati-b0a447be5bf4d43898a98ee7531d10eb6.dblock.zip.aes, please run verify; message: File length is invalid
```
as well as others when backing up, repairing, etc.

This might not be due to changes in this pull request, as it appears that others encounter similar issues when their backups get interrupted (see issue #3644 for example).  However, I'm wondering if fixing pausing/stopping will actually increase the occurrence of this annoying database issue, which is difficult to recover from.  

I'd like to get other people's thoughts on this.  It is really nice to have working pausing/stopping.  However, if this comes at the expense of more users encountering unrecoverable database issues, then perhaps we should fix the database issues first?

seantempleton:
I'm torn on this. There are complaints that pausing and stopping don't work, but then yes if the functionality is there but then users start encountering issues, that won't make anyone happy either. I think pausing should be okay in terms of database corruption and other errors as it actually allows processes to finish and start again gracefully.

Did you encounter any of those errors when stopping after upload or just stopping immediately?

On the other hand, I assume it is going to take a very long time to fix up the code to allow it to stop and abort without causing errors.

warwickmm:
I don't think I encountered any errors when using pause/stop after upload.  

I'm not sure what the best course of action is.  Would it be unreasonable to implement these fixes, but _disable_ the "Stop Now" option until the database issues are resolved (this could take a long time, like you mention)?  One case I can think of where a user would prefer "Stop Now" over "Stop after upload" is if the volume size is large and bandwidth issues are a concern.

kenkendk:
I agree that the database corruption issues are not related to the stopping functionality, but an error that will also appear with many kinds of hard-kill scenarios.

But I think this one is somehow caused by shutdown logic not working correctly:
```
2019-04-29 19:43:04 -07 - [Warning-Duplicati.Library.Main.Operation.FilelistProcessor-MissingRemoteHash]: remote file duplicati-b0a447be5bf4d43898a98ee7531d10eb6.dblock.zip.aes is listed as Uploaded with size 13030122 but should be 52428925, please verify the sha256 hash
"07ZE5Q9/gRI0i6N/xlcYXTSdxvspmApTlW/nHAzb5LQ="
```

For this to happen, the volume is added to the and put in the `uploading` state and transfer starts. The backup is then stopped during upload, but for some reason the database record is updated, marking the file as `uploaded` before stopping. 

Once the file has reached the `uploaded` state, Duplicati assumes the upload has completed without error, and this gives the error message. But since the file was not written fully, we get the error message.

The `uploading` state is set here:
https://github.com/duplicati/duplicati/blob/57b28182cb714c46226645b47fa80cb3d23f1f79/Duplicati/Library/Main/Operation/Backup/BackendUploader.cs#L361

And the `uploaded` state is set here:
https://github.com/duplicati/duplicati/blob/57b28182cb714c46226645b47fa80cb3d23f1f79/Duplicati/Library/Main/Operation/Backup/BackendUploader.cs#L392

If we hard-kill the process we should not see this particular error, because line 392 is not executed, leaving the database with only the `uploading` state, and Duplicati will delete the partial file on next run. My guess is that the `Stop` command does not produce an exception from line [L382](https://github.com/duplicati/duplicati/blob/57b28182cb714c46226645b47fa80cb3d23f1f79/Duplicati/Library/Main/Operation/Backup/BackendUploader.cs#L382) and/or [L385](https://github.com/duplicati/duplicati/blob/57b28182cb714c46226645b47fa80cb3d23f1f79/Duplicati/Library/Main/Operation/Backup/BackendUploader.cs#L385) causing it to continue and incorrectly mark the file as being fully uploaded.

seantempleton:
I would not be opposed to specifically disabling the Stop Now functionality and leaving the others intact. That would at least give users some control.

Thanks for the tip @kenkendk. I will look and see if that is indeed what is happening for that specific error.

ts678:
Anybody know if the transfer rate computing is related to the throttling? I was kind of wondering how throttling works under parallel uploads, and my initial tests worry me in terms of current effectiveness.

The backstory here is that I think parallel uploads worsen timeouts (already an issue) due to link filling. Throttling became a suspect when I couldn't get a portable test case built using artificial rate throttles.

That's a follow-up to "A task was canceled" and big corruption from [#3756](https://github.com/duplicati/duplicati/issues/3756#issuecomment-487568353) which doesn't need parallel uploads but is maybe easier to hit with them. Let's not get off-topic, but rate/time wisdom would help.

seantempleton:
Yes, they are related in that they both are computed in the same class, `FileProgressThrottler`. The throttling is pretty dumb and could probably use a better approach. If the transfer rate is too high it loops through the streams and reduces the rate for the first stream by either the amount over the limit or by a percent of the difference (with the percent calculated by the number of streams). If the rate still needs to be reduced then it moves on to the next stream, etc.

The rate calculation shown to the user is done a bit differently at the moment as the `FileProgressThrottler` calculates the rate for all current uploads and calls the `BackendProgressUpdater` and it does some calculations too. I hope to have a change for that in the not too distant future.

seantempleton:
I wasn't able to reproduce the error you were seeing @warwickmm after clicking stop now. It was always throwing an exception for me so it would skip the line where it marks the file as uploaded. I think I see how it could happen so I added a check for it.

I just now remembered that we want to disable stop now for the time being. I'll add another commit for that.

warwickmm:
Thanks @seantempleton.  One thing I noticed is if I:
1. Run a backup and wait for uploads to begin.
2. Click Pause.
3. I can then click Stop, and the menu still has the Stop Now option:

image

If I then click Stop Now, the backup is immediately interrupted.  When testing this, I at one point ran into the Unexpected difference in fileset error, but this did not happen every time.  Perhaps it depends on exactly when the process is terminated.

Are you still able to see this Stop Now option when clicking Stop during a paused backup?

Other than that, this is working pretty well for me.  It would be nice to get a few more eyes and testing on this.

seantempleton:
Yes, I can see the Stop now button when paused. I had assumed that dialog would only be shown when the backup was running. I still have a lot to learn about how everything works.

I can remove it, basically changing the dialog to say the same thing as when the backup is running. I'll have to make some more changes as there's some state somewhere that's not quite right when you click pause and then click stop after upload (same thing as stop after the current file). It stops but the stop button remains visible like a backup is in progress.

I did encounter the unexpected difference in fileset error when starting a backup, immediately hitting stop after upload, it stopped immediately, then I hit run, then after a couple seconds hit stop after upload again. Then when I would start it would give the unexpected different error. There must be a bug somewhere before files begin uploading in regards to stopping.

I'm not sure where we want go with this change. It seems like there are a bunch of little bugs in regards to database operations that need to be fixed to make for a truly error free experience when stopping and pausing. I have a feeling that it will take a while to track them down.

warwickmm:
Yes, I'm not what the best way forward is.  It's not clear if the database issues are due to the same bug, or if there are several bugs that exhibit the same symptoms.  Again, I do not think that they are due to the changes made here, but I'm still a bit wary of possibly making these issues more commonplace.

Perhaps someone else might have an opinion on what the best approach should be here.

Duplicati/Library/Main/Operation/Backup/BackendUploader.cs
```diff
@@ -205,12 +181,12 @@ public Task Run()
                         }
                     }
                 }
-                catch (Exception ex) when (!ex.IsRetiredException())
+                catch (Exception ex)
```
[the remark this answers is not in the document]
kenkendk: Yes, the exception is rethrown so we do not care about the instance here.

kenkendk:
Following up on my own comments:
> I think I want to change CoCoL to allow it to report an error close and a normal close, like NormalRetiredException and ErrorRetiredException, where the latter has a ReasonException property (names are not final, but the general idea is the same).

I tried this, but it turns out to be a bad idea. If the channel closes, that means that no more work should be done and the process should should shut down without errors(*). If there was indeed an error that cause the chain-reaction shutdown of channels, that `Task` will now be marked as failed, where all other `Task`s will be non-failed, allowing us to pinpoint the exact cause of the error.

(*) no process should buffer data, if this data collection would result in data loss.

seantempleton:
> I tried this, but it turns out to be a bad idea.

That's unfortunate @kenkendk. It would really be nice to know whether the exception needs to be reported to the user or not.

On another note, do you think I should abandon this PR until other fixes have been made? Rather than leaving it open for a long time...

warwickmm:
We've since addressed one "Unexpected difference in fileset version" error (pull request #3801), and a few issues with the `ThrottledStream` class (pull requests #3802 and #3811).  When I get a chance, I'll run some more tests on this branch.  Thanks for your patience.

In the meantime, can you merge in the upstream `master` to resolve the indicated potential merge conflicts?

ts678:
Regarding the probably-multiple-causes-possible "Unexpected difference", I wonder whether [2.0.4.16 and on silently drop uploads getting TaskCanceledException from HTTP timeout #3772](https://github.com/duplicati/duplicati/issues/3772) comments or its pull request [Throw TimeoutException upon HTTP timeout #3778](https://github.com/duplicati/duplicati/pull/3778) might have some relevance here in terms of cause here?

warwickmm:
I locally resolved the merge conflicts and performed some more testing.   Here's what I experienced with the SFTP backend:
* If I wait until files are actually being uploaded (determined by watching my network bandwidth), starting and stopping the backup worked without issues.  I did this approximately 50 times, both during an initial backup as well as a subsequent backup after modifying a large file.  :smiley: 
* However, if I stop the backup _before_ the file uploads begin (e.g., during an initial "Verifying backend data" step), I can encounter a "Unexpected difference in fileset version 0: 7/13/2019 10:03:42 AM (database id: 1), found 4 entries, but expected 12" error.  This occurs regardless if it's an initial backup or a later backup.

seantempleton:
Wow @warwickmm, going above and beyond trying that 50 times!
Sorry, it is ingrained in me to do a rebase to resolve conflicts and that's what I did without thinking. If you would like I can go back and merge master in so you can easily see how the conflicts were resolved.

I'll see if I can reproduce what you were talking about with stopping the backup before it actually starts uploading.

warwickmm:
Rebase is fine.  :+1:

seantempleton:
I setup a local Minio server and was able to reproduce the unexpected difference in fileset error. I'll see if I can figure out what is going on.

My random thought is maybe the first backup is stopping before all of the source files are inserted into the database and then on the 2nd backup it says hey there are $Total files (since it actually finished locating all of the files this time) but only X were in the database.

ts678:
> I'll see if I can figure out what is going on.

"Unexpected difference" is checked at the start of a backup, so one needs to go backwards to see how it arose. You might read ["Unexpected difference in fileset" test case and code clue #3800](https://github.com/duplicati/duplicati/issues/3800) for more on that, including detailed technical notes on that test case, and screen shots of the DB as error arose. I used this with --run-script-before, and either ignored or fixed the "cannot find the file" while DB history trail filled:

~~~
SET DB="C:\ProgramData\Duplicati\duplicati-2.0.4.22_canary_2019-06-30\RUN\BSUYJQFBGS"
MOVE /Y %DB%.4.sqlite %DB%.5.sqlite
MOVE /Y %DB%.3.sqlite %DB%.4.sqlite
MOVE /Y %DB%.2.sqlite %DB%.3.sqlite
MOVE /Y %DB%.1.sqlite %DB%.2.sqlite
COPY /Y %DB%.sqlite %DB%.1.sqlite
EXIT 0
~~~

Here's the SQL that gets run. Sometimes cut-down versions help pinpoint where in the DB the issue lies:

~~~
-- From 2.0.4.5 beta Duplicati/Library/Main/Database/LocalDatabase.cs
-- Expanding expandedCmd to study "Unexplained difference in fileset"

SELECT COUNT(*) FROM (SELECT DISTINCT "Path" FROM (
-- [ LIST_FILESETS
SELECT
    "L"."Path", 
    "L"."Lastmodified", 
    "L"."Filelength", 
    "L"."Filehash", 
    "L"."Metahash", 
    "L"."Metalength",
    "L"."BlocklistHash", 
    "L"."FirstBlockHash",
    "L"."FirstBlockSize",
    "L"."FirstMetaBlockHash",
    "L"."FirstMetaBlockSize",
    "M"."Hash" AS "MetaBlocklistHash"
FROM
    (
    SELECT 
        "J"."Path", 
        "J"."Lastmodified", 
        "J"."Filelength", 
        "J"."Filehash", 
        "J"."Metahash", 
        "J"."Metalength",
        "K"."Hash" AS "BlocklistHash", 
        "J"."FirstBlockHash",
        "J"."FirstBlockSize",
        "J"."FirstMetaBlockHash",
        "J"."FirstMetaBlockSize",
        "J"."MetablocksetID"
    FROM 
        (
        SELECT 
	        "A"."Path" AS "Path", 
	        "D"."Lastmodified" AS "Lastmodified", 
	        "B"."Length" AS "Filelength", 
	        "B"."FullHash" AS "Filehash", 
	        "E"."FullHash" AS "Metahash", 
	        "E"."Length" AS "Metalength",
	        "A"."BlocksetID" AS "BlocksetID",
	        "F"."Hash" AS "FirstBlockHash",
	        "F"."Size" AS "FirstBlockSize",
	        "H"."Hash" AS "FirstMetaBlockHash",
	        "H"."Size" AS "FirstMetaBlockSize",
	        "C"."BlocksetID" AS "MetablocksetID"
        FROM 
	        "File" A	
        LEFT JOIN "Blockset" B
          ON "A"."BlocksetID" = "B"."ID" 
        LEFT JOIN "Metadataset" C  
          ON "A"."MetadataID" = "C"."ID"
        LEFT JOIN "FilesetEntry" D
          ON "A"."ID" = "D"."FileID"
        LEFT JOIN "Blockset" E
          ON "E"."ID" = "C"."BlocksetID"
        LEFT JOIN "BlocksetEntry" G
          ON "B"."ID" = "G"."BlocksetID"
        LEFT JOIN "Block" F 
          ON "G"."BlockID" = "F"."ID"  
        LEFT JOIN "BlocksetEntry" I
          ON "E"."ID" = "I"."BlocksetID"
        LEFT JOIN "Block" H 
          ON "I"."BlockID" = "H"."ID"
        WHERE 
          "A"."BlocksetId" >= 0 AND
-- Change FilesetID to desired ID from Fileset table
          "D"."FilesetID" = 99 AND
          ("I"."Index" = 0 OR "I"."Index" IS NULL) AND  
          ("G"."Index" = 0 OR "G"."Index" IS NULL)
        ) J
    LEFT OUTER JOIN 
        "BlocklistHash" K 
    ON 
        "K"."BlocksetID" = "J"."BlocksetID" 
    ORDER BY "J"."Path", "K"."Index"
    ) L

LEFT OUTER JOIN
    "BlocklistHash" M
ON
    "M"."BlocksetID" = "L"."MetablocksetID"
-- ] LIST_FILESETS
) UNION SELECT DISTINCT "Path" FROM (
-- [ LIST_FOLDERS_AND_SYMLINKS
SELECT
    "G"."BlocksetID",
    "G"."ID",
    "G"."Path",
    "G"."Length",
    "G"."FullHash",
    "G"."Lastmodified",
    "G"."FirstMetaBlockHash",
    "H"."Hash" AS "MetablocklistHash"
FROM
    (
    SELECT
        "B"."BlocksetID",
        "B"."ID",
        "B"."Path",
        "D"."Length",
        "D"."FullHash",
        "A"."Lastmodified",
        "F"."Hash" AS "FirstMetaBlockHash",
        "C"."BlocksetID" AS "MetaBlocksetID"
    FROM
        "FilesetEntry" A, 
        "File" B, 
        "Metadataset" C, 
        "Blockset" D,
        "BlocksetEntry" E,
        "Block" F
    WHERE 
        "A"."FileID" = "B"."ID" 
        AND "B"."MetadataID" = "C"."ID" 
        AND "C"."BlocksetID" = "D"."ID" 
        AND "E"."BlocksetID" = "C"."BlocksetID"
        AND "E"."BlockID" = "F"."ID"
        AND "E"."Index" = 0
-- FOLDER_BLOCKSET_ID = -100
-- SYMLINK_BLOCKSET_ID = -200
        AND ("B"."BlocksetID" = -100 OR "B"."BlocksetID" = -200) 
-- Change FilesetID to desired ID from Fileset table
        AND "A"."FilesetID" = 99
    ) G
LEFT OUTER JOIN
   "BlocklistHash" H
ON
   "H"."BlocksetID" = "G"."MetaBlocksetID"
ORDER BY
   "G"."Path", "H"."Index"
-- ] LIST_FOLDERS_AND_SYMLINKS
));
~~~

Thanks for digging into this!

ts678:
> due to issues of corrupting the database when aborting,
> the stop now button has been removed until the corruption issues are
> fixed.

I mentioned that in [Fatal error: Detected non-empty blocksets with no associated blocks](https://forum.duplicati.com/t/fatal-error-detected-non-empty-blocksets-with-no-associated-blocks/5938/43?u=ts678) but I'm not sure if that's what you saw. Regardless, there's a chase underway now to see if anybody can reproduce it reliably.

warwickmm:
That's interesting.  I swear that during my tests I was using the "Stop Now" button.  Now, I don't see it anymore.  Perhaps browser caching issues?  However, when attempting to stop during a "Verifying backend data' step, a "Stop Now" was visible at one point.

When I have time, I'll try and run some more tests.

BlueBlock:
@seantempleton do I need to be a collaborator to contribute to your branch?

seantempleton:
@BlueBlock I think that is correct. If someone wants to push directly to another repository they have to be a collaborator, otherwise they have to fork the repository and create a pull request. What are you looking to do?

BlueBlock:
> @BlueBlock I think that is correct. If someone wants to push directly to another repository they have to be a collaborator, otherwise they have to fork the repository and create a pull request. What are you looking to do?

I worked off your branch and incorporated my solution. I'd like to commit it to your branch since it seems to make sense to work within the same branch.

BlueBlock:
I think I've got a good solution which merges our 2 efforts.

Your work involves the "stop now" and "pause". Mine only involves the "stop after the current file".

seantempleton:
I encountered the "unexpected difference in fileset" issue and then after preventing that from happening, the "Detected non-empty blocksets with no associated blocks" cropped up. I fixed both of them in my testing but I would definitely like someone more knowledgeable of the database to look at the changes. They seem to make sense to me but at the same time I'm not sure if it's an indication of a deeper problem.

The delete queries that are there have been happening that way for 2 to 3 years so if deleting from several tables by bsIdsSubQuery was a problem I would have hoped that the problem would have been noticed since then.

@BlueBlock My PR actually removes the Stop Now button so that's not a problem for now and is focused on the "Stop after Upload" and "Pause" buttons. I looked at your commit and I see you added a CancelToken to a couple places. Unless I'm missing something I don't believe that should be needed as the taskReader.ProgressAsync call will throw an exception, which should be caught, and stop the thread.

BlueBlock:
> @BlueBlock My PR actually removes the Stop Now button so that's not a problem for now and is focused on the "Stop after Upload" and "Pause" buttons. I looked at your commit and I see you added a CancelToken to a couple places. Unless I'm missing something I don't believe that should be needed as the taskReader.ProgressAsync call will throw an exception, which should be caught, and stop the thread.

The cancel token cleanly allows the current file to complete and leaves with orphan files.

I say your solution revolves around "stop now" because your stopping the task before the current file gets to complete but you do still allow the backend storage files queued to complete. This all has the effect of not letting the current file finish but leaving things in a better state than the previous "stop now" which forced the task to terminate.

seantempleton:
@BlueBlock I think part of my problem here is that I'm getting confused by stop now and stop after upload, what I thought they did previously (stop now being a thread abort, and stop after upload not starting any new uploads and letting current uploads finish), and your statement of letting the current file complete. Current file meaning the current file being split into blocks, compressed, etc. or the current file being uploaded? I am under the impression that any files not uploaded will get deleted on the next backup and those files would be split into blocks all over again, meaning it wouldn't matter if it finished or not. Maybe I'm wrong on that.

I'll get your changes and try them out, maybe that will help me understand.

seantempleton:
Ah, I see now what you're saying BlueBlock.

I think the question now boils down to what functionality do we want? One or the other, or maybe both? Stopping after the current uploads finish transferring will stop quicker overall. Stopping after finishing with the current file and also uploading it could take a while but on the other hand if you're 75% done processing a 2GB file for instance, you might want it to finish and then upload before all operations stop.

BlueBlock:
I think both would be nice to have. Plus both capabilities are now implmeneted. Step-now is quite useful especially when testing with large files and you want to stop as soon as possible and cleanly.

seantempleton:
I integrated your code with mine for testing purposes and I think you're right, I shouldn't have a problem with rebasing it.

I'll wait to update this PR until after your PR goes through.

BlueBlock:
Great.. glad to hear it!

BlueBlock:
I might be missing the strategy.. is the idea to be able to pause a backup, close Duplicati, and then resume a paused backup after restarting Duplicati? So a paused state would be remembered when restarting Duplicati?

PR #3713, opened 2019-04-04
Added folder .Cache for filter group Cache on Linux

flafleur:
I have added ".cache" for Linux to exclude ".cache" when Cache filter group is selected.


warwickmm:
Since the location of the cache files on Mac OS X is still a bit unclear, can you edit the pull request description to just mention Linux?  We can then leave issue #3633 as partially resolved, and deal with the Mac filters separately.

warwickmm:
:+1:   I think a further improvement would be to first check the `XDG_CACHE_HOME` [environment variable](https://specifications.freedesktop.org/basedir-spec/basedir-spec-latest.html), but this is a good first step.

Issue #2612, opened 2017-07-26
Clicking Windows 10 notification does not open Duplicati


aybe:
I have:
- [x] searched open and closed issues for duplicates

----------------------------------------

### Version info

**Duplicati Version:** Duplicati - 2.0.1.73_experimental_2017-07-15
**Operating System:** Microsoft Windows Enterprise [Version 10.0.14393]
**Backend:** Local drive

### Bug description
In Windows 10, when a notification appears, you can click on it and it will open the associated program pointing to a more detailed explanation (or whatever) of interest.

Example: when Windows Update has installed some updates and shows a notification, clicking it will take you to a detailed report of what's been installed.

In the case of Duplicati, I'd expect it to open the web interface where I could get more details, currently nothing happens.

### Steps to reproduce
- run a backup manually or wait for one scheduled to run
- when backup has finished, clicking the notification does not open the web interface

**Actual result:** Duplicati web interface does not open
**Expected result:** Duplicati web interface should open so one can further check what this notification was all about

### Screenshots

![2017-07-26 22_06_44-rr60 - microsoft visual studio](https://user-images.githubusercontent.com/1537591/28641548-f712316a-724f-11e7-9d7c-be28e6200d70.png)

(in this case, I have one locked file that couldn't be backed up, when clicked I would expect that Duplicati web interface opens and take me to the log where there is a detailed explanation; currently I have to manually open it using the icon in the tray)

PR #3719, opened 2019-04-06
Open web interface when click on Windows notification

flafleur:
Open web interface when click on Windows notification

Fix notifications appear like error when they are warning only

#2612


flafleur:
Codacy/PR wants static method, not sure why it should be static.  Other methods are not, I did same that other methods in that case.

kenkendk:
Not sure about Codacy here, but it looks good to me.

PR #3723, opened 2019-04-08
Revert to alphabetic db names

kenkendk:
Fixed an issue where the randomly generated database names were numeric instead of alphabetic.

This has no impact on existing database, as the names are stored once generated.

### DIFF
--- a/Duplicati/Library/Main/DatabaseLocator.cs
+++ b/Duplicati/Library/Main/DatabaseLocator.cs
@@ -199,7 +199,7 @@ namespace Duplicati.Library.Main
 
             System.Text.StringBuilder backupName = new System.Text.StringBuilder();
             for (var i = 0; i < 10; i++)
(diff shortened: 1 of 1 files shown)


Pectojin:
Looks good. It's essentially a direct reversal of the right hand side logic removed (accidentally) by @warwickmm 
https://github.com/duplicati/duplicati/commit/0b016cea2cabb7001940395866746f031571a877#diff-ffe10b896e479b969e57a96bf1e58107L208

warwickmm:
Sorry about that.  Thanks for fixing.

Pectojin:
Pretty easy mistake to make :)

I wouldn't have caught it.

kenkendk:
> I wouldn't have caught it.

I reviewed it after knowing that it generated numeric outputs, and _still_ did not catch it :)

PR #3727, opened 2019-04-09
Fix file-read-buffer-size typo

JonMikelV:
"how many bytes a read from a file" ->"how many bytes are read from a file"


JonMikelV:
Sorry - my mistake.

PR #3729, opened 2019-04-09
Fix German Retention Policy description

JonMikelV:
"nächsten 12 Monate" -> "nächsten 36 Monate"


sfahrenholz:
Hello @JonMikelV 
thanks for the correction. Is not [transifex](https://www.transifex.com/duplicati/duplicati/translate/) for the translation? I updating the string in transifex.

kenkendk:
Yes, this needs to go into Transifex, otherwise it will be removed on the next localization update.

JonMikelV:
Thanks @sfahrenholz for doing it in Transfex!  I guess that's what I get for trying to work on a translation when I only know 1 language.   :-)

And thanks to your comment, I now belong to Duplicati on Transifex so hopefully won't make such a silly mistake in the future.  :-D

@kenkendk , I don't know why the Travis CI build failed, but since this local change isn't really needed should it just be Closed/rejected?

warwickmm:
Closing this since the fix has been made in Transifex.

PR #3743, opened 2019-04-19
Show vacuum progress in status message

drwtsn32x:
I have a pretty large database and I use the auto-vacuum option.  I noticed that during the vacuum step (which takes several minutes) the main Duplicati status message would still read "verifying files" or similar.

This was an attempt to have it show when the database is being vacuumed after a backup operation.  Seems to work.

What do you all think?

BTW, I am not sure how to handle the localization files. Also there are a couple other locations in the code where a vacuum operation may be triggered that I didn't touch.


warwickmm:
Works for me.  :+1:

PR #3754, opened 2019-04-28
Remove unused members from Duplicati.GUI.TrayIcon project

warwickmm:
There is a lot of unused code in the source, and I think it is worth cleaning it up.  This removes unused members from the `Duplicati.GUI.TrayIcon` project.

### DIFF
--- a/Duplicati/GUI/Duplicati.GUI.TrayIcon/HttpServerConnection.cs
+++ b/Duplicati/GUI/Duplicati.GUI.TrayIcon/HttpServerConnection.cs
@@ -491,6 +491,12 @@ namespace Duplicati.GUI.TrayIcon
             ExecuteAndNotify("POST", "/serverstate/resume", req);
         }
 
+        public void AbortTask(long id)
+        {
+            var req = new Dictionary<string, string>();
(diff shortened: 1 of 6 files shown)


Duplicati/GUI/Duplicati.GUI.TrayIcon/HttpServerConnection.cs
```diff
@@ -495,32 +491,6 @@ public void Resume()
             ExecuteAndNotify("POST", "/serverstate/resume", req);
         }
 
-        public void StopTask(long id)
```
kenkendk: The methods here are not used, but I think there is some value in leaving them here, as they make it easier to add this functionality to a tray-icon client in the future.
On the other hand, the implementation may change in the meantime leaving them with incorrect code...

warwickmm: My preference would be to remove all unused code, but I see the value of leaving some things in.  If the implementation changes and these become incorrect, then either the next usage will fix them, or we will ultimately remove them.  I'll revert these removals for now.

PR #3755, opened 2019-04-28
Remove unused members from Duplicati.CommandLine project

warwickmm:
There is a lot of unused code in the source, and I think it is worth cleaning it up. This removes unused members from the `Duplicati.CommandLine` project.


warwickmm:
The `ConsoleLogTarget` class was introduced in revision 842fd965439c6f1e6aac2d9ecaedfac1981e16c9 without any usages.  I'm not sure if it should be used, or if it is no longer necessary.  @kenkendk, do you recall the intent of `ConsoleLogTarget`?

Duplicati/CommandLine/Strings.cs
```diff
@@ -31,6 +31,5 @@ internal static class Program {
         public static string QuietConsoleOptionLong { get { return LC.L(@"If this option is set, progress reports and other messages that would normally go to the console will be redirected to the log."); } }
         public static string QuietConsoleOptionShort { get { return LC.L(@"Disable console output"); } }
         public static string SkippingSourceArgumentsOnNonBackupOperation { get { return @"The --source argument was specified in the parameter file, but the current operation is not a backup operation, so the argument is ignored"; } }
-        public static string YouMayGetMoreHelpHere(string url) { return LC.L(@"This link may provide additional information: {0}", url); }
```
kenkendk: This is a placeholder message that is intended to be used with the code for each error and log message. The idea is that the commandline will use the code to provide a link to the forum, searching for the code, such that it is easy to maintain a list of help links.

kenkendk:
The `ConsoleLogTarget` was intended to provide better console output.
It is possible that it is no longer needed as we now have the option `LogLevelConsole` which provides the same.

Duplicati/CommandLine/Strings.cs
warwickmm: Thanks for the explanation.  I reverted the removal and added a comment to explain the currently unused method.

Issue #3747, opened 2019-04-23
Empty source file can make Recreate download all dblock files fruitlessly with huge delay


ts678:
- [x] I have searched open and closed issues for duplicates.

----------------------------------------

## Environment info

- **Duplicati version**: 2.0.4.5 beta
- **Operating system**: Windows 10
- **Backend**: Local folder

## Description

This issue is a bug that is part of open [Recreating broken DB takes forever [$25] #2889](https://github.com/duplicati/duplicati/issues/2889) and [Database recreate runs forever #3668](https://github.com/duplicati/duplicati/issues/3668) problems. Here, an empty source file leads to [Transferring/Reading Entire dblocks instead of dindex for rebuild #1928](https://github.com/duplicati/duplicati/issues/1928). There are many forum reports. Topic [Recreating database logic/understanding/issue/slow](https://forum.duplicati.com/t/recreating-database-logic-understanding-issue-slow/6496) is a good overview and includes much technical analysis and materials. [This](https://forum.duplicati.com/t/recreating-database-logic-understanding-issue-slow/6496/37?u=ts678) is much of what I would have put in a GitHub issue if I hadn't been responding to a user post too.

I tested 2.0.4.17 canary first to see if it showed the problem (it did), then went to 2.0.4.5 beta because it's widely used, I know it better and it avoids canary regression questions.

Beyond the initial test described in the forum, I also tried backing up a Linux ISO, recreate, add 1-byte-file, backup, recreate, add 0-byte file, backup, recreate. All downloaded only dindex files until the 0-byte file entered, at which time the recreate downloaded all of the dblock files after downloading the dindex.

From a resource point of view, this is not just wasted downloads, but also wasted potential decryption, unzip, and database operations due the missing block upload, thus the later inability to find its volume.

The countMissingInformation failure to drop to 0, as mentioned in [Repair is downloading dblock files](https://forum.duplicati.com/t/repair-is-downloading-dblock-files/2780/14?u=ts678) (responding to @kenkendk) and the big [Recreating database logic/understanding/issue/slow](https://forum.duplicati.com/t/recreating-database-logic-understanding-issue-slow/6496/6?u=ts678) analysis expands to SQL below. Try that on `lengthtest - length 1 backup delete repair\bugreport after delete repair\log-database.sqlite` in the zip file cited in "Debug log" below, and watch it return 1 due to VolumeID -1. Assembling and pretty-printing SQL suggests why a negative VolumeID upset it.

~~~
SELECT 
[...]

PR #3758, opened 2019-05-05
Check block size on recreate

Pectojin:
As discussed in #3747 

Check the block size to avoid scavenging behavoir when Duplicati can't locate the 'empty' block which doesn't exist

### DIFF
--- a/Duplicati/Library/Main/Database/LocalRecreateDatabase.cs
+++ b/Duplicati/Library/Main/Database/LocalRecreateDatabase.cs
@@ -498,8 +498,8 @@ namespace Duplicati.Library.Main.Database
                         blocksize / hashsize
                     );
 
-                var missingBlockInfo = 
-                    @"SELECT ""VolumeID"" FROM ""Block"" WHERE ""VolumeID"" < 0 ";
+                var missingBlockInfo =
+                    @"SELECT ""VolumeID"" FROM ""Block"" WHERE ""VolumeID"" < 0 AND SIZE > 0";
             
                 var missingBlocklistVolumes = string.Format(
                     @"SELECT ""VolumeID"" FROM ""Block"", (" +


kenkendk:
This at least fixes the problem, but ideally we should prevent empty files from being created, and purging empty files when they are encountered.

ts678:
> This at least fixes the problem, but ideally we should prevent empty files from being created, and purging empty files when they are encountered.

In [Repair is downloading dblock files](https://forum.duplicati.com/t/repair-is-downloading-dblock-files/2780/14?u=ts678), I asked "I think one design question would be how an empty file is supposed to be represented." I'm still not sure how suggested non-create and purge should should look. Users expect empty files to work as usual, including restore with metadata, so this is an internal question.

Asking the question because [Fatal error: Detected non-empty blocksets with no associated blocks](https://forum.duplicati.com/t/fatal-error-detected-non-empty-blocksets-with-no-associated-blocks/5938/33?u=ts678) may be pointing out that it may be possible to give the DB blockset information without any block. See comment from kenkendk in [Fixed empty block handling in concurrent_processing branch #2637](https://github.com/duplicati/duplicati/pull/2637/files#r132818395), and right above it there is the code change that lets the core loop run 0 times and output no block. I wouldn't think a "non-empty blocksets" error can come from an empty blockset (plus I've made lots of those), but it's good that someone's trying to find the code issue. I can claim limited "by design", but what's the rest of the design?

Pectojin:
I'm not sure I understand the question entirely. Are you asking how we differentiate between a blockset where all blocks were lost and a blockset consisting of empty files? Or just a more practical how/where purging would be implemented?

ts678:
The simple model that Duplicati used to use was that an empty file was like any other file. From a DB view, File table, Blockset table, BlocksetEntry table, and Block table all have entries, just as for usual non-empty.

[Repair is downloading dblock files](https://forum.duplicati.com/t/repair-is-downloading-dblock-files/2780/14?u=ts678) quoted extensively might help clarify the question (at least the DB part):

> I think one design question would be how an empty file is supposed to be represented. I’ve seen many ways, including one where I added an empty file late to a backup without this problem (hoping to induce it). Instead what happened was that the ID of the empty Blockset got no BlocksetID in BlocksetEntry, breaking the chain, whereas on the troubled database, the BlocksetID in BlocksetEntry has a BlockID leading to -1 in Block table, assuming I’m correctly understanding how these things link together. I wonder if there’s good documentation?

From a survey of my backup DBs that I took in January, I categorized them as follows. I got some of each, however only the ones that led to VolumeID -1 in Block table actually caused looping. Others were "OK" in that respect, but it's not clear if they were "OK" per the design because I'm not sure which "look" below is correct, or perhaps none are correct. For example, folders have metadata but don't even start on blockset. I think -100 for BlocksetID in the File table is how it avoids going down the data-file-as-usual path. But now that an empty file is no longer a data-file-as-usual, what's the intended way for DB to look to reflect them?

~~~
(first three lines of my survey -- the rest were survey results)
Looking for empty Blockset then -1 VolumeID
0 OK 1 ends with a 0 length block in dblock
0 OK 2 ends with Blockset. No BlocksetEntry
Bad 1 ends with -1 VolumeID
~~~

In terms of the question the user is raising, I suspect the concern is that if an empty file arrives, it skips the  [AddBlockToOutputAsync](https://github.com/duplicati/duplicati/blob/0d4b8bb27d29f4c8099b4164d04683968c058f04/Duplicati/Library/Main/Operation/Backup/StreamBlockSplitter.cs#L148) but does the [AddBlocksetAsync](https://github.com/duplicati/duplicati/blob/0d4b8bb27d29f4c8099b4164d04683968c058f04/Duplicati/Library/Main/Operation/Backup/StreamBlockSplitter.cs#L162), which is different than a non-empty file, maybe inconsistent, but is it what is preferred? Once having left the standard model, there are lots of variations to choose from on how to represent the empty file in the DB. Whether or not Duplicati proactively prevents or removes, knowing what representation it's aiming at would lead to a greater chance of getting it someday.

We can't really tell the user that what they worry about is by design (definitely by code change though), unless the design is known, so I thought I'd raise the question again of how an empty file should "look".

Thanks!

PR #3759, opened 2019-05-05
added boolean check to make the json output use the subject template …

Pectojin:
…instead of always returning the result as a json payload

See https://forum.duplicati.com/t/json-email-uses-json-output-at-subject-line/6989


kenkendk:
Yes, I think this was the intention :)

PR #3765, opened 2019-05-12
Remove unused members from Duplicati.Library.Backend.AzureBlob project

warwickmm:
This removes unused members from the `Duplicati.Library.Backend.AzureBlob` project.

### DIFF
--- a/Duplicati/Library/Backend/AzureBlob/AzureBlobBackend.cs
+++ b/Duplicati/Library/Backend/AzureBlob/AzureBlobBackend.cs
@@ -78,11 +78,6 @@ namespace Duplicati.Library.Backend.AzureBlob
             get { return "azure"; }
         }
 
-        public bool SupportsStreaming
-        {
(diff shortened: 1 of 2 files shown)


Duplicati/Library/Backend/AzureBlob/AzureBlobBackend.cs
```diff
@@ -21,19 +21,25 @@
...
         private readonly AzureBlobWrapper _azureBlob;
 
+        // ReSharper disable once UnusedMember.Global
+        // This constructor is needed by the BackendLoader.
         public AzureBlobBackend()
         {
         }
 
+        // ReSharper disable once UnusedMember.Global
+        // This constructor is needed by the BackendLoader.
```
warwickmm: I know many people don't use ReSharper or the Rider IDE.  Let me know if you prefer not to see these types of comments that disable warnings.

Issue #3772, opened 2019-05-20
2.0.4.16 and on silently drop uploads getting TaskCanceledException from HTTP timeout


ts678:
- [x] I have searched open and closed issues for duplicates.

----------------------------------------

## Environment info

- **Duplicati version**: 2.0.4.18
- **Operating system**: Windows 10 Version 1809
- **Backend**: Microsoft OneDrive v2

## Description

A backup which calls itself a success may have numerous missing files, very unusual database conditions (and possibly worse). These then break later runs. Older versions through 2.0.4.15 would detect an error (maybe by accident), and could easily be restarted. Not so with newest. It's hard to "fix" missing dblocks.

The basic way of getting into this issue is to run an HTTP upload (driven by Duplicati code) which takes longer than what's allowed by the not-very-well-tuned Duplicati defaults which can be overridden with --http-operation-timeout. The easiest fail is to set that really low (e.g. 1 second). Somewhat challenging would be slowing down uploads using --throttle-upload or throttle in GUI, partly because this is a little buggy and can throttle downloads too (hitting a "Hash mismatch" bug on post-backup verification), or sometimes have no effect. Slowing uploads at the router avoids the verification bug. One can probably also set a huge --dblock-size (as some people do). There are many ways to exceed the permitted time, which is 100 seconds per Microsoft's default, which I think is what Duplicati's OneDrive v2 uses. Some other backend timeouts have increased, but there's no auto-adjustment to actual upload speed AFAIK.

Problem was first noticed and mentioned in [Some backends are failing with parallel upload code #3756](https://github.com/duplicati/duplicati/issues/3756#issuecomment-487568353), with a followup question in [Fix pausing and stopping after upload #3712](https://github.com/duplicati/duplicati/pull/3712#issuecomment-493543915) about throttling in a parallel upload situation.  Maybe running the default four uploads in parallel puts one at greater risk of hitting a timeout for any given upload. Running on a cable modem, I'm generally able to almost fill my link using old-style one-at-a-time uploads, therefore slowdowns due to parallel uploads might explain my failures.

[HttpClient.SendAsync Method](https://docs.microsoft.com/en-us/dotnet/api/system.net.http.httpclient.sendasync?view=netframework-4.8)
[HttpClient.Timeout Property](https://docs.microsoft.com/en-us/dotnet/api/system.net.http.httpclient.timeout?view=netframework-4.8)

[...]

PR #3778, opened 2019-05-26
Throw TimeoutException upon HTTP timeout

warwickmm:
The `HttpClient.SendAsync` method throws an `OperationCanceledException` when the timeout is exceeded.  In order to provide a more informative exception, we will detect this case and throw a `TimeoutException` instead.  This will also allow the `BackendUploader` to differentiate between cancellations requested by the user and those generated by an HTTP timeout.

Previously, a `ThreadAbortException` or `OperationCanceledException` would be [rethrown](https://github.com/duplicati/duplicati/blob/0805b5a53c4856f3b27630098805397771f4e190/Duplicati/Library/Main/Operation/Common/BackendHandler.cs#L419) in the `BackendHandler.DoWithRetry` method.  However, the modifications made in revision [61f511c087](https://github.com/duplicati/duplicati/commit/61f511c087ae0e3fbb6295484a8bd9e36720f1b7#diff-2c0222615cad62c2e1d04706b61c7269R290) from pull request #3684 appear to have let this case [slip through](https://github.com/duplicati/duplicati/blob/c46dbbe58f3aaeba9aa32ed9d2c7a5af11881c6c/Duplicati/Library/Main/Operation/Backup/BackendUploader.cs#L293).

See the following links for more information:

https://stackoverflow.com/questions/10547895/how-can-i-tell-when-httpclient-has-timed-out 
https://thomaslevesque.com/2018/02/25/better-timeout-handling-with-httpclient

We also took this opportunity to provide a `CancellationToken` to some `async` methods that may support cancellation.

This addresses issue #3772.


warwickmm:
Previously, an HTTP timeout with the Microsoft OneDrive backend would yield the following result with no error reported to the user (note the backup size for a source that was not compressible):
image

With these changes, we get
image

PR #3779, opened 2019-05-27
Remove unused members from Duplicati.Library.Backend.Dropbox project

warwickmm:
This removes unused members from the `Duplicati.Library.Backend.Dropbox` project.

### DIFF
--- a/Duplicati/Library/Backend/Dropbox/DropboxHelper.cs
+++ b/Duplicati/Library/Backend/Dropbox/DropboxHelper.cs
@@ -339,8 +339,6 @@ namespace Duplicati.Library.Backend
         [JsonProperty(".tag")]
         public string tag { get; set; }
         public string name { get; set; }
-        public string path_lower { get; set; }
-        public string path_display { get; set; }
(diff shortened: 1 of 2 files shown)


Duplicati/Library/Backend/Dropbox/DropboxHelper.cs
```diff
@@ -336,16 +339,19 @@ public class MetaData
...
 
+        // While this is unused, the Dropbox API v2 documentation does not
+        // declare this to be optional.
+        // ReSharper disable once UnusedMember.Global
+        public string id { get; set; }
+
+        // While this is unused, the Dropbox API v2 documentation does not
+        // declare this to be optional.
+        // ReSharper disable once UnusedMember.Global
+        public string rev { get; set; }
```
warwickmm: While everything still appears to work when the `id` and `rev` fields are removed, I left them in because the documentation does not claim that they are optional (in contrast to the fields I removed above):

https://www.dropbox.com/developers/documentation/http/documentation#files-upload

Issue #3334, opened 2018-08-08
public usage statistics is not referenced


sfahrenholz:
- [x] I have searched open and closed issues for duplicates.

----------------------------------------

## Environment info

- **Duplicati version**: 2.0.3.9_canary_2018-06-30
- **Operating system**: Windows 10

## Description

When i click on the link to the Usage statistics for Duplicati, so the new tab opend with a reference to http://localhost:8200/ngax/index.html#/settings

## Steps to reproduce

1. Settings
2. Usage statistics
3. Click on public usage statistics

- **Actual result**:
reference to http://localhost:8200/ngax/index.html#/settings

![grafik](https://user-images.githubusercontent.com/6977657/43860440-fdfa8670-9b53-11e8-8779-6dadf222672c.png)

- **Expected result**:
The Tab shows the content of https://usage-reporter.duplicati.com

PR #3799, opened 2019-06-16
Fix usage statistics link

warwickmm:
This fixes the usage statistics link in the Settings page of the UI.  Previously, clicking on the link would just redirect to `http://localhost:8200/ngax/index.html` instead of the usage reporter page.

This fixes issue #3334.


warwickmm:
This works for me, but I'm not sure if this breaks the translations.  For example, this no longer matches the  `msgid` [here](https://github.com/duplicati/duplicati/blob/master/Localizations/webroot/localization_webroot-es.po#L2753).  Any guidance here would be much appreciated.

Pectojin:
We'll have to break the translation, but it should be fine.

Issue #3800, opened 2019-06-19
"Unexpected difference in fileset" test case and code clue


ts678:
- [x] I have searched open and closed issues for duplicates.

----------------------------------------

## Environment info

- **Duplicati version**: 2.0.4.5 beta
- **Operating system**: Windows 10
- **Backend**: Local folder

## Description

"Unexpected difference in fileset version 0...". I investigated, found seeming cause, then built steps. Current test case does not perfectly reproduce my actual failure, but it's easy to run, and it's a start.

The initial debug was basically a profiling log with --profile-all-database-queries, and a --run-script-before to keep a trail of DB copies. After trials showing compact seemed essential, I held off on that until the usual (small) number of backups-to-failure, ran compact, got issue, checked logs and DBs.

One basic problem is that the wasted space calculator ignores blocks that been deleted, even if the block gets used again in a different file (or a reapparance of the same file), so I found that a dblock containing a file data block got deleted as empty. Here's an example  test that checks hash and size:

https://github.com/duplicati/duplicati/blob/a03cd08bad4003e2fd10556c06c0ec6d0e17971a/Duplicati/Library/Main/Database/LocalDeleteDatabase.cs#L126

Maybe the above decision to ignore a block's existence needs another qualifier (not fully tested) like
> AND ""Block"".""VolumeID"" = ""DeletedBlock"".""VolumeID""

A second suspect problem is described in [Consistent Unexpected difference in fileset version](https://forum.duplicati.com/t/consistent-unexpected-difference-in-fileset-version/7101/7?u=ts678) where most of the tables get cleanup based on BlocksetID, but in addition to that, File cleans up based on MetadatasetID without mapping it to BlocksetID. This seems able to do random additional File table deletions compared to the other tables, causing "Unexpected difference" in the differently built lists. There's no test case yet for that variant of the issue, but below are tech notes behind the listed case:

~~~
save file A
backup to VolumeID 3 with file A data and metadata, folder metadata
delete file A
backup to VolumeID 6 with folder metadata
(delete version 1 loads DeletedBlock table including file A data)
save file A
backup to VolumeID 9 with file A data and metadata, folder metadata
(delete version 1 just deletes older metadata row in Block table)
delete file A
save file A
backup to VolumeID 12 with new metadata. The old blockset is reused
(delete version 1 leaves file A data as the only block in VolumeID 9)
storedlist is 2
expandedlist is 2
		SELECT SUM("Block"."Size") AS "ActiveSize"
[...]

PR #3801, opened 2019-06-22
Fix database issue where block was not recognized as being used

warwickmm:
On behalf of @ts678, here is a proposed fix to an occurrence of the "Unexpected difference in fileset version" error, due to a deleted block having the same hash as one that is currently in-use.  See issue #3800 for more details.

This fixes issue #3800.


kenkendk:
I understand the problem now. A block is deleted but also in use (in two different volumes) and I was not expecting that when I wrote the query. Great work @ts678 that was really hard to track down!

Issue #3787, opened 2019-06-03
Upload throttle corrupts backup, especially OneDrive. Analyzed, with code proposed.


ts678:
- [x] I have searched open and closed issues for duplicates.

----------------------------------------

## Environment info

- **Duplicati version**: 2.0.4.18
- **Operating system**: Windows 10 Version 1809
- **Backend**: OneDrive or local folder

## Description

Issue is that throttle can lose data, seemingly by internal overwrite, resulting in bad remote volume. Proposed fix is [ThrottledStream](https://github.com/duplicati/duplicati/blob/b85b817d674cc63e0bec4c6201b1ddf01be71bac/Duplicati/Library/Utility/ThrottledStream.cs) Read should advance its buffer offset as it reads through its Stream.

This is on volume uploads, so is permanent damage, unlike [Hash mismatch download corruption, likely from throttle buffer offset being reused](https://github.com/duplicati/duplicati/issues/3782) where removing the download throttle resolves the issue. This problem is also more serious because it happens at reasonable levels of throttling (at least on OneDrive) but (like the other) can even happen on local files when throttle is extremely low. It's not proven that the problems at the two speeds are identical, however the symptoms sure look the same, and are reversed from the download case because this is from ThrottledStream [Read](https://github.com/duplicati/duplicati/blob/b85b817d674cc63e0bec4c6201b1ddf01be71bac/Duplicati/Library/Utility/ThrottledStream.cs#L124), not [Write](https://github.com/duplicati/duplicati/blob/b85b817d674cc63e0bec4c6201b1ddf01be71bac/Duplicati/Library/Utility/ThrottledStream.cs#L155). In both cases the buffer offset is not adjusted when data is read or written, but here the bug is reading newer data to the initial buffer offset, whereas in the other issue it was writing the initial buffer offset repeatedly to the output.

Both issues produce a file with the right size but scrambled content, however the upload case may not be immediately noticed after backup because correct size will be considered Verified in RemoteVolume table. The default [verification](https://duplicati.readthedocs.io/en/latest/03-using-the-graphical-user-interface/#verifying-backend-files) of one [sample](https://duplicati.readthedocs.io/en/latest/04-using-duplicati-from-the-command-line/#the-test-command) trio of files might catch the issue as a "Hash mismatch", but it's not guaranteed and can also be disabled. If enabled, load-balancing might favor checking new files.

[...]

Issue #3782, opened 2019-05-27
Hash mismatch download corruption, likely from throttle buffer offset being reused


ts678:
- [x] I have searched open and closed issues for duplicates.

----------------------------------------

## Environment info

- **Duplicati version**: 2.0.4.5
- **Operating system**: Windows 10 or Linux
- **Backend**: File or other

## Description

Throttling downloads low enough gets a "Hash mismatch" error and gives up after 5 (default) tries, so breaks things like restores and remote volume verifies done after backups. The bug is probably in the ThrottledStream [Write](https://github.com/duplicati/duplicati/blob/c46dbbe58f3aaeba9aa32ed9d2c7a5af11881c6c/Duplicati/Library/Utility/ThrottledStream.cs#L155) code where the same input buffer and offset are reused if calling conditions set that up. It's not a network timeout, and even local files will do this. This is not a new issue, but reliably getting it appears to take very low throttle rates, so I hadn't looked at it before. There's a maybe more serious corruption on throttled uploads making files with missing data. Unclear if and how it's related.

One would expect buffer reuse to output the same data repeatedly, instead of getting new data. This is what's observed using the 4MiB known-data file from [Issue 3772](https://github.com/duplicati/duplicati/issues/3772). It contains lines starting with BLOCK NUMBER ##. Below are the original and copied file, to the left and right of the slash respectively. It looks like source data gets output repeatedly at fixed intervals, instead of new data (offset not moved). To get copied /tmp file required Linux (OSX could probably also do this) because Windows locks out file reading while it's in writing. Helping to make sure I got the whole file, I put the copy in a loop and let it go until file vanished. After that was some work with hexdump -Cv and less as the viewer. I searched for BLOCK and noted all addresses (rounded to 0x10) then NUMBER to do the same thing, avoiding the text splits.

~~~
BLOCK NUMBER 00 000001b0 / 000001b0 000051b0 0000a1b0 0000f1b0
BLOCK NUMBER 01 00040260 / 00040260 00045260 0004a260 0004f260
BLOCK NUMBER 02 0004e2b0 / missing
BLOCK NUMBER 03 00075310 / missing
BLOCK NUMBER 04 00083360 / 00083360 0008d360
BLOCK NUMBER 05 000aa3c0 / missing
BLOCK NUMBER 06 000b8420 / missing
BLOCK NUMBER 07 000df470 / missing
BLOCK NUMBER 08 000ed4d0 / missing
[...]

PR #3802, opened 2019-06-24
Fix ThrottledStream Read buffer overwrite

ts678:
Adjust buffer offset by amount read so far in the internal Read loop,
so newly read data goes after old data instead of going on top of it.

This fixes issue #3787 read buffer corruptions when throttle is used.

### DIFF
--- a/Duplicati/Library/Utility/ThrottledStream.cs
+++ b/Duplicati/Library/Utility/ThrottledStream.cs
@@ -133,7 +133,7 @@ namespace Duplicati.Library.Utility
 				var chunksize = (int)Math.Min(remaining, m_readspeed <= 0 ? remaining : m_readspeed * 2);
 				DelayIfRequired(ref m_readspeed, chunksize, ref m_last_read_sample, ref m_current_read_counter, ref m_current_read_speed);
 
-				var actual = m_basestream.Read(buffer, offset, chunksize);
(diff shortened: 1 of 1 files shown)


warwickmm:
It looks like prior to revision 0190a9994e55c0e4c6b7c7eaac43ec25dd811fc5, the `DelayIfRequired` method would increment the `offset` variable to advance through the stream.  When we modified the implementation in 0190a9994e55c0e4c6b7c7eaac43ec25dd811fc5, we may have forgotten to modify the `offset` accordingly.

Do we have a similar issue in the implementation of the `Write` method?

ts678:
> Do we have a similar issue in the implementation of the `Write` method?

Yes. [Hash mismatch download corruption, likely from throttle buffer offset being reused #3782](https://github.com/duplicati/duplicati/issues/3782) was actually filed first, but seemed more obscure in terms of impact (also reversible, assuming it only affects downloads, which it seemed to, but given the confusion over which throttle controls which direction, I'm not sure if we can totally assume that). I've not yet tried a code fix to see if I could get super-slow downloads with safety. Fix should be equally trivial, but would deserve effectiveness testing. Pull we're in now did fix my uploads...

kenkendk:
Oops, that mistake was mine I think.

PR #3811, opened 2019-07-08
Fix implementation of ThrottledStream.Write

warwickmm:
This fixes an issue with the `ThrottledStream.Write` method, where the offset into the source buffer was not advanced properly.  Prior to revision 0190a9994e, the `DelayIfRequired` method would increment the offset variable to advance through the buffer.  When we modified the implementation in revision 0190a9994e, we may have forgotten to modify the offset accordingly.  

We also add unit tests for `ThrottledStream.Write` as well as `ThrottedStream.Read` (see issue #3787, which was fixed in pull request #3802).

This might fix issue #3782.

### DIFF
--- a/Duplicati/Library/Utility/ThrottledStream.cs
+++ b/Duplicati/Library/Utility/ThrottledStream.cs
@@ -147,9 +147,9 @@ namespace Duplicati.Library.Utility
         }
 
 		/// <summary>
-		/// Write the specified buffer, offset and count.
+		/// Write from a buffer to this stream.
 		/// </summary>
-		/// <param name="buffer">The buffer to write to.</param>
+		/// <param name="buffer">The buffer to read from.</param>
 		/// <param name="offset">The offset into the buffer.</param>
 		/// <param name="count">The number of bytes to write.</param>
         public override void Write(byte[] buffer, int offset, int count)
--- a/Duplicati/Library/Utility/ThrottledStream.cs
+++ b/Duplicati/Library/Utility/ThrottledStream.cs
@@ -116,9 +116,9 @@ namespace Duplicati.Library.Utility
         }
 
 		/// <summary>
-		/// Read the specified buffer, offset and count.
+		/// Read from this stream into a buffer.
 		/// </summary>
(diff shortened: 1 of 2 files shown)


Duplicati/Library/Utility/ThrottledStream.cs
```diff
@@ -147,24 +147,25 @@ public override int Read(byte[] buffer, int offset, int count)
...
 
 		/// <summary>
-		/// Write the specified buffer, offset and count.
+		/// Write from a buffer to this stream.
 		/// </summary>
-		/// <param name="buffer">The buffer to write to.</param>
+		/// <param name="buffer">The buffer to read from.</param>
 		/// <param name="offset">The offset into the buffer.</param>
 		/// <param name="count">The number of bytes to write.</param>
         public override void Write(byte[] buffer, int offset, int count)
```
warwickmm: This method uses a mixture of spaces and tabs for the indentation.  I decided to not fix the indentation to keep the diff here cleaner looking.

Duplicati/UnitTest/UtilityTests.cs
```diff
@@ -204,6 +205,74 @@ public void ParseBool()
                 Assert.IsFalse(Utility.ParseBool(value, returnsFalse));
             }
         }
+
+        [Test]
+        [Category("Utility")]
+        public static void ThrottledStreamRead()
```
warwickmm: This test fails without the changes from pull request #3802.

Duplicati/UnitTest/UtilityTests.cs
```diff
@@ -204,6 +205,74 @@ public void ParseBool()
...
+                    {
+                        Assert.AreEqual(default(byte), destinationBuffer[k]);
+                    }
+                }
+            }
+        }
+
+        [Test]
+        [Category("Utility")]
+        public static void ThrottledStreamWrite()
```
warwickmm: This test fails without the changes from revision 5d087791d02fe44879decda7ab78dd8d4aa95fa4.

PR #3815, opened 2019-07-15
Make german text more readable

colinfrei:


Pectojin:
Hi @colinfrei,

Thanks for taking the time to improve the translations, but can I get you to make the changes in Transifex? https://www.transifex.com/duplicati/duplicati/translate/

The reason being that these .po files are actually auto-generated from there, so they get overwritten every time we update our translations.

You can log in with your Github account so it's quick, and then you just pick the webroot file and German translation. You can then search for the english text to be translated, e.g. 
image

colinfrei:
I adjusted the text in transifex.

Pectojin:
Thanks, @colinfrei ! :smile:

PR #3825, opened 2019-07-20
A collection of fixes and updates

BlueBlock:
- Modify/correct the backup request of 'Stop after upload' to stop the backup cleanly after the current file is done
- Fix creation of symlinks to honor no overwrite
- Update all nuget packages to current versions; many are very old
- Update all projects to framework 4.7.2
- Update included gpg version
- Add logic for the path of gpg.exe being used to give priority to the gpg4win version if installed by the user


warwickmm:
Thanks @BlueBlock for your contributions, it looks like you put in a lot of hard work.  Unfortunately, this pull request consists of so many different changes that it is difficult to review.  Can you split this up into several smaller pull requests where each one is focused on one particular change?  It would really allow the reviewers to provide a more thoughtful review.

For example, I would recommend a separate pull request for each of the following:

* Fix creation of symlinks to honor no overwrite
* Update all nuget packages to current versions; many are very old
* Update all projects to framework 4.7.2 (this may need to coincide with the nuget package updates)
* Update included gpg version and add logic for the path of gpg.exe being used to give priority to the gpg4win version if installed by the user.

Also, a lot of work is being done on fixing the stopping behavior in pull request #3712.  Can you take a look there and see how your approach can be reconciled with the changes there?

BlueBlock:
I just happened to stumble on that thread for also fixing the stopping behavior. Is it a working solution?

Thanks for the recommendation for separating the pull requests. Though I've never done that before so I'll have to look at how to accomplish it.

warwickmm:
> I just happened to stumble on that thread for also fixing the stopping behavior. Is it a working solution?

Yes, although it has revealed some potential timing issues elsewhere in the code that may cause database corruption.  We're trying to be very careful here as the database corruption issues are unacceptable for users, and unfortunately are the most difficult to reproduce and fix.

> Thanks for the recommendation for separating the pull requests. Though I've never done that before so I'll have to look at how to accomplish it.

I would recommend creating a separate branch for each pull request.  If you're comfortable with git, you can create the branches by carefully cherry-picking the commits that you've already created.

BlueBlock:
I think my solution resolves the database issue you might be seeing. I use a cancellation token to cleanly stop the job after the current file is complete.

BlueBlock:
> I would recommend creating a separate branch for each pull request. If you're comfortable with git, you can create the branches by carefully cherry-picking the commits that you've already created.

I'll attempt that. Thanks.

warwickmm:
> I think my solution resolves the database issue you might be seeing. I use a cancellation token to cleanly stop the job after the current file is complete.

I believe that's the approach taken in #3712 as well?  The most recent testing seems to indicate that the corruption issues occur when stopping _prior_ to any file uploads.  When stopping during file uploads, everything appears to work fine.

Perhaps instead of duplicating the effort, would you be interested in helping us improve, review, and test the changes in #3712?  That pull request includes related changes for pausing, as well as related UI changes to notify the user of the pausing/stopping status.

BlueBlock:
> Perhaps instead of duplicating the effort, would you be interested in helping us improve, review, and test the changes in #3712? That pull request includes related changes for pausing, as well as related UI changes to notify the user of the pausing/stopping status.

Definitely. I'll backout and save my solution for it and start testing your branch. I need to see how I can merge your branch with all the other changes I've made though.

warwickmm:
I wouldn't worry too much about merging any changes into your work, especially if the changes are still evolving.  I would just base your branches off the current upstream `master` branch and we can worry about merge conflicts at the time of merge.

BlueBlock:
> I wouldn't worry too much about merging any changes into your work, especially if the changes are still evolving. I would just base your branches off the current upstream `master` branch and we can worry about merge conflicts at the time of merge.

Got it.  Thanks.

PR #3826, opened 2019-07-20
 Add check for test data requirement to unittest

BlueBlock:


Duplicati/UnitTest/CommandLineOperationsTests.cs
```diff
@@ -198,6 +204,10 @@ private void DoRunCommands(string target)
                 if (Duplicati.CommandLine.Program.RealMain((new string[] { "test", target, "all" }.Union(opts)).ToArray()) != 0)
                     throw new Exception("Failed during final remote verification");
 
+            foreach (var datafolder in datafolders)
+            {
+                Directory.Delete(datafolder, true);
```
warwickmm: I think this might make more sense in a `[OneTimeTearDown]`, but we can clean this up along with the other tests separately.

Issue #3822, opened 2019-07-16
Update nuget packages?


BlueBlock:
Most nuget packages are fairly old.  I can update them unless there is a compatibility reason they are on older versions?

ts678:
Hmm. Is this offer retracted? I was just about to suggest FluentFTP because it just got a rather important [fix](https://forum.duplicati.com/t/problems-after-upgrade-to-2-0-4-21/7469/18?u=ts678) that may require an update-and-test with some FTP server (FileZilla Server would do if you're on Windows).

Getting more ambitious (and maybe less urgent) there's a Mega.nz [issue](https://forum.duplicati.com/t/mega-nz-error-when-check-connection/5316/8?u=ts678) that an update might help. Nobody on that post seems interested in a DIY test, and recently I'm suspecting a Duplicati rebuild might be needed.

Then there's everything else that's getting old and possibly having issues (e.g. security) that's not so obvious.

If you're still interested, somebody else would have to advise you on the mechanics. I'm not really a dev here.

If you just go ahead on your own and do pull requests, separate ones might be best, in case issues arise. Sometimes compatibility issues aren't apparent until the new package gets into canary for field testing...

Thanks for the offer!

BlueBlock:
Not retracted. I thought my question might be a little too simple so I closed it.

I've updated all the packages and will be submitting them all shortly.

I've also included a fix to a big problem I've been having when trying to stop big backup jobs. I would try using "Stop after upload" but it had no effect in stopping the job. I've changed it so 'Stop after upload' will stop the backup after the current file is complete.

Issue #13, opened 2014-08-05
Split Setup into Backup/Restore in the context menu


kenkendk:
_From [wiktor.w...@gmail.com](https://code.google.com/u/109935308661018396116/) on March 09, 2009 17:56:26_

This is my first ticket, so I'd like to write that I've just installed 
Duplicati and it seems that it is exactly what I was searching for, keep 
up the good work :-)

Anyway, here's the first minor thing that was not intuitive for me: once I 
have set up a test backup, I tried to restore a file. It took me just a 
couple of seconds to find it because there are only a couple of items, 
however I think it would be more intuitive to have a separate item in the 
context menu to restore files called "Restore" (for example).

_Original issue: http://code.google.com/p/duplicati/issues/detail?id=13_

PR #3827, opened 2019-07-20
Update nuget packages

BlueBlock:


warwickmm:
Thanks @BlueBlock.  It looks like the changes in revision 22ca352 are already included in pull request #3826.  If they're not required for updating the nuget packages, can you remove that commit from this pull request?

warwickmm:
FYI, if you mention a related issue (e.g., #3822) in a pull request comment or in a commit message, GitHub will automatically create a link in the issue to the pull request.  This helps to relate changes in the code with feature requests, bug fixes, etc.

BlueBlock:
Commit reverted for issue #3822. Thanks for your help.

BlueBlock:
As suggested, moved framework upgrade commits from #3829 to here.

ts678:
I don't do much C#, but is 4.7.2 required to get some of these updates? What does 4.7.2 mean for systems such as Windows 7, or mono installs where 4.7.2 support was seemingly added December 2018 for [5.18.0](https://www.mono-project.com/docs/about-mono/releases/5.18.0/)?

[Target frameworks](https://docs.microsoft.com/en-us/dotnet/standard/frameworks) sounds like raising the target might require users to upgrade .NET Framework or mono.

[.NET Framework version history](https://en.wikipedia.org/wiki/.NET_Framework_version_history) shows that Windows 7 can at least be updated. I'm not sure whether that's automatic. [Duplicati won't run on Windows XP](https://forum.duplicati.com/t/duplicati-is-work-on-windows-xp/3899) due to the need for .NET Framework 4.5. Vista can run that. XP and Vista are old, but Windows 7 will likely linger even if it's [unsupported by Microsoft](https://www.microsoft.com/en-ww/windowsforbusiness/end-of-windows-7-support) in January 2020.

[https://www.duplicati.com/](https://www.duplicati.com/)

> Duplicati runs under Windows, Linux, MacOS. It requires .NET 4.5 or Mono.

Source for above note might be at [https://github.com/duplicati/duplicati.github.io/blob/master/index.md](https://github.com/duplicati/duplicati.github.io/blob/master/index.md), and if (after consideration) system requirement moves, then it should get publicized, and transition should not wind up with something like a Duplicati automatic update breaking Duplicati for unsuspecting users...

Or if I got this all wrong, then please tell me what the implications are of the change of target. Thank you!

As a side question, anyone know if there's any automation for smoke-testing (or better) of backend code? The general question is whether such a big change can safely get to beta soon, or needs a big canary test. [Here](https://forum.duplicati.com/t/release-2-0-4-23-beta-2019-07-14/7522/20?u=ts678) is me wondering how to simultaneously speed up beta rates, and also make sure it's of good quality.

BlueBlock:
Given this is still in beta updating the framework seems reasonable and the requirements page can be updated. 

I'll look to see if the updater checks for requirements. It would be nice for it to abort an upgrade if the framework requirements are not met.

warwickmm:
I do share @ts678's concern about compatibility.  Unless we absolutely need a new API or C# language feature, then perhaps we should not update the target framework unnecessarily.  I'm no expert in this, but it seems that the recommendation is to set the target framework to be the minimum that is required.

@BlueBlock, I apologize for going back and forth on this multiple times, but do you think it's ok to upgrade the nuget packages without updating the target framework version?

BlueBlock:
I understand the hesitation. 

There are C# advances that are useful in forward frameworks since the C# language closely corresponds to the franework.

I'll need to check if some of the features like default value parameters are supported by 4.5

IMHO 4.6 is the minimum to target.

BlueBlock:
The priority is the packages.  So I'll close this PR and have another just for the packages.

BlueBlock:
@warwickmm  I'd like to upgrade the target framework to at least 4.5.2 sound good?

warwickmm:
I think moving to 4.6 is reasonable.  There was a strange move by Microsoft a few years ago to [end support](https://devblogs.microsoft.com/dotnet/support-ending-for-the-net-framework-4-4-5-and-4-5-1/) for 4.5 and 4.5.1 (but not some older versions) for some reason.

@kenkendk, @Pectojin, do you see any problems with updating the target framework to 4.6?

Duplicati/UnitTest/GeneralBlackBoxTesting.cs
```diff
@@ -36,13 +37,30 @@ protected IEnumerable<string> TestFolders
...
+                        BasicSetupHelper.ProgressWriteLine($"Missing data in '{SOURCE_FOLDERS}' with required minimum of 1 sub-folder with the naming convention of '{rx}'");
+                    }
+
+                    return testFolders;
+                }
+                catch (DirectoryNotFoundException)
+                {
+                    BasicSetupHelper.ProgressWriteLine($"Missing data directory '{SOURCE_FOLDERS}'");
+                    return null;
+                }
```
warwickmm: These changes seem unrelated to the nuget and framework update.  Were these changes intended for a different pull request?

Duplicati/Library/Common/IO/SystemIOWindows.cs
```diff
@@ -547,16 +547,15 @@ public string PathCombine(params string[] paths)
...
             if (FileExists(symlinkfile) || DirectoryExists(symlinkfile))
-                throw new System.IO.IOException(string.Format("File already exists: {0}", symlinkfile));
+                throw new System.IO.IOException($"File already exists: {symlinkfile}");
 
             Alphaleonis.Win32.Filesystem.File.CreateSymbolicLink(PrefixWithUNC(symlinkfile),
-                                                                 target,
-                                                                 asDir ? Alphaleonis.Win32.Filesystem.SymbolicLinkTarget.Directory : Alphaleonis.Win32.Filesystem.SymbolicLinkTarget.File,
-                                                                 AlphaFS.PathFormat.LongFullPath);
+                target,
+                AlphaFS.PathFormat.LongFullPath);
```
warwickmm: Is this change because the old method overload has been deprecated?

http://alphafs.alphaleonis.com/doc/2.2/api/html/8863EE02.htm

Duplicati/UnitTest/GeneralBlackBoxTesting.cs
BlueBlock: Yes.. it can be removed from this PR.

BlueBlock: This PR is a bit messy. I'm thinking we just close it and if agreed I'll commit the 4.6.2 to a new PR.

Duplicati/Library/Common/IO/SystemIOWindows.cs
BlueBlock: Correct because it is deprecated, but it isn't related to this any way.

Let's close this PR once we're done discussing 4.6.2

I have have all the 4.6.2 changes in a new branch. I can put it into a new PR if we get the green light.

Pectojin:
> do you see any problems with updating the target framework to 4.6?
Looks like it wouldn't be a problem. Mono compatibility with 4.6 has been good for a while.

I'm having a hard time telling exactly how far back in Mono releases .NET 4.6 support goes, but considering the age of .NET 4.6 it should be safe. Especially with Both Debian 10 and RHEL 8 being out this year with much newer packages.

BlueBlock:
> > do you see any problems with updating the target framework to 4.6?
> > Looks like it wouldn't be a problem. Mono compatibility with 4.6 has been good for a while.
> 
> I'm having a hard time telling exactly how far back in Mono releases .NET 4.6 support goes, but considering the age of .NET 4.6 it should be safe. Especially with Both Debian 10 and RHEL 8 being out this year with much newer packages.

The minimum mono version would be "Mono 4.8.0 (22 Feb 2017)" which includes support for Standard 2.0

warwickmm:
Sounds good to me.  @BlueBlock, did you want to continue with this pull request or open a new one?

BlueBlock:
A new one. Thanks.

I'll close this when I submit the new one.

ts678:
I'm not sure I can agree with @Pectojin that new Linux distro versions avoid issues. What of old ones?

Many "LTS" (Long Term Support) distros go for 5 years (Debian/Ubuntu/Mint) and Red Hat Enterprise Linux (and its CentOS rebuild) go for 10. Is Duplicati going to demand that users upgrade their Linux?

While mono can be upgraded directly from mono-project, it (or its dependencies) might break things.

> The SharpAesCrypt nuget package has moved to Standard 2.0 which would require at least 4.6

[convert to netstandard #9](https://github.com/kenkendk/sharpaescrypt/pull/9) by @mnaiman was accepted over a year ago, as part of a possible longer-term Duplicati goal of moving to .NET Core. Meanwhile, Duplicati remains on 1.3.1, presumably due to no reason to upgrade, and support reason not to. NuGet says the following about the newer versions:

> 1.3.2 is a rebuild targeting .Net 4.5 instead of 2.0.
> 1.3.3 is a rebuild targeting .net standard 2.0.

> The SharpAesCrypt will present a problem going forward as we're at a standstill. We can either move forward to framework 4.6 or we need to replace the SharpAesCrypt with another solution.

I don't think I understand this. What's wrong with staying on 1.3.1? If we can, are we only left with this:

> Some updated nuget libraries are requiring at least 4.5.2

Anybody know what release this large (and much appreciated, even though I prefer to limit the fallout) update is heading for, and how testing (beyond automated unit tests) will happen? I'd think something of this magnitude should be run by users in canary awhile, but we're now trying to get a beta out, BUT "FTP (Alternative)" is broken (lib bug) and needs a fix (and test) as described in [Problems after upgrade to 2.0.4.21](https://forum.duplicati.com/t/problems-after-upgrade-to-2-0-4-21/7469/18?u=ts678) and mentioned in [Update nuget packages? #3822](https://github.com/duplicati/duplicati/issues/3822) which was the early generous offer from @BlueBlock that led to the pull request we're planning. There was also a Mega.nz update proposed to support current Mega users with new accounts. Mega changed things. [The lib caught up](https://forum.duplicati.com/t/mega-nz-error-when-check-connection/5316/8?u=ts678) but we didn't.

Another need which becomes more apparent on a larger jump is updating all the install dependencies. We now ask for mono 3.0 (which seems wrong). [Updated to .Net framework version 4.5](https://github.com/duplicati/duplicati/commit/0d7cf7203bae4a4fe4307889d9d328c4d10bd661) hints Duplicati used v4.0 before v4.5. [upstream #13](https://github.com/mnaiman/duplicati/pull/13) gives some samples of what changed for past move to mono 3.0.

> I'm having a hard time telling exactly how far back in Mono releases .NET 4.6 support goes

I have to agree that mono only occasionally comments about equivalence to .NET Framework versions.
I worked hard on this yesterday (along with trying to work out what distros had old mono, and whether we care based on forum presence of those distros). [Survey for Linux users, which version of Mono do you have installed?](https://forum.duplicati.com/t/survey-for-linux-users-which-version-of-mono-do-you-have-installed/7605) seemed like it would get little response. Too few people read forum when they don't have an issue, so I just used Google to look for mentions (typically an issue). I can post survey if you like.

Thanks to all who are working on this. Although I only recalled two functional upgrades needed, there might be others I missed, and it might be unknown what less obvious and security fixes we're missing...

What do people think of doing an immediate-need update for beta while long-term plan gets planned?

mnaiman:
Better to join forces to migrate to .NET core and will get rid of FW dependency all together on all platforms ;)

BlueBlock:
The PR moving everything to Standard2.0 doesn't look to be ready yet (unit tests don't complete). By upgrading to 4.6.2 we can then gradually move each library to Standard2.0 rather that as a whole, all the while other work continues. We'll have 1 library already at Standard2.0 and the work done on the other PR is not lost a it can be used to help port to existing libraries to Standard2.0

Sitting on old frameworks is not really a good option especially when you integrate so many 3rd-party libraries. Libraries need to be very current, if not just for security fixes. The AWS packages are updated daily.

mnaiman:
There are some changes on tests that's needs to be done, yes, that's why I wrote to join forces - 98% is done already.

Migration one by one will be complete pain for users, there are 20 libraries, you cant upgrade them, because some of them rely on new async API.
Then after every library upgrade new issues emerge and tests needs to be done. Better is to do big bang, new experimental release, catch all related bugs and then release to beta channel for users.

But I agree that libraries should be upgraded regularly with security in mind.
.NET core 3 is marked as production ready, so we could aim on that with libraries as netstandard2.0 or 2.1

Its up to @kenkendk to decide strategy for our PR to be merged :)

BlueBlock:
I thought a slow approach might be better but I get your points. I'm all for combining resources and going to Standard2.0. For the complete move to Standard2.0 what are the biggest issues blocking it?

BlueBlock:
I'm going to be closing this PR and opening a new one for 4.6.2

warwickmm:
Sounds good to me.  We can continue the discussion in the new pull request.  I'm not sure how difficult it is to upgrade mono in a LTS distribution, but I guess that's one of the tradeoffs one makes when deciding to use a LTS distro.  Given that Microsoft ended support for .NET 4.5.1 on Jan 12, 2016, and Mono 4.8 has been out for almost 2.5 years, it's not that unreasonable to update the target framework to 4.6.2.

BlueBlock:
This Pr is replaced with (#3844)

PR #3829, opened 2019-07-20
Update framework to 4.7.2

BlueBlock:


warwickmm:
Thanks @BlueBlock.  It looks like the following commits are part of #3830. 

*  7343ead (Use gpg library method to determine gpg exe path)
* 37357e9 (Add ability for gpg encryption to use the Windows gpg4win installed gpg.exe if it exists)

 If they're not needed to update the framework to 4.7.2, can you remove them from this pull request?

BlueBlock:
I reverted those 2.  Thanks.

warwickmm:
It looks like both AppVeyor and Travis are showing build failures.  Can you take a look and see what the cause may be?  Is it possible that these changes need to occur alongside the nuget package updates?

BlueBlock:
I setup both AppVeyor and Travis on my account and they build clean all together and all unit tests past; except for the last Travis gui test which looks like it requires a paid saucelabs account.

I separated them as suggested but then they might have dependencies on each other. I think the various commits I've made are dependent upon each other; before thinking I'll need to separate them.

BlueBlock:
That error says a nuget package is missing on the server. The AppVeyor server might have had a problem downloading the nuget package.

warwickmm:
I checked out this branch and have the same build issues that Travis is having.

I'm not certain, but it looks like the commits in this pull request depend on those in #3827.  The changes made here to `Duplicati/UnitTest/Duplicati.UnitTest.csproj` reference NUnit 3.12, which is updated in #3827 but not here.

Perhaps it's easier to recombine this pull request with #3827?  It's understandable to upgrade both the nuget packages and the framework version simultaneously.

BlueBlock:
Moved these commits to #3827 and I'll close this PR.

Issue #3818, opened 2019-07-15
Make bundled GPG binaries a facultative component in Windows installer


Magissia:
- [X] I have searched open and closed issues for duplicates.

----------------------------------------

## Environment info

- **Duplicati version**: Not relevant
- **Operating system**: Windows
- **Backend**: Not relevant

## Description

GPG is already installed on the machine using GPG4Win installation package, Duplicati's Windows installation package installs another copy of GPG while it already exist.

## Steps to reproduce

1. Install GPG4Win package
2. Install Duplicati

- **Actual result**:
GPG binary is present twice and there are no switch in the installer to not install GPG binaries
- **Expected result**:
Being able to refuse duplicati's GPG binaries and let it find the ones I already installed from PATH

BlueBlock:
I've included this in my submission. If gpg4win is installed then it will be used instead of the included gpg version. Though I've also updated the included gpg version to the latest.

kenkendk:
Duplicati no longer bundles the GPG binaries. 

While it may be easier for some users to have it bundled, GPG is a security critical component and should be managed by the user.

PR #3830, opened 2019-07-20
Update GPG library and path method

BlueBlock:
Added method to determine whether to use the user installed gpg.exe or the included gpg.exe.


warwickmm:
This concerns issue #3818.

Duplicati/Library/Encryption/GPGEncryption.cs
```diff
@@ -272,5 +273,14 @@ private void Runner(object x)
             Utility.Utility.CopyStream((Stream)tmp[0], (Stream)tmp[1]);
             ((Stream)tmp[1]).Close();
         }
+
+        /// <summary>
+        /// Determines the path to the GPG exe
+        /// </summary>
+        public static string GetGpgExePath()
+        {
+            var gpg4win = System.IO.Path.GetFullPath(@"C:\Program Files (x86)\GnuPG\bin\gpg.exe");
```
warwickmm: It appears that this is likely the default install location, but can a user install GnuPG to a different path?  I'm not very familiar with Windows, but is there a way to ask the registry if this is installed and what the corresponding path is?

Duplicati/CommandLine/Program.cs
```diff
@@ -48,16 +48,29 @@ public static int RealMain(string[] args)
             FROM_COMMANDLINE = true;
             try
             {
-                //If we are on Windows, append the bundled "win-tools" programs to the search path
+                //If we are on Windows, append the gpg4win path and append the bundled "win-tools" programs to the search path
                 //We add it last, to allow the user to override with other versions
                 if (Platform.IsClientWindows)
                 {
+                    var ExternalGpgPath = @"C:\Program Files (x86)\GnuPG\bin";
+                    var gpg4win = System.IO.Path.Combine(ExternalGpgPath, "gpg.exe");
```
warwickmm: This path also appears in the `GetGpgExePath` method.  It would be nice if we can consolidate this by removing the duplicated code.

BlueBlock:
Gog4win doesn't give the user an option of the install path but of course that counts always change.  We could look at both a registry entry, the default path and the environment path,  but any portable install would be on the user to specify. As I think the is an option to override the gpg option for duplucati.

Of course the option of a system wide search for gpg.exe but that is intensive.

Duplicati/Library/Encryption/GPGEncryption.cs
BlueBlock: Latest commit will use the registry to check for the install.

Duplicati/CommandLine/Program.cs
BlueBlock: Latest commit refactors the code.

Duplicati/Library/Utility/WinTools.cs
```diff
@@ -0,0 +1,49 @@
...
+            if (!Platform.IsClientWindows) return null;
+
+            // return gpg4win if it exists
+            var location32 = Library.Utility.RegistryUtility.GetDataByValueName(@"SOFTWARE\WOW6432Node\GnuPG", "Install Directory");
+            var location64 = Library.Utility.RegistryUtility.GetDataByValueName(@"SOFTWARE\GnuPG", "Install Directory");
+            var gpg4winLocation = string.IsNullOrEmpty(location64) ? location32 : location64;
+
+            if (!string.IsNullOrEmpty(gpg4winLocation))
+            {
+                return string.IsNullOrEmpty(gpg4winLocation) ? null : System.IO.Path.Combine(gpg4winLocation, "bin", "gpg.exe");
```
warwickmm: Is this ternary expression redundant given the enclosing `if` condition?

Duplicati/Library/Utility/WinTools.cs
```diff
@@ -0,0 +1,49 @@
...
+{
+    public static class WinTools
+    {
+        public static string GetWindowsGpgExePath()
+        {
+            if (!Platform.IsClientWindows) return null;
+
+            // return gpg4win if it exists
+            var location32 = Library.Utility.RegistryUtility.GetDataByValueName(@"SOFTWARE\WOW6432Node\GnuPG", "Install Directory");
+            var location64 = Library.Utility.RegistryUtility.GetDataByValueName(@"SOFTWARE\GnuPG", "Install Directory");
```
warwickmm: If the user doesn't have permissions to read the registry key, will this throw an exception?  If so, should we handle this more gracefully?

https://docs.microsoft.com/en-us/dotnet/api/microsoft.win32.registrykey.opensubkey?view=netframework-4.8#Microsoft_Win32_RegistryKey_OpenSubKey_System_String_

Duplicati/Library/Utility/WinTools.cs
BlueBlock: Definitely. Good catch. Fixed.

Duplicati/Library/Utility/WinTools.cs
BlueBlock: Yes.. I need to make it ReadOnly.  And then I'll test this further on a vanilla-regular-user Win7/8/10 machines.

BlueBlock: This last commit makes sure it degrades gracefully. And if either gpg.exe is found the front-end already handles it and says the encryption method is not available.

Duplicati/Library/Encryption/GPGEncryption.cs
```diff
@@ -95,8 +98,8 @@ public class GPGEncryption : EncryptionBase
         /// <summary>
         /// The PGP program to use, should be with absolute path
         /// </summary>
-        private readonly string m_programpath = "gpg";
-
+        private string m_programpath { get; set; } = GetGpgProgramPath();
```
warwickmm: Is this called often (e.g., for each encrypt/decrypt operation), and is it potentially expensive?  Would it make sense to use a Lazy initializer for this?

warwickmm:
Is anyone with access to a Windows machine able to help test these changes?

Duplicati/Library/Encryption/GPGEncryption.cs
BlueBlock: The check is fast and we would want to handle the case when a user installs GPG while Duplicati is running.

kenkendk:
It looks fine to me. I am a bit worried that `Library.Utility` now depends on `Library.Common`, because I am not sure the assembly redirects are configured for this. However, I think the benefits from this PR outweighs any potential downsides.

PR #3832, opened 2019-07-21
Auto compact interval

drwtsn32x:
This adds a customizable delay between automatic compactions.

Reason for change: On a couple of my systems the compaction effort takes a long time. Specifically the GetWastedSpaceReport SQL query that takes over 1 hour on my NAS (it has a lot of data and not a very powerful CPU).  I thought it might be useful to have auto compaction run only once a week on these systems.

High level overview of changes:

Runner.cs - added code so that Last Compaction time is stored in the main database (metadata)
Controller.cs - overrides no-auto-compact flag temporarily (sets it to "true") if auto-compact-interval has not been satisfied

It works as expected on two of my test systems.

Feedback appreciated. Not sure if Controller.cs Backup() is the best place to do this check, or if temporarily overriding no-auto-compact is the best approach.

### DIFF
--- a/Duplicati/Library/Main/Controller.cs
+++ b/Duplicati/Library/Main/Controller.cs
@@ -83,6 +83,11 @@ namespace Duplicati.Library.Main
         private ControllerMultiLogTarget m_logTarget;
 
         /// <summary>
+        /// Time of last compact operation
+        /// </summary>
+        private DateTime m_lastcompact;
+
+        /// <summary>
         /// Constructs a new interface for performing backup and restore operations
         /// </summary>
         /// <param name="backend">The url for the backend to use</param>
@@ -111,7 +116,13 @@ namespace Duplicati.Library.Main
             Library.UsageReporter.Reporter.Report("USE_BACKEND", new Library.Utility.Uri(m_backend).Scheme);
             Library.UsageReporter.Reporter.Report("USE_COMPRESSION", m_options.CompressionModule);
             Library.UsageReporter.Reporter.Report("USE_ENCRYPTION", m_options.EncryptionModule);
-            
+
+            if (!m_options.NoAutoCompact && (m_lastcompact > DateTime.MinValue) && (m_lastcompact.Add(m_options.AutoCompactInterval) > DateTime.Now))
+            {
+                Logging.Log.WriteInformationMessage(LOGTAG, "CompactResults", "AutoCompactInterval not satisfied, skipping auto compaction for this backup");
+                m_options.RawOptions["no-auto-compact"] = "true";
+            }
+
             return RunAction(new BackupResults(), ref inputsources, ref filter, (result) => {
 
                 using (var h = new Operation.BackupHandler(m_backend, m_options, result))
@@ -1095,6 +1106,12 @@ namespace Duplicati.Library.Main
         {
             get { return m_options.MaxDownloadPrSecond; }
             set { m_options.MaxDownloadPrSecond = value; }
+        }
+
+        public DateTime LastCompact
+        {
+            get { return m_lastcompact; }
+            set { m_lastcompact = value; }
         }
 
         #region IDisposable Members
--- a/Duplicati/Library/Main/Controller.cs
+++ b/Duplicati/Library/Main/Controller.cs
@@ -117,7 +117,7 @@ namespace Duplicati.Library.Main
             Library.UsageReporter.Reporter.Report("USE_COMPRESSION", m_options.CompressionModule);
             Library.UsageReporter.Reporter.Report("USE_ENCRYPTION", m_options.EncryptionModule);
 
-            if (!m_options.NoAutoCompact && (m_lastcompact > DateTime.MinValue) && (m_lastcompact.Add(m_options.AutoCompactInterval) >= DateTime.Now))
+            if (!m_options.NoAutoCompact && (m_lastcompact > DateTime.MinValue) && (m_lastcompact.Add(m_options.AutoCompactInterval) > DateTime.Now))
             {
                 Logging.Log.WriteInformationMessage(LOGTAG, "CompactResults", "Skipping auto compaction until {0}", m_lastcompact.Add(m_options.AutoCompactInterval));
                 m_options.RawOptions["no-auto-compact"] = "true";
--- a/Duplicati/Library/Main/Controller.cs
+++ b/Duplicati/Library/Main/Controller.cs
@@ -117,9 +117,9 @@ namespace Duplicati.Library.Main
             Library.UsageReporter.Reporter.Report("USE_COMPRESSION", m_options.CompressionModule);
             Library.UsageReporter.Reporter.Report("USE_ENCRYPTION", m_options.EncryptionModule);
 
-            if (!m_options.NoAutoCompact && (m_lastcompact > DateTime.MinValue) && (m_lastcompact.Add(m_options.AutoCompactInterval) > DateTime.Now))
+            if (!m_options.NoAutoCompact && (m_lastcompact > DateTime.MinValue) && (m_lastcompact.Add(m_options.AutoCompactInterval) >= DateTime.Now))
             {
-                Logging.Log.WriteInformationMessage(LOGTAG, "CompactResults", "AutoCompactInterval not satisfied, skipping auto compaction for this backup");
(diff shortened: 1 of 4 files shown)


Duplicati/Library/Main/Controller.cs
```diff
@@ -1097,6 +1108,12 @@ public long MaxDownloadSpeed
             set { m_options.MaxDownloadPrSecond = value; }
         }
 
+        public DateTime LastCompact
+        {
+            get { return m_lastcompact; }
+            set { m_lastcompact = value; }
+        }
```
warwickmm: This isn't very critical, but can we convert this to an auto-implemented property and remove the explicit `m_lastcompact`
backing field?  The behavior is the same, but it would clean up the code a very tiny bit.

drwtsn32x: Done... let me know if that's what you were looking for.  Thanks for the suggestion!

warwickmm: :+1: Thanks!

drwtsn32x: If you think this approach is ok, I am planning on doing similar with an auto vacuum interval. Was waiting to see what you guys thought.

Duplicati/Library/Main/Controller.cs
```diff
@@ -111,7 +111,13 @@ public Duplicati.Library.Interface.IBackupResults Backup(string[] inputsources,
             Library.UsageReporter.Reporter.Report("USE_BACKEND", new Library.Utility.Uri(m_backend).Scheme);
             Library.UsageReporter.Reporter.Report("USE_COMPRESSION", m_options.CompressionModule);
             Library.UsageReporter.Reporter.Report("USE_ENCRYPTION", m_options.EncryptionModule);
-            
+
+            if (!m_options.NoAutoCompact && (LastCompact > DateTime.MinValue) && (LastCompact.Add(m_options.AutoCompactInterval) > DateTime.Now))
+            {
+                Logging.Log.WriteInformationMessage(LOGTAG, "CompactResults", "Skipping auto compaction until {0}", LastCompact.Add(m_options.AutoCompactInterval));
+                m_options.RawOptions["no-auto-compact"] = "true";
```
warwickmm: It feels a bit strange to be toggling the `no-auto-compact` option to achieve the desired behavior.  Would it make more sense to modify the [BackupHandler.CompactIfRequired](https://github.com/duplicati/duplicati/blob/master/Duplicati/Library/Main/Operation/BackupHandler.cs#L243) method?  That way, the logic that determines when the delete or compact operations occur reside in the same place.

Duplicati/Server/Runner.cs
```diff
@@ -747,6 +758,9 @@ private static void UpdateMetadata(Duplicati.Server.Serialization.Interface.IBac
                 backup.Metadata["LastBackupFinished"] = Library.Utility.Utility.SerializeDateTime(result.EndTime.ToUniversalTime());
                 backup.Metadata["LastBackupDuration"] = r.Duration.ToString();
 
+                if (r.CompactResults != null)
+                    UpdateMetadata(backup, r.CompactResults);
```
warwickmm: Does this result in some metadata (e.g., `LastBackupFinished`) being written twice to the database?  For example, we set `LastBackupFinished` a few lines above.  Then, if `r.CompactResults` is not `null`, we update the metadata.  Once that is done, we continue along here and end up updating the metadata again.

Would it suffice to just set the new metadata (`LastCompactDuration`, etc.) here in this `if` statement?

Duplicati/Library/Main/Controller.cs
drwtsn32x: At first I was going to put it there, but then I saw that the DeleteHandler DoRun may also call Compact.  Intent of this change was not to defer the Deletion phase - just the Compaction phase.

So I was hunting for a single location where I could put the check.

I think there was a second reason that CompactIfRequired wasn't ideal but I cannot recall at the moment.

Duplicati/Server/Runner.cs
drwtsn32x: Nope, it is not written to the database twice.  The reason I did it this way was so that also an intentional Compact only operation would update the LastCompact time metadata.

For an intentional Compact operation, ICompactResults type is passed to UpdateMetadata.  But for an automatic compaction, it is the parent type IBackupResults that is passed. (presumably because a Backup operation was the main one kicked off).

I was mimicking some previous code in UpdateMetadata that seemed to use the same approach.

Duplicati/Library/Main/Controller.cs
warwickmm: Is the `CompactHandler.DoCompact` method a feasible location?

Duplicati/Server/Runner.cs
warwickmm: Sorry, I'm not that familiar with this part of the code.  I haven't looked at it too closely, but it still seems to me that `SetMetadata` at the end of this method can be invoked twice in the case where `IBackupResults` is provided, and `r.CompactResults` is not `null`.

The call to `UpdateMetadata` in the `IBackendStatsticsReporter` case above calls a different overload that simply adds entries to `backup.Metadata`, but doesn't actually insert the entries into the database.  Was that the behavior you were trying to mimic?

drwtsn32x: Yes I saw the one that calls the different overload, but I thought the second one did not. But now looking more closely I see it also calls the different overload.  So yeah good catch!

Because I need to update stats for either ICompactResults (intentional compact) or IBackupResults.CompactResults != nil (auto compact), I could expand the If test higher up instead of calling the function again. It should have the same end result.

I will go ahead and revise....

Duplicati/Library/Main/Controller.cs
drwtsn32x: Problem with that spot is we get there for intentional compact OR auto compact.  Need to be able to differentiate how we got there.  Also, we would have to make that determination even before db.GetCompactReport (first line of function) because that is what ultimately calls the GetWastedSpaceReport which is the root reason for this change - that WastedSpace report is what takes a long time on some systems.

That's ultimately why I thought temporarily flipping the no-auto-compact option near the start of the entire backup job was a decent approach.

What do you think?

warwickmm: Hmmm...what are the different scenarios where compact can occur, and which ones are you trying to skip with this option?  I can think of 
* Manually run (via command-line or UI)
* Backup operation
* Delete operation
* Purge operation

drwtsn32x: Those are the only three that I know of, and the intent of this change was to defer ANY automatic compaction so they don't happen too frequently.  Intentional compact operation will still run.

The alternative to flipping the m_options value temporarily would be to add an additional check to see if it should be deferred in all three of these locations. I could try to rework it that way if you think it would be better.

drwtsn32x: Not sure if this will affect your opinion, but thought I'd mention in case you weren't aware.... the use of RawOptions to temporarily change a flag is not unprecedented. In fact I got the idea from a usage elsewhere in the code.  See `PurgeBrokenFilesHandler.cs`  (in fact they change the same no-auto-compact flag there)

So maybe it's not so bad to use this method? ;)

warwickmm: Yes, I just saw that as well.  I think what you have here may be the cleanest implementation. :+1: 

One thing I would eventually like to pursue is to rewrite how we handle all the options.  For example, it would be nice if we can hide the internal `Dictionary` and avoid knowledge of the underlying strings elsewhere in the code.

Duplicati/Server/Runner.cs
drwtsn32x: Ok I have reworked this a bit.

It would have been messy to add another overload to `UpdateMetadata` that takes an `ICompactResults` object.  Doing so would have resulted in an intentional compact operation calling the new overload directly, so the Metadata would not actually be written to the database (unless I copied that code from the main `UpdateMetadata` function).

Instead I added a new `UpdateLastCompactMetadata` function and then call it two different ways from the main `UpdateMetadata` function.  One way is used when an intentional compact operation is run, the other is used when compact happened automatically after a backup.

The end result is the same as it was before, but now `UpdateMetadata` doesn't call itself.

Do you like this better?

warwickmm: Looks good to me.  Thanks!

Issue #2506, opened 2017-05-27
Expected there to be a temporary fileset for synthetic filelist


RubbeeDuck:
I have:
- [x] searched open and closed issues for duplicates

----------------------------------------

### Version info

**Duplicati Version:** 2.0.1.56
**Operating System:** QNAP - Linux
**Backend:** Linux

### Bug Description 

I got this error when I run an Backup. 

### Steps to reproduce
- Starr a Backup

**Actual result:** Describe here what happens after you run the steps above (i.e. the buggy behaviour)
**Expected result:** Describe here what should happen after you run the steps above (i.e. what would be the correct behaviour)

### debug log
DeletedFiles: 0
DeletedFolders: 0
ModifiedFiles: 1321
ExaminedFiles: 2792
OpenedFiles: 1395
AddedFiles: 74
SizeOfModifiedFiles: 102369729
SizeOfAddedFiles: 267267633
SizeOfExaminedFiles: 15278576024
SizeOfOpenedFiles: 8032101584
NotProcessedFiles: 0
AddedFolders: 0
TooLargeFiles: 0
FilesWithError: 0
ModifiedFolders: 0
ModifiedSymlinks: 0
AddedSymlinks: 0
DeletedSymlinks: 0
PartialBackup: False
Dryrun: False
MainOperation: Backup
CompactResults: null
DeleteResults: null
RepairResults: null
TestResults:
    MainOperation: Test
    Verifications: [
        Key: duplicati-20170104T135943Z.dlist.zip.aes
        Value: [],
        Key: duplicati-i376bbec0de2d48b1a635bc1e92272b1f.dindex.zip.aes
        Value: [],
        Key: duplicati-bba40cac5e8e04492b714bdd38f512ba1.dblock.zip.aes
        Value: []
    ]
    ParsedResult: Success
    EndTime: 5/26/2017 11:12:49 PM
    BeginTime: 5/26/2017 11:12:09 PM
    Duration: 00:00:40.0580100
    BackendStatistics:
        RemoteCalls: 10
        BytesUploaded: 146015489
        BytesDownloaded: 105765895
        FilesUploaded: 5
        FilesDownloaded: 3
        FilesDeleted: 0
        FoldersCreated: 0
        RetryAttempts: 0
        UnknownFileSize: 0
        UnknownFileCount: 0
        KnownFileCount: 311
        KnownFileSize: 14579977579
        LastBackupDate: 5/27/2017 1:00:00 AM
        BackupListCount: 15
[...]

PR #3836, opened 2019-07-23
Fix 'stop after current file'

BlueBlock:
The current 'stop after upload' fails to actually stop the backup process. This PR changes the operation name to 'stop after the current file' and correctly stops the backup operation after the current file backup is complete.

A complimentary PR 'Fix pausing and stopping after upload' (#3712) will fix the 'stop now' as it stops the backup operation in the middle of the current file.

Together, these two PR's will resolve and offer the ability to stop and pause a backup cleanly.


BlueBlock:
Travis broken test should be rerun.

seantempleton:
One thing I noticed while testing both ways of stopping was that if I had a large amount of files in my source folder, hit stop after current file, then when it stopped I would get a warning for every single file that it failed to process the metadata for. I can see the possibility of that being useful as you could know what files were not backed up, though it's not very obvious from the warning that the user could know that. I would assume most people would be scared by it.

The Output channel in the FileEnumerationProcess doesn't get closed so then MetadataPreProcess continues in it's infinite loop and fails on every Output.WriteAsync, as it's channel is retired, and logs it.

BlueBlock:
> One thing I noticed while testing both ways of stopping was that if I had a large amount of files in my source folder, hit stop after current file, then when it stopped I would get a warning for every single file that it failed to process the metadata for. I can see the possibility of that being useful as you could know what files were not backed up, though it's not very obvious from the warning that the user could know that. I would assume most people would be scared by it.
> 
> The Output channel in the FileEnumerationProcess doesn't get closed so then MetadataPreProcess continues in it's infinite loop and fails on every Output.WriteAsync, as it's channel is retired, and logs it.

This was corrected in the last commit regarding FileEnum.

When you test, search the for "cancel" which marks the point the users cancels. All files listed in the log following that point are the FileEnum process still running until it hits a point when it checks for the cancel token.

seantempleton:
> This was corrected in the last commit regarding FileEnum.

Ah, excellent. Sorry I didn't check that there was an update first.

BlueBlock:
> > This was corrected in the last commit regarding FileEnum.
> 
> Ah, excellent. Sorry I didn't check that there was an update first.

I need to remember to place a comment after committing to a PR so everyone knows there was a change.

warwickmm:
The changes to the translations will need to be fixed in Transifex (https://www.transifex.com/duplicati/duplicati/translate/).  The `.po` files are auto-generated, so the modifications to them here will get overwritten the next time the translations are generated.  Can you revert the changes to the `.po` files here?

BlueBlock:
Ah.. ok. Thanks.

BlueBlock:
Should I update the phrase in the single .pot file though?

warwickmm:
I started running some tests and noticed one thing.  If I "stop after current file" during the initial backup, the operation does stop after the current uploads finish.  The backup job correctly lists "0 bytes / 0 versions"
image

but the restore menu indicates that there are some files (but not all the source files) that can be restored:

image

If I then run another backup and allow it to finish, the backup job lists 2 versions:
image

and the restore menu indicates that there are 2 versions that can be restored from.  One version is incomplete (from the interrupted initial backup like above), and the second version is complete:
image

This leads me to a few questions:
1. Is this behavior desired?
2. Will the existence of the incomplete backup version mess with retention settings?  In other words, is it possible that retention settings will discard a complete backup in favor of an incomplete one?

BlueBlock:
The first backup is incomplete as expected but remains because while it is incomplete, it is still a valid backup since those files were successful. Removing the backup completely when cancelled doesn't seem desirable since the backup can essentially now be resumed just be running it again. 

For retention, those files were backed up so I would expect ant retention rules to apply just as they would for any other backup.

The 0 doesn't seem correct and I've not encountered that for a local file destination.

BlueBlock:
UpdateStorageStatsFromDatabase() or PostBackupVerification() is called when a backup ends completely. They were being skipped when a backup was cancelled. The stats will then display correctly.

This wasn't an issue previously since the "stop" wasn't working and just let the backup completely finish.

warwickmm:
I'm now quite confused, and a little concerned that people use "stop after upload" with different intentions. For example, if one interrupts the backup during the first of thousands of files, most likely they wouldn't want this version to be considered for retention.  But do they want this first file to be immediately restorable?  Or do they think this is a long pause, where the next backup simply resumes the interrupted backup?

I've asked this in the forums with the hope of getting some more clarity on this.

https://forum.duplicati.com/t/stop-after-upload/3753/25

BlueBlock:
> I'm now quite confused, and a little concerned that people use "stop after upload" with different intentions. For example, if one interrupts the backup during the first of thousands of files, most likely they wouldn't want this version to be considered for retention. But do they want this first file to be immediately restorable? Or do they think this is a long pause, where the next backup simply resumes the interrupted backup?
> 
> I've asked this in the forums with the hope of getting some more clarity on this.
> 
> https://forum.duplicati.com/t/stop-after-upload/3753/25

The phrase is changed to "Stop after the current file". The backup would be a valid partial backup after the current file has completed. 

The "Stop now" will is being handled by another devs PR.

warwickmm:
My original thought was that both "Stop after the current file" and "Stop now" should behave like a "Pause", except that it allows for one to perform other tasks before resuming.  For example one can stop a huge backup in favor of running a smaller faster backup, and later resume the huge backup while avoiding re-uploading blocks that were already uploaded.  None of these would record a new backup version for a partial backup.

My fear with recording a backup version for an incomplete backup is the following:
* A user's retention settings indicates that one backup version should be kept each month
* A backup operation runs every day of the month
* On the last day of the month, the user stops the backup for some reason
* When the retention settings take effect, the last (partial) backup is kept, and no complete backups remain for that month.

Perhaps my interpretation is wrong, but I don't think partial backups should be recorded as a valid version.  I don't see a scenario where a user is ok with a backup version containing only a subset of the source files.  Is there a different use-case that I am not seeing?

BlueBlock:
Pausing is good to have but there must be a way for a user to stop/interrupt the backup cleanly. In fact we should probably be listening for a system shutdown event and then begin stopping the job.

Partial backups could be recorded as partial and the retention policies could take then that into account.

And on the Restore list we could also place "(partial)" to the right of the datetime.

ts678:
FWIW [Stop after Upload](https://forum.duplicati.com/t/stop-after-upload/3753/30?u=ts678) is my (partly code-level) comment on the revived forum discussion on stop.

BlueBlock:
> FWIW [Stop after Upload](https://forum.duplicati.com/t/stop-after-upload/3753/30?u=ts678) is my (partly code-level) comment on the revived forum discussion on stop.

Does this functionality currently work? Is the stop-now supposed to allow a backup to be interrupted and later continued? When I test this I keep getting additional backups added to the restore dropdown and they don't seem to resume a previous backup.

ts678:
From the code section there (read further down), I think the intent was for an interrupted backup to be later continued (not automatically, but on the next run), starting with verification and repair if required -- user might be asked), doing the synthetic filelist (except a bug is blocking getting into that), then on to usual backup. I might misread something. Can you read up from here instead of down from the other:

https://github.com/duplicati/duplicati/blob/4e1f94bd3479ec10ef812f005266c70d3c6da37b/Duplicati/Library/Main/Operation/BackupHandler.cs#L476-L478

[How does Duplicati Resume After Interruption?](https://forum.duplicati.com/t/how-does-duplicati-resume-after-interruption/724/3?u=ts678) talks about interruptions, and I assume it includes stop.

I forget which variety (if any) of stop works. I generally avoid them because I know it's a trouble point...

You could maybe turn on some modest level of logging (live or file) to see what path you're taking, e.g.

https://github.com/duplicati/duplicati/blob/4e1f94bd3479ec10ef812f005266c70d3c6da37b/Duplicati/Library/Main/Operation/Backup/UploadSyntheticFilelist.cs#L73

Above will be one way to tell if your dropdown item was a synthetic filelist (despite the earlier bug -- or maybe bug means it WILL be bad -- I'm not sure how well it works even if it somehow avoids that [bug](https://github.com/duplicati/duplicati/issues/2506#issuecomment-463622742)).

Later on maybe I'll try some testing (does it need this code or will regular canary do?) and study forum.

BlueBlock:
Thanks for the info.

It seems like the synthetic features purpose is more of a recovery option for when a backup job is in a bad state. I think that can be useful when a crash of Duplicati or the OS occurs and an attempt can be made to repair or close-out the cleanly. Currently there isn't a way to actually stop a backup cleanly.. the current method of killing the thread is not a good option.

So while I do think the synthetic piece will be useful when it is ready, I think a feature for the user to be able to stop a backup is needed.

So this is what I'm thinking. Partial backups (backups stopped cleanly by the user or by the OS) will be marked as "partial". The retention policy will not remove delete partial backups that precede a full backup since essentially the full-backup+preceding-backups equal an entire completed full backup.

The Restore dropdown will identify the partial backups:
Capture

BlueBlock:
Note, the job database will be upgraded and have a new column added to fileset. Use a new backup job if you don't want an existing one upgraded while testing.

warwickmm:
It looks like some tests are failing.  Can you take a look when you get a chance?

BlueBlock:
I think the codacy issue can be ignored.  It complains about the sql but the sql is correct.

warwickmm:
I've done some initial testing and things are looking good so far.  :+1: 

I did notice something that was unexpected for me.   Consider the following scenario:
1. Configure a backup with "Keep a specific number of backups" set to 1.
2. Run a full backup.
3. Modify a file, run a backup, and click "Stop after current file".  As expected, there are 2 backup versions, and the restore list indicates that the most recent version is a partial backup.
4. Modify a file and run a backup.  Note that there are 2 versions (the partial one from Step 3, and the one that just completed).  The initial backup from Step 2 has been removed.

I was expecting that the partial backup from Step 3 would have been removed as well.  Is my intuition incorrect here?

BlueBlock:
> I've done some initial testing and things are looking good so far. 👍
> 
> I did notice something that was unexpected for me. Consider the following scenario:
> 
> 1. Configure a backup with "Keep a specific number of backups" set to 1.
> 2. Run a full backup.
> 3. Modify a file, run a backup, and click "Stop after current file".  As expected, there are 2 backup versions, and the restore list indicates that the most recent version is a partial backup.
> 4. Modify a file and run a backup.  Note that there are 2 versions (the partial one from Step 3, and the one that just completed).  The initial backup from Step 2 has been removed.
> 
> I was expecting that the partial backup from Step 3 would have been removed as well. Is my intuition incorrect here?

It correctly removed backup #1.

So you're scenario is this:
3) full backup
2) partial backup
1) full backup

#1 is a full and standalone complete backup. While the next full backup is actually the combined backup of #2 and #3.

To restore, you would restore #2 and #3.

Keeping in mind the purpose of the partial backup is to remove the need of the full backup to backup those files already in the partial backup.

Though maybe I'm seeing a possible improvement though I'm not sure of the complexity behind it. What if full backup #3 were to incorporate the previous partial backup(s)? so the next full backup would cause the partials to not display. Possibly combing those backups into a single backup within the database. when the full backup completes.

warwickmm:
> So it really doesn't make sense to display any partials expect those that are immediate, until a full backup is made.

That makes sense.  Once a full backup is completed, I think the previous partial is somewhat redundant.

With the most recent changes, I did notice an inconsistency between the number of versions displayed on the main page and the number of versions in the restore list.

1. Configure a backup with "Keep a specific number of backups" set to 1.
2. Run a full backup.
3. Modify a file, run a backup, and click "Stop after current file".  At this point, the main page shows 2 versions, and the restore list shows 2 versions.
4. Modify a file and run a full backup. Now, the main page still shows 2 versions, while the restore list shows only 1 version available.

Are you able to reproduce this?

Thanks for your patience.  It seems that time for reviewing pull requests is a bit scarce these days.

Duplicati/Library/Main/Database/LocalListDatabase.cs
```diff
@@ -367,13 +373,22 @@ public IEnumerable<IFileset> QuickSets
                                 @"SELECT DISTINCT ""ID"", ""IsFullBackup"" FROM ""Fileset"" ORDER BY ""Timestamp"" DESC ")
                         )
                         {
+                            // partial backup sets are only included until the first full backup is encountered
+                            var isFullBackupEncountered = false;
                             while (rd.Read())
                             {
                                 var id = rd.GetInt64(0);
-                                var isFullBackup = rd.GetInt32(1);
+                                var isFullBackup = rd.GetInt32(BackupType.FULL_BACKUP);
```
warwickmm: To me, the name `isFullBackup` seems more fitting for a `bool` instead of an `int`.  Would it be clearer to define this as
```
var isFullBackup = rd.GetInt32(BackupType.FULL_BACKUP) == BackupType.FULL_BACKUP;
```

warwickmm: Is the argument provided to `rd.GetInt32` supposed to be the ordinal of the column we are accessing?  Given the above query, I think this should be `rd.GetInt32(1)`, and the resulting value should be compared to `BackupType.FULL_BACKUP`.  Although this works since `BackupType.FULL_BACKUP = 1`, I don't think we should be providing `BackupType.FULL_BACKUP` as the argument to `rd.GetInt32`.

Together with the above comment, we would have something like
```
var isFullBackup = rd.GetInt32(1) == BackupType.FULL_BACKUP;
```

BlueBlock: Good catch on the "1". 

It could be a bool but I need the value when creating the Fileset().

BlueBlock:
> With the most recent changes, I did notice an inconsistency between the number of versions displayed on the main page and the number of versions in the restore list.
> Are you able to reproduce this?

I can thanks. So they should match.  The count be the versions of full backups + the displayed partial backups.
 
> Thanks for your patience. It seems that time for reviewing pull requests is a bit scarce these days.
No a problem!

Duplicati/Library/Main/Database/LocalListDatabase.cs
BlueBlock: I agree that an "is" should be a bool. I'll change the variable from isFullBackup to backupType

warwickmm:
It looks like some tests are failing. Can you take a look when you get a chance?

BlueBlock:
Two issues fixed. 

- A check as missing for the fileset ID's within the dict[], this also occurs within QuickSets.
- a purge unittest was expecting 4 filesets but now first 3 of the 4 are partials so the unittest should just expect 1 fileset now

BlueBlock:
PR #3712 is waiting on this PR to go through.

Duplicati/Library/Main/Operation/DeleteHandler.cs
```diff
@@ -173,45 +174,92 @@ where toDelete.Contains(n.Item2)
...
 
-            // Remove backups via rentention policy option
-            toDelete.AddRange(ApplyRetentionPolicy(allBackups));
+            // Remove backups via retention policy option
+            toDelete.AddRange(ApplyRetentionPolicy(db, allBackups));
 
-            // Check how many backups will be remaining after the previous steps
+            // Check how many full backups will be remaining after the previous steps
             // and remove oldest backups while there are still more backups than should be kept as specified via option
             var backupsRemaining = allBackups.Except(toDelete).ToList();
```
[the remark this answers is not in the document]
warwickmm: I think that's a good idea.  It's not clear in the `FilesetTimes` getter that some other code is depending on the ordering so an uninformed change to a SQL query could lead to disastrous deletion of backup sets.

Duplicati/Library/Main/Operation/DeleteHandler.cs
```diff
@@ -173,45 +174,92 @@ where toDelete.Contains(n.Item2)
...
             var keepVersions = m_options.KeepVersions;
-            if (keepVersions > 0 && keepVersions < backupsRemaining.Count())
-                toDelete.AddRange(backupsRemaining.Skip(keepVersions));
-
-            var toDeleteDistinct = toDelete.Distinct().OrderByDescending(x => x.ToUniversalTime()).AsEnumerable();
+            var versionsKeptCount = 0;
+            if (keepVersions > 0 && keepVersions < backupsRemaining.Count)
+            {
+                // skip full versions for the count of keepVersions
```
warwickmm: This comment is a little confusing to me.  Are we actually skipping partial versions when incrementing `versionsKeptCount`?

Duplicati/Library/Main/Operation/ListControlFilesHandler.cs
```diff
@@ -68,7 +68,7 @@ public void Run(IEnumerable<string> filterstrings = null, Library.Utility.IFilte
                                     if (Library.Utility.FilterExpression.Matches(filter, cf.Key))
                                         files.Add(new ListResultFile(cf.Key, null));
                             
-                            m_result.SetResult(new Library.Interface.IListResultFileset[] { new ListResultFileset(fileversion.Key, fileversion.Value.Time, -1, -1) }, files);
+                            m_result.SetResult(new Library.Interface.IListResultFileset[] { new ListResultFileset(fileversion.Key, 0, fileversion.Value.Time, -1, -1) }, files);
```
warwickmm: What's the meaning of `0` here?  Is it meant to represent `BackupType.PARTIAL_BACKUP`?

Duplicati/UnitTest/PurgeTesting.cs
```diff
@@ -146,7 +146,7 @@ public void PurgeTest()
                 Console.WriteLine("Files:");
                 Console.WriteLine("  " + string.Join(Environment.NewLine + "  ", files.Select(x => string.Format("{0}: {1}", x.Path, string.Join(" - ", x.Sizes.Select(y => y.ToString()))))));
 
-                Assert.AreEqual(4, filesets.Length, "Incorrect number of filesets after final backup");
+                Assert.AreEqual(1, filesets.Length, "Incorrect number of filesets after final backup");
```
warwickmm: Sorry, I'm not familiar with the expectations here.  Does this mean that a purge operation marks the existing backup sets as partial?

warwickmm:
@BlueBlock, thanks for your patience with this.  This is great work, but it's a lot of changes to review so it's taking quite some time.

Duplicati/Library/Main/Operation/DeleteHandler.cs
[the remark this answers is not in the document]
BlueBlock: Definitely. And you've made a -really- good find. Unintendended consequences abound from such assumptions in a method. Well done.

We're not programming pace makers but to anyone that loses data, well, the attention to detail is desired.

I really appreciate your attention to such detail which directly impacts the robustness of the code. In that it can then effectively correct future programming mistakes such as your example.

Duplicati/Library/Main/Operation/DeleteHandler.cs
BlueBlock: Partials are skipped and considered to be part of the proceeding full backup. They are to be considered together a complete backup.

Duplicati/Library/Main/Database/LocalDatabase.cs
```diff
@@ -1283,13 +1314,30 @@ public virtual long CreateFileset(long volumeid, DateTime timestamp, System.Data
...
             }
         }
-        
+
+        /// <summary>
+        /// Updates a backup operation and marks it as a completed full backup.
+        /// </summary>
+        /// <param name="filesetid">The ID of the fileset to update</param>
+        /// <param name="transaction">An optional external transaction</param>
+        public virtual long SetFilesetFullBackupComplete(long filesetid, System.Data.IDbTransaction transaction = null)
```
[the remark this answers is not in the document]
BlueBlock: It is not used. Now removed.

Duplicati/Library/Main/Operation/RestoreHandler.cs
```diff
@@ -18,8 +38,9 @@ internal class RestoreHandler
         private readonly string m_backendurl;
         private readonly Options m_options;
         private byte[] m_blockbuffer;
-        private readonly RestoreResults m_result;
+        private readonly RestoreResults m_result;
         private static readonly string DIRSEP = Util.DirectorySeparatorString;
+        public CancellationTokenSource cancellationTokenSource = new CancellationTokenSource();
```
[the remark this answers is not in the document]
BlueBlock: Unused and now removed.

Duplicati/Library/Main/Operation/ListControlFilesHandler.cs
BlueBlock: Correct and now fixed.

Duplicati/UnitTest/PurgeTesting.cs
BlueBlock: The partial backups are removed when the next most recent full backup is removed. The 4 was from when all the backups were not differentiated.

Duplicati/Library/Main/Operation/DeleteHandler.cs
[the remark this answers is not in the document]
BlueBlock: Change is in.  It now enforces the sorting.

BlueBlock:
> @BlueBlock, thanks for your patience with this. This is great work, but it's a lot of changes to review so it's taking quite some time.

Thanks. It is a lot of changes so thanks for reviewing it.

Duplicati/Library/Main/Operation/DeleteHandler.cs
warwickmm: That makes sense, but the confusing part to me is the comment says "skip _full_ versions"?

Duplicati/Library/Main/Operation/ListFilesHandler.cs
```diff
@@ -177,17 +193,17 @@ orderby n.Key
...
-            return filelistFilter(parsedlist).ToList();                
+            return filelistFilter(parsedlist).ToList();
         }
-        
+
         public static IEnumerable<Library.Interface.IListResultFileset> CreateResultSequence(IEnumerable<KeyValuePair<long, Volumes.IParsedVolume>> filteredList)
         {
-            return (from n in filteredList select (Library.Interface.IListResultFileset)(new ListResultFileset(n.Key, n.Value.Time.ToLocalTime(), -1, -1))).ToArray();
-        }        
+            return (from n in filteredList select (Library.Interface.IListResultFileset)(new ListResultFileset(n.Key, 0, n.Value.Time.ToLocalTime(), -1, -1))).ToArray();
```
warwickmm: Should this also be `BackupType.PARTIAL_BACKUP` instead of `0`?

Duplicati/Library/Main/Operation/DeleteHandler.cs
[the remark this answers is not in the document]
warwickmm: :+1:

Duplicati/UnitTest/PurgeTesting.cs
warwickmm: I'm not super familiar with this test, but why are partials appearing?  My understanding of a purge operation is that a user has decided that a file shouldn't have been included in the backup sets (e.g., for security reasons) and hence that file should be removed from all backup sets.  However, I think these are still considered by the user to be "full" backups.

Duplicati/Library/Main/Operation/ListFilesHandler.cs
warwickmm: Also, why are these considered "partial"?  Or, is it that the backup type doesn't matter and we have to choose _some_ value?

Sorry, I'm not that familiar with this part of the code...

Duplicati/Library/Main/Operation/DeleteHandler.cs
BlueBlock: keepVersion is the count of the full backups and does not include the partial backups. And backupsRemaining is also only the full backups. 

Partials don to get removed on their own, only when the matching full backup is removed.

Should I add "full" to the variables to make it more understandable?

Duplicati/UnitTest/PurgeTesting.cs
BlueBlock: The test is checking for the fileset count and is not looking at particular files in the fileset.

Duplicati/Library/Main/Operation/DeleteHandler.cs
warwickmm: > Should I add "full" to the variables to make it more understandable?

No, perhaps I am just confused.  At a quick glance, I was expecting the comment to be
```
// skip partial versions for the count of keepVersions
```
I will think about this some more...

BlueBlock: It's not skipping partials at all. It is determining which full backups to keep.

Duplicati/Library/Main/Operation/ListFilesHandler.cs
BlueBlock: > BackupType.PARTIAL_BACKUP

Yes.. it should be.

BlueBlock: > Also, why are these considered "partial"? Or, is it that the backup type doesn't matter and we have to choose _some_ value?
> 
> Sorry, I'm not that familiar with this part of the code...

It's the beginning for that object and the fileset so we default it to being a partial backup. It only becomes a full backup when the backup is completed without interruption.

Duplicati/Library/Main/Operation/DeleteHandler.cs
warwickmm: Ah, I get it now.  Thanks for the explanation.

Duplicati/Library/Main/Operation/ListFilesHandler.cs
warwickmm: Got it.  Thanks!

Duplicati/Library/Main/Operation/DeleteHandler.cs
BlueBlock: No problem. I think renaming the variables will help.

Duplicati/Library/Main/Database/LocalListDatabase.cs
```diff
@@ -373,24 +404,37 @@ public IEnumerable<IFileset> Sets
...
+                                var id = rd.GetInt64(0);
+                                if (!dict.ContainsKey(id))
+                                {
+                                    continue;
+                                }
+                                var isFullBackup = rd.GetInt32(1);
+                                var e = dict[id];
+
+                                var filecount = rd.ConvertValueToInt64(1, -1L);
+                                var filesizes = rd.ConvertValueToInt64(2, -1L);
```
warwickmm: Since we added the `IsFullBackup` column to the result set, should we increment these column indices?

Duplicati/UnitTest/PurgeTesting.cs
warwickmm: Forgive my confusion, but I'm still trying to understand this.  When running this test, I notice that
1. 3 backups are run.  After this, `c.List` shows 3 filesets and indicates that they are full backups.
2. A purge is run on a file that exists in all filesets.  After this, `c.List` shows 3 filesets and indicates that they are partial backups.

In step 2, I would have expected the result of `c.List` to have 3 filesets (each with one less file than previously), but still marked as full backups.  Is my intuition incorrect?  Why are these now considered partial after a purge?

Duplicati/Library/Main/Database/LocalListDatabase.cs
BlueBlock: Definitely. Good catch. Change is in.

Duplicati/UnitTest/PurgeTesting.cs
BlueBlock: I think you're right. They should be full backups. It looks like purge creates new filesets. I need to have it carry over the isFullBackup value when recreating a fileset.

BlueBlock: PurgeFilesHandler re-creates filesets the value of isFullBackup was then always set to partial. The value of isFullBackup is now carried over by PurgeFilesHandler.

I checked PurgeBroken and it is fine since it calls PurgeFilesHandler.

Commits are in.

Duplicati/Library/Main/Operation/PurgeFilesHandler.cs
```diff
@@ -123,7 +125,9 @@ private void DoRun(Database.LocalPurgeDatabase db, Library.Utility.IFilter filte
                                 break;
                         }
 
-                        var ts = filesets[ix].Value.AddSeconds(secs);
+                        var tsOriginal = filesets[ix].Value;
+                        var ts = tsOriginal.AddSeconds(secs);
```
warwickmm: These changes look ok to me (although I do have questions about the original logic that we have left intact).
1. Does this method assume that `filesets` and `versions` are ordered?  
2. Can the definition of `ts` here result in timestamps in the future?  If so, is that problematic?

Duplicati/Library/Main/Database/LocalDatabase.cs
```diff
@@ -1283,8 +1314,8 @@ public virtual long CreateFileset(long volumeid, DateTime timestamp, System.Data
             using (var cmd = m_connection.CreateCommand())
             using (var tr = new TemporaryTransactionWrapper(m_connection, transaction))
             {
-                cmd.Transaction = tr.Parent;                
-                var id = cmd.ExecuteScalarInt64(@"INSERT INTO ""Fileset"" (""OperationID"", ""Timestamp"", ""VolumeID"") VALUES (?, ?, ?); SELECT last_insert_rowid();", -1, m_operationid, NormalizeDateTimeToEpochSeconds(timestamp), volumeid);
+                cmd.Transaction = tr.Parent;
+                var id = cmd.ExecuteScalarInt64(@"INSERT INTO ""Fileset"" (""OperationID"", ""IsFullBackup"", ""Timestamp"", ""VolumeID"") VALUES (?, ?, ?, ?); SELECT last_insert_rowid();", -1, m_operationid, BackupType.PARTIAL_BACKUP, NormalizeDateTimeToEpochSeconds(timestamp), volumeid);
```
warwickmm: The `CreateFileset` method is also invoked by the `RecreateDatabaseHandler`.  Have we ensured that the `IsFullBackup` field is properly restored after a database recreate operation?

BlueBlock: I think it will currently default to partial. We can't know what the original state was since it isn't saved in the remote backup so it should default to full.

Duplicati/Library/Main/Operation/PurgeFilesHandler.cs
BlueBlock: 1. I think it does assume that.  Should I enforce it here too?
2. Not the future. It looks like it is expecting the new fileset to just not be the same timestamp as the original?

Duplicati/Library/Main/Database/LocalDatabase.cs
warwickmm: That's true, we cannot recreate the original state since the backup type is not preserved.  

However, I think this does lead to some strange behavior that users might not anticipate.  For example:
1. Create a full backup.
2. Create partial backup.
3. Recreate database (now both filesets are marked as a full backups)
4. Retention policy (or something else) removes the first backup, leaving only the partial backup. 

Also, are you able to reproduce the following?

1. Create a partial backup.  The main page shows 1 backup version, and the restore list shows 1 (partial) version.
2. Create a full backup.  The main page shows 1 backup version, and the restore list shows 1 (full) version.
3. Recreate the database.  The main page now shows 1 backup version, but the restore list shows 2 versions.

warwickmm:
I also noticed the following behavior.  Can you reproduce this?
1. Start a backup and click "Stop after current file."  Note that the progress bar displays "Stopping after current the file".
2. While the current file is still being uploaded, refresh the page.  The progress bar reverts to a normal state and does not indicate anymore that the backup will be stopped after the current file.

Duplicati/Library/Main/Operation/PurgeFilesHandler.cs
BlueBlock: I think the ts at most would get set to the current time. 

We can look at enforcing sorting within methods but this seems outside the scope of this PR.

BlueBlock:
> I also noticed the following behavior. Can you reproduce this?
> 
> 1. Start a backup and click "Stop after current file."  Note that the progress bar displays "Stopping after current the file".
> 2. While the current file is still being uploaded, refresh the page.  The progress bar reverts to a normal state and does not indicate anymore that the backup will be stopped after the current file.

It looks like it is a pre-existing issue in the master branch. Possibly relating to the state handling in angular.

warwickmm:
> It looks like it is a pre-existing issue in the master branch. Possibly relating to the state handling in angular.

Ok, let's leave that for a separate pull request.

warwickmm:
It looks like merging in `master` in revision c8bb889f6dfc438e5df6cd1f863dd0017a1d4e0e reverted some of your changes to the `CreateFileset` method in `LocalDatabase.cs`.  I think this is the cause of the AppVeyor and Travis failures.

Duplicati/Library/Main/Database/LocalDatabase.cs
warwickmm: I've finally made it through all the changes.  I'll start to run some more tests to see if anything unexpected reveals itself.  Thanks again for your patience @BlueBlock.

I do have a few remaining questions:
1. Any thoughts about the two situations described above?
2. Do you think this should be included in the next beta?  I believe this will address some of the causes of database corruption, but my only concern is the amount of testing given the large number of changes.  Given that the next beta is likely not _that_ soon, one approach is to create more canary releases to get these changes out there for users to test.

warwickmm:
Here's a question about user expectations.  Consider the following example:
1. Perform a full backup of files `a`, `b`, and `c`.  These files need to be large enough that their contents cannot all fit within a single volume (e.g., `b` is 100 MB and `c` is 250 MB).
2. Modify files `b` and `c`.
3. Start a backup and click on "Stop after current file".  Suppose that `b` is in progress (and will be allowed to finish), while the changes in `c` are not backed up.
4. In the restore list, the most recent backup is correctly marked as "partial".  Files `a` and `b` appear in the restore list, but file `c` does not appear in the list of files.

Do you think users would expect to see `c` in the restore list for this partial version?  In some sense `c` is similar to `a` in that no changes were backed up in this version.

BlueBlock:
> Do you think users would expect to see `c` in the restore list for this partial version? In some sense `c` is similar to `a` in that no changes were backed up in this version.

I would not expect the changes for 'c' for show since the backup was interrupted before 'c' could be processed.

Partial backups are backups that have not completed the processing of all files for a backup and will only show those files which have been processed. Partials are cumulative, so one can keep making partials and the displayed backup files will contain all files for the previous partials as well. The final full backup works the same way and is cumulative of all the partials and the final full backup.

Duplicati/Library/Main/Database/LocalDatabase.cs
BlueBlock: All those are a result of not having the state stored at the remote end. To avoid those scenarios it may be worthwhile to see what it would take to store the state though.. I'll have a look. Once that gets decided then I think we can include it in a canary then beta.

warwickmm: I believe the number of versions does sync up once the next backup is run...perhaps this is a minor issue?

ts678:
Thanks for the user-oriented explanation. This probably deserves some release notice or docs publicity. Does it have any impact on the currently-buggy synthetic filelist code mentioned [earlier in this thread](https://github.com/duplicati/duplicati/pull/3836#issuecomment-518381838)? [--disable-synthetic-filelist](https://duplicati.readthedocs.io/en/latest/06-advanced-options/#disable-synthetic-filelist) may be the only mention in the manual though it's been posted a few times.

> --disable-synthetic-filelist = false
> If Duplicati detects that the previous backup did not complete, it will generate a filelist that is a merge of the last completed backup and the contents that were uploaded in the incomplete backup session.

BlueBlock:
I haven't touched to synthetic filelist code. That would still come into play if a backup is stopped in an unclean way like a reboot or system crash.

We can take it to a forum thread but I'd be interested to know what the buggyness issues are with the synthetic filelist. And ways to reproduce those bugs.

BlueBlock:
To save state it looks like it would make sense to store it in the dlist manifest file.

warwickmm:
Testing has gone well so far.  I haven't encountered any issues yet.  Very nice job @BlueBlock.  :+1: 

Do you think saving the backup type in the remote files is a blocker, or it's something that can be addressed later?

BlueBlock:
> Testing has "throw new NotImplementedException();"gone well so far. I haven't encountered any issues yet. Very nice job @BlueBlock. 👍
> 
> Do you think saving the backup type in the remote files is a blocker, or it's something that can be addressed later?

I've been working on this part. I currently can save and retrieve IsFullBackup within the manifest. 

The only issue I currently have is that I'm at the point where because I'm adding a new interface method I need to add it to the 7zip lzma compression class. Is the 7zip compression currently usable or can I leave the new 7zip method with "throw new NotImplementedException();" ?

BlueBlock:
> Do you think saving the backup type in the remote files is a blocker, or it's something that can be addressed later?

Maybe to be addressed first.

BlueBlock:
It looks like adding to the manifest file will not work easily and would be unnecessarily complex since it is created at the start. We want to record the fileset state at the end of the backup.

Like file "manifest", I'm adding a file to dlist called "fileset".  This file for now will just contain the backup state IsFullBackup.  This file can be also used in the future for other data not accessible at the start of a backup.

Let me know if there is a better name than "fileset".

BlueBlock:
I'm running into a problem adding a file to the dlist.zip file and it might be an existing problem. There is the user option ability to add "extra" files to the dlist file and might not be used much if at all since it seems to me that trying to do so causes the filelist in the zip to be invalid.

As seen here when using the "control-files" to add a file:

Annotation 2019-09-10 101825

Annotation 2019-09-10 101705

BlueBlock:
I have this fixed along with 'control-files" The stream for filelist was kept open for writing, then 'control-files' were added to the zip, then the filelist was still being written and eventually closed. This made the filelist file in the zip split within the zip.

The solution is to not stream the filelist to the zip but to write to a memorystream and write to the zip once the filelist file is complete.

BlueBlock:
The commits are in for handling storing the IsFullBackup state hon the backend.

For example, taking 2 partial backups, then 1 full backup results in backend files in the image. Deleting the database and recreating will then show 3 filesets correctly set as 2 partial and 1 full.

Annotation 2019-09-10 124556

Annotation 2019-09-10 131539

BlueBlock:
@warwickmm  Can Travis be kicked off again?  Thanks.

warwickmm:
I restarted the Travis build.

BlueBlock:
> I restarted the Travis build.

Thanks. It looks good now.

warwickmm:
@BlueBlock, this looks good to me.  Testing has gone well.  Awesome job.  Can you think of anything else that we should do before I merge?

ts678:
> Partial backups could be recorded as partial and the retention policies could take then that into account. Partial backups would essentially belong to the next most-recent non-partial backup. When it is removed by retention then the partials are as well.

Is this how it was done? My outcome is different. First test deleted one-and-only full, and made its partials visible to run manual deletion. Second test with retention set to one version found first full removed by retention when second full was made, but partials leading to first full were left behind.

~~~
old
ID      OperationID     VolumeID        IsFullBackup    Timestamp
1       1               1               0               1571960139
2       7               10              0               1571960251
3       8               19              0               1571960329
4       12              28              0               1571960486
5       13              43              0               1571960642
6       14              52              0               1571960730
7       18              56              0               1571960789
8       19              65              0               1571960865
9       23              96              0               1571961237
10      63              187             1               1572011665

new
ID      OperationID     VolumeID        IsFullBackup    Timestamp
1       1               1               0               1571960139
2       7               10              0               1571960251
3       8               19              0               1571960329
4       12              28              0               1571960486
5       13              43              0               1571960642
6       14              52              0               1571960730
7       18              56              0               1571960789
8       19              65              0               1571960865
9       23              96              0               1571961237
11      64              226             1               1572012797
~~~

[add logic to Filesettimes for isFullBackup](https://github.com/duplicati/duplicati/commit/1cf8864193262dec127a4e55ed7decc13d41fd9b) is possibly where partials get made invisible by any later full. This may explain partials "unhide" in first test, and the "stay hidden" in second. It seemed time to ask. Hoping for help in ironing out any remaining issues like this that arise in canary test. Thanks so much.

warwickmm:
I think that was an initial suggestion that was abandoned for a different implementation.

It's difficult to follow the very long conversation above, but see the comments [here](https://github.com/duplicati/duplicati/pull/3836#issuecomment-518441912) and [here](https://github.com/duplicati/duplicati/pull/3836#issuecomment-522169136).

ts678:
I think that's just stating the usual way that a Duplicati backup version might build on all prior versions. Partials appear to follow this rule too, with subsequent partials adding only the changes from the prior.

One might claim that the partials outlived their usefulness once their full existed, and so should vanish. Possibly these should be deleted right then and there, otherwise they will probably hold blocks forever.

Having FilesetTimes shrink suddenly when the full finally happens (due to the hiding of the partials) will prevent access to the hidden partials, e.g. through --version in the UI, or even internally in some places.

Some other places will go directly to the DB via SQL, get a different view (without partials hidden), then get in an unsolvable bind when they need to work with code or user having access only to filtered view.

For one example of above, test the `affected` command on a dblock from a partial after its full is there. The result is it crashes in LocalListAffectedDatabase probably because the actual fileset ID of the partial can't map to the smaller range after hiding. Mapping back and forth can now no longer be counted on.

Another suspect is the infamous "Unexpected difference in fileset" (hopefully less common now) where code was added in [Improved error message for faults in database filelists](https://github.com/duplicati/duplicati/commit/2826dbe6f04497bcfdb88a9da96ea7e65338f055) to map into helpful --version instead of giving the internal fileset ID (used for loop [here](https://github.com/duplicati/duplicati/blob/f23bcdcac9dd2b068143c9ed06700c67ac798373/Duplicati/Library/Main/Database/LocalDatabase.cs#L851)). One can't map partial to external --version.

A suspected gap is the purge command, due to use of FilesetTimes [here](https://github.com/duplicati/duplicati/blob/master/Duplicati/Library/Main/Operation/PurgeFilesHandler.cs#L95).  Can this purge partial filesets?

What other gaps might there be? Having two different low-level views of what filesets exist seems risky.

warwickmm:
> One might claim that the partials outlived their usefulness once their full existed, and so should vanish.

That was my initial intuition as well, but it sounds like the partial backups may still play a role.  Perhaps @BlueBlock can provide some more insight?

ts678:
Permanent loss of destination storage used by partials that go hidden-but-really-there seems confirmed, testing with one-version retention, two partials and a full, big trimming of the source area, and a backup. Full version has 1 byte source. Destination has two partials too. Purge [errors out](https://github.com/duplicati/duplicati/blob/8015171be32eae3fa2fe9f85e6502d9c7b7e92e9/Duplicati/Library/Main/Operation/PurgeFilesHandler.cs#L121) from view inconsistency:

~~~
System.InvalidProgramException: Fileset was reported with id 1, but could not be found?
   at Duplicati.Library.Main.Operation.PurgeFilesHandler.DoRun(LocalPurgeDatabase db, IFilter filter, Action`3 filtercommand, Single pgoffset, Single pgspan)
   at Duplicati.Library.Main.Operation.PurgeFilesHandler.Run(IFilter filter)
   at Duplicati.Library.Main.Controller.RunAction[T](T result, String[]& paths, IFilter& filter, Action`1 method)
   at Duplicati.Library.Main.Controller.PurgeFiles(IFilter filter)
   at Duplicati.CommandLine.Commands.PurgeFiles(TextWriter outwriter, Action`1 setup, List`1 args, Dictionary`2 options, IFilter filter)
   at Duplicati.CommandLine.Program.ParseCommandLine(TextWriter outwriter, Action`1 setup, Boolean& verboseErrors, String[] args)
   at Duplicati.CommandLine.Program.RunCommandLine(TextWriter outwriter, TextWriter errwriter, Action`1 setup, String[] args)
Return code: 100
~~~

PR #3840, opened 2019-07-30
Add auto vacuum interval option

drwtsn32x:
Similar to https://github.com/duplicati/duplicati/pull/3832, this adds a customizable delay between automatic vacuum operations.

Reason for change: I like having auto vacuum enabled, but it can take a long time on some of my systems. Instead of turning off auto vacuum entirely I wanted to just limit how often it would run.

This uses the same approach as https://github.com/duplicati/duplicati/pull/3832 - code added to store LastVacuum information in metadata part of main database.  If the time interval has not been satisfied, the `auto-vacuum` option flag is overridden (set to "false") temporarily.

There were two situations where auto vacuum may occur that I wanted to be impacted by this code:  at end of a backup job, and after an explicit compact operation. An explicit vacuum operation is not affected (it will always run).

In a couple locations I changed a direct call to `db.Vacuum()` to instead use the VacuumHandler so that results could be properly captured so that the timestamp can be saved as metadata.  Also I renamed some members to be more consistent with other parts of the code.

### DIFF
--- a/Duplicati/Server/Runner.cs
+++ b/Duplicati/Server/Runner.cs
@@ -492,6 +492,12 @@ namespace Duplicati.Server
                     ((RunnerData)data).Controller = controller;
                     data.UpdateThrottleSpeed();
 
+                    if (backup.Metadata.ContainsKey("LastCompactFinished"))
+                        controller.LastCompact = Library.Utility.Utility.DeserializeDateTime(backup.Metadata["LastCompactFinished"]);
+
+                    if (backup.Metadata.ContainsKey("LastVacuumFinished"))
+                        controller.LastVacuum = Library.Utility.Utility.DeserializeDateTime(backup.Metadata["LastVacuumFinished"]);
+
                     switch (data.Operation)
                     {
                         case DuplicatiOperation.Backup:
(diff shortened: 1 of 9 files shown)


Duplicati/Server/Runner.cs
```diff
@@ -764,7 +789,10 @@ private static void UpdateMetadata(Duplicati.Server.Serialization.Interface.IBac
                 backup.Metadata["LastBackupDuration"] = r.Duration.ToString();
 
                 if (r.CompactResults != null)
-                    UpdateLastCompactMetadata(backup, r.CompactResults);
+                    UpdateMetadataLastCompact(backup, r.CompactResults);
+
+                if (r.VacuumResults != null)
+                    UpdateMetadataLastVacuum(backup, r.VacuumResults);
```
warwickmm: Is it possible for `r.CompactResults.VacuumResults` to be non-null here?

As an aside, I would like to eventually try and change how we deal with all the different results.  Instead of results containing other results, perhaps a list or tree of results can be processed in a more streamlined manner.

drwtsn32x: Nope - it will always be null.  When a backup triggers a compact operation, compact will never then trigger a vacuum.  I dug into that because at first I thought there was the possibility of a double vacuum under normal operation, but there is not.

An automatic compact operation after backup triggers `CompactHandler.DoCompact` which does not call any sort of vacuum operation.

An intentional compact operation triggers `CompactHandler.Run`, which in turn calls `CompactHandler.DoCompact` and then triggers a vacuum.

Took me a little while to wrap my head around all this.

PR #3846, opened 2019-08-02
fix bug in GetLargestPrefix() which adds an ending slash to file name

BlueBlock:
Do not apply AppendDirSeparator() to files.

For the section of code which handles files, AppendDirSeparator() is being applied to files when it should n't be at all. This results in all file paths to have an ending slash. This adjusted path causes the sql query to have 0 results.

For example:
GetLargestPrefix() would change D:\\myfiles\\photo1.jpg to be D:\\myfiles\\photo1.jpg\\

When the adjusted path is searched in the database it has 0 hits.

When debugging this can be tested by hitting the Restore and getting the file list for an existing backup job in the GUI.

### DIFF
--- a/Duplicati/Library/Main/Database/LocalListDatabase.cs
+++ b/Duplicati/Library/Main/Database/LocalListDatabase.cs
@@ -163,9 +163,8 @@ namespace Duplicati.Library.Main.Database
     
                     while (filecount != foundfiles && maxpath.Length > 0)
                     {
-                        var mp = Platform.IsClientPosix ? Util.AppendDirSeparator(maxpath, dirsep) : maxpath;
-                        cmd.SetParameterValue(0, mp.Length);
-                        cmd.SetParameterValue(1, mp);
+                        cmd.SetParameterValue(0, maxpath.Length);
+                        cmd.SetParameterValue(1, maxpath);
                         
                         foundfiles = cmd.ExecuteScalarInt64(0);
 
--- a/Duplicati/Library/Main/Database/LocalListDatabase.cs
+++ b/Duplicati/Library/Main/Database/LocalListDatabase.cs
@@ -163,17 +163,9 @@ namespace Duplicati.Library.Main.Database
     
                     while (filecount != foundfiles && maxpath.Length > 0)
                     {
-                        if (Platform.IsClientWindows)
-                        {
-                            cmd.SetParameterValue(0, maxpath.Length);
-                            cmd.SetParameterValue(1, maxpath);
-                        }
-                        else
-                        {
-                            var mp = Util.AppendDirSeparator(maxpath, dirsep);
-                            cmd.SetParameterValue(0, mp.Length);
-                            cmd.SetParameterValue(1, mp);
-                        }
+                        var mp = Platform.IsClientPosix ? Util.AppendDirSeparator(maxpath, dirsep) : maxpath;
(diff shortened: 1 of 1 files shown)


warwickmm:
@BlueBlock, can you clarify how to reproduce the issue here?  I haven't had any issues with restore and getting the file list.

BlueBlock:
> @BlueBlock, can you clarify how to reproduce the issue here? I haven't had any issues with restore and getting the file list.

There doesn't appear to be any negative consequence to the bug. 

Running with the debugger and put a break on the line shown in an image below. Use an existing backup and select the Restore operation. The code will be hit when the file list gets pulled.

Without the fix, watch the values of "mp" and "foundfiles". "mp" will have the filename and an ending slash, which is invalid. So "foundFiles" will be 0.

Without fix:


With fix:

warwickmm:
It looks like that line was introduced in revision 9a849a2b58943c53eeba995c7b78d39bfda95dac to fix some issues with cross-os restores.  Would this introduce a regression?

BlueBlock:
> It looks like that line was introduced in revision [9a849a2](https://github.com/duplicati/duplicati/commit/9a849a2b58943c53eeba995c7b78d39bfda95dac) to fix some issues with cross-os restores. Would this introduce a regression?

The safest thing to do then would be to place this fix in an if/else for the OS.

I think I may need to try and confirm all the cases when DirectorySeparator is applied, at least under Windows, and see if it adds the slash elsewhere on files.

warwickmm:
Unfortunately, I only have a Linux machine so I cannot help much with testing here.

BlueBlock:
Added logic to apply it only to Windows

warwickmm:
The original commit (9a849a2) mentions cross-os restores where the backup was created on Windows and the restore was performed on Mac/Linux.  Do we know how these changes might affect the other direction?  That is, backup on Mac/Linux and restore on Windows?

BlueBlock:
> The original commit ([9a849a2](https://github.com/duplicati/duplicati/commit/9a849a2b58943c53eeba995c7b78d39bfda95dac)) mentions cross-os restores where the backup was created on Windows and the restore was performed on Mac/Linux. Do we know how these changes might affect the other direction? That is, backup on Mac/Linux and restore on Windows?

It wouldn't affect another platform since it is only reading data and the code is restricted to Windows with the original code for other platforms. So it seems there would be no affect?

warwickmm:
I don't understand why the directory separator is needed in any case, so the main concern is introducing a regression.  Even though the modifications are restricted to Windows, it's not clear to me if this affects how a Linux backup is restored on Windows.

@BlueBlock, is your setup able to test some simple cross-os backup/restores?

BlueBlock:
I can try with a linux dev machine I have setup.

BlueBlock:
I tested and was able to backup/restore across Windows/Linux without a problem.

warwickmm:
That's great!  In fact, would you mind running your tests with the fix applied to all platforms (not just Windows)?  I don't understand why the directory separator is needed in any case.

BlueBlock:
I'll give it a test.

BlueBlock:
The issue also happens on linux so the fix should apply to all systems. New commit added.

warwickmm:
Sounds good to me.  Thanks @BlueBlock.

PR #3847, opened 2019-08-03
Fix negative bytes to go issue

drwtsn32x:
Under certain circumstances the bytes to go will be shown as negative:

dbc696f948ba515d252bb040ff869eb6a16e6bde

This seems to happen most notably when backing up a large file that is a significant portion of the total backup set size.  It is easiest to reproduce if you create a new backup set and choose a single large file to back up.

After some testing it appears the root cause is that `m_filesizeprocessed` is incremented when the file is done being processed, but some time may pass before `m_curfileoffset` is set back to zero.  If the status bar updates during this brief time window, the size of the file is counted twice when the `sizeleft` value is calucated in `AppUtils.js`:

`var sizeleft = $scope.state.lastPgEvent.TotalFileSize - $scope.state.lastPgEvent.ProcessedFileSize - $scope.state.lastPgEvent.CurrentFileoffset;`

This can result in `sizeleft` being a negative number if, as stated above, the file being counted twice is a significant portion (over 50%) of the total backup set size.   Even if it is not significant, it still results in an incorrect calculation.

My proposed solution is to immediately set `m_curfileoffset` to zero when `UpdatefilesProcessed` is called in `ProgressHandler.cs:72`.

In my testing it seems to work well.

### DIFF
--- a/Duplicati/Library/Main/ProgressClasses.cs
+++ b/Duplicati/Library/Main/ProgressClasses.cs
@@ -379,18 +379,13 @@ namespace Duplicati.Library.Main
             {
                 m_curfilename = filename;
                 m_curfilesize = size;
-                Logging.Log.WriteInformationMessage("DEBUG-TEMP", "DEBUG-TEMP", "StartFile - setting m_curfileoffset = 0, currently = {0}", m_curfileoffset);
-                m_curfileoffset = 0;
             }
         }
         
         public void UpdateFileProgress(long offset)
         {
             lock (m_lock)
-            {
-                Logging.Log.WriteInformationMessage("DEBUG-TEMP", "DEBUG-TEMP", "UpdateFileProgress - setting m_curfileoffset = {0}, currently = {1}", offset, m_curfileoffset);
                 m_curfileoffset = offset;
-            }
         }
         
         public void UpdatefileCount(long filecount, long filesize, bool done)
@@ -409,7 +404,6 @@ namespace Duplicati.Library.Main
             {
                 m_filesprocessed = count;
                 m_filesizeprocessed = size;
-                Logging.Log.WriteInformationMessage("DEBUG-TEMP", "DEBUG-TEMP", "UpdatefilesProcessed - setting m_curfileoffset = 0, currently = {0}", m_curfileoffset);
                 m_curfileoffset = 0;
             }
         }
--- a/Duplicati/Library/Main/ProgressClasses.cs
+++ b/Duplicati/Library/Main/ProgressClasses.cs
@@ -405,6 +405,7 @@ namespace Duplicati.Library.Main
             {
                 m_filesprocessed = count;
                 m_filesizeprocessed = size;
+                m_curfileoffset = 0;
             }
         }
         
--- a/Duplicati/Library/Main/ProgressClasses.cs
+++ b/Duplicati/Library/Main/ProgressClasses.cs
@@ -375,7 +375,7 @@ namespace Duplicati.Library.Main
         
         public void StartFile(string filename, long size)
         {
-            lock (m_lock)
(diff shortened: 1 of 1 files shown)


Pectojin:
Won't it cause issues if we set m_curfileoffset to zero both immediately and later?

BlueBlock:
> Won't it cause issues if we set m_curfileoffset to zero both immediately and later?

When tested it I saw the bytes go to 0 while on the last file. Not a big issue but not really accurate. This maybe part of it but I also saw that folders are being included in the 'file' count.

Try it with maybe 5 files and have 4 or 5 directories to see the result.

Directories should probably not be included in the count.

drwtsn32x:
> When I had tested it I saw the bytes go to 0 while on the last file.

What do you mean, the bytes to go said zero?  Isn't that correct when the last file has finished?

Regarding the file count - this change should have no bearing on that.

drwtsn32x:
> Won't it cause issues if we set m_curfileoffset to zero both immediately and later?

I don't see how it would cause an issue, but it does seem unnecessary to set it to zero both at the end of a file and at the start of the next file.  I could do testing where we remove the set to zero at the start of a file, if you wish.

Pectojin:
I guess it wouldn't cause an issue to set it to 0 twice. But it's a bit confusing.

I'd prefer it cleaner with only setting it once.

BlueBlock:
> > When I had tested it I saw the bytes go to 0 while on the last file.
> 
> What do you mean, the bytes to go said zero? Isn't that correct when the last file has finished?
> 
> Regarding the file count - this change should have no bearing on that.

From an end-user perspective I would not expect a file in-progress to reduce the bytes remaining. 0 to me says it is done.

drwtsn32x:
> I'd prefer it cleaner with only setting it once.

Ok, I will do some testing. It should work if we only set it to zero at file completion.

drwtsn32x:
> From an end-user perspective I would not expect a file in-progress to reduce the bytes remaining. 0 to me says it is done.

That's exactly right. This change sets "current file processed" to zero right at file completion, when the "total bytes processed" is incremented by that that file's size.  The file is done at that point.

drwtsn32x:
@Pectojin - ok I did testing where OperationProgressUpdater.StartFile() does NOT set m_curfileoffset = 0.  It works as expected ... there is no effective change, we just aren't setting it to zero twice between files any more.

For the very first file, m_curfileoffset is already zero since this object is newly initialized. For subsequent files, it is set to zero by OperationProgressUpdater.UpdatefilesProcessed() - which is called when a file is complete. 

I will update the pull request shortly....

Pectojin:
Ok, that looks more clean even if I guess it doesn't have an effect in practice.

drwtsn32x:
> When I had tested it I saw the bytes go to 0 while on the last file.

@BlueBlock - did you mean the per-file progress meter? I just noticed this myself.  The last file, once complete, shows 0% on its individual progress bar.

The main progress bar is corrected with this fix but messes up the individual file progress meter. Not sure of a clean way to fix this correctly without introducing another variable.  Will think on it some more...

drwtsn32x:
Screen shot showing the new problem - main status bar is correct, but individual file status bar drops back to 0% when complete (and is last file):

Capture

One solution I can think of is to create a new variable that indicates the current file is complete.  In the javascript portion we change the calculation to NOT include "currentFileOffset" in the main progress meter when that flag is true, because at that point the ProcessedFileSize has been incremented.

When the next file starts, that flag is set back to false.

Thoughts?

drwtsn32x:
Suggested fix in PR https://github.com/duplicati/duplicati/pull/3904

It rolls back the changes made by this PR and implements a newer solution....

Issue #2673, opened 2017-08-29
Google Drive backend should ignore trashed folders and files


Pectojin:
I have:
- [x] searched open and closed issues for duplicates

----------------------------------------

### Version info
**Duplicati Version:** 2.0.2.1_beta_2017-08-01
**Operating System:** Ubuntu 17.04
**Backend:** Google Drive

### Bug description
It seems that Duplicati keeps a cached list of directories created in Google Drive and if that directory is deleted the cache prevents duplicati from realizing. 

### Steps to reproduce
- Create new backup
- Set backup directory in Google Drive to be backup/weekly/server
- Test the connection
- Accept creation of folder
- Complete backup creation
- Delete backup folder from Google Drive 
- Edit backup config
- Test the connection again, which results in the error 'Failed to connect: There is more than one item named "server" in the folder'
- Subsequent backup attempts will also fail with said error
- Manually recreating the folder does not resolve the issue.

**Actual result:** I'm unable to backup to a given path or setup a new backup to the path if a duplicati folder was previously deleted from that path
**Expected result:** Duplicati should detect a folder no longer existing and recreate it - ideally even if it finds out during a backup.

### debug log
`Duplicati.Library.Interface.UserInformationException: There is more than one item named "server" in the folder "/backup/weekly/"
  at Duplicati.Library.Main.BackendManager.List () [0x00038] in <118ad25945a24a3991f7b65e7a45ea1e>:0 
  at Duplicati.Library.Main.Operation.FilelistProcessor.RemoteListAnalysis (Duplicati.Library.Main.BackendManager backend, Duplicati.Library.Main.Options options, Duplicati.Library.Main.Database.LocalDatabase database, Duplicati.Library.Main.IBackendWriter log, System.String protectedfile) [0x0000d] in <118ad25945a24a3991f7b65e7a45ea1e>:0 
  at Duplicati.Library.Main.Operation.FilelistProcessor.VerifyRemoteList (Duplicati.Library.Main.BackendManager backend, Duplicati.Library.Main.Options options, Duplicati.Library.Main.Database.LocalDatabase database, Duplicati.Library.Main.IBackendWriter log, System.String protectedfile) [0x00000] in <118ad25945a24a3991f7b65e7a45ea1e>:0 
[...]

PR #3856, opened 2019-08-12
Do not include Google Drive trashed items

BlueBlock:
In Google Drive the trashed items are being included when enumerating. If a previous backup folder or files are deleted manually by the user these items prevent Duplicati form creating a new backup folder. Duplicati also see's the trashed files as being in the active backup folder.


Pectojin:
I believe this fixes #2673

PR #3866, opened 2019-08-20
WIP AFTP backend data upload corruption fix

BlueBlock:
This is a proof of concept fix since there will likely be a more elegant solution. But this is a working fix and possibly similar is needed for the ftp backend.

Currently AFTP will seemingly transfer a file via async ftp and return 'success' even though it seems the transfer is still proceeding. The symptom looks to be an error stating the hashes do not match.

I am able to reproduce every time when not in debug mode. Debug mode seems to introduce enough delay that the error never occurs, at least not for me. 

There might be another issue going on with timing between processes; I'm not sure. The simplistic solution is to loop the check for the file size and only proceed when the file size is correct or our set time for checking has expired. 

- Can this solution be a stop-gap fix even in the event there is a more involved issue deeper in Duplicati?

- As a WIP and discussion on this fix, what modifications to the check might others suggest? A better way than sleep()?

Real data is shown below to show how after the async upload the remote file size is smaller and continues to grow until it is the correct size. This is using a FileZilla Server installed on the same machine. The data is sorted by the filename. At 250 ms you'll see that for the blocks there is generally a number of rechecks that occur.

[aftp-fix-test.csv.txt](https://github.com/duplicati/duplicati/files/3518124/aftp-fix-test.csv.txt)

### DIFF
--- a/Duplicati/Library/Backend/AlternativeFTP/AlternativeFTPBackend.cs
+++ b/Duplicati/Library/Backend/AlternativeFTP/AlternativeFTPBackend.cs
@@ -337,25 +337,23 @@ namespace Duplicati.Library.Backend.AlternativeFTP
                     throw new UserInformationException(string.Format(Strings.ErrorWriteFile, remotename), "AftpPutFailure");
                 }
 
-                if (_listVerify)
+                // check remote file size; matching file size indicates completion
+                var sleepTime = 250;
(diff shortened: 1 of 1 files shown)


BlueBlock:
We can revisit if the 5 seconds is an issue. Ideally I'd like to Polly and have a smarter retry. But at least this fixes the most immediate problem.

PR #3871, opened 2019-08-25
Improve some unittest handling

BlueBlock:
Currently the data.zip is not functional for some tests on Windows. 

- This PR adds the ability of a dev to use a alternative data.zip for testing. To use an alternative filename the file "data-alternative.zip" for it to be automatically used.

- The downloading of data.zip will only occur if the local file is missing or the remote data.zip file has been updated. 

- The CommandLineOperationsTests use the data.zip file which is very large rather than recursively copy the files, we'll move the directories instead. Each test sets everything up again.

### DIFF
--- a/Duplicati/UnitTest/CommandLineOperationsTests.cs
+++ b/Duplicati/UnitTest/CommandLineOperationsTests.cs
@@ -29,7 +29,7 @@ namespace Duplicati.UnitTest
         /// The log tag
         /// </summary>
         private static readonly string LOGTAG = Library.Logging.Log.LogTagFromType<CommandLineOperationsTests>();
-        
+
         /// <summary>
         /// The folder that contains all the source data which the test is based on
         /// </summary>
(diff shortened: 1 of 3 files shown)


Duplicati/UnitTest/CommandLineOperationsTests.cs
```diff
@@ -52,13 +57,49 @@ orderby x
         public override void OneTimeSetUp()
         {
             base.OneTimeSetUp();
-            
-            using (WebClient client = new WebClient())
+
+            if (!File.Exists($@"{BASEFOLDER}\{zipAlternativeFilename}"))
```
warwickmm: Can we use `Path.Combine` here instead of `\` here?

BlueBlock: Fixed. Thanks.

BlueBlock:
Thanks!

PR #3872, opened 2019-08-27
Fix for issue1400

BlueBlock:
There can essentially be a race between a file getting updated and Duplicati trying to back that file up. 

We'll use a 'limit' stream to limit the file streams length to be for that point in time.

As an aside, if Duplicati is processing slower than a file is being updated then a Duplicati backup will never end. We always want to be backing up a file of known length at that moment in time.

### DIFF
--- a/Duplicati/Library/Main/Operation/Backup/StreamBlock.cs
+++ b/Duplicati/Library/Main/Operation/Backup/StreamBlock.cs
@@ -17,6 +17,7 @@
 using System;
 using CoCoL;
 using System.IO;
+using System.Threading;
 using System.Threading.Tasks;
 using Duplicati.Library.Interface;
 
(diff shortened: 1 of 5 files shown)


warwickmm:
The test passed in AppVeyor, but failed in Travis.  It fails on my Linux machine as well, but I wasn't able to spend much time looking at it...

BlueBlock:
Fixed the shared file access for the unittest. Though it looks like Travis is having some issues pulling build files.

BlueBlock:
The unittest passes on Windows and Linux now.

BlueBlock:
Are we to do a small commit to force new CI runs if they fail for non-code reasons?

warwickmm:
I kicked off a re-build of the failed Travis jobs.

BlueBlock:
Thanks

warwickmm:
Nice job @ts678 and @BlueBlock.

BlueBlock:
Thanks!

ts678:
Thanks to all. I'm not sure if this is a 100% fix for all changing file scenarios, but it should help a LOT. Even better is that it finally made it into master (I wasn't expecting it would take 4 days...) so that will resolve the issue of whether or not to wait for it before turning master into canary, if that's our plan.

drwtsn32x:
Nice work on this one!

PR #3873, opened 2019-08-28
WIP Use subfolders in backends

BlueBlock:
(I'm putting this out to get early feedback. This top-post will be updated with all progress.)

**Still is a work-in-progress**

Summary:
-
This PR adds support for utilizing subfolders on the backend and is backward compatible with existing backups. The creation of subfolders is designed to be flat by filling the top-most folders first before proceeding to the next level of subfolders. 

Two options have been added to control the use of subfolders, one for the number of files per folder and a second for the number of folder within a folder.
- max-files-per-remote-subfolder:
- max-folders-per-remote-subfolder:

With a recommended setting of 2000 files per folder and 1000 folders per folder (which works with all backends), most users may not even utilize subfolders but it would then be used if needed.

[The forum discussion on the topic can be found here.](https://forum.duplicati.com/t/logic-and-usecase-for-remote-subfolders/6511/29)

Features:
-
- subfolders can be enabled on existing backups
- subfolders which become sparse through compaction will be refilled on following backups
- increasing/decreasing subfolder settings do not modify existing files but do effect the placement of new files 

The solution requires each backend to be modified to support the handling of subfolders. This is WIP because there are still backends to modify. 

The backends enabled to handle subfolders are listed below. Note, attempting to use subfolders on unsupported backends will have unpredictable results.

Sub-Folders Working:
-
- aftp
- Box.com
- Google Drive (also replaced deprecated api calls)
- local file
- Mega.nz
- OneDrive
- SharePoint (untested)

In progress:
-

To be done:
-

- Azure Blob
- BackBlaze
- CloudFiles
- Dropbox
- ftp
- HubiC
- JottaCloud
- OpenStack
- rclone
- Sia
- SSHv2
- TahoeLAFS
- WebDav

- Amazon Cloud Drive (deprecated) (implement last?)


typoworx-de:
Will there also be some kind of migration for existing storage-backends without sub-folders?

warwickmm:
Closing this due to lack of activity.  We can revisit this later if necessary.

PR #3874, opened 2019-08-28
Fix travis complaint about build.sh not being executable

BlueBlock:
Sometimes Travis will complain about build.sh not being executable. I keep needing to add this to every new branch.

The echo line is not changed, just moved to the left.


warwickmm:
Strange, I've never seen Travis complain about this.  Doesn't hurt though.

PR #3885, opened 2019-09-03
typo ellipsis, added

leela52452:
[source](https://hosted.weblate.org/projects/duplicati/webroot/source/review/?type=check%3Aellipsis)


kenkendk:
Not sure about this one. The arguments given from the Weblate docs are:

> Using the Unicode character is in most cases the better approach and looks better rendered, and may sound better with text-to-speech.

I _personally_ like the visuals of three dots taking up more space (at least in Courier fonts), but I am not strongly committed to either.

Anyone preferring the unicode single character version?

BlueBlock:
The current use of ellipsis seems to be preferred. Screen readers like JAWS will not read the ellipsis aloud and simply pause a little.  US 508 accessibility compliance prefers ellipsis.

warwickmm:
I don't have a preference, but it makes sense to adhere to the accessibility guidelines.

kenkendk:
Ok, I agree that we should follow accessibility guidelines over personal preferences.

PR #3890, opened 2019-09-05
Copy AlphaVSS dlls for debug build

BlueBlock:
The AlphaVSS dll's are not copied when building. The release build .sh script handles this for releases.

This PR will cause the dll's to be copied only for the debug build so that AlphaVSS can be tested in debug.


warwickmm:
Looks like Travis is complaining about a syntax error?

PR #3897, opened 2019-09-11
Improve logging during database recreation

drwtsn32x:
This adds "X of Y" type messages to the verbose logging mechanism while processing files during database recreation. It makes it easier to see how far along the process is.


drwtsn32x:
Example of why this change is helpful... recreating database in this case requires the download of 467 blocklists. Each takes about 15 minutes to process:

PR #3905, opened 2019-09-17
Update s3 providers

kenkendk:
This updates the list of S3 providers, and changes the style of the code to use dictionary initializers.


Duplicati/Library/Backend/S3/S3Backend.cs
```diff
@@ -218,7 +224,7 @@ public S3(string url, Dictionary<string, string> options)
 
                 //Change in S3, now requires that you use location specific endpoint
                 if (!string.IsNullOrEmpty(locationConstraint))
-                    foreach (KeyValuePair<string, string> kvp in DEFAULT_S3_LOCATION_BASED_HOSTS)
+                    foreach (var kvp in DEFAULT_S3_LOCATION_BASED_HOSTS)
```
warwickmm: Minor point, but can we now perform a single lookup via `Dictionary.TryGetValue` instead of enumerating the entries?

Duplicati/Library/Backend/S3/S3Backend.cs
```diff
@@ -218,7 +224,7 @@ public S3(string url, Dictionary<string, string> options)
 
                 //Change in S3, now requires that you use location specific endpoint
                 if (!string.IsNullOrEmpty(locationConstraint))
-                    foreach (KeyValuePair<string, string> kvp in DEFAULT_S3_LOCATION_BASED_HOSTS)
+                    foreach (var kvp in DEFAULT_S3_LOCATION_BASED_HOSTS)
                         if (kvp.Key.Equals(locationConstraint, StringComparison.OrdinalIgnoreCase))
```
warwickmm: Minor point, but can we now perform a single lookup via `Dictionary.TryGetValue` instead of enumerating the entries?

kenkendk: Yes, you are right. I have added a new commit that uses `TryGetValue` instead of the looping code.

PR #3908, opened 2019-09-18
Remove unused FileReadBufferSize property

warwickmm:
The last reference was removed in revision 20f7b42945.


kenkendk:
Then we should also either remove the commandline option, or add a deprecation warning (and remove it later):
https://github.com/duplicati/duplicati/blob/89191a6caf95e47dc545a6f206a3578a692cf31d/Duplicati/Library/Main/Options.cs#L349

warwickmm:
Good catch.  I added a deprecation message in revision 7090501c0b001d12df9d2284680eb63d00256686.

Issue #3891, opened 2019-09-05
CommandLine test-filters fails on RegExp pipe (or-function) on linux 


arberg:
- [X ] I have searched open and closed issues for duplicates.

----------------------------------------

## Environment info
- **Duplicati version**: 2.0.4.23_beta_2019-07-14
- **Operating system**: Linux docker, [linuxserver](https://hub.docker.com/r/linuxserver/duplicati)

## Description
Commandline regExp fails to match if it contains the selection group with `|`, for example `(x|y)`

## Steps to reproduce

All of the command examples below are identical except the ending in the regExp.

* This matches /mnt/user/media/TestDuplicati/A/
`docker exec duplicati mono /app/duplicati/Duplicati.CommandLine.exe test-filters /mnt/user/media/TestDuplicati/ --exclude='[/mnt/user/media/Test(A)/]'`

* This does not (where we now have (A|A)
`docker exec duplicati mono /app/duplicati/Duplicati.CommandLine.exe test-filters /mnt/user/media/TestDuplicati/ --exclude='[/mnt/user/media/Test(A|A)/]'`

Interestingly this 
`docker exec duplicati mono /app/duplicati/Duplicati.CommandLine.exe test-filters /mnt/user/media/TestDuplicati/ --exclude='[/mnt/user/media/TestDuplicati/(A\|Test1)/]'`

gives the output 

    Including path as no filters matched: /mnt/user/media/TestDuplicati/A/
    Excluding path due to filter: /mnt/user/media/TestDuplicati/Test1/ => ([/mnt/user/media/TestDuplicati/(A\\|Test1)/])

which hints that the other problem might be some escaping. The above should not have excluded Test1, as it should match (I think) the folder name 'A|Test1'.

## UI
Entering filters with (A|B) in UI works as expected.

## Docs
I have commented about it in the docs here: https://github.com/duplicati/duplicati/wiki/How-do-filters-work%3F

ts678:
[Bug Regular Expression Filter operator ‘|’ (or) does not work inside parenthesis ()](https://forum.duplicati.com/t/bug-regular-expression-filter-operator-or-does-not-work-inside-parenthesis/7950/8?u=ts678) supports a theory of a general command line escaping issue, with possible workaround of disabling autoupdater mechanism.

arberg:
Identical (earlier) issue: https://github.com/duplicati/duplicati/issues/3135

warwickmm:
Fixed in pull request #3911.

PR #3911, opened 2019-09-19
Fixed incorrect escaping of pipe characters

kenkendk:
Based on forum discussion, the regexp is incorrect:
https://forum.duplicati.com/t/bug-regular-expression-filter-operator-or-does-not-work-inside-parenthesis/7950/21

### DIFF
--- a/Duplicati/Library/Utility/Utility.cs
+++ b/Duplicati/Library/Utility/Utility.cs
@@ -1385,7 +1385,7 @@ namespace Duplicati.Library.Utility
         /// <summary>
         /// Special characters that needs to be escaped on Linux
         /// </summary>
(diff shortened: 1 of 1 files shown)


warwickmm:
This fixes #3891.

warwickmm:
Looks good to me.  Although, I'm not sure why the GUI tests are failing in Travis now.  I re-ran them a few times and they failed consistently.

warwickmm:
I think the GUI test is failing due to the ellipsis changes from pull request #3885.  This is something that should be fixed separately.
```
Traceback (most recent call last):
  File "guiTests/guiTest.py", line 148, in <module>
    [n for n in driver.find_elements_by_xpath("//span[contains(text(),'Restore files ...')]") if n.is_displayed()][0].click()
IndexError: list index out of range
Error on line 51, stopping build.
```

PR #3917, opened 2019-09-20
Fix search for ellipsis in GUI tests

warwickmm:
The three periods were replaced by an ellipsis in revision 7ff50a30e1 (pull request #3885).


kenkendk:
Great work!

PR #3922, opened 2019-09-21
Fix cleanup of Remotevolume records that use the old datetime method

BlueBlock:
Currently the old deleted records will remain since they are in milliseconds. This PR will remove the old records.

Also, the where clause now has 'State' first since it should be considered first so we limit the set size for the other parts of the clause.

### DIFF
--- a/Duplicati/Library/Main/Database/LocalDatabase.cs
+++ b/Duplicati/Library/Main/Database/LocalDatabase.cs
@@ -135,7 +135,7 @@ namespace Duplicati.Library.Main.Database
             m_removeremotevolumeCommand.CommandText = @"DELETE FROM ""Remotevolume"" WHERE ""Name"" = ? AND (""DeleteGraceTime"" < ? OR ""State"" != ?)";
             m_removeremotevolumeCommand.AddParameters(3);
 
-            m_removedeletedremotevolumeCommand.CommandText = $@"DELETE FROM ""Remotevolume"" WHERE ""DeleteGraceTime"" < ? AND ""State"" == ""{RemoteVolumeState.Deleted.ToString()}""";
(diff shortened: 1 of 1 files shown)


Duplicati/Library/Main/Database/LocalDatabase.cs
```diff
@@ -135,7 +135,7 @@ private LocalDatabase(System.Data.IDbConnection connection)
             m_removeremotevolumeCommand.CommandText = @"DELETE FROM ""Remotevolume"" WHERE ""Name"" = ? AND (""DeleteGraceTime"" < ? OR ""State"" != ?)";
             m_removeremotevolumeCommand.AddParameters(3);
 
-            m_removedeletedremotevolumeCommand.CommandText = $@"DELETE FROM ""Remotevolume"" WHERE ""DeleteGraceTime"" < ? AND ""State"" == ""{RemoteVolumeState.Deleted.ToString()}""";
+            m_removedeletedremotevolumeCommand.CommandText = $@"DELETE FROM ""Remotevolume"" WHERE ""State"" == ""{RemoteVolumeState.Deleted.ToString()}"" AND ""DeleteGraceTime"" < ? OR LENGTH(""DeleteGraceTime"") > 12 "; // >12 is to handle removal of old records that were in milliseconds
```
warwickmm: Was the old format actually in ticks (not milliseconds)?

BlueBlock: I think you're right. But the code doesn't change. Just the comment.

warwickmm: Also, I think `AND` has precedence over `OR`.  Does that mean that the condition here is equivalent to
```
("State" == "Deleted" AND "DeleteGraceTime" < ?) OR LENGTH("DeleteGraceTime") > 12
```
Should it be instead
```
"State" == "Deleted" AND ("DeleteGraceTime" < ? OR LENGTH("DeleteGraceTime") > 12)
```

kenkendk: I prefer explicit parenthesis, that makes it easier to understand the intent.

Issue #3924, opened 2019-09-25
Purge operations result in broken dlist files


warwickmm:
**Duplicati version**: 2.0.4.29

## Description

A purge operation results in broken dlist files, as well as possibly other issues.

## Steps to reproduce

1. Run a backup.
2. Purge a file.  This _appears_ to run without error.  However, notice that most of the dlist files are 541 bytes in size.
3. Recreate the database.  This should result in an exception and reveal that the purge did not behave properly:
```
Failed to process file: duplicati-20190925T020411Z.dlist.zip.aes => Value cannot be null.
Parameter name: stream
Failed to process file: duplicati-20190925T020414Z.dlist.zip.aes => Value cannot be null.
Parameter name: stream
  Downloading file (1.39 KB) ...
```

This appears to have been introduced in pull request #3836, possibly between revisions e8f3bccef8da469f3b00f228a77b1965b16d0781 and 153a97285315d9562f21384365fc45b783d862ec.

See forum discussion [here](https://forum.duplicati.com/t/purge-resulted-in-damaged-dlist-files/8114).

warwickmm:
I can get my tests to pass if I add
```
vol.AddFilelistFile();
```
to the `PurgeFileHander.DoRun` method [here](https://github.com/duplicati/duplicati/blob/master/Duplicati/Library/Main/Operation/PurgeFilesHandler.cs#L165).  However, it would be more satisfying if we didn't have to remember to invoke this everywhere.

@BlueBlock, any suggestions here?

warwickmm:
Revision e0899a2a91aab8e982e9674461c0504907ef5845 appears to have been the one that made creation of the filelist non-mandatory.  However, the commit message mentions that it fixes a different issue regarding the "control-files" that can be optionally added.

ts678:
2.0.4.29 has the same problem if using repair to re-upload a missing dlist file. Possibly similar cause?

[...]

PR #3925, opened 2019-09-28
Ensure that filelist is always included in dlist files

warwickmm:
This fixes a regression introduced around revisions e0899a2a91 and 3b5af1cf01, where the filelist was not written by all usages of the `FilesetVolumeWriter` (these revisions did fix an issue with including the control-files).  Previously, one would have to invoke `AddFilelistFile` with each usage.  Now, we simply do so when the `FilesetVolumeWriter` is closed.

We also modified the `UploadSyntheticFilelist` class so that the usage of the `FilesetVolumeWriter` is contained in a `using` statement, which will ensure that it is disposed.

This fixes issue #3924.

### DIFF
--- a/Duplicati/Library/Main/Operation/Backup/UploadRealFilelist.cs
+++ b/Duplicati/Library/Main/Operation/Backup/UploadRealFilelist.cs
@@ -62,9 +62,6 @@ namespace Duplicati.Library.Main.Operation.Backup
                             return;
 
                         await db.WriteFilesetAsync(filesetvolume, filesetid);
-
-                        filesetvolume.AddFilelistFile();
-
                         filesetvolume.Close();
 
                         if (!await taskreader.ProgressAsync)
--- a/Duplicati/Library/Main/Operation/Backup/UploadSyntheticFilelist.cs
+++ b/Duplicati/Library/Main/Operation/Backup/UploadSyntheticFilelist.cs
@@ -88,7 +88,6 @@ namespace Duplicati.Library.Main.Operation.Backup
 
                 var prevId = prevs.Length == 0 ? -1 : prevs.Last();
 
-                    FilesetVolumeWriter fsw = null;
                     try
                     {
                         var s = 1;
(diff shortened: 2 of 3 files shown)


drwtsn32x:
Initial test looks like it fixes the issue where dlist files are saved without the filelist.

But - the purge operation appears stalled after the dlist files are processed.  Duplicati status bar says "Running task: PurgeFiles_Compact" but nothing seems to be happening.  Has been stuck here for 4 hours.  No activity in Live (Profiling) log.  CPU usage for Duplicati process is maxing out a single core.  No file activity on local disk and no transfers happening with back end storage.

warwickmm:
Hmm...I haven't encountered any issues with purge stalling.  In fact, I reproduced issue #3927 with this branch.

How large is the source directory in your test and how many files are you purging?

drwtsn32x:
I'm testing with a copy of my largest backup sets, and I am purging a lot of data.  Before this patch the same test purge didn't hang and appeared to complete successfully, but of course the dlist files were invalid.

Let me redo the test by purging a much smaller set of data to see what happens.

warwickmm:
If it's not too much trouble, can you see if the purge stalls with version 2.0.4.28 (before the dlist files were broken)?

Thanks for your help with testing this.

drwtsn32x:
Absolutely, let me try that.... will report back

warwickmm:
By the way, how large are your dlist files?  I believe that as of 2.0.4.29, the filelist json is kept in memory before being written to a file, whereas previous versions might have streamed to a file.  I'm not completely sure about that though.

drwtsn32x:
my dlist files are about 15MB each

ts678:
Could someone also please see if `repair` after deleting a dlist file has gone back to recreating it OK?

warwickmm:
> Could someone also please see if repair after deleting a dlist file has gone back to recreating it OK?

I deleted a dlist file and repair was able to recreate it.

drwtsn32x:
Ok, a small purge worked properly - no hang at the PurgeFiles_Compact phase. Perhaps I just need to wait long enough if it's a really large purge?

I'm going to test a larger folder next....

drwtsn32x:
First small test reduced back end from 713GB to 710GB - number of files from 29677 to 29503.
Second larger test reduced back end from 710GB to 666GB - number of files from 29503 to 27687.

Second test didn't hang at the Compact phase either.

Starting even larger third test now....

drwtsn32x:
Third test reduced back end from 666GB to 272GB - number of files from 27687 to 11519. No problems.

Going to now do the the original purge that appeared to hang on compact phase. Will let you know.

drwtsn32x:
Purge worked ok! I want to do additional tests: recreate local db from scratch from remote data, among others.

warwickmm:
Thanks @drwtsn32x for all the testing!

drwtsn32x:
Database recreation was successful. I want to re-test original conditions that appeared to cause purge to hang. Not sure if it was just an anomaly or not....

drwtsn32x:
Same issue - it hangs at `PurgeFiles_Compact` if I do a large purge first instead of in stages.

I think it's a different issue though, and not related to this bug and your pull request.  This PR solves the issue of purge creating invalid dlist files in my testing!

drwtsn32x:
I have confirmed that the "hang after purge" has nothing to do with this fix. I was able to reproduce it on 2.0.4.30 without your fix applied.

warwickmm:
Thanks @drwtsn32x for all the testing!

PR #3930, opened 2019-09-29
Ensure that instances of VolumeWriterBase are disposed

warwickmm:
This encloses some usages of `BlockVolumeWriter` and `IndexVolumeWriter` in `using` statements to ensure that they are disposed of after their use.


warwickmm:
The inline diff is a bit messy due to whitespace changes. I recommend viewing the [side-by-side diff](https://github.com/duplicati/duplicati/pull/4047/files?utf8=%E2%9C%93&diff=split&w=1).

Issue #1870, opened 2016-08-02
tempdir not respected


huste:
I am using 2.0.1.11_experimental_2016-04-08 and specified a tempdir variable in duplicati's general settings: D:\temp

I know that sqlite doesn't use this, but I'd expect all the zip-files that Duplicati creates to be created there. Instead, they are still produced in %TEMP%.
##

kenkendk:
I think this is related to the general settings not being propagated to each backup.
If you set the tmpdir on the individual backup it should work.
Can you confirm that the problem is only with general settings?

mr-flibble:
I can confirm that - problem is witch general setting only. If i set tmpdir for individual backup, everything is OK as expected.

huste:
Yes, I can confirm that, too.

ts678:
Is this still reproducible? I've tested as far back as 2.0.2.1_beta_2017-08-01, and Settings `Default options` --tempdir successfully moved the `dup-*` files for the backup to the specified location. I could not find a claimed fix in the records, but it seems OK now...

huste:
No, it works with a newly created backup job and a specifically set tempdir in the general options with 2.0.5.111. I'll close it now

PR #3936, opened 2019-10-05
Fix ignoring of --tempdir in CommandLine.exe

drwtsn32x:
The `--tempdir` command line option was being ignored by `Duplicati.CommandLine.exe` and the default system temp path was being used instead.


warwickmm:
I think this fixes #1870.

Issue #3932, opened 2019-09-30
dblock put retry corrupts dindex, using old dblock name for index -- canary regression


ts678:
- [x] I have searched open and closed issues for duplicates.

----------------------------------------

## Environment info

- **Duplicati version**: 2.0.4.30
- **Operating system**: Windows 10 Version 1809
- **Backend**: various

## Description

An upload (put) of a dblock file that fails is retried after a dblock file rename (as expected) but the vol file in the dindex file is never updated and stays at the original dblock file name, which does not exist on the destination. The database tables are correct and hide the damage until a direct restore is done (which fails) or a Recreate is done, which picks up the dangling reference which can then break future restores and possibly other operations. The problem seems to have been introduced in 2.0.4.16 which did have a dindex file problem. Maybe the fix was incorrect. [Here](https://forum.duplicati.com/t/issues-to-address-to-get-out-of-beta/7871/77?u=ts678) is an earlier issue on a Backblaze B2 backup, along with a pointer to code that used to update the index name when dblock got renamed.

## Steps to reproduce

1. Have a dblock upload failure retry and your backup will break. Retries are normal on networks.
2. Test steps below used FileZilla Server on Windows, but there are many ways to stop an upload.
3. Set backup destination to not allow upload, e.g. turn server off, change folder permissions, etc.
4. Set --no-backend-verification=true if the destination disabling will not allow file listings either.
5. Set About --> Show log --> Live --> Retry in a different browser tab to see names and actions.
6. Backup two (to avoid a bug) tiny files (1 byte is enough), without encryption for easier analysis.
7. Allow the dblock upload to finish, e.g. turn the server on, add write permission onto folder, etc.
8. Look in the dindex zip file vol folder at the file with a dblock file name. This has the index for it.

- **Actual result**:
The dblock file name is the original dblock file name, but that file doesn't exist.
- **Expected result**: 
The dblock file name is the dblock file name that was uploaded and that exists.

## Screenshots

![image](https://user-images.githubusercontent.com/40771998/65915840-3e3e7b00-e3a2-11e9-9e7f-5089b5b1f627.png)

![image](https://user-images.githubusercontent.com/40771998/65915601-c5d7ba00-e3a1-11e9-89a0-91ac0b792c64.png)

![image](https://user-images.githubusercontent.com/40771998/65915738-020b1a80-e3a2-11e9-8340-236ce07724de.png)

[...]

PR #3938, opened 2019-10-05
Ensure that dindex references correct dblock file after renames

warwickmm:
When support for parallel uploads was implemented, we overlooked a case where a failed put of a dblock file is retried with a different filename.  In this case, an instance of `TemporaryIndexVolume` was created in the `DataBlockProcessor` or `SpillCollectorProcess` prior to the attempted put.  The put would fail and then be retried with a different filename.  While the `Remotevolume` table in the database would reflect the new filename, the `TemporaryIndexVolume` (and resulting dindex file) still contained a reference to the old dblock filename.  This would cause issues in direct restores, etc., when the referenced dblock file could not be found.  

Now, the `TemporaryIndexVolume` is only created after the dblock put has completed.

This addresses issue #3932.

### DIFF
--- a/Duplicati/Library/Main/Operation/Backup/BackendUploader.cs
+++ b/Duplicati/Library/Main/Operation/Backup/BackendUploader.cs
@@ -75,15 +75,17 @@ namespace Duplicati.Library.Main.Operation.Backup
     {
         public FileEntryItem BlockEntry { get; }
         public BlockVolumeWriter BlockVolume { get; }
-        public FileEntryItem IndexEntry { get; }
-        public IndexVolumeWriter IndexVolume { get; }
+        public TemporaryIndexVolume IndexVolume { get; }
+        public Options Options { get; }
+        public BackupDatabase Database { get; }
 
-        public VolumeUploadRequest(BlockVolumeWriter blockvolume, FileEntryItem blockEntry, IndexVolumeWriter indexVolume, FileEntryItem indexEntry)
+        public VolumeUploadRequest(BlockVolumeWriter blockvolume, FileEntryItem blockEntry, TemporaryIndexVolume indexVolume, Options options, BackupDatabase database)
         {
             BlockVolume = blockvolume;
             BlockEntry = blockEntry;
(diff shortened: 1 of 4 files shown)


warwickmm:
This was testing by doing the following:
1. Configure a backup with a local destination.
2. Remove write permissions to the destination folder.
3. But a breakpoint [here](https://github.com/duplicati/duplicati/blob/c412bc9cae0db9780f7c219bed34c17bef5ef1c8/Duplicati/Library/Main/Operation/Backup/BackendUploader.cs#L363), and run the backup.
3. Note that the dblock file gets renamed:
        * `oldname`: "duplicati-babd58e212a18456eaed725138dbebcdc.dblock.zip"
        * `newname`: "duplicati-bedf0e82c6ed944dca25f5cdf1f2f05ad.dblock.zip"
4. Restore write permissions to the destination folder and allow the backup to complete.
5. Inspect the dlist file and verify that the new dblock file is referenced:
```
$ unzip -t duplicati-i5aada53b6e7e430fae928023554d6008.dindex.zip
Archive:  duplicati-i5aada53b6e7e430fae928023554d6008.dindex.zip
    testing: manifest                 OK
    testing: vol/duplicati-bedf0e82c6ed944dca25f5cdf1f2f05ad.dblock.zip   OK
    testing: list/mzEpIPXU7lkQxtVliTHOjxryITWC82eYt_4CJHqvFrw=   OK
    testing: list/cHiKWHMLLxYEy_yOJajbXWDiQmEVx5VTqXHeLoTT4LE=   OK
No errors detected in compressed data of duplicati-i5aada53b6e7e430fae928023554d6008.dindex.zip.
```

As before, the `Remotevolume` table in the database reflects the new dblock
filename:
Removevolume

kenkendk:
Awesome debugging!
One of the tests uses a randomly failing uploader, which is intended to catch errors such as this. 
But it has not caught this one, because it requires that we ignore the database contents, and only rely on the remote data.

anamarinapalmela:
helo

Issue #3861, opened 2019-08-15
Feature request:  Additional tray icon status styles


drwtsn32x:
Currently the tray icon will display a "Red X" style not only for error conditions, but also for warning and informational events.

Suggest we add two more styles:
* "Yellow exclamation" style to be used for Warnings (eg, backup job completed with warnings).
* "Blue information" style to be used for Informational events (eg, new version available).

If multiple alerts are active, it should use the style for the most severe alert.

Thoughts?

warwickmm:
What platform?  The GTK notification icons were supposed to be fixed in #3739.

drwtsn32x:
I am thinking of Windows - I had no idea that GTK had additional icons like that!  Would be nice if Windows did as well.

drwtsn32x:
PR 3943 implements the main feature I was looking for

PR #3943, opened 2019-10-08
Add TrayIcon warning state

drwtsn32x:
This PR adds a yellow(ish) exclamation warning icon:

normal-warning-yellow-5

To replace the red error icon currently used for the "IdleWarning" state. Here's what it looks like in action:

Capture

Zoomed:
Capture2

Notes:

* I did my best to edit the existing SVG icons and then save a multi-size ICO. I don't doubt that someone can do better. Please feel free to enhance.
* I only attempted to update this for the Windows platform.  I don't run a GUI on the Linux based systems I use, and I have no access to Mac OS X platform.


drwtsn32x:
If you think this should be a different shade of yellow let me know RGB values and I can update.  I tried a lighter yellow but there isn't enough contrast with the white exclamation mark.

Or maybe this should be orange instead - that seems to be the color used in Duplicati Web UI for warnings.

Thoughts?

warwickmm:
I think this partially addresses issue #3861.

drwtsn32x:
@warwickmm - yeah, I opened that issue (really a feature request) a couple months back so this was on my mind for a while. :)

I may have been wrong in that request about informational alerts triggering a red X.  I thought it happened in the past but I haven't been able to reproduce anything.  Version upgrade informational alerts, for instance, do not change the icon from the neutral/OK look.

Issue #3726, opened 2019-04-08
UI (dark theme): file name very hard to read when importing configurations


KommX:
- [x] I have searched open and closed issues for duplicates.

----------------------------------------

## Environment info

- **Duplicati version**: 2.0.4.16
- **Operating system**: Windows 10
- **Backend**: doesn't matter

## Description

I can't read the file name in the "Import backup configuration" section because it's dark on dark.

## Steps to reproduce

1. Activate the dark theme.
2. Choose "Add backup".
3. Choose "Import from file".
4. Click on the "Choose File" button
5. Select a file and click Open.

- **Actual result**:
The file name is located next to the "Choose File" button but it's very hard to read.
- **Expected result**:
I can read it, because a brighter color is used for the file name.

## Screenshots

![image](https://user-images.githubusercontent.com/7859245/55752505-81014500-5a48-11e9-96fa-238c4b11e9fc.png)

kees-z:
It looks like this in Chrome and Firefox. Internet Explorer and Edge have a different view:

![image](https://user-images.githubusercontent.com/24669568/56353988-fef4e700-61d2-11e9-85de-bfadee00eca8.png)

mithileshz:
Hey, can I pick this up?

warwickmm:
Of course.  Thanks @mithileshz.

PR #3946, opened 2019-10-13
UI (dark theme): file name very hard to read when importing configurations #3726

mithileshz:


warwickmm:
Looks good to me on Brave (Chromium). @mithileshz, there is a [comment](https://github.com/duplicati/duplicati/issues/3726#issuecomment-484438041) on issue #3726 on how Internet Explorer and Edge showed dark text on a light background.  What browsers did you test with?

mithileshz:
> Looks good to me on Brave (Chromium). @mithileshz, there is a [comment](https://github.com/duplicati/duplicati/issues/3726#issuecomment-484438041) on issue #3726 on how Internet Explorer and Edge didn't have this issue. What browsers did you test with?

I tested it with Chrome and it was fine on there. I checked it on IE and Edge and those browsers implement their own file pickers so this CSS change does not affect them.

Just a side question; there are the inputs for Passphrase for example, which still has a white background for the input. Should I try to make it black background with white text in them?

warwickmm:
> I checked it on IE and Edge and those browsers implement their own file pickers so this CSS change does not affect them.

That's great.  Thanks!

> Just a side question; there are the inputs for Passphrase for example, which still has a white background for the input. Should I try to make it black background with white text in them?

I think it would be nice to make things more consistent.  Let's do this in a separate pull request.

PR #3956, opened 2019-10-21
Feature/aftp update

kenkendk:
Possible fix for aftp relying on `SIZE` command which has non-compliant behavior on some servers.
Related forum post:
https://forum.duplicati.com/t/issues-to-address-to-get-out-of-beta/7871/75


warwickmm:
Looks reasonable to me.  Unless there are some other details to consider, I think we can merge and get this out for users to test.

Issue #3955, opened 2019-10-21
[2.0.4.31] Issue: cannot create a new backup job


sfahrenholz:
- [x] I have searched open and closed issues for duplicates.

----------------------------------------

## Environment info

- **Duplicati version**: 2.0.4.31
- **Operating system**: Windows 10
- **Backend**: Folder

## Description
After updating to 2.0.4.31_canary_2019-10-19 i want to add a new backup job. In the last step of the wizard, the information, "Unable to save schedule or backup object" are showing. The Job are not created.

## Steps to reproduce

1. Add a new Backup
2. Set the schedule plan
3. finished the backup job

## Screenshots
![image](https://user-images.githubusercontent.com/6977657/67178518-5880dd00-f3d3-11e9-80e8-27d631c781a2.png)

![image](https://user-images.githubusercontent.com/6977657/67178506-4dc64800-f3d3-11e9-8535-b2d5be435186.png)

there are no logs..

DarkSilence:
Same here. Using latest build

ts678:
Working here on Windows 10 on fresh Duplicati install with fresh job DB. I'm not well set up to upgrade, but I don't know if that's the key difference. Can anyone give me more details about the path you took?

ts678:
Google Translate thinks the error message from the image is German and translates to English as "The UTC time displayed after applying the offset must be between 0 and 10000". I noticed several Google search hits suggesting that the standard English error is "The UTC time represented when the offset is applied must be between year 0 and 10,000". Someone who's more familiar with C# time should look.

Is the @DarkSilence error the same message in German, or in some other language, or very different?

DarkSilence:
@ts678 I'll provide more details. My OS is Windows 7 (all latest patches installed). English (US). Bug appears when I create a new backup and on tab #4 I check „Automatically run backups.“ checkbox. From my searchings it doesn't matter what date\time I put below the checkbox. If the checkbox is set I get this error, when checkbox is unset there's no error. I get this message in english. It says: „The UTC time represented when the offset is applied must be between year 0 and 10,000.´“. Bug only appears only on the latest build of Duplicati. I installed the previous build and couldn't reproduce the bug there.

DarkSilence:
@ts678  Maybe this #3919 have some connections with the bug...

sfahrenholz:
The path of source and destination is not important in this case. 
for example
destination : C:\BackupFolder
source: C:\SourceFolder

gives the message.

[...]

PR #3960, opened 2019-10-24
Avoid possible invalid DateTimeOffset values

warwickmm:
When creating a new backup, a value of `DateTime.MinValue` can be provided to the `NormalizeDateTimeToEpochSeconds` method.  If `DateTimeOffset` is used, it will convert the provided `DateTime` to the UTC equivalent, which may result in a value outside of the allowable bounds.

See [this post](https://stackoverflow.com/questions/13799214/the-utc-time-represented-when-the-offset-is-applied-must-be-between-year-0-and-1/13799438#13799438) for more details.

This reverts the changes from pull request #3919.
This fixes issue #3955.

### DIFF
--- a/Duplicati/Library/Utility/Utility.cs
+++ b/Duplicati/Library/Utility/Utility.cs
@@ -957,7 +957,12 @@ namespace Duplicati.Library.Utility
 	/// <param name="input">The input time</param>
         public static long NormalizeDateTimeToEpochSeconds(DateTime input)
         {
-            return (new DateTimeOffset(input)).ToUnixTimeSeconds(); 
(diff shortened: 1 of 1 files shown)


ts678:
Looks good, and thanks for the comment explaining why the method being reverted has a pitfall.
[DateTimeOffset.MinValue Field](https://docs.microsoft.com/en-us/dotnet/api/system.datetimeoffset.minvalue?redirectedfrom=MSDN&view=netframework-4.8) Remark is still same as quoted in the [2012 Stack Overflow answer](https://stackoverflow.com/a/13799438).

kenkendk:
Great find! These bugs are really hard to find if you are not in an affected timezone (and apparently our CI server is not).

Issue #2220, opened 2016-12-31
"Exclude files whose names contain" causes error


moltenform:
Hi,
I'm seeing an error message when I try to use a "Exclude files whose names contain" filter.

### Version info

**Duplicati Version:** both 2.0.1.35 and github latest (e1a989ca3...)
**Operating System:** Win 8.1 x64
**Backend:** any

### Steps to reproduce
- Start duplicati.server.exe
- From the web interface,
- Menu->Add backup
- Enter a name, choose "No encryption"
- Choose a local folder path destination
- Use "Add a path directly" to provide a source path, include trailing \.
- Add filter of type
- "Exclude files whose names contain" and value "example"
- Click next, using default settings for all other settings.
- Click Run Now.

**Actual result:**

A red error immediately appears saying:

    Error while running simple-test-exclude
    parsing ".*example.*[^\]" - Unterminated [] set.

Here are the command line arguments that were created

    "...\Duplicati.CommandLine.exe" backup "file://C:\testduplic\simpletestexcludes" "C:\testduplic\testexcludes--input\\" --backup-name="simple-test-exclude" --dbpath="C:\...\Duplicati\Server\bin\Debug\OXXLAXDONK.sqlite" --encryption-module= --compression-module="zip" --dblock-size="50mb" --keep-time="3M" --no-encryption="true" --disable-module="console-password-input" --exclude="[.*example.*[^\]]"

I did look at AppUtils.js to investigate. Problem looks straightforward, the dirsep character \ is inadvertently escaping the ]. I would also have expected the generated regex to look something like

    .*example[^!]*

(using AppUtils.js's convention that ! is replaced with dirsep)

I am able to work around this by using a raw regular expression but thought I would notify that this feature doesn't seem to be working as intended,

agsfbm6:
After just having installed version 2.0.2.1 on Windows 10 64bit I have replicated the above error.

Cyberwizzard:
Version 2.0.3.3 still has this issue; the automatically generated regex is invalid

ts678:
Some forum discussion was in [A bug on regular expression when creating filters (Windows platform)](https://forum.duplicati.com/t/a-bug-on-regular-expression-when-creating-filters-windows-platform/4543?u=ts678) with reference citations and a proposed one-character fix, but it would of course need multi-platform testing.

downpoured:
[...]

PR #3967, opened 2019-10-27
Possible fix for issue #2220

kenkendk:
This PR escapes the directory separator character if that happens to be a backslash (as it is on Windows) as it would otherwise break the regular expressions that use backslash as an escape character.

It also changes the expression generated for the "Exclude files whose names contain" to use `.*x[^/]*` (where `x` is the expression to find and `/` is the directory separator) based on feedback from @drwtsn32x.

### DIFF
--- a/Duplicati/Server/webroot/ngax/scripts/services/AppUtils.js
+++ b/Duplicati/Server/webroot/ngax/scripts/services/AppUtils.js
@@ -131,8 +131,8 @@ backupApp.service('AppUtils', function($rootScope, $timeout, $cookies, DialogSer
             name: gettextCatalog.getString('Exclude files whose names contain'),
             key: '-file*',
             prefix: '-[.*',
-            suffix: '.*[^!]]',
-            rx: '\\-\\[\\.\\*([^\\!]+)\\.\\*\\[\\^\\!\\]\\]'
+            suffix: '[^!]*]',
+            rx: '\\-\\[\\.\\*[\\^\\!\\]\\*\\]'
         }, {
             name: gettextCatalog.getString('Exclude folder'),
             key: '-folder',
@@ -545,6 +545,10 @@ backupApp.service('AppUtils', function($rootScope, $timeout, $cookies, DialogSer
         }
 
         function matches(txt, n) {
+            // We need to escape dirsep if it is the backslash character
+            if (dirsep == '\\')
+                dirsep = '\\\\';
+
             var pre = apputils.replace_all(n.prefix || '', '!', dirsep);
             var suf = apputils.replace_all(n.suffix || '', '!', dirsep);
 
@@ -582,6 +586,10 @@ backupApp.service('AppUtils', function($rootScope, $timeout, $cookies, DialogSer
             return body;
 
         body = body || '';
+
+        // We need to escape dirsep if it is the backslash character
+        if (dirsep == '\\')
+            dirsep = '\\\\';
 
         var f = this.filterTypeMap[type];
         var pre = this.replace_all(f.prefix || '', '!', dirsep);


warwickmm:
I was testing this with the following source folder:
```
- test_source
  - 100M
  - 10M
  - 1M
  - 250M
  - backend_test.txt
  - bye.txt
  - dir
    - bye.txt
    - hello.txt
  - empty
  - hello.txt
```

The strange thing is, with version 2.0.4.32 I can't get "Exclude files whose names contain" to work with _any_ pattern on Linux.  My initial thought was that issue #2220 only affected Windows.  Are you seeing issues with this type of exclusion on Linux with 2.0.4.32?

With this branch, using "Exclude files whose names contain" with the pattern `100M`, the `test-filters` command yields the expected result:
```
Including source path: /home/warwickmm/no_backup/duplicati_tests/test_source/
Including path as no filters matched: /home/warwickmm/no_backup/duplicati_tests/test_source/dir/
Including path as no filters matched: /home/warwickmm/no_backup/duplicati_tests/test_source/bye.txt
Excluding path due to filter: /home/warwickmm/no_backup/duplicati_tests/test_source/100M => ([.*100M[^/]*])
Including path as no filters matched: /home/warwickmm/no_backup/duplicati_tests/test_source/hello.txt
Including path as no filters matched: /home/warwickmm/no_backup/duplicati_tests/test_source/10M
Including path as no filters matched: /home/warwickmm/no_backup/duplicati_tests/test_source/empty
Including path as no filters matched: /home/warwickmm/no_backup/duplicati_tests/test_source/250M
Including path as no filters matched: /home/warwickmm/no_backup/duplicati_tests/test_source/bacwarwickmmd_test.txt
Including path as no filters matched: /home/warwickmm/no_backup/duplicati_tests/test_source/1M
Including path as no filters matched: /home/warwickmm/no_backup/duplicati_tests/test_source/dir/bye.txt
Including path as no filters matched: /home/warwickmm/no_backup/duplicati_tests/test_source/dir/hello.txt
Matched 9 files (261.02 MB)
Return code: 0
```

Using "Exclude files whose names contain" with the pattern `1*M`, the `test-filters` command yields
```
Including source path: /home/warwickmm/no_backup/duplicati_tests/test_source/
Including path as no filters matched: /home/warwickmm/no_backup/duplicati_tests/test_source/dir/
Including path as no filters matched: /home/warwickmm/no_backup/duplicati_tests/test_source/bye.txt
Excluding path due to filter: /home/warwickmm/no_backup/duplicati_tests/test_source/100M => ([.*1*M[^/]*])
Including path as no filters matched: /home/warwickmm/no_backup/duplicati_tests/test_source/hello.txt
Excluding path due to filter: /home/warwickmm/no_backup/duplicati_tests/test_source/10M => ([.*1*M[^/]*])
Including path as no filters matched: /home/warwickmm/no_backup/duplicati_tests/test_source/empty
Excluding path due to filter: /home/warwickmm/no_backup/duplicati_tests/test_source/250M => ([.*1*M[^/]*])
Including path as no filters matched: /home/warwickmm/no_backup/duplicati_tests/test_source/bacwarwickmmd_test.txt
Excluding path due to filter: /home/warwickmm/no_backup/duplicati_tests/test_source/1M => ([.*1*M[^/]*])
Including path as no filters matched: /home/warwickmm/no_backup/duplicati_tests/test_source/dir/bye.txt
Including path as no filters matched: /home/warwickmm/no_backup/duplicati_tests/test_source/dir/hello.txt
Matched 6 files (24.61 KB)
Return code: 0
```
I think the exclusion of the file `250M` makes sense to me from a C# regex perspective, but I'm not sure if that coincides with what a user may expect, given the [documentation](https://duplicati.readthedocs.io/en/latest/appendix-d-filters/#syntax) which states

> \* specifies none or multiple characters

as opposed to the [.NET documentation](https://docs.microsoft.com/en-us/dotnet/standard/base-types/regular-expression-language-quick-reference), which says

> \* Matches the previous element zero or more times

Or, am I confused?

drwtsn32x:
> Or, am I confused?

It looks correct to me.  With regex `1*M` pattern says to match a literal `1` zero or more times followed by a literal `M`.  In such a simple pattern the `1*` basically does nothing, so your regex matched anything containing `M`.

The correct regex for your test might be `1.*M` which means match anything with a literal `1` followed by zero or more other characters of any type, and then followed by a literal `M`.  This would match `1M`, `10M`, `100M`, but not `250M`.

warwickmm:
Thanks @drwtsn32x.  The behavior of the regex pattern `1*M` makes sense to me.  I just feel that our [documentation](https://duplicati.readthedocs.io/en/latest/appendix-d-filters/#syntax) might be a bit misleading.  Since I didn't explicitly include `[]`, the documentation seems to imply that `*` follows globbing rules, and not regex rules.

> If you want to use file globbing to specify rules, ? and * are allowed placeholders. ? matches any single character. * specifies none or multiple characters. For example, *.txt would specify all .txt files.

> Rules can also be specified as regular expression. In this case put the regular expression (using .NET syntax) into hard brackets []. For example, [.*/IMG_\d{4}\.jpg] would specify all files named IMG_0000.jpg through IMG_9999.jpg.

drwtsn32x:
> Since I didn't explicitly include [], the documentation seems to imply that * follows globbing rules, and not regex rules.

That's exactly how I understand it, too.  If you didn't put the brackets around the pattern, it shouldn't have used regex. Maybe a bug?

warwickmm:
Either a bug, or the documentation is wrong.  Since it wasn't introduced in this branch, we can deal with that separately.

This looks good to me on Linux, but I don't have a Windows machine to check the problem described in #2220.  @drwtsn32x, would you be able to test this on Windows?

drwtsn32x:
@warwickmm  tested on Windows and it does seem to fix the problem.  Had to delete the broken filter generated before the update was applied, but once I readded the filter it worked as expected.

I also confirmed the behavior you saw. I entered "1*b" and it filtered out all files that contained just a "b".  So it does seem to be forcing regex matching on this field. Not sure if that is by design or an oversight.

drwtsn32x:
It is obvious once I look at the source. Even before the changes made in this PR, the prefix and suffix (for "exclude files whose names contain") have the square brackets.

"Exclude directories whose name contain" does not have those brackets in the prefix/suffix.  Can/should they be removed from the file-based one, too?  That way the user can choose whether or not to use regex.

warwickmm:
> "Exclude directories whose name contain" does not have those brackets in the prefix/suffix. Can/should they be removed from the file-based one, too? That way the user can choose whether or not to use regex.

That's a good idea.  The only other ones that include `[]` in the prefix/suffix are "Exclude regular expression" and "Include regular expression", which makes sense.

@kenkendk, what are your thoughts about removing the square brackets from prefix/suffix for the "Exclude files whose names contain" case?

kenkendk:
> @kenkendk, what are your thoughts about removing the square brackets from prefix/suffix for the "Exclude files whose names contain" case?

My intention of the feature was that it would only support plain names, meaning no wildcards or regexps.

The code is supposed to regex encode the user input, but currently does not:
https://github.com/duplicati/duplicati/blob/master/Duplicati/Server/webroot/ngax/scripts/services/AppUtils.js#L590

If we had this implemented, the example would have matched none of the files, but only a file with literal `1*M` in the name. We cannot remove the brackets, because the filter use a negative match (`[^/]`) which cannot be expressed with wildcards.

I am not sure what I would want to happen if the user enters a wildcard (or regex construct).

> The strange thing is, with version 2.0.4.32 I can't get "Exclude files whose names contain" to work with any pattern on Linux. My initial thought was that issue #2220 only affected Windows. Are you seeing issues with this type of exclusion on Linux with 2.0.4.32?

The expression generated in 2.0.4.32 is also wrong on Linux, because it can match anywhere in the path, not just the filename. That should let it exclude more files than expected. Based on my testing it did seem to filter correctly when building from source (equivalent to 2.4.32).

drwtsn32x:
@kenkendk - ok that makes sense. If a user wants to use wildcards or regex they can use some of the other filtering options anyway... 'exclude file' or 'exclude folder'.

The documentation should probably be updated to say the "whose name contain" options are designed to only accept literal strings and not wildcards/regex.

In any case this PR looks good to me for fixing issue 2220!

kenkendk:
> The documentation should probably be updated to say the "whose name contain" options are designed to only accept literal strings and not wildcards/regex.

Agreed! There should also be some sanitizing of the filter, otherwise it is easy to mess it up (i.e. typing `*.txt` will create an invalid filter).

> In any case this PR looks good to me for fixing issue 2220!

Yes, it is not solving all the issues, but at least it improves the current state.

ts678:
I'm wondering (although I don't see exactly how) this broke at least Windows "Exclude file".
Test began at Home on an existing backup with no filters, added one, looked in Edit as text.
I haven't actually saved and tested, or run the test command, so this is just an editing issue.
I'm not following the thread too closely on ideas like wildcards (or not), full paths, and docs.
Some of this might be change for the better, but I'm mainly picking on `Exclude file` here.

```
Exclude folders whose names contain string
2.0.4.32
-[.*string.*[^\]]
2.0.4.33
-[.*string[^\\]*] <-- the Windows fix for #2220
Exclude file string
2.0.4.32
-string
2.0.4.33
-string\\ <-- Don't trailing backslashes make the file exclude a folder exclude?
Exclude folder string
2.0.4.32
-*string*\ <-- Not certain how many Windows needs. Maybe this needed double too?
2.0.4.33
-string\\ <-- And we now have doubled, but changed the wildcard behavior as well.
```

drwtsn32x:
I think the problem is that the backslash should only be escaped when used in a regular expression.

Issue #3957, opened 2019-10-23
Permission issues on macOS Catalina


gfrewqpoiu:
- [x] I have searched open and closed issues for duplicates.

----------------------------------------

## Environment info

- **Duplicati version**: 2.0.4.31_canary_2019-10-19
- **Operating system**: macOS 10.15
- **Backend**: local (External Media)

## Description

On macOS Catalina, Apple has added new protections so that you have to grant Apps access to external storage devices, as well as to folders like Documents and Downloads separately. When I try to configure a job with a storage target of an external USB disk, Duplicati reports `System.UnauthorizedAccessException: Access to the path '/Volumes/MacSSD/Duplicati' is denied. ---> System.IO.IOException: Operation not permitted` instead of asking for access. I gave Duplicati full disk access, as well as mono and mono-sgen but that didn't change anything.

## Steps to reproduce

1. Create a new backup job
2. Select /Volumes/ANY_EXTERNAL_DRIVE/ as the target
3. Try to test the connection.

- **Actual result**:
Error message: `Failed to connect: Access to the path '/Volumes/MacSSD' is denied.`
- **Expected result**:
A popup asking for permission. Or Full Disk Access to work.

## Debug log

```C#
23. Okt. 2019 13:24: Request for http://localhost:8200/api/v1/remoteoperation/test gave error
System.UnauthorizedAccessException: Access to the path '/Volumes/MacSSD' is denied. ---> System.IO.IOException: Operation not permitted
   --- End of inner exception stack trace ---
  at System.IO.Enumeration.FileSystemEnumerator`1[TResult].CreateDirectoryHandle (System.String path, System.Boolean ignoreNotFound) [0x00032] in <80a26ecfb1f54efa9ad70b5be7520ce5>:0 
  at System.IO.Enumeration.FileSystemEnumerator`1[TResult]..ctor (System.String directory, System.IO.EnumerationOptions options) [0x00048] in <80a26ecfb1f54efa9ad70b5be7520ce5>:0 
  at System.IO.Enumeration.FileSystemEnumerable`1+DelegateEnumerator[TResult]..ctor (System.IO.Enumeration.FileSystemEnumerable`1[TResult] enumerable) [0x00000] in <80a26ecfb1f54efa9ad70b5be7520ce5>:0 
[...]

PR #3844, opened 2019-07-31
Update framework to 462

BlueBlock:
This PR upgrades all projects to .net 4.6.2 with support of standard2.0 and with all nuget packages now current.

- Upgrades all projects to 4.6.2
- Upgrade Travis and Appveyor unittesting package references
- Update all nuget packages

This is a good step to take to keep the projects and components moving forward while we also work on moving towards standard2.0

Why migrate to 4.6.2?

Standard and Core is definitely the end-game and this is a worthwhile effort but more long-term. Now that I've spent a good number of hours with the migration branch towards standard2.0 I'm of the opinion there is a lot more time needed to get it complete. They'll eventually they'll get resolved but there are still a number of hurdles.

For the **short-term** here is my summary of the choice that needs to be made:
- **stay on .net 4.5**
  - cannot upgrade the Azure nuget package (requires 4.5.2)
  - cannot upgrade the SharpAESCrypt nuget package (requires 4.6.2)
- **migrate to 4.5.2**
  - cannot upgrade the SharpAESCrypt nuget package (requires 4.6.2)
- **migrate to 4.6.2**
   - marginally increasing the minimum version requirement for Mono

I'm still of the opinion that moving to 4.6.2 is the best short-term strategy. With an long-term goal of Standard2.0

.NET 4.6.2 moves us just forward enough that all nuget packages can be updated and sets us up to transition to Standard2.0

IMHO

PR #3969, opened 2019-10-28
Ported `run-with-mono.sh` bash script to Objective C version

kenkendk:
This now works better with MacOS Catalina security features


Installer/OSX/launchers/duplicati-cli.m
```diff
@@ -0,0 +1,13 @@
+#import "run-with-mono.h"
+
+NSString * const ASSEMBLY = @"Duplicati.CommandLine.exe";
+NSString * const APP_NAME = @"Duplicati.CommandLine";
+int const MONO_VERSION_MAJOR = 4;
```
warwickmm: I think the changes in pull request #3844 that updated the target framework to 4.6.2 require a minimum mono version of  5.0.0 (see comment [here](https://github.com/duplicati/duplicati/pull/3844#issuecomment-525982565)).  Does this value need to be consistent with that?

kenkendk: Yes, I will update it to 5.0.

kenkendk:
The solution does not work for the commandline. The [NSTask documentation](https://developer.apple.com/documentation/foundation/nstask/1407627-standardoutput?language=objc) states that it should forward the output to the parent, but I see no output when invoking from the commandline.

warwickmm:
I don't have a Mac, so I can't provide much feedback here.  @kenkendk, please merge if you think this is good to go.

Also, do you think this addresses issue #3957?

kenkendk:
> Also, do you think this addresses issue #3957?

Yes. This fixes it.

PR #3980, opened 2019-11-09
Fix Passphrase input in dark theme

mithileshz:
Passphrase input in the dark theme had a light background and I noticed that so just adding this as a fix for it


warwickmm:
Looks good to me.

mithileshz:
Thank you @warwickmm

Issue #3982, opened 2019-11-10
"Stop after current file" has various problems with partial backups


ts678:
- [x] I have searched open and closed issues for duplicates.

----------------------------------------

## Environment info

- **Duplicati version**: 2.0.4.34 canary
- **Operating system**: Windows 10 Version 1903
- **Backend**: Local folder

## Description

Various issues around partial backup. Rather than file separate issues now, this is a collection of issues.

["Stop after current file" with --usn-policy may never back up remaining files #3971](https://github.com/duplicati/duplicati/issues/3971) was separately filed, and possibly has a different cause than the ones here which are primarily on how backups are now split into full and partial, with record-keeping in the database and backup, and attempts to make things like retention work right by hiding partial backups after a full exists. This breaks things because some code is going by the view-after-hiding, some is going with the raw view, and translating between may break.

[Fix 'stop after current file' #3836](https://github.com/duplicati/duplicati/pull/3836#issuecomment-546687741) has technical info. This issue is more of a hopefully-reproducible tour.

* Partial backup produces a Warning result without indication of why. Unknown if this is intentional.
* Partial backup Repair regenerates missing dlist file without a `fileset` file to show "partial" status.
* Partial backup marking is on all Direct restore versions, whether that version is really full or partial.
* Hidden partial backups hold backend storage forever.
* Hidden partial backups cannot be referenced by user.
* Hidden partial backups might not be reachable by Duplicati, unless direct queries to DB are used.
* Hidden partial backups break Duplicati when direct DB queries are translated to version for users.
One is probably in the [number translator](https://github.com/duplicati/duplicati/commit/2826dbe6f04497bcfdb88a9da96ea7e65338f055) for "Unexpected difference in fileset" errors (not tested).
* There's no telling how much other code is now (or may someday be) broken due to the dual map. Having only one way to map Fileset ID to its version (and vice versa) just didn't have this problem.

## Steps to reproduce

1. Make a C:\stop test folder holding two Duplicati executables zip files. Mine are about 14 MB each.
2. Slow with Remote Volume Size at 1 MB, --throttle-download and --throttle-upload at 100 KByte/s
3. Set Backup retention to keep 1 backup.
4. To spot when to Stop, and see other info, open a tab to About --> Show log --> Live --> Verbose
5. "Run now" Backup, go to live log tab, wait for uploading to begin, then do "Stop after current file"
6. "Run now" Backup. There will be a third backup later, but partial-then-full is a good starting point.

[...]

PR #3984, opened 2019-11-10
Issue warning when backup is cancelled by user

warwickmm:
This issues a warning when a backup is cancelled by the user (e.g., by clicking "Stop after current file"). 
 Previously, a yellow warning icon would be displayed but the logs did not contain any details regarding the cause of the warning.

This concerns issue #3982.


warwickmm:
Now, the log view shows:

image

PR #3989, opened 2019-11-13
OneDrive v2: Wrap requests, responses, and contents in using blocks

tygill:
This is a cleanup change to the MicrosoftGraphBackend (the base for the OneDrive v2 and SharePoint v2 backends). It adds explicit using blocks to dispose the requests, responses, and various content objects.

I also fixed a few bugs I noticed while doing this, including one content object which wasn't being attached to the request (to set behavior for an upload session - apparently the default was working before), and a few cases where lazy execution might have been preventing exceptions from being properly caught and rethrown (since on Mono the auto create logic for creating the base folder hasn't been working for me).

My hope is that this may help with the memory leaks on Mono on Debian reported here:
https://forum.duplicati.com/t/onedrive-v2-memory-leak/6112/26

### DIFF
--- a/Duplicati/Library/Backend/OneDrive/MicrosoftGraphBackend.cs
+++ b/Duplicati/Library/Backend/OneDrive/MicrosoftGraphBackend.cs
@@ -173,9 +173,11 @@ namespace Duplicati.Library.Backend
                     UploadSession uploadSession = this.Post<UploadSession>(string.Format("{0}/root:{1}{2}:/createUploadSession", this.DrivePrefix, this.RootPath, NormalizeSlashes(dnsTestFile)), null);
 
                     // Canceling an upload session is done by sending a DELETE to the upload URL
-                    var request = new HttpRequestMessage(HttpMethod.Delete, uploadSession.UploadUrl);
-                    var response = this.m_client.SendAsync(request).Await();
-                    this.CheckResponse(response);
+                    using (var request = new HttpRequestMessage(HttpMethod.Delete, uploadSession.UploadUrl))
+                    using (var response = this.m_client.SendAsync(request).Await())
+                    {
+                        this.CheckResponse(response);
+                    }
 
                     this.dnsNames = new[]
                         {
@@ -269,21 +271,17 @@ namespace Duplicati.Library.Backend
 
         public IEnumerable<IFileEntry> List()
         {
-            try
+            foreach (DriveItem item in this.Enumerate<DriveItem>(string.Format("{0}/root:{1}:/children", this.DrivePrefix, this.RootPath)))
             {
-                return this.Enumerate<DriveItem>(string.Format("{0}/root:{1}:/children", this.DrivePrefix, this.RootPath))
-                    .Where(item => item.IsFile && !item.IsDeleted) // Exclude non-files and deleted items (not sure if they show up in this listing, but make sure anyway)
-                    .Select(item =>
-                        new FileEntry(
-                            item.Name,
-                            item.Size ?? 0, // Files should always have a size, but folders don't need it
-                            item.FileSystemInfo?.LastAccessedDateTime?.UtcDateTime ?? new DateTime(),
-                            item.FileSystemInfo?.LastModifiedDateTime?.UtcDateTime ?? item.LastModifiedDateTime?.UtcDateTime ?? new DateTime()));
-            }
-            catch (DriveItemNotFoundException ex)
-            {
-                // If there's an 'item not found' exception here, it means the root folder didn't exist.
-                throw new FolderMissingException(ex);
+                // Exclude non-files and deleted items (not sure if they show up in this listing, but make sure anyway)
+                if (item.IsFile && !item.IsDeleted)
+                {
+                    yield return new FileEntry(
+                        item.Name,
+                        item.Size ?? 0, // Files should always have a size, but folders don't need it
+                        item.FileSystemInfo?.LastAccessedDateTime?.UtcDateTime ?? new DateTime(),
+                        item.FileSystemInfo?.LastModifiedDateTime?.UtcDateTime ?? item.LastModifiedDateTime?.UtcDateTime ?? new DateTime());
+                }
             }
         }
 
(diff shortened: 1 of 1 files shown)


Duplicati/Library/Backend/OneDrive/MicrosoftGraphBackend.cs
```diff
@@ -269,21 +271,17 @@ public void CreateFolder()
...
-                return this.Enumerate<DriveItem>(string.Format("{0}/root:{1}:/children", this.DrivePrefix, this.RootPath))
-                    .Where(item => item.IsFile && !item.IsDeleted) // Exclude non-files and deleted items (not sure if they show up in this listing, but make sure anyway)
-                    .Select(item =>
-                        new FileEntry(
-                            item.Name,
-                            item.Size ?? 0, // Files should always have a size, but folders don't need it
-                            item.FileSystemInfo?.LastAccessedDateTime?.UtcDateTime ?? new DateTime(),
-                            item.FileSystemInfo?.LastModifiedDateTime?.UtcDateTime ?? item.LastModifiedDateTime?.UtcDateTime ?? new DateTime()));
-            }
-            catch (DriveItemNotFoundException ex)
```
verhoek: Can you briefly explain why this catch vanished?

Duplicati/Library/Backend/OneDrive/MicrosoftGraphBackend.cs
```diff
@@ -353,89 +356,95 @@ public async Task PutAsync(string remotename, Stream stream, CancellationToken c
...
-
-                // If the stream's total length is less than the chosen fragment size, then we should make the buffer only as large as the stream.
-                int bufferSize = (int)Math.Min(this.fragmentSize, stream.Length);
-
-                byte[] fragmentBuffer = new byte[bufferSize];
-                int read = 0;
-                for (int offset = 0; offset < stream.Length; offset += read)
+                using (HttpRequestMessage createSessionRequest = new HttpRequestMessage(HttpMethod.Post, string.Format("{0}/root:{1}{2}:/createUploadSession", this.DrivePrefix, this.RootPath, NormalizeSlashes(remotename))))
+                /* Indicate that we want to replace any existing content with this new data we're uploading */
+                using (createSessionRequest.Content = this.PrepareContent(new UploadSession() { Item = new DriveItem() { ConflictBehavior = ConflictBehavior.Replace } }))
```
verhoek: This was forgotten before I guess as you mentioned in your PR.

Duplicati/Library/Backend/OneDrive/MicrosoftGraphBackend.cs
tygill: This catch didn't vanish, but was moved into the Enumerate method at the bottom of this file (this is the only place Enumerate is called).

The reason for this is mainly to make sure that the lazy evaluation of the select (and of any other requests) here don't cause the exception mapping from the internal one to Duplicati's FolderNotFound one to be missed. This also moves the exception handling closer to the source.

Duplicati/Library/Backend/OneDrive/MicrosoftGraphBackend.cs
tygill: That's what it seems like (which is particularly embarrassing as I was the one to originally write this code). I'm also not sure why I missed all the disposing then..

Duplicati/Library/Backend/OneDrive/MicrosoftGraphBackend.cs
verhoek: I understand, but I don't see how the Enumerate method is connected to the List method. Can you explain?

tygill: List now calls `foreach (DriveItem item in this.Enumerate<DriveItem>(...))` to iterate rather than returning `this.Enumerate<DriveItem>(...).Where(...).Select(...)`

Duplicati/Library/Backend/OneDrive/MicrosoftGraphBackend.cs
```diff
@@ -491,33 +500,45 @@ protected T Get<T>(string url)
...
 
         private T SendRequest<T>(HttpMethod method, string url, T body) where T : class
         {
-            var request = new HttpRequestMessage(method, url);
-            if (body != null)
+            using (var request = new HttpRequestMessage(method, url))
+            using (request.Content = this.PrepareContent(body))
             {
-                request.Content = this.PrepareContent(body);
+                return this.SendRequest<T>(request);
```
warwickmm: Is it now possible for `request.Content` to be `null`?  Does that have any side effects?

tygill: It was previously possible for it to be null - the check was just made before calling Prepare Content then and now is done inside, so that the using block can capture the content to dispose it (it could be done in a try/finally, but using is smart enough to do a null check before attempting to dispose). There shouldn't be any change in behavior beyond that - this method is used both for requests with and without bodies (IIRC, this and the method it calls with are the two used for all requests for this backend).

warwickmm: You're right.  Sounds good.

warwickmm:
Looks good to me.  Thanks, and it's nice to see you guys again.  :grinning:

Duplicati/Library/Backend/OneDrive/MicrosoftGraphBackend.cs
verhoek: I meant the enumerate function below in the code, where the driveitemnotfound exception is caught and you referred to earlier. It involves a GraphCollection.

warwickmm:
A user has reported an issue with Sharepoint using v2.0.4.35 with these changes (v2.0.4.34 appears to work fine):

https://forum.duplicati.com/t/bad-request-error-from-request-nameconflictbehavior-must-be-fail-replace-or-rename/8578

PR #3991, opened 2019-11-20
Only warn if backend delete failure is confirmed

drwtsn32x:
Current versions of Duplicati will flag the job with a warning when a back-end file deletion results in an exception (fails).  When that happens, Duplicati smartly double checks if this failure is legitimate by doing a directory listing to test for the presence of the file.  If the file is not there, the deletion either worked or is unnecessary.  If the file IS present, then a failure did occur and the exception is thrown.

The problem with the current code is that the Warning flag is set prior to the double checking if the delete failure is valid.  This change moves the Warning so that it only is triggered when the failure is confirmed (the file is still present on the back end).

In my testing this has solved some spurious warning messages I was getting occasionally with backups to the B2 back-end.

### DIFF
--- a/Duplicati/Library/Main/BackendManager.cs
+++ b/Duplicati/Library/Main/BackendManager.cs
@@ -1095,7 +1095,7 @@ namespace Duplicati.Library.Main
 
                 if (isFileMissingException || (wr != null && wr.StatusCode == System.Net.HttpStatusCode.NotFound))
                 {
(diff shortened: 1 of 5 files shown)


drwtsn32x:
Discussion here:  https://forum.duplicati.com/t/warning-at-end-of-job-after-upload-retry/8480

Thanks ts678 for assistance with troubleshooting.

drwtsn32x:
oops, this got fouled up when I merged another pull request with my fork. will resubmit

PR #3992, opened 2019-11-20
Fix issues related to partial backups

warwickmm:
This fixes some issues regarding how we treated partial backups.  Several inconsistencies arose by hiding partial backups from restore lists, etc., which resulted in some backup versions begin inaccessible to certain operations.  The approach here is to treat partial backups almost like full backups, except for some special cases (e.g., in the retention policy so that we do not discard a full backup in favor of a partial one).

Some notable changes:
* For "keep a specific number of backups", a partial backup can be deleted if it is followed by a full backup.
* For a "retention rule" (e.g., `1W:1D,4W:1W,12M:1M`), partials are not removed, and a full backup will not be discarded in favor of a partial one.
* For "delete backups that are older than", backups are kept if they are recent enough.  If all of the recent backups are partial, we continue to keep backup versions until we encounter a full backup.
* When restoring without a local database ("direct restore"), the partial/full state is correctly displayed in the restore list.
* When repairing the database due to missing `dlist` files, the newly created `dlist` files are no longer missing the `fileset` file.
* When recreating `dlist` files after a purge operation, the new `dlist` files are no longer missing the `fileset` file.
* Some simple unit tests were added to verify the behaviors of the various retention strategies.

This addresses issue #3982.

### DIFF
--- a/Duplicati/Library/Main/Operation/Backup/UploadSyntheticFilelist.cs
+++ b/Duplicati/Library/Main/Operation/Backup/UploadSyntheticFilelist.cs
@@ -111,6 +111,7 @@ namespace Duplicati.Library.Main.Operation.Backup
                                 foreach (var p in options.ControlFiles.Split(new char[] {System.IO.Path.PathSeparator}, StringSplitOptions.RemoveEmptyEntries))
                                     fsw.AddControlFile(p, options.GetCompressionHintFromFilename(p));
 
+                            fsw.CreateFilesetFile(false);
                             var newFilesetID = await database.CreateFilesetAsync(fsw.VolumeID, fileTime);
                             await database.LinkFilesetToVolumeAsync(newFilesetID, fsw.VolumeID);
                             await database.AppendFilesFromPreviousSetAsync(null, newFilesetID, prevId, fileTime);
--- a/Duplicati/Library/Main/Operation/Backup/UploadSyntheticFilelist.cs
+++ b/Duplicati/Library/Main/Operation/Backup/UploadSyntheticFilelist.cs
@@ -111,6 +111,8 @@ namespace Duplicati.Library.Main.Operation.Backup
                                 foreach (var p in options.ControlFiles.Split(new char[] {System.IO.Path.PathSeparator}, StringSplitOptions.RemoveEmptyEntries))
                                     fsw.AddControlFile(p, options.GetCompressionHintFromFilename(p));
 
+                            // We declare this to be a partial backup since the synthetic filelist is only created
+                            // when a backup is interrupted.
                             fsw.CreateFilesetFile(false);
                             var newFilesetID = await database.CreateFilesetAsync(fsw.VolumeID, fileTime);
                             await database.LinkFilesetToVolumeAsync(newFilesetID, fsw.VolumeID);
--- a/Duplicati/Library/Main/Operation/DeleteHandler.cs
+++ b/Duplicati/Library/Main/Operation/DeleteHandler.cs
@@ -221,34 +221,41 @@ namespace Duplicati.Library.Main.Operation
             var fullVersionsKeptCount = 0;
             if (fullVersionsToKeep > 0 && fullVersionsToKeep < backupsRemaining.Count)
             {
-                // keep the number of full backups specified in fullVersionsToKeep.
-                // once the last full backup t okeep is found, also keep the partials immediately after it the full backup.
-                // add the remainder of full and partial backups to toDelete
-                bool foundLastFullBackupToKeep = false;
+                ISet<DateTime> intermediatePartials = new HashSet<DateTime>();
+                bool haveFullBackup = false;
+
+                // Keep the number of full backups specified in fullVersionsToKeep.
+                // Remove partial backups that are surrounded by full backups.
+                // Once enough versions are kept, delete all older backups.
                 foreach (var backup in backupsRemaining)
                 {
-                    bool isFullBackup;
+                    bool isFullBackup = db.IsFilesetFullBackup(backup);
+                    if (isFullBackup)
+                    {
+                        if (haveFullBackup)
+                        {
+                            toDelete.AddRange(intermediatePartials);
+                            intermediatePartials.Clear();
+                        }
+                        haveFullBackup = true;
+                    }
+                    else
+                    {
+                        intermediatePartials.Add(backup);
+                    }
+
                     if (fullVersionsKeptCount < fullVersionsToKeep)
                     {
-                        isFullBackup = db.IsFilesetFullBackup(backup);
                         // count only a full backup
                         if (fullVersionsKeptCount < fullVersionsToKeep && isFullBackup)
                         {
                             fullVersionsKeptCount++;
                         }
-                        continue;
                     }
-                    // do not include any partial backup that precedes the last full backup
-                    if (!foundLastFullBackupToKeep)
+                    else
                     {
-                        isFullBackup = db.IsFilesetFullBackup(backup);
-                        if (!isFullBackup)
-                        {
-                            continue;
-                        }
-                        foundLastFullBackupToKeep = true;
+                        toDelete.Add(backup);
                     }
-                    toDelete.Add(backup);
                 }
             }
 
--- a/Duplicati/Library/Main/Operation/DeleteHandler.cs
+++ b/Duplicati/Library/Main/Operation/DeleteHandler.cs
@@ -254,8 +254,7 @@ namespace Duplicati.Library.Main.Operation
 
                     if (fullVersionsKeptCount < fullVersionsToKeep)
                     {
-                        // count only a full backup
-                        if (fullVersionsKeptCount < fullVersionsToKeep && isFullBackup)
+                        if (isFullBackup)
                         {
                             fullVersionsKeptCount++;
                         }
--- a/Duplicati/Library/Main/Operation/DeleteHandler.cs
+++ b/Duplicati/Library/Main/Operation/DeleteHandler.cs
@@ -239,16 +239,17 @@ namespace Duplicati.Library.Main.Operation
                     bool isFullBackup = db.IsFilesetFullBackup(backup);
                     if (isFullBackup)
                     {
-                        if (haveFullBackup)
-                        {
-                            toDelete.AddRange(intermediatePartials);
-                            intermediatePartials.Clear();
-                        }
+                        toDelete.AddRange(intermediatePartials);
+                        intermediatePartials.Clear();
                         haveFullBackup = true;
                     }
                     else
                     {
-                        intermediatePartials.Add(backup);
+                        // Only consider a partial backup for deletion if we have already encountered a full backup.
+                        if (haveFullBackup)
+                        {
+                            intermediatePartials.Add(backup);
+                        }
                     }
 
                     if (fullVersionsKeptCount < fullVersionsToKeep)
--- a/Duplicati/Library/Main/Operation/DeleteHandler.cs
+++ b/Duplicati/Library/Main/Operation/DeleteHandler.cs
@@ -232,7 +232,7 @@ namespace Duplicati.Library.Main.Operation
                 bool haveFullBackup = false;
 
                 // Keep the number of full backups specified in fullVersionsToKeep.
-                // Remove partial backups that are surrounded by full backups.
+                // Remove partial backups that are followed by full backups.
                 // Once enough versions are kept, delete all older backups.
                 foreach (var backup in backupsRemaining)
                 {
--- a/Duplicati/Library/Main/Operation/DeleteHandler.cs
+++ b/Duplicati/Library/Main/Operation/DeleteHandler.cs
@@ -225,43 +225,29 @@ namespace Duplicati.Library.Main.Operation
             // and remove oldest backups while there are still more backups than should be kept as specified via option
             var backupsRemaining = sortedAllBackups.Except(toDelete).ToList();
             var fullVersionsToKeep = m_options.KeepVersions;
-            var fullVersionsKeptCount = 0;
             if (fullVersionsToKeep > 0 && fullVersionsToKeep < backupsRemaining.Count)
             {
+                int fullVersionsKept = 0;
                 ISet<DateTime> intermediatePartials = new HashSet<DateTime>();
-                bool haveFullBackup = false;
 
-                // Keep the number of full backups specified in fullVersionsToKeep.
-                // Remove partial backups that are followed by full backups.
-                // Once enough versions are kept, delete all older backups.
-                foreach (var backup in backupsRemaining)
+                // Enumerate the collection starting from the most recent full backup.
+                foreach (DateTime backup in backupsRemaining.SkipWhile(x => !db.IsFilesetFullBackup(x)))
                 {
-                    bool isFullBackup = db.IsFilesetFullBackup(backup);
-                    if (isFullBackup)
+                    if (fullVersionsKept >= fullVersionsToKeep)
                     {
+                        // If we have enough full backups, delete all older backups.
+                        toDelete.Add(backup);
+                    }
+                    else if (db.IsFilesetFullBackup(backup))
+                    {
+                        // We can delete partial backups that are surrounded by full backups.
                         toDelete.AddRange(intermediatePartials);
                         intermediatePartials.Clear();
-                        haveFullBackup = true;
+                        fullVersionsKept++;
                     }
                     else
                     {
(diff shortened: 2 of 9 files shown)


Duplicati/Library/Main/Operation/DeleteHandler.cs
```diff
@@ -204,11 +204,18 @@ private DateTime[] GetFilesetsToDelete(Database.LocalDeleteDatabase db, DateTime
...
 
-            // Remove backups that are older than date specified via option
+            // Remove backups that are older than date specified via option while ensuring
+            // that we always have at least one full backup.
             var keepTime = m_options.KeepTime;
             if (keepTime.Ticks > 0)
             {
-                toDelete.AddRange(sortedAllBackups.SkipWhile(x => x >= keepTime));
+                bool haveFullBackup = false;
+                toDelete.AddRange(sortedAllBackups.SkipWhile(x =>
```
verhoek: What about 
                toDelete.AddRange(sortedAllBackups.Where(x => x <= keepTime)
                    .Where(x =>
                    {
                        haveFullBackup = haveFullBackup || db.IsFilesetFullBackup(x);
                        return haveFullBackup;
                    }));
?

warwickmm: Since `sortedAllBackups` are ordered (descending), I liked how `SkipWhile` would avoid enumerating the entire sequence.

Also, I'm not sure if I'm reading it correctly, but does your example only look for "old" backups that are full?  If one of the "recent" backups is full, then there's no need to keep any backups prior to `keepTime`.

verhoek: That's true, then your code is correct. I had to think about your code yesterday evening for a second, I understood it the way you explained. Could you write a test for it that we make sure that logic is preserved?
(btw. I need more time to review this since this PR is a bit more delicate)

Duplicati/Library/Main/Operation/ListFilesHandler.cs
```diff
@@ -203,9 +204,20 @@ orderby n.Key
             return filelistFilter(parsedlist).ToList();
         }
 
-        public static IEnumerable<Library.Interface.IListResultFileset> CreateResultSequence(IEnumerable<KeyValuePair<long, Volumes.IParsedVolume>> filteredList)
+        private static IEnumerable<IListResultFileset> CreateResultSequence(IEnumerable<KeyValuePair<long, IParsedVolume>> filteredList, BackendManager backendManager, Options options)
         {
-            return (from n in filteredList select (Library.Interface.IListResultFileset)(new ListResultFileset(n.Key, BackupType.PARTIAL_BACKUP, n.Value.Time.ToLocalTime(), -1, -1))).ToArray();
+            List<IListResultFileset> list = new List<IListResultFileset>();
+            foreach (KeyValuePair<long, IParsedVolume> entry in filteredList)
```
verhoek: Can you add some comments to this function what is happening?

warwickmm: Good idea.  I added a comment in revision fa3ffb75400f4e954e8b1f62ab4bc050ea9d330a to explain why the `dlist` files are being downloaded, as well as the resulting impact on performance.

Duplicati/Library/Main/Operation/DeleteHandler.cs
warwickmm: > I need more time to review this since this PR is a bit more delicate

No problem.  Thanks for your help reviewing.  In case you haven't noticed, there is a lot of related discussion in issue #3982 that may help explain some of the changes made here.

Duplicati/Library/Main/Operation/ListFilesHandler.cs
warwickmm: Revision afecf891dab745a7f06e74e9b47687e12a09e3a1 adds a test for the `--keep-time` retention logic.  I'll try and write a test for the `--keep-versions` logic soon.

Duplicati/Library/Main/Operation/DeleteHandler.cs
warwickmm: Revision afecf891dab745a7f06e74e9b47687e12a09e3a1 adds a test for the `--keep-time` retention logic.  I'll try and write a test for the `--keep-versions` logic soon.

warwickmm:
I suspect that there is a bug in my `keep-versions` logic.  I'll have to investigate some more...

warwickmm:
Ok, I think I've fixed the bug in the `keep-versions` logic.  Previously, if the most recent backups were partial, they were deleted when they should have been kept.  A partial backup should only be deleted if it is followed by a full backup.

I tested this by doing the following:
1. Run a full backup.
2. Run a partial backup.
3. Run a partial backup.
4. Run a full backup.
5. Run a partial backup.
6. Set `keep-versions` to `2`.
7. Run a partial backup and notice that the backups from steps 1, 4, 5, and 7 were kept.
8. Run a full backup and notice that the backups from steps, 4 and 8 were kept.

I also added a unit test to help verify the behavior.

verhoek:
>
> 8. Run a full backup and notice that the backups from steps, 4 and 8 were kept.
> 
> I also added a unit test to help verify the behavior.

Point 8 I cannot find in your unit tests, the rest I can follow. Do you still want to add that? Very good that you added these tests.

warwickmm:
Revision 68e7844daeb1a4c9cfb1e3ac404d5ae23df34368 adds a test for the case described in Step 8, where the last backup was full.

Duplicati/Library/Main/Operation/DeleteHandler.cs
```diff
@@ -221,34 +228,41 @@ private DateTime[] GetFilesetsToDelete(Database.LocalDeleteDatabase db, DateTime
             var fullVersionsKeptCount = 0;
             if (fullVersionsToKeep > 0 && fullVersionsToKeep < backupsRemaining.Count)
```
verhoek: Does this make sense? If fullVersionsToKeep is like 50 and backupsRemaining is 3, with 1 full, 1 partial and one full, then we still want to clean the partial in between I guess? Then we should still enter the clean up logic within the if.

Duplicati/Library/Main/Operation/DeleteHandler.cs
```diff
@@ -221,34 +228,41 @@ private DateTime[] GetFilesetsToDelete(Database.LocalDeleteDatabase db, DateTime
             var fullVersionsKeptCount = 0;
             if (fullVersionsToKeep > 0 && fullVersionsToKeep < backupsRemaining.Count)
             {
-                // keep the number of full backups specified in fullVersionsToKeep.
-                // once the last full backup t okeep is found, also keep the partials immediately after it the full backup.
-                // add the remainder of full and partial backups to toDelete
-                bool foundLastFullBackupToKeep = false;
+                ISet<DateTime> intermediatePartials = new HashSet<DateTime>();
```
verhoek: I think it is good to split this function in a part for keep time and keep version. I think it is excellent you wrote tests for them. If the logic is contained in separate functions, one can write tests just against such functions instead of going through the entire duplicati stack with controller etc. to reach them.

Duplicati/Library/Main/Operation/DeleteHandler.cs
```diff
@@ -221,34 +228,41 @@ private DateTime[] GetFilesetsToDelete(Database.LocalDeleteDatabase db, DateTime
...
                 {
-                    bool isFullBackup;
-                    if (fullVersionsKeptCount < fullVersionsToKeep)
+                    bool isFullBackup = db.IsFilesetFullBackup(backup);
+                    if (isFullBackup)
                     {
-                        isFullBackup = db.IsFilesetFullBackup(backup);
-                        // count only a full backup
-                        if (fullVersionsKeptCount < fullVersionsToKeep && isFullBackup)
+                        toDelete.AddRange(intermediatePartials);
```
verhoek: Maybe there is a fancier way to do the iterating in a few lines like you did with keepTime, using linq. Like get the last n or n-1 full backups and throw away the rest, plus some pending partial ones.

warwickmm:
I'm beginning to suspect that the logic in [ApplyRetentionPolicy](https://github.com/duplicati/duplicati/blob/f3bdd6c732328bd9c3360be5b48f2c6220593ad6/Duplicati/Library/Main/Operation/DeleteHandler.cs#L271) is incorrect (and has been incorrect for a while).  Since `clonedBackupList` is [sorted by ascending time](https://github.com/duplicati/duplicati/blob/f3bdd6c732328bd9c3360be5b48f2c6220593ad6/Duplicati/Library/Main/Operation/DeleteHandler.cs#L313), in the case where `backup - lastKept.Value` is a small enough timespan, shouldn't we delete `lastKept` (the older backup) instead of `backup` [here](https://github.com/duplicati/duplicati/blob/f3bdd6c732328bd9c3360be5b48f2c6220593ad6/Duplicati/Library/Main/Operation/DeleteHandler.cs#L345)?  @verhoek, can you provide a second opinion on this?

Duplicati/Library/Main/Operation/DeleteHandler.cs
warwickmm: My thinking was that if there are fewer backups remaining than the number we want to keep, we might as well keep them all (assuming that storage space is the motivation for the limit).  The partial "might" still be useful in some cases.  I think the existence of partials should be relatively rare, so it doesn't hurt to be conservative and not delete them unless we can make a strong case for doing so.

Duplicati/Library/Main/Operation/DeleteHandler.cs
warwickmm: I think that's a great idea.  However, it sounds like a larger refactoring that might require rewriting the tests to create entries in the database without actually running the backups.  If it's ok with you, I'd like to leave this for a future pull request.

Going through the controller does have a slight appeal to me, as it exercises the handling of the cancellation tokens...

Duplicati/Library/Main/Operation/DeleteHandler.cs
warwickmm: I tried several different approaches and ended up with the one in revision 20eb785f098dfa5315f270687a6824b4374748ee.  Let me know if it looks correct and if you think it is more readable.

warwickmm:
I added a test for the `retention-policy` logic in revision 30fddd4b913afdb7debd5b2ab3d3aea80fb2afd7.  Sorry if this pull request is getting too large.  If it's difficult to review, I can move some of these tests to future pull requests.

Duplicati/Library/Main/Operation/DeleteHandler.cs
```diff
@@ -218,37 +225,30 @@ private DateTime[] GetFilesetsToDelete(Database.LocalDeleteDatabase db, DateTime
...
-                // keep the number of full backups specified in fullVersionsToKeep.
-                // once the last full backup t okeep is found, also keep the partials immediately after it the full backup.
-                // add the remainder of full and partial backups to toDelete
-                bool foundLastFullBackupToKeep = false;
-                foreach (var backup in backupsRemaining)
+                int fullVersionsKept = 0;
+                ISet<DateTime> intermediatePartials = new HashSet<DateTime>();
+
+                // Enumerate the collection starting from the most recent full backup.
+                foreach (DateTime backup in backupsRemaining.SkipWhile(x => !db.IsFilesetFullBackup(x)))
```
verhoek: With this we do not get partial backups, but within the foreach you try to remove partial backups?

warwickmm: Sorry, I'm not sure I understand.  `SkipWhile`will skip elements until the first occurrence of `false`, and then return the rest of the sequence.  So, I think this will only skip partial backups at the beginning of the sequence?

verhoek: You're right, I misread the description of skipwhile.

Duplicati/Library/Main/Operation/Backup/UploadSyntheticFilelist.cs
```diff
@@ -111,6 +111,7 @@ orderby n.Key
                                 foreach (var p in options.ControlFiles.Split(new char[] {System.IO.Path.PathSeparator}, StringSplitOptions.RemoveEmptyEntries))
                                     fsw.AddControlFile(p, options.GetCompressionHintFromFilename(p));
 
+                            fsw.CreateFilesetFile(false);
```
verhoek: The comment of the commit with this change is useful to put in here as a comment too - if code is refactored that information is lost in the git history.

verhoek:
The changes you made so far seem to make sense, I did not test the actual build though. I read in the issue thread that @drwtsn32x tested some intermediate state, maybe @drwtsn32x can test this state again? :) 

It would be really really nice if after this PR you could create another PR where you chop these huge functions in smaller ones following more or less the single responsibility principle, where the function name describes the purpose. I have the feeling you're quite into these functions at the moment, and it would be good for the code quality to do this. I agree right now it is not a good idea, because diffing would get more and more difficult.

Duplicati/Library/Main/Operation/Backup/UploadSyntheticFilelist.cs
warwickmm: Good point.  Revision 0f30604046b8f4eb8481118e42324128470bdaf2 adds a comment explaining the decision here.

warwickmm:
Thanks for your very helpful review @verhoek.

> It would be really really nice if after this PR you could create another PR where you chop these huge functions in smaller ones following more or less the single responsibility principle, where the function name describes the purpose. I have the feeling you're quite into these functions at the moment, and it would be good for the code quality to do this. I agree right now it is not a good idea, because diffing would get more and more difficult. 

I think that's a good idea.  There are many places in the code that would benefit from some cleanup.  Given the current release process, I might wait until the next beta is released (hopefully very soon) before making too many changes just to be safe since I think this might be the last change blocking the beta release.  At that point, I'm thinking that we should focus a bit more on code quality and testing.  Breaking things up into smaller methods should make the code easier to understand and facilitate testing as well.

I suspect that there is a lot of unused code as well (at least my IDE tells me so) that I would like to remove as well.

warwickmm:
@drwtsn32x and @ts678, I don't have an planned changes left for this branch.  Can you think of any additional tests to run or things to consider here?

I have a local branch that cleans up some of the retention code, but I will wait for after the beta release to post it.

drwtsn32x:
@warwickmm I can't think of anything.... do you want me to also run some tests? I might have time this weekend.  Nice work!

warwickmm:
If you don't mind running a few more tests, that would be very much appreciated!  I don't think I broke anything since you last tested, but it would be good to double-check to be safe.

ts678:
> Can you think of any additional tests to run or things to consider here?

Nothing specific. I might take a shot at it when a Canary (or nightly if we get those going) emerges.
You're commendably diligent about the fix and the testing. I was hoping this could release to allow progress towards Beta, but now there's a possible 2.0.4.35 SharePoint regression, and [analysis clash](https://forum.duplicati.com/t/release-2-0-4-35-canary-2019-11-29/8567/4?u=ts678) against [user's report](https://forum.duplicati.com/t/bad-request-error-from-request-nameconflictbehavior-must-be-fail-replace-or-rename/8578/2?u=ts678). Anyone got SharePoint? I might ask other Staff, but I'll ask you while I'm here.

verhoek:
For me it is mergeable, a canary is a canary and another PR can be put on top if something breaks, but we can wait for drwtsn32x .

drwtsn32x:
I merged this branch into a fresh fork of duplicati (containing all merged changes so far).

First I re-ran all the previous tests outlined by @ts678 in his issue thread and confirmed that everything still works as well as the last time I did the tests.  So subsequent pushes to this branch haven't re-broken anything.

Then ran some new tests.

Set `keep-versions` to `1`:
* Ran a full backup
* Ran a partial backup

Observed: Both versions are retained - good that the older "full" is not deleted.

Removed `keep-versions`:
* Ran a partial backup
* Ran a full backup
* Ran a partial backup

Observed: I see all 5 backup versions as expected:  0: partial, 1: full, 2: partial, 3: partial, 4: full.

Set `keep-versions` to `2`:
* Ran a partial backup

Observed: 2 oldest partial backups were deleted.  Left with 2 fulls followed by 2 partials.

* Ran a full backup

Observed: oldest full backup was deleted, and 2 partials were deleted.  Left with 2 most recent full backups.


Looks good to me! Let me know if I should run any other type of tests.

warwickmm:
Great!  Thanks @drwtsn32x for all your help testing this.

Issue #3411, opened 2018-10-07
Duplicati downloads OneDrive online only files


veleek:
- [X] I have searched open and closed issues for duplicates.

----------------------------------------

## Environment info
- **Duplicati version**:  2.0.3.3_beta_2018-04-02
- **Operating system**: Windows 10
- **Backend**: N/A

## Description
When you configure Duplicati to backup a folder that includes a folder being synced by OneDrive, it will cause all of the files to start being downloaded even if they are configured as Online Only (or Files On-Demand - https://www.microsoft.com/en-us/microsoft-365/blog/2017/05/11/introducing-onedrive-files-on-demand-and-additional-features-making-it-easier-to-access-and-share-files/).

## Steps to reproduce

1. Setup a OneDrive sync'd folder
2. Configure it as only (should be able to force it by selecting Free Up Space)
3. Setup Duplicati to backup this folder.  In my case this was backed up as part of the backup for my user folder, but would also be the case for people doing full drive backups.

- **Actual result**:
Duplicati attempts to download all files including those that are marked as not available offline.
- **Expected result**:
Duplicati should respect the file status if there's anyway to determine whether it's online or offline only.

veleek:
Files which haven't been downloaded yet by OneDrive will have [FileAttribute.Offline](https://docs.microsoft.com/en-us/dotnet/api/system.io.fileattributes?view=netframework-4.7.2).  Duplicati should just ignore these files by default.  Maybe there should be an additional option to force these types of files to download.

ts678:
Which Windows 10 version is this? Version 1803 ("April 2018 Update") seems to have brought more issues although it's conceivable that these users had managed to skip the 2017 "Fall Creators Update" somehow.

[OneDrive backups no longer work after Windows April 2018 Update](https://forum.duplicati.com/t/onedrive-backups-no-longer-work-after-windows-april-2018-update/3526)

veleek:
I'll give --exclude-file-attributes a shot.  I'm going to be out of the country for about a month so I likely won't have a chance to fiddle around with this until I get back so don't take that as my ignoring this.

I'm using 1803 still (might be 1809 now, but it was 1803 when I first ran into the issue).

ts678:
I just noticed that there's a forum poll at [Cloud drive integration (files on-demand)](https://forum.duplicati.com/t/cloud-drive-integration-files-on-demand/4246) on how to handle these.

parksj10:
[...]

PR #3998, opened 2019-11-22
Disable file attribute masking on Windows 10

drwtsn32x:
Fix (partial?) for Issue https://github.com/duplicati/duplicati/issues/3411

Discussion here:  https://forum.duplicati.com/t/exclude-offline-files-broken-with-onedrive/5196/13

Root cause:
Starting with Windows 10 1803, the operating system may mask the process's view of some file attributes such as reparse, offline, and sparse.

See:
https://docs.microsoft.com/en-us/windows-hardware/drivers/ddi/ntifs/nf-ntifs-rtlqueryprocessplaceholdercompatibilitymode

Proposed change below disables this masking.

Demonstration:

Screen shot showing that the "Offline.txt" file has been flagged in OneDrive as available online only. You can see the cloud icon for its status:
Capture

Now when I run test-filter the file is properly excluded.  Before it was included due to the file attribute masking:
```
Duplicati.CommandLine.exe test-filter "C:\Users\xxx\OneDrive\Test" --exclude-files-attributes=Offline

Including source path: C:\Users\xxx\OneDrive\Test\
Excluding path due to attribute filter: C:\Users\xxx\OneDrive\Test\Offline.txt
Including path as no filters matched: C:\Users\xxx\OneDrive\Test\Online.txt
Matched 1 files (6 bytes)
```

For a full fix of issue 3411 maybe Duplicati defaults need to change to skip Offline files by default? If decision to go that route, I think it should be in a separate PR.

### DIFF
--- a/Duplicati/Library/Main/ProcessController.cs
+++ b/Duplicati/Library/Main/ProcessController.cs
@@ -237,7 +237,7 @@ namespace Duplicati.Library.Main
         /// <summary>
         /// Expose all filesystem attributes
         /// </summary>
-        private void ExposeAllFilesystemAttributes()
+        private static void ExposeAllFilesystemAttributes()
         {
             // Starting with Windows 10 1803, the operating system may mask the process's view of some
             // file attributes such as reparse, offline, and sparse.
--- a/Duplicati/Library/Main/ProcessController.cs
+++ b/Duplicati/Library/Main/ProcessController.cs
@@ -250,15 +250,11 @@ namespace Duplicati.Library.Main
             {
                 try
                 {
-                    var currentPcm = Win32.RtlQueryProcessPlaceholderCompatibilityMode();
-                    if (currentPcm == Win32.PHCM_VALUES.PHCM_DISGUISE_PLACEHOLDER)
-                    {
-                        Win32.RtlSetProcessPlaceholderCompatibilityMode(Win32.PHCM_VALUES.PHCM_EXPOSE_PLACEHOLDERS);
-                    }
+                    Win32.RtlSetProcessPlaceholderCompatibilityMode(Win32.PHCM_VALUES.PHCM_EXPOSE_PLACEHOLDERS);
                 }
                 catch
                 {
-                    // Ignore exceptions - not supported on this version of Windows
+                    // Ignore exceptions - not applicable on this version of Windows
                 }
             }
         }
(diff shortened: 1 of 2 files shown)


Duplicati/Library/Main/ProcessController.cs
```diff
@@ -234,6 +234,35 @@ private void ActivateBackgroundIOPriority()
...
+            // See https://docs.microsoft.com/en-us/windows-hardware/drivers/ddi/ntifs/nf-ntifs-rtlqueryprocessplaceholdercompatibilitymode
+
+            if (Platform.IsClientWindows)
+            {
+                try
+                {
+                    var currentPcm = Win32.RtlQueryProcessPlaceholderCompatibilityMode();
+                    if (currentPcm == Win32.PHCM_VALUES.PHCM_DISGUISE_PLACEHOLDER)
+                    {
+                        Win32.RtlSetProcessPlaceholderCompatibilityMode(Win32.PHCM_VALUES.PHCM_EXPOSE_PLACEHOLDERS);
```
verhoek: Just call the setter without the if, and remove the getter. Then we also don't need the import of the getter.

drwtsn32x: Yep I can do that if you prefer. I will submit an update later today.  Thanks

warwickmm:
@drwtsn32x, do you think this should be included in the next beta, or should we wait until after pull request #3992 is merged?

drwtsn32x:
@verhoek - changed code as you requested!

drwtsn32x:
@warwickmm - 3992 won't make it to the next beta? Bummer, I think that one is valuable and should be included, unless you think it needs more testing.

This PR here is pretty basic and should be very low risk to merge. That being said it probably doesn't help or affect most users.  I don't know how many are trying to exclude Offline files.

I still think we need to discuss if Offline files should be excluded by default, but I'd like that to be out of scope for this PR.

warwickmm:
Sorry, I was suggesting that #3992 be merged first so that we can optionally create a release from there before other changes are integrated.  #3992 should definitely be part of the next release!

drwtsn32x:
Ok, gotcha!  I think both can be merged for the next release. If this one needs Canary testing then it can wait.  I know one user is affected by the bug that fixes this but haven't heard of any other people complaining.

verhoek:
Both could go in, I will review #3992 later, this change here looks rather harmless.

PR #4005, opened 2019-12-02
Feature  - Optimize build system using docker

verhoek:
Importing dockerized CI system from the duplicatipipeline project.

WIth the new system, we:

- do not rebuild duplicati in every test job anymore
- do not run tests if the build fails
- are independent of the c# containers enforced by Travis
- have lower startup times of containers
- use caching between build and test jobs
- have timestamps in front of every line to measure build times
- can rebuild the system easily locally with just docker installed
- cut down the CI time by some time
- can extend this as done in the duplicatipipeline project to build nightly builds


warwickmm:
> do not rebuild duplicati in every test job anymore

Great!  It was extremely expensive to start a new worker for each test category.

pipeline/jobs/unittest5_job.sh
```diff
@@ -0,0 +1,10 @@
+#!/bin/bash
+SCRIPT_DIR="$( cd "$(dirname "$0")" ; pwd -P )"
+. "${SCRIPT_DIR}/params.sh"
+
+${ROOT_DIR}/pipeline/stage_unittests/trigger.sh \
+${FORWARD_OPTS[@]} \
+--sourcedir $BUILD_DIR \
+--targetdir $TEST_DIR \
+--testcategories Filter,Targeted,Purge,Serialization,WebApi,Utility,UriUtility,IO,ImportExport,Disruption | ts
```
warwickmm: It would be nice if we didn't have to remember to add an entry here every time we add a new test category.  Is there any way for this to be a "catch-all" by specifying which categories to exclude instead?  That way, we can say run all test categories except the ones in the other jobs.

verhoek: I moved them to the travis yaml and local start.sh script. Maybe later in another PR we can have such a mechanism that can do that, because you're right, it is a bit redundant to specify them again here unless we want to dedicate a travis worker to it.

pipeline/start.sh
```diff
@@ -0,0 +1,12 @@
...
+SCRIPT_DIR="$( cd "$(dirname "$0")" ; pwd -P )"
+. "${SCRIPT_DIR}/docker-run/error_handling.sh"
+. "${SCRIPT_DIR}/.local_config.sh"
+
+${ROOT_DIR}/pipeline/jobs/build_job.sh
+${ROOT_DIR}/pipeline/jobs/unittest_job.sh --testcategories BulkNormal --testdata data.zip
+${ROOT_DIR}/pipeline/jobs/unittest_job.sh --testcategories BulkNoSize --testdata data.zip
+${ROOT_DIR}/pipeline/jobs/unittest_job.sh --testcategories SVNDataLong,SVNData --testdata DSMCBE.zip
+${ROOT_DIR}/pipeline/jobs/unittest_job.sh --testcategories Border
+${ROOT_DIR}/pipeline/jobs/unittest_job.sh --testctegories Filter,Targeted,Purge,Serialization,WebApi,Utility,UriUtility,IO,ImportExport,Disruption
```
warwickmm: Typo in `testctegories`?  Also, it's not clear to me how this script is used.  It looks like the test categories are duplicated here as well as in `.travis.yml`?

warwickmm:
Can we try a quick experiment?  Can you introduce a deliberate test failure just to make sure that Travis detects and reports the test failure correctly

pipeline/start.sh
verhoek: Fixed it. This script is not used by travis but a script for your local environment and your convenience - it does exactly that what runs on travis, so you can execute integration tests etc from docker containers.

verhoek:
> Can we try a quick experiment? Can you introduce a deliberate test failure just to make sure that Travis detects and reports the test failure correctly

You can do that locally with the start.sh script, you will see. The only thing in Travis is that you will see what job failed and click that to see the error.

.travis.yml
```diff
@@ -1,80 +1,89 @@
...
+          - $BUILD_DIR
+      script:
+        - ${ROOT_DIR}/pipeline/jobs/unittest_job.sh --testcategories Border
+    - stage: tests
+      cache:
+        directories:
+          - $HOME/download/
+          - $BUILD_DIR
+      script:
+        - ${ROOT_DIR}/pipeline/jobs/unittest_job.sh --testctegories Filter,Targeted,Purge,Serialization,WebApi,Utility,UriUtility,IO,ImportExport,Disruption
```
warwickmm: Typo here?  `testctegories`

verhoek: Thanks for checking, and in fact travis doesn't bail out on errors, I have to look into that.

verhoek: Rootcause was refactoring from yesterday such that a trap was not set correctly. It works now. You can see an example unit test fail as the experiment.

warwickmm: Thanks.  Does a failure in one category prevent the subsequent categories from being tested?  For example, the failure was in `IO`, so the `ImportExport` and `Disruption` tests don't run?  It would be nice if all test failures could be determined in a single build, similar to if we ran all the tests in one worker.

verhoek: Yes it prevents it. Often it is like that in CI/CD systems, fail fast. I know what you mean, and we can change this by caching the return values and check the aggregate later on, but do we really want that?

verhoek: Right now we have a dedicated worker for each test, the upshot is you see what fails but it's very expensive computational-wise :) I think it is not needed that much, if you do a PR it is a quality control for others to see there are no errors. Best practice would be, before a PR is made, that all tests locally run fine and travis shows the proof of that. With the start.sh script, you can now run these tests locally detached from your dev environment.

warwickmm: That sounds reasonable.  Moreover, AppVeyor is running all the tests, so that would provide some more complete feedback as well.

By the way, I think this is great work.  The way we were distributing tests in Travis was really slow and frustrating.

Want to revert the deliberate test failure and we can see the tests pass again?

I think we're trying to impose a moratorium on merges so that we can incorporate the fix in pull request #4010 and hopefully ship a beta.  But since this only affects the Travis builds, I think this can be merged as well.

pipeline/stage_unittests/job.sh
```diff
@@ -0,0 +1,36 @@
+#!/bin/bash
+. /pipeline/docker-run/markers.sh
+. /pipeline/shared/duplicati.sh
+
+function get_and_extract_test_zip () {
+    travis_mark_begin "DOWNLOADING TEST DATA $CAT"
+    # test if zip file exists and contains no errors, otherwise redownload
+    unzip -t ~/download/"${CAT}"/"${testdata}" &> /dev/null || \
+    wget --progress=dot:giga "https://s3.amazonaws.com/duplicati-test-file-hosting/${testdata}" -O ~/download/"${CAT}"/"${testdata}"
```
warwickmm: I believe the tests now have setup methods that download the test data as needed (see pull request #3864).  The intent was to avoid having to perform the download separately in each test environment.  If that's working correctly, is this function needed?

verhoek: I wasn't aware of that change, but yes, that means we can remove all that stuff from the pipeline script. I will do that and let's see what happens.

pipeline/shared/duplicati.sh
```diff
@@ -0,0 +1,9 @@
+#!/bin/bash
+. "$( cd "$(dirname "$0")" ; pwd -P )/../docker-run/error_handling.sh"
+
+export DUPLICATI_ROOT="/application/"
```
warwickmm: I see a few references to `/application` in the other scripts. Was the intent to reference `DUPLICATI_ROOT` if possible?

pipeline/shared/duplicati.sh
```diff
@@ -0,0 +1,9 @@
...
+declare -a FORWARD_OPTS
+
+export UPDATE_SOURCE="${DUPLICATI_ROOT}/Updates/build/${BUILDTAG}_source"
+export UPDATE_TARGET="${DUPLICATI_ROOT}/Updates/build/${BUILDTAG}_target"
```
warwickmm: I don't see `UPDATE_SOURCE` or `UPDATE_TARGET` referenced anywhere.  Is there a future use for these?

.travis.yml
```diff
@@ -1,80 +1,89 @@
...
 
-    - env: CATEGORY=BulkNoSize
+
+    - stage: tests
       cache:
         directories:
-          - $HOME/download/bulk
-
-    - env: CATEGORY=SVNData
+          - $HOME/download/
```
warwickmm: Is caching of `$HOME/download/` still needed now that we aren't downloading the test data?

pipeline/shared/duplicati.sh
verhoek: Yes, I only added a part of the pipeline I constructed. Next stages are packaging, signing and deploying to docker. These are not relevant for PR's but will be for CD. I am awaiting feedback from Kenneth for this.

.travis.yml
verhoek: No, I can also remove that.

pipeline/shared/duplicati.sh
warwickmm: :+1:

pipeline/shared/duplicati.sh
verhoek: Okay I need to check out how the entire logic was again :) The /application is the mount point within the docker container, where things are actually happening. I think this was especially relevant to the packaging of rpm's, deb's etc. The duplicati.sh script should be important for things that run within the container.

verhoek: There was one occurrence that I changed now. As said, the duplicati.sh script is for the job files in the stages subdirs.

.travis.yml
warwickmm: Everything else looks pretty good to me.  Were there any other changes planned?  I think we can merge after removing these lines.

verhoek: Not in this PR. Removed the download cache.

warwickmm: :+1:

PR #4007, opened 2019-12-04
Fix long path issue when 259 or 260 characters

drwtsn32x:
Duplicati currently fails when backing up files or folders that are exactly 259 or 260 characters in length.  The cause appears to be a faulty test used to decide when to use the alternate API.

Right now it tests for `Length > 260` (MAX_PATH), but this doesn't leave room for the null terminator, which is [documented](https://docs.microsoft.com/en-us/windows/win32/fileio/naming-a-file#maximum-path-length-limitation) as counting towards the 260 limit.

Changing the test to `Length > 259` solves some but not all issues.  Because the string type is UTF-16 I believe we need to leave 2 bytes for the null terminator.  Using `Length > 258` as the test resolves all issues in my testing.

Discussion in the forum here:  https://forum.duplicati.com/t/filepath-longer-than-256-bytes-cause-fileaccesserror-under-windows-7

Thanks @ts678 for spotting the root cause and providing a testing formula.

### DIFF
--- a/Duplicati/Library/Common/IO/SystemIOWindows.cs
+++ b/Duplicati/Library/Common/IO/SystemIOWindows.cs
@@ -35,6 +35,7 @@ namespace Duplicati.Library.Common.IO
 
         private static bool IsPathTooLong(string path)
         {
+            // Use 258 for length check instead of 260 (MAX_PATH) - we need to leave room for the 16-bit (wide) null terminator
             return path.StartsWith(UNCPREFIX, StringComparison.Ordinal) || path.StartsWith(UNCPREFIX_SERVER, StringComparison.Ordinal) || path.Length > 258;
         }
 
--- a/Duplicati/Library/Common/IO/SystemIOWindows.cs
+++ b/Duplicati/Library/Common/IO/SystemIOWindows.cs
@@ -35,7 +35,7 @@ namespace Duplicati.Library.Common.IO
 
         private static bool IsPathTooLong(string path)
         {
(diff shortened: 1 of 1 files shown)


warwickmm:
Nice work!  Can you include a comment in the code explaining why we compare with `258`?

drwtsn32x:
@warwickmm sure thing, will submit an update shortly

ts678:
> exactly 259 or 260 characters in length

would include

```
C:\deeptest\123456789012345678901234567890123456789012345678901234567890123456789012345678901234567890123456789012345678901234567890123456789012345678901234567890123456789012345678901234567890123456789012345678901234567890123456789012345678901234567890\12.txt
```

yet it's seemingly backed up. Discrepancy between file treatment and folder treatment was noted [here](https://forum.duplicati.com/t/filepath-longer-than-256-bytes-cause-fileaccesserror-under-windows-7/8390/13?u=ts678), however I think switching to the alternate method a character early is safe. AFAIK it should handle that.

Thanks for the fix! We also seem to have a user-confirmed workaround in current release, by using VSS.

klorinczi:
It it released?
I don't see it on the canary update channel, after clicking to "Check for updates now".

drwtsn32x:
@klorinczi this change was only just merged, so you'll need to wait for the next canary release.

The latest is 2.0.4.35, not sure why your install isn't finding it.  But as mentioned this particular fix is not in 2.0.4.35 - need to wait for the next one.

drwtsn32x:
@ts678 - yeah that discrepancy is puzzling. I'll need to dig deeper.  But the test changing to `> 258` resolves it for both files and folders, even if it could technically be `> 259` for files.

PR #4009, opened 2019-12-07
Ensure files are selected for restore

drwtsn32x:
Currently it is possible to submit a restore job without selecting anything to restore.  This causes ALL files to be restored.

To reproduce:

1. In the Web UI, start the restore process either from existing backup set or direct restore.
2. When you get to the Select Files step, do not select anything.
3. Click the Continue button and observe the check that ensures something is selected.
4. Instead click the circled step number for the Restore Options step (2 - in the case of restore from backup set, 4 - in the case of direct restore).
5. Observe that the same check from step 3 was not performed in this case.
6. The job can be submitted and all files will be restored.

The proposed change performs the check when the ~~Restore button is clicked~~ Restore Options step is clicked, but I'm not convinced this is the preferred way to handle it.

Another way to fix this is to disallow clicking the Restore Options step number to force the user to click Continue, similar to how you cannot click Step 3 or 4 when you first start a Direct Restore from files.  I am thinking the latter more consistent with the rest of the UI, but want feedback first.

From my observations in the UI, if `< Previous` button is present you can also click the previous step number directly.  Likewise if `Next >` button is shown, you can directly click the next step number.

Instead if `Continue` or `Connect` is shown, you cannot click future step numbers. If `Back` is shown, you cannot click previous step numbers.

Thoughts?

### DIFF
--- a/Duplicati/Server/webroot/ngax/scripts/controllers/RestoreController.js
+++ b/Duplicati/Server/webroot/ngax/scripts/controllers/RestoreController.js
@@ -298,6 +298,12 @@ backupApp.controller('RestoreController', function ($rootScope, $scope, $routePa
     };
 
     $scope.onStartRestore = function() {
+        if ($scope.Selected.length == 0) {
+            DialogService.dialog(gettextCatalog.getString('No items selected'), gettextCatalog.getString('No items to restore, please select one or more items'));
+            $scope.restore_step = 0;
+			return;
+        }
+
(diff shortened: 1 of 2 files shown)


drwtsn32x:
New commit - instead of performing test at start of restore, I just changed the onclick event for "Restore Options" so that it performs the test at that point, just as if they clicked "Continue".

I think this is cleaner and also eliminates the duplication where the test was performed in two spots.

warwickmm:
Looks good to me.  I think allowing a user to click on the Restore Options step in the legend is ok and consistent with how the other menus work.

PR #4010, opened 2019-12-07
Fix regression with SharePoint backend

warwickmm:
In pull request #3989, we set the conflict resolution behavior for creating new items to be `"Replace"`. 
While this worked for OneDrive, SharePoint apparently is case-sensitive and expects `"replace"`. 
Since the default behavior is to replace, we will simply remove the specification to avoid possible conflicts.

See the forum discussion below:

https://forum.duplicati.com/t/bad-request-error-from-request-nameconflictbehavior-must-be-fail-replace-or-rename/8578


warwickmm:
Thanks @tygill.  We'll try and get this released soon.

PR #4013, opened 2019-12-09
Fixes and improvements to TrayIcon warning state

drwtsn32x:
Changes to previous PR https://github.com/duplicati/duplicati/pull/3943:

* Better icon color - it now matches the warning color used in the web UI (hex ffa500). I was never really happy with previous icon color.
* Fix handling of TrayIcons.IdleWarning state on non-Windows platforms - Previous PR ignored non-Windows runners, inadvertantly causing them to show a "normal" tray icon instead of an "error" icon when there was a warning condition (due to Switch statement not handling the new state).
* Now works on Linux/Gtk (as well as Windows)
* Fix Duplicati.GUI.TrayIcon.csproj - new warning icon was unnecessarily being copied to the output directory

CocoaRunner and RumpsRunner currently will continue to show the error icon when a warning occurs. Maybe we can get a volunteer to generate a normal-warning.png for them. I don't have the ability to test on anything besides Windows and Linux.

**New icon:**
normal-warning-orange-2

**Screen shot:**
cap1

**Zoomed in tray icon on Win7:**
cap2


warwickmm:
Looks good to me on Linux.

Issue #3983, opened 2019-11-10
Cannot disable auto updater


drwtsn32x:
Instructions for disabling the auto update mechanism don't seem to work.  I have seen suggestions to set one or the other following environment variables:

```
AUTOUPDATER_Duplicati_POLICY=Never
AUTOUPDATER_Duplicati_SKIP_UPDATE=1
```

However even after setting both of them I am still alerted when updates are released.

When looking at source I noticed that the SKIP_UPDATE option value isn't even considered, only if it is not null/whitespace:

https://github.com/duplicati/duplicati/blob/master/Duplicati/Library/AutoUpdater/UpdaterManager.cs#L59

Is this correct or should this be doing a more thorough boolean evaluation of the supplied value?

Also, what is the correct way to disable the automatic updater?

warwickmm:
Is it possible that the check is performed below, without regard to the environment variables?

https://github.com/duplicati/duplicati/blob/master/Duplicati/Server/UpdatePollThread.cs#L149

drwtsn32x:
@warwickmm perhaps.  I am willing to debug and fix this issue, but I am not sure how a user is actually instructed (if at all?) to disable the update mechanism.  Is it one of those two env vars or some other method?

warwickmm:
I think setting `AUTOUPDATER_Duplicati_SKIP_UPDATE` has been the recommended way.

https://forum.duplicati.com/t/how-to-disable-check-for-new-duplicati-version/3354

ts678:
> However even after setting both of them I am still alerted when updates are released.

sounds like it's the main issue. The title of the issue could be interpreted in several ways.

[...]

PR #4017, opened 2019-12-10
Add option to disable Update notification

drwtsn32x:
This adds a new environment variable that can disable the update notification prompt.  Other mechanisms of the update engine are unaffected:  manual check, automatic periodic check, triggering install, etc.

`AUTOUPDATER_Duplicati_NOTIFY`

Defaults to `True` if undefined, so standard Duplicati behavior remains unchanged.
If this environment variable is set to `False`, `No`, `Off`, or `0` then the popup UI notification will be suppressed.

User will still be able to see the version identified in the upgrade check by looking at the web UI About page.

This PR also makes the `AUTOUPDATER_Duplicati_POLICY` check case insensitive.  I noticed it wasn't when I was doing testing.

Can close issue https://github.com/duplicati/duplicati/issues/3983 once merged.

### DIFF
--- a/Duplicati/Library/AutoUpdater/UpdaterManager.cs
+++ b/Duplicati/Library/AutoUpdater/UpdaterManager.cs
@@ -86,7 +86,7 @@ namespace Duplicati.Library.AutoUpdater
         /// <summary>
         /// Show UI notification when update is found - default is True
         /// </summary>
-        public static bool ShowUpdateNotification { get { return Duplicati.Library.Utility.Utility.ParseBool(System.Environment.GetEnvironmentVariable(string.Format(UPDATE_NOTIFY_ENVNAME_TEMPLATE, APPNAME)), true); } }
+        public static bool ShowUpdateNotification { get { return Utility.Utility.ParseBool(Environment.GetEnvironmentVariable(string.Format(UPDATE_NOTIFY_ENVNAME_TEMPLATE, APPNAME)), true); } }
 
         /// <summary>
         /// Gets the original directory that this application was installed into
--- a/Duplicati/Library/AutoUpdater/UpdaterManager.cs
+++ b/Duplicati/Library/AutoUpdater/UpdaterManager.cs
@@ -77,10 +77,16 @@ namespace Duplicati.Library.AutoUpdater
         private const string RUN_UPDATED_FOLDER_PATH = "AUTOUPDATER_LOAD_UPDATE";
         private const string SLEEP_ENVNAME_TEMPLATE = "AUTOUPDATER_{0}_SLEEP";
         internal const string UPDATE_STRATEGY_ENVNAME_TEMPLATE = "AUTOUPDATER_{0}_POLICY";
+        internal const string UPDATE_NOTIFY_ENVNAME_TEMPLATE = "AUTOUPDATER_{0}_NOTIFY";
         private const string UPDATE_MANIFEST_FILENAME = "autoupdate.manifest";
         private const string README_FILE = "README.txt";
         private const string INSTALL_FILE = "installation.txt";
         private const string CURRENT_FILE = "current";
+
+        /// <summary>
+        /// Show UI notification when update is found - default is True
+        /// </summary>
+        public static bool ShowUpdateNotification { get { return Duplicati.Library.Utility.Utility.ParseBool(System.Environment.GetEnvironmentVariable(string.Format(UPDATE_NOTIFY_ENVNAME_TEMPLATE, APPNAME)), true); } }
 
         /// <summary>
         /// Gets the original directory that this application was installed into
--- a/Duplicati/Library/AutoUpdater/UpdaterManager.cs
+++ b/Duplicati/Library/AutoUpdater/UpdaterManager.cs
@@ -850,7 +850,7 @@ namespace Duplicati.Library.AutoUpdater
         {
             string optstr = Environment.GetEnvironmentVariable(string.Format(UPDATE_STRATEGY_ENVNAME_TEMPLATE, APPNAME));
             AutoUpdateStrategy strategy;
-            if (string.IsNullOrWhiteSpace(optstr) || !Enum.TryParse(optstr, out strategy))
+            if (string.IsNullOrWhiteSpace(optstr) || !Enum.TryParse(optstr, true, out strategy))
                 strategy = defaultstrategy;
 
             System.Threading.Thread backgroundChecker = null;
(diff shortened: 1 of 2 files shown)


warwickmm:
I don't have a good understanding of how the existing options work.  If someone doesn't want anything related to updates enabled (no checking, no downloading, no notificatoins), how many environment variables need to be set?

drwtsn32x:
I think we only need two. One to disable the feature entirely and another to just make it be not so noisy (the latter being what this PR adds).

I think one of the existing options is SUPPOSED to disable the update feature entirely, but from my testing it doesn't seem to work

See the issue thread where we discussed. I still have a lot of uncertainty with how the other options are supposed to work, their original intent, etc.

drwtsn32x:
These are the ones that have uncertainty surrounding them:

`AUTOUPDATER_Duplicati_POLICY=Never`
`AUTOUPDATER_Duplicati_SKIP_UPDATE=1`

Neither actually disables the periodic update checking or subsequent notification, so this new option is not superfluous IMO and has value.

Duplicati/Library/AutoUpdater/UpdaterManager.cs
```diff
@@ -77,11 +77,17 @@ public static class UpdaterManager
...
+        internal const string UPDATE_NOTIFY_ENVNAME_TEMPLATE = "AUTOUPDATER_{0}_NOTIFY";
         private const string UPDATE_MANIFEST_FILENAME = "autoupdate.manifest";
         private const string README_FILE = "README.txt";
         private const string INSTALL_FILE = "installation.txt";
         private const string CURRENT_FILE = "current";
 
+        /// <summary>
+        /// Show UI notification when update is found - default is True
+        /// </summary>
+        public static bool ShowUpdateNotification { get { return Duplicati.Library.Utility.Utility.ParseBool(System.Environment.GetEnvironmentVariable(string.Format(UPDATE_NOTIFY_ENVNAME_TEMPLATE, APPNAME)), true); } }
```
warwickmm: This line is quite long.  Can we remove the redundant `Duplicati.Library` and `System` qualifiers?

We can also simplify this by using the expression body syntax, but perhaps it's best to remain consistent with the rest of the file (we can refactor this separately).

drwtsn32x: Ok will submit an update tomorrow. Thanks

drwtsn32x: Ok done.  I agree about expression body syntax - I prefer it but I try to match the existing coding style.

I could also make long lines multiline if desired. I'm not usually a big fan of that style but can do it if desired.

drwtsn32x:
By the way - did more digging and testing with:

`AUTOUPDATER_Duplicati_SKIP_UPDATE=xxx`

If that value is defined and set to anything, Duplicati will not check the 'updates' folder or attempt to launch a secondary TrayIcon process at startup.

Updates will still be checked for, can still be installed, and the user can still attempt to "activate" it.  But the updated version will not be used while that env var is present.  Only the main Duplicati TrayIcon (original installed version) will run.

drwtsn32x:
Also, did more digging on this env var too:

`AUTOUPDATER_Duplicati_POLICY=xxx`

It, too, has no real effect on the updating mechanism. Setting it may affect if updates are checked or installed for at startup.  But it has no effect on manual or automatic checking, notifications, etc.

The reason setting it to `Never` seems to have no effect is because it is already the default setting for the following:

```
Duplicati.GUI.TrayIcon
Duplicati.Library.AutoUpdater
Duplicati.Server
Duplicati.Service
WindowsService
```

For the following, the default is `CheckDuring`

```
Duplicati.CommandLine
Duplicati.CommandLine.BackendTester
Duplicati.CommandLine.BackendTool
Duplicati.CommandLine.RecoveryTool
```

kenkendk:
You are correct, the update checker for the UI will always check, with no disable option available.

Your fix will keep asking for updates, but never use the information.
From a resource optimization perspective, I think it would be better to fix [line 128](https://github.com/duplicati/duplicati/blob/master/Duplicati/Server/UpdatePollThread.cs#L128) such that it just does not check unless a forced check is performed.

Looking at [line 113](https://github.com/duplicati/duplicati/blob/master/Duplicati/Server/UpdatePollThread.cs#L113) and [line 118](https://github.com/duplicati/duplicati/blob/master/Duplicati/Server/UpdatePollThread.cs#L118), it should be possible to either use a dummy value (i.e. `-1`) or add a new setting that disables the notification. I would prefer this over an environment variable, as it should only affect the GUI/Server process.

The `AUTOUPDATER_Duplicati_SKIP_UPDATE=1` setting should prevent installing updates (maybe fix in another PR).

drwtsn32x:
@kenkendk 

> Your fix will keep asking for updates, but never use the information.

Not exactly true... the About screen will still show the detected update, and allow the user to trigger installation there.  This change just stops the tray and WebUI popup notification, which some have expressed a desire for.

> or add a new setting that disables the notification. I would prefer this over an environment variable, as it should only affect the GUI/Server process.

I can certainly look at implementing it that way. Environment variables are less friendly to work with from a user's point of view.

> The `AUTOUPDATER_Duplicati_SKIP_UPDATE=1` setting should prevent installing updates (maybe fix in another PR).

Should it also disable the automatic (and maybe explicit) checking mechanism?  Agree that it could be done in a different PR.

drwtsn32x:
@kenkendk 

By the way, `AUTOUPDATER_Duplicati_SKIP_UPDATE` currently goes into effect when it is set to ANY value.  Should this be changed so that it goes into effect with boolean value (1, true, on, yes)?

drwtsn32x:
@kenkendk  - what do you think about adding an option to the main settings page, in or near the 'Update channel' section.  Could be a dropdown option where the user can select from:

* Automatically Check for Updates and Notify
* Automatically Check for Updates but do not Notify
* Do not automatically check

Or something like that?

Or we could have two check boxes:

* Automatically Check for Updates
* Notify when Update is Found

Changing this PR to WIP until there is consensus on how to handle.....

warwickmm:
> By the way, AUTOUPDATER_Duplicati_SKIP_UPDATE currently goes into effect when it is set to ANY value. Should this be changed so that it goes into effect with boolean value (1, true, on, yes)?

I think this should be changed.  When this was introduced in revision 55a0e0bfb57a1e695e77337935ea6917aad7f6a5, all the usages provided a value so it looks like that was the original intent.

drwtsn32x:
> I think this should be changed. When this was introduced in revision 55a0e0b, all the usages set the value to 1 so it looks like that was the original intent.

I'll submit a separate PR to implement just that one change.

drwtsn32x:
Canceling this PR as it needs a lot more thought on how to accomplish.

PR #4018, opened 2019-12-12
Refactor fileset deletion code and add tests

warwickmm:
This extracts the logic for the different fileset deletion strategies into separate classes.  This simplifies the code and allows us to more easily add tests that exercise various corner cases.

The unified diff doesn't communicate the changes very well.  I recommended viewing the changes using the split [side-by-side](https://github.com/duplicati/duplicati/pull/4018/files?utf8=%E2%9C%93&diff=split&w=1) diff.


Duplicati/Library/Main/Operation/DeleteHandler.cs
```diff
@@ -156,222 +158,233 @@ public void DoRun(Database.LocalDeleteDatabase db, ref System.Data.IDbTransactio
...
+    {
+        public SpecificVersionsRemover(Options options) : base(options)
+        {
+        }
 
-            List<DateTime> toDelete = new List<DateTime>();
+        public override IEnumerable<IListResultFileset> GetFilesetsToDelete(IEnumerable<IListResultFileset> filesets)
+        {
+            ISet<long> versionsToDelete = new HashSet<long>(this.Options.Version ?? new long[0]);
+            return filesets.Where(x => versionsToDelete.Contains(x.Version));
```
warwickmm: This is causing the `PurgeBrokenFilesTest` test to fail.  It appears that the `Fileset` table in a `LocalDeleteDatabase` has a state I did not expect, where the newest `DateTime` has the largest version number (typically, version `0` represents the newest backup).  In the test, the options specify that version `2` should be deleted.  While the new implementation here will select fileset version `2` for deletion, the old implementation would take a `DateTime[]` (ordered descending) and remove the 3rd (index `2`) element.  However, the 3rd element is not fileset version `2` in the database.

Either the "version" in `Options.Version` does not mean what I think it means, or there may be a bug somewhere.

warwickmm: It turns out I was building the `IListResultFilesets` incorrectly.  The `Version` should not be the `ID` field of the `Fileset` table.  Rather, it should be the backup version exposed to the user (integers where `0` is the most recent backup version.

Revision 60b4c8479cc9bf7982e8909dc45881a4e3e3202e should fix this.

Duplicati/Library/Main/Database/LocalDeleteDatabase.cs
```diff
@@ -109,18 +109,24 @@ private void InitializeCommands()
...
+        /// </summary>
+        internal IEnumerable<IListResultFileset> FilesetsWithBackupVersion
         {
             get
             {
                 List<IListResultFileset> filesets = new List<IListResultFileset>();
                 using (IDbCommand cmd = this.m_connection.CreateCommand())
                 {
-                    using (IDataReader reader = cmd.ExecuteReader(@"SELECT ""ID"", ""IsFullBackup"", ""Timestamp"" FROM ""Fileset"" ORDER BY ""Timestamp"" DESC "))
+                    using (IDataReader reader = cmd.ExecuteReader(@"SELECT ROW_NUMBER() OVER (ORDER BY ""Timestamp"" DESC) Version, ""IsFullBackup"", ""Timestamp"" FROM ""Fileset"""))
```
warwickmm: I'm not sure why the `PurgeBrokenFilesTest` is still failing.  The test passes for me when I run the tests locally.

Could there be an issue with using `ROW_NUMBER()` on Travis?  I think we are using the same versions of `System.Data.SQLite` (1.0.111).

verhoek: What happens if you run the test with the pipeline/start.sh script? Because this should be travis independent.

warwickmm: I haven't used the `pipeline/start.sh` script, but the test passes for me in both MonoDevelop and Rider.  I noticed that Travis appears to be using mono 6.6.0.161, whereas I am using mono 6.4.0 locally.  Could that be an issue?

Duplicati/Library/Main/Operation/DeleteHandler.cs
```diff
@@ -92,19 +88,25 @@ public void DoRun(Database.LocalDeleteDatabase db, ref System.Data.IDbTransactio
...
+                List<IListResultFileset> versionsToDelete = new List<IListResultFileset>();
+                versionsToDelete.AddRange(new SpecificVersionsRemover(this.m_options).GetFilesetsToDelete(filesets));
+                versionsToDelete.AddRange(new KeepTimeRemover(this.m_options).GetFilesetsToDelete(filesets));
+                versionsToDelete.AddRange(new RetentionPolicyRemover(this.m_options).GetFilesetsToDelete(filesets));
 
-                if (!m_options.AllowFullRemoval && sets.Length == toDelete.Length)
+                // When determining the number of full versions to keep, we need to ignore the versions already marked for removal.
+                versionsToDelete.AddRange(new KeepVersionsRemover(this.m_options).GetFilesetsToDelete(filesets.Except(versionsToDelete)));
+
+                DateTime[] toDelete = versionsToDelete.Select(x => x.Time.ToUniversalTime()).Distinct().OrderByDescending(x => x).ToArray();
```
verhoek: Did you fix a bug here since the ordering wasn't there before? Anyway, if you only order to keep the latest one that you do within the if, then only do the sorting in the if. Can't you keep working with the variable versionsToDelete?

Duplicati/Library/Main/Operation/DeleteHandler.cs
```diff
@@ -156,222 +158,233 @@ public void DoRun(Database.LocalDeleteDatabase db, ref System.Data.IDbTransactio
...
                     new CompactHandler(m_backendurl, m_options, (CompactResults)m_result.CompactResults).DoCompact(db, true, ref transaction, sharedManager);
                 }
-                
-                m_result.SetResults(
-                    from n in filesetNumbers
-                    where toDelete.Contains(n.Item2)
-                    select n, 
-                    m_options.Dryrun);
+
+                m_result.SetResults(versionsToDelete.Select(v => new Tuple<long, DateTime>(v.Version, v.Time)), m_options.Dryrun);
```
verhoek: Is versionsToDelete correct, or toDelete?

Duplicati/Library/Main/Database/LocalDeleteDatabase.cs
verhoek: Maybe. As a test you could change the docker image to be used to mono 6.4. Should be changed in stage_build/jobs.sh iirc.

Duplicati/Library/Main/Operation/DeleteHandler.cs
warwickmm: This ordering was on line 255 of the old `GetFilesetsToDelete` method.  I wasn't sure if the `DropFilesetsFromTable` method assumed that the array was sorted.  I can look at that after I figure out why the test is failing.

Duplicati/Library/Main/Operation/DeleteHandler.cs
warwickmm: I think `versionsToDelete` is correct.  The old implementation used the following (from line 97 of `DoRun`)
```
var filesetNumbers = db.FilesetTimes.Zip(Enumerable.Range(0, db.FilesetTimes.Count()), (a, b) => new Tuple<long, DateTime>(b, a.Value)).ToList();
```
which has local timestamps, similar to `versionsToDelete`.

Duplicati/Library/Main/Database/LocalDeleteDatabase.cs
warwickmm: I tested this with mono 6.4 in the docker image and the test still fails.  Looking at the nuget packages folder in the `/tmp/.duplicati_build` and `/tmp/.duplicati_test` folders, they match what is in my local development folder.

Are you able to see if the `PurgeBrokenFilesTest` in this branch is passing in Rider for you?

warwickmm: This is really strange.  When I run `pipeline/start.sh`, I see the test failure:
```
Dec 14 21:14:56 1) Error : Duplicati.UnitTest.PurgeTesting.PurgeBrokenFilesTest
Dec 14 21:14:56 Mono.Data.Sqlite.SqliteException : SQLite error
Dec 14 21:14:56 near "(": syntax error
Dec 14 21:14:56   at Mono.Data.Sqlite.SQLite3.Prepare (Mono.Data.Sqlite.SqliteConnection cnn, System.String strSql, Mono.Data.Sqlite.SqliteStatement previous, System.UInt32 timeoutMS, System.String& strRemain) [0x001f8] in <1c97a93cf49e412e9e991ba4fa395bc9>:0 
Dec 14 21:14:56   at Mono.Data.Sqlite.SqliteCommand.BuildNextCommand () [0x000d3] in <1c97a93cf49e412e9e991ba4fa395bc9>:0 
Dec 14 21:14:56   at Mono.Data.Sqlite.SqliteCommand.GetStatement (System.Int32 index) [0x00008] in <1c97a93cf49e412e9e991ba4fa395bc9>:0 
Dec 14 21:14:56   at (wrapper remoting-invoke-with-check) Mono.Data.Sqlite.SqliteCommand.GetStatement(int)
Dec 14 21:14:56   at Mono.Data.Sqlite.SqliteDataReader.NextResult () [0x000b3] in <1c97a93cf49e412e9e991ba4fa395bc9>:0 
Dec 14 21:14:56   at Mono.Data.Sqlite.SqliteDataReader..ctor (Mono.Data.Sqlite.SqliteCommand cmd, System.Data.CommandBehavior behave) [0x0004e] in <1c97a93cf49e412e9e991ba4fa395bc9>:0 
Dec 14 21:14:56   at (wrapper remoting-invoke-with-check) Mono.Data.Sqlite.SqliteDataReader..ctor(Mono.Data.Sqlite.SqliteCommand,System.Data.CommandBehavior)
Dec 14 21:14:56   at Mono.Data.Sqlite.SqliteCommand.ExecuteReader (System.Data.CommandBehavior behavior) [0x00006] in <1c97a93cf49e412e9e991ba4fa395bc9>:0 
Dec 14 21:14:56   at Mono.Data.Sqlite.SqliteCommand.ExecuteDbDataReader (System.Data.CommandBehavior behavior) [0x00000] in <1c97a93cf49e412e9e991ba4fa395bc9>:0 
Dec 14 21:14:56   at System.Data.Common.DbCommand.System.Data.IDbCommand.ExecuteReader () [0x00000] in <ae96b2e871e140cf907c5d70e1fdda5a>:0 
Dec 14 21:14:56   at Duplicati.Library.Main.Database.ExtensionMethods.ExecuteReader (System.Data.IDbCommand self, System.Boolean writeLog, System.String cmd, System.Object[] values) [0x0005e] in <3ae981f3f23d409abaef5211e2158058>:0 
Dec 14 21:14:56   at Duplicati.Library.Main.Database.ExtensionMethods.ExecuteReader (System.Data.IDbCommand self, System.String cmd, System.Object[] values) [0x00000] in <3ae981f3f23d409abaef5211e2158058>:0 
Dec 14 21:14:56   at Duplicati.Library.Main.Database.LocalDeleteDatabase.get_FilesetsWithBackupVersion () [0x0001d] in <3ae981f3f23d409abaef5211e2158058>:0 
Dec 14 21:14:56   at Duplicati.Library.Main.Operation.DeleteHandler.DoRun (Duplicati.Library.Main.Database.LocalDeleteDatabase db, System.Data.IDbTransaction& transaction, System.Boolean hasVerifiedBacked, System.Boolean forceCompact, Duplicati.Library.Main.BackendManager sharedManager) [0x00056] in <3ae981f3f23d409abaef5211e2158058>:0 
Dec 14 21:14:56   at Duplicati.Library.Main.Operation.PurgeBrokenFilesHandler.Run (Duplicati.Library.Utility.IFilter filter) [0x0039d] in <3ae981f3f23d409abaef5211e2158058>:0 
Dec 14 21:14:56   at Duplicati.Library.Main.Controller+<>c__DisplayClass35_0.<PurgeBrokenFiles>b__0 (Duplicati.Library.Main.PurgeBrokenFilesResults result) [0x0001c] in <3ae981f3f23d409abaef5211e2158058>:0 
Dec 14 21:14:56   at Duplicati.Library.Main.Controller.RunAction[T] (T result, System.String[]& paths, Duplicati.Library.Utility.IFilter& filter, System.Action`1[T] method) [0x0026f] in <3ae981f3f23d409abaef5211e2158058>:0 
Dec 14 21:14:56   at Duplicati.Library.Main.Controller.RunAction[T] (T result, System.Action`1[T] method) [0x00009] in <3ae981f3f23d409abaef5211e2158058>:0 
Dec 14 21:14:56   at Duplicati.Library.Main.Controller.PurgeBrokenFiles (Duplicati.Library.Utility.IFilter filter) [0x0001a] in <3ae981f3f23d409abaef5211e2158058>:0 
Dec 14 21:14:56   at Duplicati.UnitTest.PurgeTesting.PurgeBrokenFilesTest () [0x0044f] in <1c9d8eaba09e43bba2809428372a5926>:0 
Dec 14 21:14:56   at (wrapper managed-to-native) System.Reflection.RuntimeMethodInfo.InternalInvoke(System.Reflection.RuntimeMethodInfo,object,object[],System.Exception&)
Dec 14 21:14:56   at System.Reflection.RuntimeMethodInfo.Invoke (System.Object obj, System.Reflection.BindingFlags invokeAttr, System.Reflection.Binder binder, System.Object[] parameters, System.Globalization.CultureInfo culture) [0x0006a] in <d2ec5c92492f4d6ba8c422bdf574b786>:0 
Dec 14 21:14:56 
Dec 14 21:14:56 Run Settings
Dec 14 21:14:56     DisposeRunners: True
Dec 14 21:14:56     WorkDirectory: /application
Dec 14 21:14:56     NumberOfTestWorkers: 1
Dec 14 21:14:56     ImageRuntimeVersion: 4.0.30319
Dec 14 21:14:56     ImageTargetFrameworkName: .NETFramework,Version=v4.6.2
Dec 14 21:14:56     ImageRequiresX86: False
Dec 14 21:14:56     ImageRequiresDefaultAppDomainAssemblyResolver: False
Dec 14 21:14:56 
Dec 14 21:14:56 Test Run Summary
Dec 14 21:14:56   Overall result: Failed
Dec 14 21:14:56   Test Count: 2, Passed: 1, Failed: 1, Warnings: 0, Inconclusive: 0, Skipped: 0
Dec 14 21:14:56     Failed Tests - Failures: 0, Errors: 1, Invalid: 0
Dec 14 21:14:56   Start time: 2019-12-15 05:14:07Z
Dec 14 21:14:56     End time: 2019-12-15 05:14:56Z
Dec 14 21:14:56     Duration: 49.765 seconds
```
If I then `cd` to `/tmp/.duplicati_test` and run
```
nunit3-console Duplicati/UnitTest/bin/Release/Duplicati.UnitTest.dll --where:cat==Purge
```
the tests pass:
```
Run Settings
    DisposeRunners: True
    WorkDirectory: /tmp/.duplicati_test
    NumberOfTestWorkers: 1
    ImageRuntimeVersion: 4.0.30319
    ImageTargetFrameworkName: .NETFramework,Version=v4.6.2
    ImageRequiresX86: False
    ImageRequiresDefaultAppDomainAssemblyResolver: False

Test Run Summary
  Overall result: Passed
  Test Count: 2, Passed: 2, Failed: 0, Warnings: 0, Inconclusive: 0, Skipped: 0
  Start time: 2019-12-15 05:17:55Z
    End time: 2019-12-15 05:18:48Z
    Duration: 52.673 seconds
```

warwickmm: I don't quite understand how the sqlite dependencies work.  Do we depend on system sqlite libraries for runtime?  On my machine, I have sqlite 3.30, but Debian stretch has [sqlite 3.16.2](https://packages.debian.org/stretch/sqlite3), which doesn't support the `ROW_NUMBER` window function.  We are compiling with the sqlite 1.0.111 nuget package, which supports sqlite 3.28.  This might explain the different behavior.

verhoek: Good question, I can change the docker config or add in the setup of the container a newer version of sqlite, but if we do not depend on system libraries, then we should use the included libs.

warwickmm: I'm not certain, but I have a feeling that in some cases we do fall back to the system libraries.  This concerns me, since several distributions (e.g., Debian) ship with outdated versions of sqlite.  I think the ideal case is to to ship the correct libraries and reduce the dependency on system libraries.  There's also the `Microsoft.Data.Sqlite` package, but I'm not sure how that differs from the `System.Data.Sqlite` and `Mono.Data.Sqlite` ones.

Perhaps the best approach is to not use `ROW_NUMBER` here, and deal with our sqlite dependencies as a separate task?

warwickmm: Reivision 610fa4ffeae90640c50f6dd801810b433add9fee removes the use of the `ROW_NUMBER` window function and generates the backup version numbers separately.  I expect that the tests should pass now.

I also created issue #4024 to remind us to solidify the SQLite dependencies.

Duplicati/Library/Main/Operation/DeleteHandler.cs
warwickmm: Revision a7ab5f6f00450dc39149addd028a24dbba66372f removes the unnecessary array creation and also removes some redundant `Timestamp` conversion (the conversion to universal time occurs in the `DropFilesetsFromTable` method.

kenkendk:
Looks good!

PR #4021, opened 2019-12-14
Fix spelling errors in comments

warwickmm:
This fixes several spelling errors in comments.  I tried to only fix obvious ones and did not address differences due to spelling preferences of different English speaking locales.

In doing so, this also normalized several line endings.


verhoek:
Did you do this in an automated way, and if so, how or manually?

warwickmm:
I used Rider to search for spelling errors in comments, and then manually decided whether to fix the error (using its suggestion), or skip the warning.

warwickmm:
Viewing the diff with "Hide whitespace changes" will provide a better view of the actual changes.

Issue #4028, opened 2019-12-19
MetadataWriteFailed (warning) when selecting Restore read/write permissions


mediedata:
- [ x] I have searched open and closed issues for duplicates.

----------------------------------------

## Environment info
- **Duplicati version**: 2.0.4.37_canary_2019-12-12
- **Operating system**: Win 10 pro
- **Backend**: B2

## Description

Got a Warning about null-value when restoring a backup, when the "Restore read/write permissions" option is checked. It happens only on one folder.

It doesn't happen in 2.0.4.23_beta_2019-07-14.

## Steps to reproduce

1. Restore a backup.
2. Select "Restore read/write permissions" and start restore

- **Actual result**:
Got a warning about null-value.
- **Expected result**:
No warning. The affected folder has the same read/write permissions as the other folders in the backup.

## Debug log

```
 "Warnings": [
    "2019-12-19 16.58.35 +01 
- [Warning-Duplicati.Library.Main.Operation.RestoreHandler-MetadataWriteFailed]: 
Failed to apply metadata to file: \"D:\\system\\Desktop\\restore test2\\_My Documents\\\", 
message: Value cannot be null.\r\nParameter name: sddlForm"
  ],
```

warwickmm:
I can't reproduce this on Linux.  The parameter name `sddlForm` does not appear in the Duplicati source code, so I believe this is coming from somewhere else (perhaps in .NET somewhere).  A quick search seems to indicate that this may be Windows-specific (typically related to SQL server somehow).

mediedata:
Yes it's definently a Windows-only thing. 

Digging a bit further, it looks like the issue is with reading/writing Windows' Security permissions on that folder.

The affected folder's permissions has an additional Deny-permission set for one user-account.
![Image1](https://user-images.githubusercontent.com/59032089/71290616-91581680-2370-11ea-99b9-d0b13bf4ef9b.jpg)

![Image2](https://user-images.githubusercontent.com/59032089/71290516-4fc76b80-2370-11ea-8902-2bed65e7096e.jpg)

It's most likely related to SetSecurityDescriptorSddlForm:
https://docs.microsoft.com/en-us/windows/win32/secauthz/security-descriptor-string-format
https://docs.microsoft.com/en-us/dotnet/api/system.security.accesscontrol.objectsecurity.setsecuritydescriptorsddlform?view=netframework-4.8

[...]

PR #4030, opened 2019-12-24
Fix issue 4028 - Unable to restore file permissions on Windows

drwtsn32x:
Fix issue https://github.com/duplicati/duplicati/issues/4028

File permissions are not restored on Windows. Instead the following error is produced for each file:

```
- [Warning-Duplicati.Library.Main.Operation.RestoreHandler-MetadataWriteFailed]: 
Failed to apply metadata to file: \"D:\\PATH\\FILE", 
message: Value cannot be null.\r\nParameter name: sddlForm"
```

Bug introduced in 2.0.4.32 by PR https://github.com/duplicati/duplicati/pull/3953/
Read only fields not settable by deserializer.

Adding `[JsonProperty]` overcomes this as an alternative to removing `readonly`.

### DIFF
--- a/Duplicati/Library/Common/IO/SystemIOWindows.cs
+++ b/Duplicati/Library/Common/IO/SystemIOWindows.cs
@@ -23,6 +23,7 @@ using System.Linq;
 
 using AlphaFS = Alphaleonis.Win32.Filesystem;
 using Duplicati.Library.Interface;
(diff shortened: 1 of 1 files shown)


warwickmm:
Great!  I prefer to use `readonly` wherever we can to prevent inadvertent assignments.

PR #4031, opened 2019-12-26
Back up and restore inheritance breaks

drwtsn32x:
When Duplicati backs up ACLs on Windows, it does not record information related to inheritance breaks.  When restoring files with ACLs this causes incorrect results for files or folders where inheritance was originally broken.

With the below change, inheritance break states are backed up with the file/folder metadata.  When restoring, the inheritance break state is restored just before the ACL.

I have done some testing and it works well for me, but would really like some others to test.

### DIFF
--- a/Duplicati/Library/Common/IO/SystemIOWindows.cs
+++ b/Duplicati/Library/Common/IO/SystemIOWindows.cs
@@ -101,17 +101,6 @@ namespace Duplicati.Library.Common.IO
             }
         }
 
-        private class FileSystemAccessProtected
-        {
-            [JsonProperty]
-            public readonly bool IsProtected;
-
-            public FileSystemAccessProtected(bool Protected)
-            {
-                IsProtected = Protected;
-            }
-        }
-
         private static Newtonsoft.Json.JsonSerializer _cachedSerializer;
 
         private Newtonsoft.Json.JsonSerializer Serializer
@@ -498,7 +487,7 @@ namespace Duplicati.Library.Common.IO
                 objs.Add(new FileSystemAccess((FileSystemAccessRule)f));
 
             dict["win-ext:accessrules"] = SerializeObject(objs);
-            dict["win-ext:accessrulesprotected"] = SerializeObject(new FileSystemAccessProtected(rules.AreAccessRulesProtected));
+            dict["win-ext:accessrulesprotected"] = rules.AreAccessRulesProtected.ToString();
 
             return dict;
         }
@@ -514,10 +503,10 @@ namespace Duplicati.Library.Common.IO
 
                 if (data.ContainsKey("win-ext:accessrulesprotected"))
                 {
-                    var content = DeserializeObject<FileSystemAccessProtected>(data["win-ext:accessrulesprotected"]);
-                    if (rules.AreAccessRulesProtected != content.IsProtected)
+                    bool IsProtected = bool.Parse(data["win-ext:accessrulesprotected"]);
+                    if (rules.AreAccessRulesProtected != IsProtected)
                     {
-                        rules.SetAccessRuleProtection(content.IsProtected, false);
+                        rules.SetAccessRuleProtection(IsProtected, false);
                     }
                 }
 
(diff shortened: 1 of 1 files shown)


Duplicati/Library/Common/IO/SystemIOWindows.cs
```diff
@@ -496,31 +497,44 @@ public void SetMetadata(string path, Dictionary<string, string> data, bool resto
...
-
-                Exception ex = null;
+                FileSystemSecurity rules = isDirTarget ? GetAccessControlDir(targetpath) : GetAccessControlFile(targetpath);
 
-                foreach (var r in content)
+                if (data.ContainsKey("win-ext:accessrulesprotected"))
                 {
-                    // Attempt to apply as many rules as we can
-                    try
+                    bool IsProtected = bool.Parse(data["win-ext:accessrulesprotected"]);
```
warwickmm: We typically begin variable names with a lowercase letter.

warwickmm:
Sorry I can't help test since this only affects Windows.

I don't understand what the exact issue is, but is it something that can be tested programmatically?

Duplicati/Library/Common/IO/SystemIOWindows.cs
drwtsn32x: Ok, that's a simple change.

drwtsn32x:
> but is it something that can be tested programmatically?

Possibly, but I was hoping another dev could set up a custom folder structure with complicated permissions including inheritance breaks to see if it works for them.  It works for the scenarios I tested...

warwickmm:
No problem.  Let's wait to see if someone with Windows can confirm the results.

What is an inheritance break, exactly?  When a subfolder does not have the same permissions as the parent?

drwtsn32x:
Yes, normally permissions flow downward but you can explicitly stop that on a particular folder or file. IT staff I know refer to this as "breaking inheritance".  Modern APIs seem to call it "Protecting Access Rules" - in other words, they won't be affected by any changes higher up that would normally apply to them.  Here's the check box on Windows:

Capture


This particular file has inheritance broken.

warwickmm:
Thanks!  I'm guessing that creating a test for this won't be too difficult.  Basically, create a folder structure with broken inheritance, run a backup, restore, and compare the restored permissions with the original source?  Perhaps something like the following?

https://stackoverflow.com/questions/6574196/programatically-removing-include-inheritable-permissions-from-this-objects-par

warwickmm:
I was doing some light investigation, but cannot fully test without Windows.  Can you see if [this test](https://github.com/warwickmm/duplicati/commit/ccbd05f1175abbd8900d8edf91159d69a76a8cfd) fails without your fix, and passes with your fix?  I think that is the case in AppVeyor.

drwtsn32x:
@warwickmm I will look into it the next day or two...thanks

drwtsn32x:
@warwickmm - looking at this now. i'm unfamiliar with the UnitTesting portion and not sure how I go about running these tests manually. Looks like they are normally run by Travis when a PR is submitted?

warwickmm:
Waiting for AppVeyor (Windows) and Travis (Linux) is one way, but it would be much faster to run them locally.  What IDE are you using?  Most have the ability to run the tests (you can select a specific test to run instead of running them all).

drwtsn32x:
Nice, I have never messed with Unit Testing.  Had to install the NUnit testing adapter for Visual Studio 2019 before it would work.

Ran your RestoreInheritanceBreaks test without this PR applied and it failed.  Then ran it again with this PR applied and it succeeded.  So it appears you did a good job building the test for this.

Thanks

warwickmm:
Great!  Feel free to cherry pick f147a94eb3c7aa096662f9f2c89c5400682ada60 onto your branch.

Do you want to wait some more to see if anyone else can perform some more manual testing or do you think it's ok to merge?  The changes look reasonable to me.

drwtsn32x:
We are good to merge in my opinion. Thanks!

warwickmm:
Ok.  Do you want to add the test to your branch, or I can create a separate pull request?

drwtsn32x:
I attempted to do a cherry-pick but your commit didn't apply completely to my branch.  The UnitTest files did, but .travys.yml and pipeline/start.sh did not.

If you are able to do a separate PR that might be quicker.

Thanks

warwickmm:
> If you are able to do a separate PR that might be quicker.

Will do.  No problem.

PR #4034, opened 2019-12-31
Add test for listing filesets without local database

warwickmm:
This adds a test to verify the backup type associated with the filesets obtained when performing a list operation without the local database.

Prior to revision 8ab64071f006033c64e2f8acc3682230e915e2fe, all filesets would have been marked as partial when performing a direct restore without the local database.

This concerns issues that were raised in #3982.


warwickmm:
This hasn't garnered much feedback, but since it simply adds a test I'll go ahead and merge it.

PR #4035, opened 2019-12-31
Add test to verify that failed put operations result in exception

warwickmm:
This adds a test to verify that an exception is thrown when the put operations fail.

This test would fail without the fixes from pull request #3720.

This concerns issue #3673.


warwickmm:
This test fails in AppVeyor and in Travis because it runs with administrator/root privileges.  We cannot force the put to fail by making the directory read-only.

PR #4038, opened 2020-01-01
Add test for restoring permissions with inheritance breaks

warwickmm:
This adds a simple test for restoring permissions with inheritance breaks on Windows.

This concerns pull request #4031.  Some related discussion occurs in issue #4028.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -69,7 +69,7 @@ jobs:
         directories:
           - $BUILD_DIR
       script:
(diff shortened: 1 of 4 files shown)


drwtsn32x:
I tested this and am comfortable with merging. Let me know if you would like to wait for other input.

warwickmm:
Thanks @drwtsn32x.  I think we can merge, especially since it simply adds a test.

PR #4039, opened 2020-01-02
Windows Filters cleanup

drwtsn32x:
Changed hard-coded `/Windows` paths to use Environment.SpecialFolder instead.
Moved `Recycle.Bin` filter to SystemFiles for consistency (that's where `Recycler` was).
Added `Recycled` - used for FAT volumes on some OS versions.
Removed some filter duplication.

I was also going to change AppData to use SpecialFolder, but then realized this may not be ideal as it would only return the current user's AppData folder.  Leaving the paths hard coded (with wildcards) will match all users AppData folders.

### DIFF
--- a/Duplicati/Library/Utility/FilterGroups.cs
+++ b/Duplicati/Library/Utility/FilterGroups.cs
@@ -324,7 +324,7 @@ namespace Duplicati.Library.Utility
                 yield return FilterGroups.CreateWildcardFilter(@"*/Microsoft*/Windows/Cookies*");
                 yield return FilterGroups.CreateWildcardFilter(@"*/MSOCache*");
                 yield return FilterGroups.CreateWildcardFilter(@"*/NTUSER*");
-                yield return FilterGroups.CreateWildcardFilter(@"*UsrClass.dat*");
+                yield return FilterGroups.CreateWildcardFilter(@"*/UsrClass.dat");
                 yield return FilterGroups.CreateWildcardFilter(@"?:/hiberfil.sys");
                 yield return FilterGroups.CreateWildcardFilter(@"?:/pagefile.sys");
                 yield return FilterGroups.CreateWildcardFilter(@"?:/swapfile.sys");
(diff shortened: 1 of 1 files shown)


drwtsn32x:
I don't know if we care too much, but all of the "AppData" hard coded paths aren't valid on XP.  If we care about XP, we should probably duplicate those with their full "Application Data" equivalents.

Duplicati/Library/Utility/FilterGroups.cs
```diff
@@ -324,15 +324,15 @@ private static IEnumerable<string> CreateWindowsFilters(FilterGroup group)
...
                 yield return FilterGroups.CreateWildcardFilter(@"?:/pagefile.sys");
                 yield return FilterGroups.CreateWildcardFilter(@"?:/swapfile.sys");
+                yield return FilterGroups.CreateWildcardFilter(@"?:/$Recycle.Bin/");
+                yield return FilterGroups.CreateWildcardFilter(@"?:/Recycled/");
+                yield return FilterGroups.CreateWildcardFilter(@"?:/Recycler/");
                 yield return FilterGroups.CreateWildcardFilter(@"?:/System Volume Information/");
-                yield return FilterGroups.CreateWildcardFilter(@"?:/Windows/Installer*");
-                yield return FilterGroups.CreateWildcardFilter(@"?:/Windows/Temp*");
-                yield return FilterGroups.CreateWildcardFilter(@"*/ntuser.dat*");
+                yield return FilterGroups.CreateWildcardFilter(FilterGroups.CreateSpecialFolderFilter(Environment.SpecialFolder.Windows) + "Installer/");
```
warwickmm: It looks like `CreateSpecialFolderFilter` creates a wildcard filter.  Do we still need the call to `CreateWildcardFilter`?

warwickmm: Never mind...I think this is needed for normalization of the path separators.

PR #4040, opened 2020-01-05
improve speed of indexing in RecoveryTool

tfriedel:
I noticed Duplicati is not a good tool for huge backups over 1TB. Restoring is very slow (weeks). 
The RecoveryTool especially is not optimized. No multi threading and the algorithms don't use efficient data structures. 
One thing that was unbearably slow was the index command. It would accumulate lines in the index.txt and for each file it would merge the sorted index (which at some point was 700mb of lines) with it.
I changed this behaviour so it's using a SortedSet which always keeps the lines sorted. Indexing was done in a couple of minutes instead of weeks. The disadvantage is, the datastructure has to fit in memory. But I guess for most people it would be an improvement.

### DIFF
--- a/Duplicati/CommandLine/RecoveryTool/Index.cs
+++ b/Duplicati/CommandLine/RecoveryTool/Index.cs
@@ -41,23 +41,14 @@ namespace Duplicati.CommandLine.RecoveryTool
 
             Directory.SetCurrentDirectory(folder);
 
+            SortedSet<string> ix_sorted =
+                new SortedSet<string>(StringComparer.Ordinal);
             string ixfile;
             options.TryGetValue("indexfile", out ixfile);
             if (string.IsNullOrWhiteSpace(ixfile))
                 ixfile = "index.txt";
 
             ixfile = Path.GetFullPath(ixfile);
-            if (!File.Exists(ixfile))
-            {
-                using (File.Create(ixfile))
-                {
-                }
-            }
-            else
-            {
-                Console.WriteLine("Sorting existing index file");
-                SortFile(ixfile, ixfile);
-            }
 
             var filecount = Directory.EnumerateFiles(folder).Count();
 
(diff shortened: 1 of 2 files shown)


warwickmm:
Thanks @tfriedel.  The RecoveryTool is only meant to be used if all other options fail (`Duplicati.CommandLine.exe` should be used to restore if possible).  As such, I don't know if we should sacrifice the ability to run the tool (some users run this in low resource environments) for performance, even if the gains are substantial.

What are your thoughts on making the in-memory index the default behavior, but keeping the ability to use the file-based index via a command-line option?

tfriedel:
I have no info about the low resource environments of users. I can imagine that the actual backup tool is run on something like a raspberry pi, but the restore... I guess if you are forced to use this tool you can afford to run it on a modern PC. I think when memory becomes an issue, the slowness of the tool becomes even more of an issue. Who has the time to wait for weeks for a restore to complete?
I guess you could implement both behaviours, but it would mean extra work / higher code complexity.

warwickmm:
That's a good point.  Since this is a last resort, it's reasonable to require that this be run on a more capable machine.  

> I guess you could implement both behaviours, but it would mean extra work / higher code complexity.

Before we completely abandon the file-based approach, do you think it's overly complex to have two "Index" implementations, and an option that decides which one is used [here](https://github.com/duplicati/duplicati/blob/v2.0.5.100-2.0.5.100_canary_2020-01-18/Duplicati/CommandLine/RecoveryTool/Program.cs#L82)?  I'm still a little concerned that people have _huge_ backups, where the memory requirements are still too great for a decent machine.

By the way, I created a pull request (#4053) that adds a test to help verify the behavior of the `RecoveryTool`.  Hopefully this can help us with making improvements to the `RecoveryTool`.

kenkendk:
> Before we completely abandon the file-based approach, do you think it's overly complex to have two "Index" implementations, and an option that decides which one is used here? I'm still a little concerned that people have huge backups, where the memory requirements are still too great for a decent machine.

I would opt for having both methods. I have explictly designed this such that it can run on low-memory machines, even though it is slower (like a RPI). I can see that in your case it is way too slow, so it makes sense to trade memory for speed.

I think we can also optimize the restore process quite a lot if we just store the hashes in a dictionary, but again at the cost of memory.

tfriedel:
It's your software and you can of course implement both methods if you
like. You could also do some estimates how many files would consume how
much memory to get a better idea about the limits.

Am Do., 23. Jan. 2020 um 13:50 Uhr schrieb Kenneth Skovhede <
notifications@github.com>:

warwickmm:
@tfriedel, I took your patch and included it in pull request #4061, which builds the index in memory by default but adds an option to use the file-based approach in low-resource situations.

Thanks for your help with this!  :+1:

PR #4042, opened 2020-01-06
Fix exceptions caused by Stop Now

### LINKED ISSUE #4037: Stop now results in "Detected non-empty blocksets with no associated blocks!"
## Environment info

- **Duplicati version**: 2.0.4.38
- **Backend**: Local

## Description
I added a test [here](https://github.com/warwickmm/duplicati/commit/1158a8a68dd2fa1f47d8ba8b166723fe5cd45cc7) for interrupting a backup with "stop now" that does the following:
1. Run a complete backup.
2. Interrupt a backup with "stop now" and verify that an exception is thrown.
3. Run a complete backup.
4. Restore files from step 3 and compare the restored files with the original source.

Test results:

| Environment | Number of Trials | Number of Failures |
| ----- | ----- | ----- |
| AppVeyor (Windows) | 10 | 8* |
| Travis (Linux) | 10 | 0 |
| Local (Linux) | 100 | 0 |

*All of the AppVeyor failures occurred during Step 3 by encountering the "Detected non-empty blocksets with no associated blocks!" exception.

This seems to indicate that the issue is more prevalent on (if not specific to) Windows.

seantempleton:
The LocalDatabase class was deleting from the BlocklistHash table based on the volume id of blocks. The deletes for the Blockset and BlocksetEntry tables were using a union query which included the same query the BlocklistHash delete was using. Changed the BlocklistHash delete query to use the union as well to ensure that all three tables would use the same query to delete the necessary rows. This fixes the "Detected non-empty blocksets with no associated blocks!" error.

Moved a write call in the DataBlockProcessor to send a volume for upload after setting blockvolume to null. This prevents a null reference exception because otherwise it would be sent to the SpillCollector and it tries to add a block, but the volume has already been closed.

Added the ? operator to the RegistryUtility class to prevent a null reference exception while debugging. It was annoying catching it in the debugger.

Fixes #4037

### DIFF
--- a/Duplicati/Library/Main/Database/LocalDatabase.cs
+++ b/Duplicati/Library/Main/Database/LocalDatabase.cs
@@ -439,7 +439,7 @@ namespace Duplicati.Library.Main.Database
                 deletecmd.ExecuteNonQuery(string.Format(@"DELETE FROM ""Blockset"" WHERE ""ID"" IN ({0})", bsIdsSubQuery));
                 deletecmd.ExecuteNonQuery(string.Format(@"DELETE FROM ""BlocksetEntry"" WHERE ""BlocksetID"" IN ({0})", bsIdsSubQuery));
 
-                deletecmd.ExecuteNonQuery(string.Format(@"DELETE FROM ""BlocklistHash"" WHERE ""Hash"" IN (SELECT ""Hash"" FROM ""Block"" WHERE ""VolumeID"" IN ({0}))", volIdsSubQuery));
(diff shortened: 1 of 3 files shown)


seantempleton:
I snuck in a change to the DataBlockProcessor. I'm not 100% certain it's correct but figured maybe some people can look at it and see what they think. It's for the null reference exception that was seen while debugging this non-empty blockset exception.

warwickmm:
This looks good to me.  I'll create a pull request that adds the test for the "stop now" behavior.

PR #4043, opened 2020-01-06
Add test for interrupting backup with "stop now"

warwickmm:
This adds a test for interrupting a backup with "stop now" by doing the following:  
1. Run a complete backup.
2. Interrupt a backup with "stop now" and verify that an exception is thrown.
3. Run a complete backup.
4. Restore files from step 3 and compare the restored files with the original source.

This concerns issue #4037 and its fix in pull request #4042.

### DIFF
--- a/Duplicati/UnitTest/DisruptionTests.cs
+++ b/Duplicati/UnitTest/DisruptionTests.cs
@@ -512,5 +512,63 @@ namespace Duplicati.UnitTest
                 }
             }
         }
+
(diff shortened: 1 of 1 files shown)


seantempleton:
Thanks for writing this test warwickmm. It was and is extremely useful.

warwickmm:
Since this simply adds a test and was used in diagnosing issue #4037, I'll go ahead and merge this.

Issue #4046, opened 2020-01-10
missing dindex Repair fails -- 2.0.4.31 and up regression


ts678:
- [x] I have searched open and closed issues for duplicates.

----------------------------------------

## Environment info

- **Duplicati version**: 2.0.4.31 Canary
- **Operating system**: Windows 10 Version 1909
- **Backend**: Local folder

## Description

Missing dindex file Repair fails.

## Steps to reproduce

1. Make a backup of a short text file, using no encryption. --tempdir="C:\tmp" optional for privacy.
2. Delete the dindex file.
3. Do database Repair. This worked on 2.0.4.30 Canary and earlier and was a rare recovery method.

- **Actual result**:
Missing dindex file is not replaced. After about a minute of failed uploads to folder, popup error:
`Path cannot be null. Parameter name: path`

There is no job log made, but If you look in About --> Show log --> Stored you can see a stack trace:
```
System.ArgumentNullException: Path cannot be null.
Parameter name: path
   at Duplicati.Library.Main.BackendManager.WaitForComplete(LocalDatabase db, IDbTransaction transation)
   at Duplicati.Library.Main.Operation.RepairHandler.RunRepairRemote()
   at Duplicati.Library.Main.Operation.RepairHandler.Run(IFilter filter)
   at Duplicati.Library.Main.Controller.RunAction[T](T result, String[]& paths, IFilter& filter, Action`1 method)
   at Duplicati.Library.Main.Controller.Repair(IFilter filter)
   at Duplicati.Server.Runner.Run(IRunnerData data, Boolean fromQueue)
```
There's a side oddity on repeated Repair after the initial failure. Sometimes it believes all is fine even though it's still missing a dindex. Sometimes it actually replaces dindex. If in doubt, restart Duplicati. 

This has also at least once (maybe on different source data) errored about a Temp file it can't locate.
[Ensure that dindex references correct dblock file after renames #3938](https://github.com/duplicati/duplicati/pull/3938/files) added TemporaryIndexVolume and might be relevant. 2.0.4.31 is one of the few places where dindex file changes made changelog:

> Fixed a retry error where uploaded dindex-files would reference non-existing dblock files, thanks @warwickmm

[...]

PR #4047, opened 2020-01-10
Fix bug when performing repair with missing dindex files

warwickmm:
This fixes a bug introduced in pull request #3930 that broke the repair operation when `dindex` files were missing and no encryption was used.  Since we are keeping around a reference to the `IndexVolumeWriter` and `BlockVolumeWriter` instances, we cannot wrap their usages in `using` statements.

This also adds a test to verify the behavior.

This fixes issue #4046.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -69,7 +69,7 @@ jobs:
         directories:
           - $BUILD_DIR
       script:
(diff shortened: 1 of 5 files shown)


warwickmm:
The inline diff is a bit messy due to whitespace changes. I recommend viewing the [side-by-side diff](https://github.com/duplicati/duplicati/pull/4047/files?utf8=%E2%9C%93&diff=split&w=1).

warwickmm:
This hasn't garnered much feedback, but since it's a pretty obvious bug with a simple fix I'll go ahead and merge it.

PR #4049, opened 2020-01-11
Fix unexpected difference in fileset when using Stop Now

seantempleton:
The exception could occur due to the existing delete for FilesetEntry
deleting based on volume ids, which would only match a filelist.
The bsIdsSubQuery would return BlocksetIds that should
be deleted and would delete from the FileLookup table.

FilesetEntry would then have a higher count of rows
than what would be returned by the LEFT OUTER JOINS on all of the
other tables in VerifyConsistency() since more rows had been deleted
from FileLookup.

To fix this when rows are deleted in FileLookup, use the associated metadata id
to delete from FilesetEntry and other tables based on the blockset id for the
metadata id.

Some existing delete statements were moved, none were deleted.
The existing FileLookup delete had the match on metadata id removed as
a blockset id does not equal a metadata id.

### DIFF
--- a/Duplicati/Library/Main/Controller.cs
+++ b/Duplicati/Library/Main/Controller.cs
@@ -1076,12 +1076,11 @@ namespace Duplicati.Library.Main
         public void Stop(bool allowCurrentFileToFinish)
         {
             var ct = m_currentTask;
-            if (ct == null) return;
-            if (allowCurrentFileToFinish)
-            {
-                Logging.Log.WriteVerboseMessage(LOGTAG, "CancellationRequested", "Cancellation Requested");
-                m_cancellationTokenSource.Cancel();
-            }
+            if (ct == null)
+                return;
+
+            Logging.Log.WriteVerboseMessage(LOGTAG, "CancellationRequested", "Cancellation Requested");
+            m_cancellationTokenSource.Cancel();
             ct.Stop(allowCurrentFileToFinish);
         }
 
--- a/Duplicati/Library/Main/Database/LocalBackupDatabase.cs
+++ b/Duplicati/Library/Main/Database/LocalBackupDatabase.cs
@@ -2,10 +2,6 @@ using System;
 using System.Collections.Generic;
 using System.Data;
 using System.Linq;
-using System.Text;
-using System.IO;
-using System.Linq.Expressions;
-
 
 namespace Duplicati.Library.Main.Database
 {
@@ -842,7 +838,6 @@ namespace Duplicati.Library.Main.Database
         {
             using (var cmd = m_connection.CreateCommand(transaction))
             {
-                cmd.Transaction = transaction;
                 using (var rd = cmd.ExecuteReader(@"SELECT DISTINCT ""Fileset"".""ID"", ""Fileset"".""Timestamp"" FROM ""Fileset"", ""RemoteVolume"" WHERE ""RemoteVolume"".""ID"" = ""Fileset"".""VolumeID"" AND ""Fileset"".""ID"" IN (SELECT ""FilesetID"" FROM ""FilesetEntry"")  AND (""RemoteVolume"".""State"" = ""Uploading"" OR ""RemoteVolume"".""State"" = ""Temporary"")"))
                     while (rd.Read())
                     {
@@ -880,6 +875,23 @@ namespace Duplicati.Library.Main.Database
                     );
                 else
                     return default(RemoteVolumeEntry);
+        }
+
+        public IEnumerable<string> GetTemporaryFilelistVolumeNames(bool latestOnly, IDbTransaction transaction = null)
+        {
+            var incompleteFilesetIDs = GetIncompleteFilesets(transaction).OrderBy(x => x.Value).Select(x => x.Key).ToArray();
+
+            if (!incompleteFilesetIDs.Any())
+                return Enumerable.Empty<string>();
+
+            if (latestOnly)
+                incompleteFilesetIDs = new long[] { incompleteFilesetIDs.Last() };
+
+            var volumeNames = new List<string>();
+            foreach (var filesetID in incompleteFilesetIDs)
+                volumeNames.Add(GetRemoteVolumeFromFilesetID(filesetID).Name);
+
+            return volumeNames;
         }
 
         public IEnumerable<string> GetMissingIndexFiles(System.Data.IDbTransaction transaction)
--- a/Duplicati/Library/Main/Database/LocalDatabase.cs
+++ b/Duplicati/Library/Main/Database/LocalDatabase.cs
@@ -412,36 +412,67 @@ namespace Duplicati.Library.Main.Database
                 var volIdsSubQuery = string.Format(@"SELECT ""ID"" FROM ""{0}"" ", volidstable);
                 deletecmd.Parameters.Clear();
 
-                // If the volume is a block or index volume, this will update the crosslink table, otherwise nothing will happen
-                deletecmd.ExecuteNonQuery(string.Format(@"DELETE FROM ""IndexBlockLink"" WHERE ""BlockVolumeID"" IN ({0}) OR ""IndexVolumeID"" IN ({0})", volIdsSubQuery));
-
-                // If the volume is a fileset, this will remove the fileset, otherwise nothing will happen
-                deletecmd.ExecuteNonQuery(string.Format(@"DELETE FROM ""FilesetEntry"" WHERE ""FilesetID"" IN (SELECT ""ID"" FROM ""Fileset"" WHERE ""VolumeID"" IN ({0}))", volIdsSubQuery));
-                deletecmd.ExecuteNonQuery(string.Format(@"DELETE FROM ""ChangeJournalData"" WHERE ""FilesetID"" IN (SELECT ""ID"" FROM ""Fileset"" WHERE ""VolumeID"" IN ({0}))", volIdsSubQuery));
-                deletecmd.ExecuteNonQuery(string.Format(@"DELETE FROM ""Fileset"" WHERE ""VolumeID""  IN ({0})", volIdsSubQuery));
 
                 var bsIdsSubQuery = string.Format(
-                      @"SELECT ""BlocksetEntry"".""BlocksetID"" FROM ""BlocksetEntry"", ""Block"" "
+                      @"SELECT DISTINCT ""BlocksetEntry"".""BlocksetID"" FROM ""BlocksetEntry"", ""Block"""
                     + @" WHERE ""BlocksetEntry"".""BlockID"" = ""Block"".""ID"" AND ""Block"".""VolumeID"" IN ({0}) "
                     + @"UNION ALL "
-                    + @"SELECT ""BlocksetID"" FROM ""BlocklistHash"" "
-                    + @"WHERE ""Hash"" IN (SELECT ""Hash"" FROM ""Block"" WHERE ""VolumeID"" IN ({0}))"
+                    + @"SELECT DISTINCT ""BlocksetID"" FROM ""BlocklistHash"""
+                    + @" WHERE ""Hash"" IN (SELECT ""Hash"" FROM ""Block"" WHERE ""VolumeID"" IN ({0}))"
                     , volIdsSubQuery);
 
                 // Create a temporary table to cache subquery result, as it might take long (SQLite does not cache at all). 
                 deletecmd.ExecuteNonQuery(string.Format(@"CREATE TEMP TABLE ""{0}"" (""ID"" INTEGER PRIMARY KEY)", blocksetidstable));
                 deletecmd.ExecuteNonQuery(string.Format(@"INSERT OR IGNORE INTO ""{0}"" (""ID"") {1}", blocksetidstable, bsIdsSubQuery));
-                bsIdsSubQuery = string.Format(@"SELECT ""ID"" FROM ""{0}"" ", blocksetidstable);
+                bsIdsSubQuery = string.Format(@"SELECT DISTINCT ""ID"" FROM ""{0}"" ", blocksetidstable);
                 deletecmd.Parameters.Clear();
 
-                deletecmd.ExecuteNonQuery(string.Format(@"DELETE FROM ""FileLookup"" WHERE ""BlocksetID"" IN ({0}) OR ""MetadataID"" IN ({0})", bsIdsSubQuery));
+                // Create a temp table to associate metadata that is being deleted to a fileset
+                var metadataFilesetQuery = $@"SELECT Metadataset.ID, FilesetEntry.FilesetID
+FROM Metadataset
+INNER JOIN FileLookup ON FileLookup.MetadataID = Metadataset.ID
+INNER JOIN FilesetEntry ON FilesetEntry.FileID = FileLookup.ID
+WHERE Metadataset.BlocksetID IN ({bsIdsSubQuery})
+OR Metadataset.ID IN (SELECT MetadataID FROM FileLookup WHERE BlocksetID IN ({bsIdsSubQuery}))";
+
+                var metadataFilesetTable = @"DelMetadataFilesetIds-" + temptransguid;
+                deletecmd.ExecuteNonQuery($@"CREATE TEMP TABLE ""{metadataFilesetTable}"" (MetadataID INTEGER PRIMARY KEY, FilesetID INTEGER)");
+                deletecmd.ExecuteNonQuery($@"INSERT OR IGNORE INTO ""{metadataFilesetTable}"" (MetadataID, FilesetID) {metadataFilesetQuery}");
+
+                // Delete FilesetEntry rows that had their metadata deleted
+                deletecmd.ExecuteNonQuery($@"DELETE FROM FilesetEntry
+WHERE FilesetEntry.FilesetID IN (SELECT DISTINCT FilesetID FROM ""{metadataFilesetTable}"")
+AND FilesetEntry.FileID IN (
+	SELECT FilesetEntry.FileID
+	FROM FilesetEntry
+	INNER JOIN FileLookup ON FileLookup.ID = FilesetEntry.FileID
+	WHERE FileLookup.MetadataID IN (SELECT MetadataID FROM ""{metadataFilesetTable}""))");
+
+                // Delete FilesetEntry rows that had their blocks deleted
+                deletecmd.ExecuteNonQuery($@"DELETE FROM FilesetEntry WHERE FilesetEntry.FileID IN (
+SELECT ID FROM FileLookup
+WHERE FileLookup.BlocksetID IN ({bsIdsSubQuery}))");
+                deletecmd.ExecuteNonQuery($@"DELETE FROM FileLookup WHERE FileLookup.MetadataID IN (SELECT MetadataID FROM ""{metadataFilesetTable}"")");
+
                 deletecmd.ExecuteNonQuery(string.Format(@"DELETE FROM ""Metadataset"" WHERE ""BlocksetID"" IN ({0})", bsIdsSubQuery));
+                deletecmd.ExecuteNonQuery(string.Format(@"DELETE FROM ""FileLookup"" WHERE ""BlocksetID"" IN ({0})", bsIdsSubQuery));
                 deletecmd.ExecuteNonQuery(string.Format(@"DELETE FROM ""Blockset"" WHERE ""ID"" IN ({0})", bsIdsSubQuery));
                 deletecmd.ExecuteNonQuery(string.Format(@"DELETE FROM ""BlocksetEntry"" WHERE ""BlocksetID"" IN ({0})", bsIdsSubQuery));
-
                 deletecmd.ExecuteNonQuery(string.Format(@"DELETE FROM ""BlocklistHash"" WHERE ""BlocklistHash"".""BlocksetID"" IN ({0})", bsIdsSubQuery));
+
+                // If the volume is a block or index volume, this will update the crosslink table, otherwise nothing will happen
+                deletecmd.ExecuteNonQuery(string.Format(@"DELETE FROM ""IndexBlockLink"" WHERE ""BlockVolumeID"" IN ({0}) OR ""IndexVolumeID"" IN ({0})", volIdsSubQuery));
                 deletecmd.ExecuteNonQuery(string.Format(@"DELETE FROM ""Block"" WHERE ""VolumeID"" IN ({0})", volIdsSubQuery));
                 deletecmd.ExecuteNonQuery(string.Format(@"DELETE FROM ""DeletedBlock"" WHERE ""VolumeID"" IN ({0})", volIdsSubQuery));
+                deletecmd.ExecuteNonQuery(string.Format(@"DELETE FROM ""ChangeJournalData"" WHERE ""FilesetID"" IN (SELECT ""ID"" FROM ""Fileset"" WHERE ""VolumeID"" IN ({0}))", volIdsSubQuery));
+                deletecmd.ExecuteNonQuery($@"DELETE FROM FilesetEntry WHERE FilesetID IN (SELECT ID FROM Fileset WHERE VolumeID IN ({volIdsSubQuery}))");
+                deletecmd.ExecuteNonQuery($@"DELETE FROM Fileset WHERE VolumeID IN ({volIdsSubQuery})");
+
+                // Delete from Fileset if FilesetEntry rows were deleted by related metadata and there are no references in FilesetEntry anymore
+                deletecmd.ExecuteNonQuery($@"DELETE FROM Fileset WHERE Fileset.ID IN
+(SELECT DISTINCT FilesetID FROM ""{metadataFilesetTable}"")
+AND Fileset.ID NOT IN
+    (SELECT DISTINCT FilesetID FROM FilesetEntry)");
 
                 // Clean up temp tables for subqueries. We truncate content and then try to delete.
                 // Drop in try-block, as it fails in nested transactions (SQLite problem)
@@ -453,6 +484,7 @@ namespace Duplicati.Library.Main.Database
                     deletecmd.CommandTimeout = 2;
                     deletecmd.ExecuteNonQuery(string.Format(@"DROP TABLE IF EXISTS ""{0}"" ", blocksetidstable));
                     deletecmd.ExecuteNonQuery(string.Format(@"DROP TABLE IF EXISTS ""{0}"" ", volidstable));
+                    deletecmd.ExecuteNonQuery(string.Format(@"DROP TABLE IF EXISTS ""{0}"" ", metadataFilesetTable));
                 }
                 catch { /* Ignore, will be deleted on close anyway. */ }
 
--- a/Duplicati/Library/Main/Database/LocalDeleteDatabase.cs
+++ b/Duplicati/Library/Main/Database/LocalDeleteDatabase.cs
@@ -95,11 +95,11 @@ namespace Duplicati.Library.Main.Database
                 
                 //We save the block info for the remote files, before we delete it
                 cmd.ExecuteNonQuery(@"INSERT INTO ""DeletedBlock"" (""Hash"", ""Size"", ""VolumeID"") SELECT ""Hash"", ""Size"", ""VolumeID"" FROM ""Block"" WHERE ""ID"" NOT IN (SELECT DISTINCT ""BlockID"" AS ""BlockID"" FROM ""BlocksetEntry"" UNION SELECT DISTINCT ""ID"" FROM ""Block"", ""BlocklistHash"" WHERE ""Block"".""Hash"" = ""BlocklistHash"".""Hash"") ");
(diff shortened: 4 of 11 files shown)


seantempleton:
This could definitely use a look by someone who knows the database structure better. Even though I believe this to be correct I'm a little scared of just saying "oh delete some more data, that will fix the problem!".

I'm hoping this fixes #3644.

warwickmm:
Thanks @seantempleton.  This looks reasonable to me, but I'm not very familiar with this part of the code.  We may need some input from @kenkendk here.

kenkendk:
It has been way too long since I last looked at that part of the code.

Your new query appears to remove an file entries that have either a blocksetid or a metadataid in one of the deleted items.

I think this is how it should be, but the way I recall this, it should happen [in this query](https://github.com/duplicati/duplicati/pull/4049/files#diff-cef850fceae9a7aac64407b09da1d77bR419):
```csharp
deletecmd.ExecuteNonQuery(string.Format(@"DELETE FROM ""FilesetEntry"" WHERE ""FilesetID"" IN (SELECT ""ID"" FROM ""Fileset"" WHERE ""VolumeID"" IN ({0}))", volIdsSubQuery));
```

Do you know when/why it happens that the `FilesetEntry` is left behind?

seantempleton:
I duplicated this with a bunch of custom logging in place and I was also able to keep a copy of the database before the deletions.

The goal was to delete VolumeID 2. Block 1 belongs to Blockset 1 and the block is in volume 2.
A row in the `FileLookup` table was deleted that was matched on the `MetadataID` column of 1 because the current query is to delete anything from FileLookup where `BlocksetID` or `MetadataID` matches the list of BlocksetIDs in the temporary table. I followed the row from `FileLookup` through `FilesetEntry`, `Fileset`, and then over to `Remotevolume` and the row that was deleted from `FileLookup` belongs to volume 1.

The delete query for `FileLookup` is the only one that compares `MetadataID` to the list of `BlocksetIDs`. Is that comparison valid? Does a MetadataID always equal a BlocksetID? If not, then I suspect that part of the where clause shouldn't be there and then my fix here isn't valid either.

seantempleton:
Looking at the database of one of my own backups, the `Metadataset` table has differing `MetadataID` and `BlocksetID` values for 99% of the rows in the table. Based on that it seems logical to me that they are meant to be different values and track different things (probably the entire purpose of the `Metadataset` table, to correlate the two). I am going to try and remove the match on metadata (`OR ""MetadataID"" IN ({0})", bsIdsSubQuery))`) locally and see what happens.

seantempleton:
Thanks kenkendk for asking me about why this happens. I didn't fully understand what was going on and that is a reminder to not make any changes until I truly understand the problem. I should have done that in the first place and I apologize. I investigated deeper and now I think I really know why this happens.

I have been able to easily reproduce this by starting a backup and then hitting stop now as soon as it shows X number of files to go. After it stops, hit start again.

TLDR
The dlist file is protected from deletion during `PreBackupVerify`. (And I don't yet know why we do that.) The dblock files are deleted but all of the `FilesetEntry`s belong to the `FilesetID` that is associated to the dlist file so they do not get deleted and hence the difference in fileset exception is thrown.

Longer explanation
When the backup starts the [last dlist](https://github.com/duplicati/duplicati/blob/22d34f60c0b45bf282d784d669e1e1b312d4b7b7/Duplicati/Library/Main/Operation/BackupHandler.cs#L458) file is [protected](https://github.com/duplicati/duplicati/blob/22d34f60c0b45bf282d784d669e1e1b312d4b7b7/Duplicati/Library/Main/Operation/FilelistProcessor.cs#L308) from deletion while all of the other temporary volumes, the dblock files, are deleted.

The delete [from FilesetEntry](https://github.com/duplicati/duplicati/blob/22d34f60c0b45bf282d784d669e1e1b312d4b7b7/Duplicati/Library/Main/Database/LocalDatabase.cs#L419) doesn't delete anything because `FilesetID`s are only ever [associated with](https://github.com/duplicati/duplicati/blob/22d34f60c0b45bf282d784d669e1e1b312d4b7b7/Duplicati/Library/Main/Operation/BackupHandler.cs#L490) the volumes for dlist files.

Then when the [query](https://github.com/duplicati/duplicati/blob/22d34f60c0b45bf282d784d669e1e1b312d4b7b7/Duplicati/Library/Main/Database/LocalDatabase.cs#L837) is run in `VerifyConsistency` to compare against the count in `FilesetEntry`, since all of the blocksets, blocks, etc. have been deleted, the count returned is 0. This doesn't match the count of `FilesetEntry` rows and the exception is thrown.

In addition, I do currently believe the query that [deletes from](https://github.com/duplicati/duplicati/blob/22d34f60c0b45bf282d784d669e1e1b312d4b7b7/Duplicati/Library/Main/Database/LocalDatabase.cs#L437) `Filelookup` should have the `MetadataID` part removed. But that would indicate that when something is deleted from `Metadataset` based on `BlocksetID` that anything associated with the `MetadataID` of the deleted rows should be deleted.
I went and modified `RemoveRemoteVolumes` to track which `MetadataID`s are deleted from `Metadataset` based on `BlocksetID`, use those IDs to delete additional rows from `FileLookup`, and then use _those_ `FileID`s to delete from `FilesetEntry`. But that gets into deleting entries that belong to the protected dlist file. If we can go ahead and delete `FilesetEntry`s while the protected dlist file is still around, then that should be fine.

So ultimately I think this all boils down to the fact that in the scenario I can easily reproduce, the dlist file is protected so then we are guaranteed to have `FilesetEntry`s sitting around in the database.

Sorry if that's too much information or if some of it is unnecessary. I like to write things down in a lot of detail as otherwise I'll forget little details later on down the road if I don't.

warwickmm:
Excellent findings @seantempleton.  Having an easily reproducible scenario is a huge step forward.

seantempleton:
It turns out that the dlist file that is being protected is listed as a Temporary volume and the reason it is being protected is for use in building a synthetic filelist. However, a synthetic filelist is not actually generated because of [this check](https://github.com/duplicati/duplicati/blob/1dbaf360a64ca5cd0790836f4e4c8206f365c106/Duplicati/Library/Main/Operation/Backup/UploadSyntheticFilelist.cs#L54), where the volume is listed as Temporary, so I get the warning instead.

That information combined with if I delete rows from `FileLookup`, `Fileset`, and `FilesetEntry` (related to the `MetadataID` that was deleted, from my message above), allows the 2nd backup to complete and verify properly, though with the above noted warning. A restore also works properly.

seantempleton:
Are there steps to force a synthetic filelist to be created? I'm not super familiar with it's purpose yet. Also, do we know that it is in fact currently working? Unless I'm missing something it seems to me like it is impossible.

Slight pseudo code
```
if (syntbase.Name == null || syntbase.State != RemoteVolumeState.Uploaded)
    return;
if (syntbase.Name == null || (syntbase.State != RemoteVolumeState.Uploading && syntbase.State != RemoteVolumeState.Temporary))
    return;

Code down here to generate the synthetic filelist
```
1. syntbase.Name is null, then we hit the first return.
2. Name is not null but State is any other state but Uploaded, we hit the first return.
3. Name is null, we can't hit the second return because we hit the first.
4. State has to be Uploaded to get to the second if, and Upload != Uploading or Temporary so we hit the second return.

No path leads to generating the synthetic filelist.

warwickmm:
I recall hearing that there are some issues with the synthetic filelist, but I'm not sure exactly what the issues are.  There is a related discussion [here](https://github.com/duplicati/duplicati/issues/2506).

warwickmm:
I think your analysis is correct.  It appears that the code after the two `if` statements will never be reached.

seantempleton:
I think I've got that fixed. I'm also trying to knock out this fileset exception at the same time.

kenkendk:
> Are there steps to force a synthetic filelist to be created?

It _should_ create a new synthetic filelist, if you stop a backup, and let it finish uploading.

In this case, we do not know what files are missing. To avoid problems with missing files, we assume that all the files that existed in the previous backup also exist this time. The synthetic filelist is then created by making a copy of the previous filelist, and appending the files found before stopping (this may replace files, if a file was changed).

I think the syntethetic list is also created after a forced stop. In this case, there exists data (`dblock` and `dindex`) on the remote backend that has no reference files.

At the time, it was the best idea I had for supporting "partial backups". However, the code has changed a bit since then, as there is a problem where creating multiple "partial backups" can eventually cause files to be removed as old versions are deleted.

The "new" approach is to mark some backups as full and others as partial, such that the removal process can take this information into account, when choosing which backups to remove.

So, it is possible that the synthetic filelists are no longer being created in any circumstance.

> It turns out that the dlist file that is being protected is listed as a Temporary volume and the reason it is being protected is for use in building a synthetic filelist.

Yes, that makes sense. The synthetic filelists needs to be based on the previous (incomplete) filelist, which is appended to the last backup before that (as explained above).

> Looking at the database of one of my own backups, the Metadataset table has differing MetadataID and BlocksetID values for 99% of the rows in the table. Based on that it seems logical to me that they are meant to be different values and track different things (probably the entire purpose of the Metadataset table, to correlate the two).

Yes, that is correct. Each `FileLookup` (originally just `File`) contains "`Path`, `BlocksetID`, `MetadataID`". The `BlocksetID` points to the sequence of blocks that comprise the file contents. The `MetadataID` points to the `Metadataset`, which just points to `BlocksetID`, giving the sequence of blocks that comprise the metadata.

Ideally, metadata and contents should be treated equally: just a stream of bytes. Only reading and writing the streams should be done differently (obviously).

The original idea with `Metadataset` was to record additional information, but it never happened. Maybe there is no need for storing metadata explictly in the database. It could be used to store simple thing like "lastmodified" or similar, so the GUI can search for files with these properties, but maybe we can add this to the `FileLookup` table instead.

I considered removing `Metadataset` a few times, as it just adds an extra lookup in the database, but there was at least one place where it became hard to keep track.

ts678:
> So, it is possible that the synthetic filelists are no longer being created in any circumstance.

2.0.5.1 certainly seems to be trying or at least is expecting. Reason it fails (still) proposed [here](https://forum.duplicati.com/t/large-backup-fails-after-first-successful-run/9250/15?u=ts678).

seantempleton:
Thanks for the explanation @kenkendk, it was helpful.

Along with #3712, I think this should get Stop Now into a usable state. There are still some edge cases where the overall behavior is bad / weird.

For instance, create a brand new backup, throttle the streams to some low number, as soon as it starts uploading block files hit Stop Now. If you then immediately go to Restore files a fileset shows up, but when you tell it to do the restore it goes into `VerifyRemoteList` first, removes the filelist, and then the restore fails with the error that there are no filesets.

I'm hoping someone can try this out and see if they run into any problems that I haven't found yet.

drwtsn32x:
@seantempleton I can try to find some time to test this out. Thanks for your efforts!

drwtsn32x:
Testing with a new backup set. Local destination, no encryption, no backup versions (yet).

I've tried a few tests so far with hitting the Stop Now button.  In each case I end up getting this error:

`CoCoL.RetiredException: The channel "BackendRequests" is retired`

I end up with dblocks and dindexes on the back end, but the dlist was never written.  Maybe this is expected for interrupted backups?

If I go to Restore files, the interrupted backup is shown in the dropdown.  Should interrupted backups appear here?  If so should they have the "(partial)" suffix?

If I attempt to restore one of the files, I get this error:

`Duplicati.Library.Interface.UserInformationException: No backup at the specified date`

If I go back to Restore files again, the interrupted backup version is no longer shown at all - the dropdown is empty.

If I re-run the backup job without interrupting it, it completes successfully and then 1 version is shown in the Restore dropdown.  Tested restores and they seemed to work.

mcronce:
I'm not here to add anything constructive, just to thank @seantempleton for tackling the issue that has made Duplicati 100% unusable for me for multiple years.  :fireworks: :)

ts678:
> throttle the streams to some low number

makes me invite a few people to see if they see or recall issue area below where I detail prior history:

[--throttle-download ignored, --throttle-upload throttles download too #4115](https://github.com/duplicati/duplicati/issues/4115)

Apologies for the diversion, and please follow up there. I thought it was a good opportunity to point.

ts678:
I'm guessing the @mcronce reference is to the below, which involves an OOM kill:

[Error / DB corruption: Unexpected difference in fileset version X: found Y entries, but expected Z #3644](https://github.com/duplicati/duplicati/issues/3644#issuecomment-498824126)

although I hoped it was the one **not** involving hard stop that's [fixed](https://github.com/duplicati/duplicati/pull/3801) in 2.0.5.1 (actually [2.0.4.22 Canary](https://github.com/duplicati/duplicati/releases/tag/v2.0.4.22-2.0.4.22_canary_2019-06-30)). That was what was making Duplicati nearly unusable for me, due to the nature of my primary backup...

mcronce:
@ts678 OOM kill was one of the causes, but not the only one.  Upload interruptions in general.

seantempleton:
Thanks for testing @drwtsn32x. Sorry, I should have spelled out what to expect.

> I've tried a few tests so far with hitting the Stop Now button. In each case I end up getting this error:
> 
> `CoCoL.RetiredException: The channel "BackendRequests" is retired`
> 
I have not tried to prevent the CoCoL.RetiredException from bubbling up yet. From looking into that in the past I think it will be tricky. The problem is that the exception is thrown when there are legitimate errors, but it is also thrown when the program is behaving properly.

> I end up with dblocks and dindexes on the back end, but the dlist was never written. Maybe this is expected for interrupted backups?

Yes, the dlist doesn't get written to the backend until everything else is finished and written.

> If I go to Restore files, the interrupted backup is shown in the dropdown. Should interrupted backups appear here? If so should they have the "(partial)" suffix?
> If I attempt to restore one of the files, I get this error:
> 
> `Duplicati.Library.Interface.UserInformationException: No backup at the specified date`

I am of the opinion if there were dblock files successfully uploaded then the interrupted backup should appear there. The RestoreHandler calls [VerifyRemoteList](https://github.com/duplicati/duplicati/blob/c5b91503b8ede912863d641d79f8c383f5b08a85/Duplicati/Library/Main/Operation/RestoreHandler.cs#L336) and does not pass in a protected file. The dlist from the interrupted backup has the state of `Temporary` in the database and hence the call deletes it right before trying to do the restore. Maybe that can do the same thing the BackupHandler [does](https://github.com/duplicati/duplicati/blob/94ea86b2845d0382db3c51a3a994bb5063b562a7/Duplicati/Library/Main/Operation/BackupHandler.cs#L456) by protecting the dlist from deletion.
I'll try that and see what might happen in that situation.

> I'm not here to add anything constructive, just to thank @seantempleton for tackling the issue that has made Duplicati 100% unusable for me for multiple years. 🎆 😄

Thanks @mcronce. Just trying to defeat some corruption issues!

mcronce:
Haha it was an emotional rollercoaster finding this the other day.  First I saw that #3644 was fixed and in a beta release and I was super excited.  Then I saw that it isn't expected to fix my problem and I was super disappointed.  Then I found my way here and saw that there's a PR open and just has some final issues being tracked down and was excited again.

Thanks again ;)

ts678:
The problem is that the error is so generic (it's an internal consistency error) that causes may vary.

> Upload interruptions in general.

Does that include things like ones due to network errors? I test that sometimes, but not rigorously.
There have been proposals for automated torture-testing, but nobody has had time to set that up.

> Thanks again ;)

Slowly trying to knock things into better shape. 3644 seemed spontaneous. Thanks to all who help.

mcronce:
> Does that include things like ones due to network errors?

Network errors were another, yeah.  I also had times where I'd be several days into an upload and it had just stalled with no explanation.  (Maybe transient network issues, maybe something else, I couldn't find any forensics available to determine the root cause.)  The stop button (at least at the time) did nothing, and restarting the service resulted in database corruption.  I expect #3644 will probably have solved that last one.

In any case, it seems like the corruption issues are dwindling down.  Once this fix is available in a canary I'm going to update the helm chart and give it another shot.

kenkendk:
@seantempleton It looks like you deleted the original commit, and added a new one.
Do you feel confident enough about the fix, that we can roll it into a canary build?

seantempleton:
Not yet @kenkendk. I am pretty confident with what I have here right now though I am almost finished with the comments from @drwtsn32x. Those changes will be pretty small.

seantempleton:
I think this is good to go now unless someone finds something.

When stop now is used the channel retired exception is handled (technically it's never thrown in the first place now), the backup is listed as partial if you try to restore from it right away, and when restoring from the partial backup it actually works.

When stopping a backup the dlist file is not uploaded. If synthetic filelists are enabled, on the next backup the previous temporary filelist (n - 1) is used along with filelist before that (n - 2) to create the new synthetic filelist, and the temporary filelist is deleted. So you won't see that filelist in the restore files backup list anymore.

If synthetic filelists are disabled, the temporary filelist is just deleted.

seantempleton:
Oops. I've been working with my 3 PRs at the same time. The build failure requires the `GetRemoteVolumeFromFilesetID` function from #4114.

warwickmm:
@seantempleton, pull request #4114 has been merged.  Can you update this branch to pick up the required dependencies?

seantempleton:
@warwickmm I'm not sure if you got a notification or not. I updated the branch.

warwickmm:
Thanks @seantempleton.  For some reason, I don't think I get notifications for force push operations.

I noticed some behavior that I don't recall happening before.  After interrupting a backup with "stop now", the restore list shows a partial version (like you described) with a subset of the files.  If I then run an uninterrupted backup and go to the restore list, the prior partial version now shows the full set of files.  Is this expected?

warwickmm:
@drwtsn32x, would you be able to run some more tests on Windows?

seantempleton:
@warwickmm Did you have a complete backup first, then interrupt the 2nd, and complete a 3rd?

For the scenario of backup 1 complete, 2 interrupted, and the 3rd backup complete:

If synthetic filelists are enabled, when the 3rd backup is started, backup 1 is appended to backup 2 and a new fileset is created. The backup time in the restore window should be different from the original 2nd backup time. If you look in the database before and after you'll see that the fileset for the 2nd backup is actually deleted and a new fileset is created.

If synthetic filelists are disabled then the interrupted backup will be deleted when the 3rd backup starts. And by that I mean the local filelist that was never uploaded and the fileset will be deleted.

warwickmm:
> Did you have a complete backup first, then interrupt the 2nd, and complete a 3rd?

Yes, I think that's what actually happened.

I think I understand your explanation (and I just saw your related comment above).  However, this behavior is a bit confusing to me, and I imagine would confuse some users as well.

For example, I encountered the following situation:
1. Run a full backup of 3 files: `a`, `b`, and `c`
2. Modify files `b` and `c`
3. Run a full backup and interrupt with stop now
4. Restore list shows a partial backup with only file `a`
5. Run a full backup
6. Restore list for the previous partial backup now shows files `a`, `b`, and `c`

If I now restore files `b` and `c` from the second (partial) backup, I get the original versions from step 1.  This to me seems misleading, since this was not the state of the files for that version.  This is also inconsistent with how "stop after current file" behaves.  In my experiments, the restore view for the partial fileset is not filled in with entries from the prior fileset when using "stop after current file".

Unfortunately, I still don't fully understand the purpose of the synthetic filelist.  The behavior described above feels a bit dangerous since the fileset (although marked partial) appears to be complete, but is actually inconsistent.

Perhaps I should have asked this earlier, but does the synthetic filelist still have a purpose now that we have a notion of a partial fileset?  Ideally for the user, "stop now" would behave just like "stop after current file", except that "stop now" would stop sooner with possibly fewer files backed up.

> The backup time in the restore window should be different from the original 2nd backup time.

I didn't observe this, but perhaps I didn't wait long enough before performing the 3rd backup?

Thanks for your patience and hard work @seantempleton.  This is one of the most annoying corruption issues and it would be a huge accomplishment to finally have it resolved.

seantempleton:
I haven't tried out "stop after current file" in this context so I'm not sure of it's behavior. I'm guessing it stops the backup properly so the dblock files and dlist file are all uploaded, no volumes are listed as temporary or uploading, and everything is recorded properly in the database. So it's like a complete backup in that everything finishes properly, it just didn't get all of the files on the local machine.

If all that's true, I'm sure that's why you don't see the files from the previous fileset after using "stop after current file" and then doing another complete backup, because no synthetic filelist is created.

> Unfortunately, I still don't fully understand the purpose of the synthetic filelist. The behavior described above feels a bit dangerous since the fileset (although marked partial) appears to be complete, but is actually inconsistent.

I do not fully understand what the synthetic filelist is used for either, though kenkendk's [explanation](https://github.com/duplicati/duplicati/pull/4049#issuecomment-587431894) helps somewhat. It seems like the main reasoning was for preventing missing files (why is file `d` not showing up in my backup when it's been on my hard drive for years?) and to not delete files when cleaning up.
I think you're right about there being confusion though. If you modify a file `c`, the backup gets interrupted and file `c` was not recorded, a synthetic filelist (synthetic backup/fileset really) gets created, then when you go to restore you see file `c` and think it will have the latest contents but in fact it will not. The contents would be from the previous backup.

> Perhaps I should have asked this earlier, but does the synthetic filelist still have a purpose now that we have a notion of a partial fileset?

I'm coming in here relatively new to the project and have less knowledge about most things than all of you. Based on kenkendk's comment it sounds like a partial backup is a newer concept than a synthetic filelist.

From my current understanding of things I would lean towards a stopped backup would be listed as partial, any files in dblock files that had been successfully uploaded would be kept in the backup, and any files not processed or simply not uploaded would be removed from the database for that fileset. 
A user could have files missing from an interrupted backup, but I think that is a relatively easy thing to explain to them. Your backup was interrupted and was not able to get to all files, hence why some of them do not appear.

I'm not sure where to go from here. If it is decided that a partial backup is enough and synthetic filelists are no longer needed, I would be happy to remove them. I don't think it would be too bad to do that. Otherwise, it seems like what you wrote about your findings @warwickmm are fundamental to how synthetic filelists work. I'm not really sure how it could be changed to not do that either without some kind of backup like process examining files to determine what should be where.

Thanks @warwickmm and thanks to everyone for helping me work through this and test. And for allowing my PRs to hang out here for a while.

warwickmm:
> From my current understanding of things I would lean towards a stopped backup would be listed as partial, any files in dblock files that had been successfully uploaded would be kept in the backup, and any files not processed or simply not uploaded would be removed from the database for that fileset.  A user could have files missing from an interrupted backup, but I think that is a relatively easy thing to explain to them. Your backup was interrupted and was not able to get to all files, hence why some of them do not appear.

This would be my preference.  In my opinion, "stop now" and "stop after current file" should yield the same state.  The only difference is how soon the operation is stopped, and how many files are captured in the fileset.  Given this approach, I think the synthetic fileset is no longer necessary.  If anyone here disagrees, please feel free to comment.

Since this approach still involves the use of a `CancellationToken`, I'm guessing that an unexpected interruption (e.g. power failure, laptop suspend, reboot) might still leave things in an undesirable state?  If so, we should tackle that separately (but hopefully soon).

ts678:
I think synthetic filelist is aimed at things other than the two stop button flavors. Do we need two if we can actually make them equivalent except for speed? I'd probably favor the one that responds sooner.

Previous descriptions include:

[Expected there to be a temporary fileset for synthetic filelist #2506](https://github.com/duplicati/duplicati/issues/2506#issuecomment-304622076)

> The "synthetic filelist" is created after a backup has been interrupted.
It is synthetic as it uses the previous (successful) backup, and then just adds whatever new/updated files it managed to upload.

[Warning: Expected there to be a temporary fileset for synthetic filelist #2329](https://github.com/duplicati/duplicati/issues/2329#issuecomment-281472917)

> The error means that:
>
>   1. Duplicati detected an unfinished backup
>   2. It attempts to upload a "synthetic filelist", which is "last succesful backup + new files"

and there I also explain my [theory](https://github.com/duplicati/duplicati/issues/2329#issuecomment-449896156) (mentioned [earlier here](https://github.com/duplicati/duplicati/pull/4049#issuecomment-589794317)) of a faulty lookup preventing processing.

Because of this bug, it's hard to tell how well the code works if it could be run, however from a user's point of view there should be a reasonable result from resuming a backup that had an untidy ending. There should be no need to (for example) avoid computer stop or reboot. Duplicati should handle it.

One particular trouble spot is on initial backup, which is often slow, therefore has more time to suffer some unexpected death. Having no backups and no dlist file gets in the way of trying to recover this. Synthetic filelist would (I think -- it's been a long time) put up the dlist for interrupted backup before continuing. I suppose the manual equivalent with stop button would be to restart for controlled stop.

warwickmm:
> Do we need two if we can actually make them equivalent except for speed?

A case can be made for having both.  If we can get both to work correctly, then that would be nice.

> There should be no need to (for example) avoid computer stop or reboot. Duplicati should handle it.

Is an unplanned interruption the only use case for the synthetic filelist (e.g., computer reboot)?  Is it possible to have such an interruption result in a partial backup without the need for a synthetic filelist?  If not, I would prefer that  "stop now" and "stop after upload" not use the synthetic filelist.  In my opinion, it can be misleading (although perhaps I'm just not used to the idea).

ts678:
> Is an unplanned interruption the only use case for the synthetic filelist (e.g., computer reboot)?

These questions deserve more expertise in the code than I have, and due to the bad lookup, aren't testable now. A question I have about fixing the bad lookup is whether the rest of the plan works...

Decision on whether to do synthetic filelist might be the below:

https://github.com/duplicati/duplicati/blob/c86ca0f38c7edc945e69d291afd63c3eb6d51ee4/Duplicati/Library/Main/Operation/BackupHandler.cs#L455-L473

https://github.com/duplicati/duplicati/blob/c86ca0f38c7edc945e69d291afd63c3eb6d51ee4/Duplicati/Library/Main/Database/LocalBackupDatabase.cs#L841-L855

>  Is it possible to have such an interruption result in a partial backup without the need for a synthetic filelist?

Another good question. I'd guess one might be able to pull this off if only persistent storage was involved, for example the database and files on a local drive, but not having the remote updated continuously would mean the backup technically is broken until something cleans up the remote.

If one wanted to invent something that would do that, maybe one just invented synthetic filelist?

> If not, I would prefer that "stop now" and "stop after upload" not use the synthetic filelist.

I'm not following "stop now" progress closely I'm afraid (but thanks to all who are working on it!), however I'm pretty sure "stop after current file" does tidy shutdowns and needs no fixing up later.

[Question concerning “Waiting for upload to finish…”](https://forum.duplicati.com/t/question-concerning-waiting-for-upload-to-finish/9383/6?u=ts678) agrees with my sad impression that stopping Duplicati cleanly can be hard. TrayIcon can do it -- I'm not certain about the other Duplicati forms.

seantempleton:
> I think synthetic filelist is aimed at things other than the two stop button flavors.

I think you're right @ts678. I don't think it was really meant for the stop buttons at all, but more for interrupted backups.
(And your theory about the bad lookup is correct. It was using the wrong id to do the lookup. Fixed in #4114)

> Do we need two if we can actually make them equivalent except for speed?

There's the obvious how fast do they want to stop, but that's a simple view of it. You can always find a reason for something to exist. For instance they might be busy uploading a 2GB volume and really want that to finish before it stops, so they would want stop after current file rather than stop now.
But then again, there is always the problem of analysis paralysis (having too many choices), confusion on how things behave, etc. Designing and writing software is hard. :)

When I was working on this I started out focusing on the unexpected difference in fileset error when using stop now and that kept expanding to fixing more things as I went. As this PR stands right now it was mainly an effort to fix that exception and then make stop now functional, without really changing the design of anything. So the backup is stopped, the dlist file is not uploaded, and hence the database and backup itself are not in a consistent state until the next backup runs, a synthetic backup is created, and the cleanup code is run.

I was then thinking about this further a few days ago and came to the conclusion that both stop buttons need to leave the database in a well known and consistent state right when it stops. It is a user controlled action and should be an expected and normal thing.
I think I'm going to try and make that happen with the stop now button. There will be some hoops to jump through. For instance the BackendUploader is told to stop and cancel any in-progress uploads, but then its thread exits and therefore it can't upload the filelist.

I need to see what stop after current file does, but I'm imagining this will turn out to be a partial backup like @warwickmm wants. Stop the backup right now, canceling anything that can be canceled, cleanup the database to match what was uploaded and not uploaded, upload the filelist, and mark it as a partial backup.
(Hopefully that won't take too long as then stop _now_ turns into stop soon.)

If we use the synthetic filelist it will always have the problem of combining the previous backup and the incomplete backup, and if file C was modified but not uploaded, then the file C that could be restored from this combined backup is from the previous backup, not the most recently modified version. Without some kind of UI to indicate that, that will definitely be misleading to the user.

warwickmm:
Thanks @seantempleton!  We really appreciate your efforts here.

Squarebro:
I also have something interesting.

I have a backup that was working fine until I installed version v2.0.5.1. 
Before I installed it my backup was working fine on old version and there were no unexpected errors. If i restore old version I can still make new backups.  I will add this backup had some Start-Stop usage but it was long time ago and until this new version nothing was bad. Were there some changes that made Duplicati more "strict" and in a way show more "Unexpected differences" errors?

@seantempleton Is there a way this pull request could fix existing backup that already got problems?

Could a experimental version of Duplicati be released with all this changes?

seantempleton:
@Squarebro No, unfortunately it will not fix any existing backups that are broken. The work for fixing the unexpected difference in fileset error happens when starting a backup after an interrupted backup. After that time it's too late.

In my opinion this could be released as is for testing if people wanted it. It can have some unexpected behavior with regards to what version of a file you might get if a restore is done using the backup that immediately follows an interrupted backup. As far as I'm aware, that's the only thing to be aware of right now.

mcronce:
FWIW, I'd be interested in a container image to test this out

PR #4053, opened 2020-01-18
Add test for restoring files using RecoveryTool

warwickmm:
This adds a simple test for recovering files using the `RecoveryTool`.  We verify that the backend files can be downloaded, decrypted, and used to restore the source files to a different location.

In doing so, we discovered an instance of `HashLookupHelper` that wasn't disposed that kept an open `FileStream` to the `index.txt` file.


warwickmm:
I'll have to investigate the AppVeyor and Travis failures.  The test passes on my local machine.

warwickmm:
The tests were failing because of some issues that prevented the teardown methods from cleaning up the test files:
* There was an instance of `HashLookupHelper` that wasn't disposed, which kept an open `FileStream` to the `index.txt` file.
* The `RecoveryTool` changes the current directory to the target directories, which prevented their removal.

These issues should now be resolved.

PR #4054, opened 2020-01-19
Write blocklists when compacting files

### LINKED ISSUE #4048: compact may fail to write useful list folder of dindex file

## Environment info

- **Duplicati version**: 2.0.3.5 to 2.0.5.0 were the ends of the tested range. Looks like a very old bug.
- **Operating system**: Windows 10 Version 1909
- **Backend**: Local folder

## Description

Compact may fail to write the list folder into the newly created dindex file created by a dblock merge. 
I think the files in this folder are performance-enhancers, controlled by [--index-file-policy](https://duplicati.readthedocs.io/en/latest/06-advanced-options/#index-file-policy), so the dlist file `blocklists` reference can read the blocklist from the dindex file without downloading its dblock. 

Downloading dblock files for DB rebuild in Recreate or direct restore seems to go against the design intent, slows things down, and might get questions or complaints from people who observe carefully:

[Repair is downloading dblock files](https://forum.duplicati.com/t/repair-is-downloading-dblock-files/2780)

Note that this is a far more subtle performance reduction than the all-dblocks-downloaded situation.

[Database recreate desperately needs improvement #4041](https://github.com/duplicati/duplicati/issues/4041) has one person who might be heading for downloading all dblocks, and one who might be hitting bad dindex files. See technical remarks there.

## Steps to reproduce

1. Prepare four test files, which are variations of a 1024 text character file, intended to force all files to get a blocklist, and all blocks to be different. Line endings shouldn't matter. My notepad had none.

```
1more.txt
1 1234567890123456789012345678901234567890123456789012345678901234567890123456789012345678901234567890123456789012345678901234567890123456789012345678901234567890123456789012345678901234567890123456789012345678901234567890123456789012345678901234567890123456789012345678901234567890123456789012345678901234567890123456789012345678901234567890123456789012345678901234567890123456789012345678901234567890123456789012345678901234567890123456789012345678901234567890123456789012345678901234567890123456789012345678901234567890123456789012345678901234567890123456789012345678901234567890123456789012345678901234567890123456789012345678901234567890123456789012345678901234567890123456789012345678901234567890123456789012345678901234567890123456789012345678901234567890123456789012345678901234567890123456789012345678901234567890123456789012345678901234567890123456789012345678901234567890123456789012345678901234567890123456789012345678901234567890123456789012345678901234567890123456789012345678901234567890123456789012345678901234

2more.txt
12 1234567890123456789012345678901234567890123456789012345678901234567890123456789012345678901234567890123456789012345678901234567890123456789012345678901234567890123456789012345678901234567890123456789012345678901234567890123456789012345678901234567890123456789012345678901234567890123456789012345678901234567890123456789012345678901234567890123456789012345678901234567890123456789012345678901234567890123456789012345678901234567890123456789012345678901234567890123456789012345678901234567890123456789012345678901234567890123456789012345678901234567890123456789012345678901234567890123456789012345678901234567890123456789012345678901234567890123456789012345678901234567890123456789012345678901234567890123456789012345678901234567890123456789012345678901234567890123456789012345678901234567890123456789012345678901234567890123456789012345678901234567890123456789012345678901234567890123456789012345678901234567890123456789012345678901234567890123456789012345678901234567890123456789012345678901234567890123456789012345678901234

3more.txt
123 1234567890123456789012345678901234567890123456789012345678901234567890123456789012345678901234567890123456789012345678901234567890123456789012345678901234567890123456789012345678901234567890123456789012345678901234567890123456789012345678901234567890123456789012345678901234567890123456789012345678901234567890123456789012345678901234567890123456789012345678901234567890123456789012345678901234567890123456789012345678901234567890123456789012345678901234567890123456789012345678901234567890123456789012345678901234567890123456789012345678901234567890123456789012345678901234567890123456789012345678901234567890123456789012345678901234567890123456789012345678901234567890123456789012345678901234567890123456789012345678901234567890123456789012345678901234567890123456789012345678901234567890123456789012345678901234567890123456789012345678901234567890123456789012345678901234567890123456789012345678901234567890123456789012345678901234567890123456789012345678901234567890123456789012345678901234567890123456789012345678901234

4more.txt
1234 1234567890123456789012345678901234567890123456789012345678901234567890123456789012345678901234567890123456789012345678901234567890123456789012345678901234567890123456789012345678901234567890123456789012345678901234567890123456789012345678901234567890123456789012345678901234567890123456789012345678901234567890123456789012345678901234567890123456789012345678901234567890123456789012345678901234567890123456789012345678901234567890123456789012345678901234567890123456789012345678901234567890123456789012345678901234567890123456789012345678901234567890123456789012345678901234567890123456789012345678901234567890123456789012345678901234567890123456789012345678901234567890123456789012345678901234567890123456789012345678901234567890123456789012345678901234567890123456789012345678901234567890123456789012345678901234567890123456789012345678901234567890123456789012345678901234567890123456789012345678901234567890123456789012345678901234567890123456789012345678901234567890123456789012345678901234567890123456789012345678901234

```
2. Backup 1more.txt and 2more.txt, using --blocksize=1 KByte and --keep-versions=1, also without encryption and with --no-auto-compact true  for step-by-step view and ease of watching dindex.
3. Add 3more.txt and 4more.txt into backup via checkbox editing or move into folder, then backup.
4. Remove 1more.txt and 3more.txt from the backup, then backup. Uncheck is slightly tidier than a delete-from-folder because it won't even need to output a dblock to contain folder info update.
5. Compact. This merges the two half-empty dblock files into a new dblock, and its new dindex file.

- **Actual result**:
New dindex file has no list folder.
- **Expected result**:
New dindex file has list folder with two files containing the blocklists for 2more.txt and 4more.txt.

I'm not sure if you'll notice, but if you're looking very carefully in the DB and filenames, you'll see that the list folder filenames are not exactly the Base64 encoding of the hash, but a character-safe variant:

https://github.com/duplicati/duplicati/blob/52929e9eb2d88fbf48fb7f235976830b81c6cb34/Duplicati/Library/Main/Volumes/IndexVolumeWriter.cs#L75-L77

One way to try to spot this problem "in the wild" is to look for compact in logs, then look at the dindex from that time. Or look for dindex files with no list folder, then see if it was by compact. Without known sources though, it's hard to know whether no list folder is from all small files thus no blocklists needed.

seantempleton:
The list folder was not being written to dindex files when performing a
compact operation. After the data blocks have been moved to a new
volume the blocklists are now read from the database and written to
the new dindex file.

Fixes #4048

### DIFF
--- a/Duplicati/Library/Main/BackendManager.cs
+++ b/Duplicati/Library/Main/BackendManager.cs
@@ -1180,7 +1180,7 @@ namespace Duplicati.Library.Main
                 throw m_lastException;
         }
 
-        public void Put(VolumeWriterBase item, IndexVolumeWriter indexfile = null, bool synchronous = false)
+        public void Put(VolumeWriterBase item, IndexVolumeWriter indexfile = null, Action indexVolumeFinishedCallback = null, bool synchronous = false)
         {
             if (m_lastException != null)
                 throw m_lastException;
@@ -1212,6 +1212,11 @@ namespace Duplicati.Library.Main
                 req2 = new FileEntryItem(OperationType.Put, indexfile.RemoteFilename);
                 req2.LocalTempfile = indexfile.TempFile;
                 req.Indexfile = new Tuple<IndexVolumeWriter, FileEntryItem>(indexfile, req2);
+
+                indexfile.FinishVolume(req.Hash, req.Size);
+                indexVolumeFinishedCallback?.Invoke();
+                indexfile.Close();
+                req.IndexfileUpdated = true;
             }
 
             try
--- a/Duplicati/Library/Main/Operation/CompactHandler.cs
+++ b/Duplicati/Library/Main/Operation/CompactHandler.cs
@@ -15,12 +15,11 @@
 //  You should have received a copy of the GNU Lesser General Public
 //  License along with this library; if not, write to the Free Software
 //  Foundation, Inc., 59 Temple Place, Suite 330, Boston, MA 02111-1307 USA
-using System;
-using System.Linq;
-using System.Collections.Generic;
 using Duplicati.Library.Main.Database;
 using Duplicati.Library.Main.Volumes;
-using System.Text;
+using System;
+using System.Collections.Generic;
+using System.Linq;
 
 namespace Duplicati.Library.Main.Operation
 {
@@ -101,7 +100,7 @@ namespace Duplicati.Library.Main.Operation
                     var backend = bk ?? sharedBackend;
                     if (!hasVerifiedBackend && !m_options.NoBackendverification)
                         FilelistProcessor.VerifyRemoteList(backend, m_options, db, m_result.BackendWriter);
-        
+
                     BlockVolumeWriter newvol = new BlockVolumeWriter(m_options);
                     newvol.VolumeID = db.RegisterRemoteVolume(newvol.RemoteFilename, RemoteVolumeType.Blocks, RemoteVolumeState.Temporary, transaction);
     
@@ -140,7 +139,7 @@ namespace Duplicati.Library.Main.Operation
                                                  where report.CompactableVolumes.Contains(v.Name)
                                                  select (IRemoteVolume)v).ToList();
 
-                        using(var q = db.CreateBlockQueryHelper(transaction))
+                        using (var q = db.CreateBlockQueryHelper(transaction))
                         {
                             foreach (var entry in new AsyncDownloader(volumesToDownload, backend))
                             {
@@ -174,15 +173,7 @@ namespace Duplicati.Library.Main.Operation
 
                                                 if (newvol.Filesize > m_options.VolumeSize)
                                                 {
-                                                    uploadedVolumes.Add(new KeyValuePair<string, long>(newvol.RemoteFilename, newvol.Filesize));
-                                                    if (newvolindex != null)
-                                                        uploadedVolumes.Add(new KeyValuePair<string, long>(newvolindex.RemoteFilename, newvolindex.Filesize));
-
-                                                    if (!m_options.Dryrun)
-                                                        backend.Put(newvol, newvolindex);
-                                                    else
-                                                        Logging.Log.WriteDryrunMessage(LOGTAG, "WouldUploadGeneratedBlockset", "Would upload generated blockset of size {0}", Library.Utility.Utility.FormatSizeString(newvol.Filesize));
-
+                                                    FinishVolumeAndUpload(db, backend, newvol, newvolindex, uploadedVolumes);
 
                                                     newvol = new BlockVolumeWriter(m_options);
                                                     newvol.VolumeID = db.RegisterRemoteVolume(newvol.RemoteFilename, RemoteVolumeType.Blocks, RemoteVolumeState.Temporary, transaction);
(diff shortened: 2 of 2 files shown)


seantempleton:
This does cause the hash of the new volume file to be calculated twice. Once in the `CompactHandler` and once in the [BackendManager](https://github.com/duplicati/duplicati/blob/0f125a6113a50e9af801621dfa048e4538d5903f/Duplicati/Library/Main/BackendManager.cs#L1204) due to the Put code calculating hashes. I couldn't figure out a nice way of preventing that without some large refactoring.

warwickmm:
Looks like there are a few test failures in AppVeyor and Travis.  Are you able to reproduce these failures locally?

seantempleton:
I forgot to run the unit tests before creating the PR. Sorry about that.
Yes, I can reproduce the failure. When getting an encrypted file its expected file size is a little smaller than the actual size. I thought that should all be handled automatically though. I will investigate.

seantempleton:
It turns out that when `FinishVolume` is called on the `IndexVolumeWriter` it needs the final hash and size of the encrypted dblocks file. But `WriteBlocklist` has to be called after `FinishVolume` is called, and `FinishVolume` is called in the `BackendManager.Put` function. Added a callback function to `Put` to allow the blocklists to be written at the right time.

warwickmm:
I tried to follow the steps described in #4048 but I still don't see the `list` folder in the `dindex` file.  I also wrote a local test that involves merging `dblock` files after a compact operation and the `list` folder still appears to be missing.

It's possible that I'm testing this incorrectly.  Are you able to see the `list` folder after performing the steps in #4048?

seantempleton:
That is curious. I do indeed have the `list` folder after following the steps.
I thought about writing a unit test but I didn't think a test that merely tested for the existence of the `list` folder would be all that useful. As opposed to testing the contents of the files in the list folder but I figured that would be fairly difficult/messy to do.

warwickmm:
My mistake, my test files did not have the correct content to induce creation of a blocklist.  I see the `list` folder now.  :+1:

seantempleton:
Good to hear. Thanks for testing the code.

Duplicati/Library/Main/BackendManager.cs
```diff
@@ -1212,6 +1212,11 @@ public void Put(VolumeWriterBase item, IndexVolumeWriter indexfile = null, bool
                 req2 = new FileEntryItem(OperationType.Put, indexfile.RemoteFilename);
                 req2.LocalTempfile = indexfile.TempFile;
                 req.Indexfile = new Tuple<IndexVolumeWriter, FileEntryItem>(indexfile, req2);
+
+                indexfile.FinishVolume(req.Hash, req.Size);
+                indexVolumeFinishedCallback?.Invoke();
```
warwickmm: Is it true that if `indexfile` is not null, then we should _always_ want to write the blocklists as defined [here](https://github.com/duplicati/duplicati/commit/04662bace1393b428e39cdc89a276ca38b44b618#diff-59539f1832ed42c3579b81f766e18bc0R301)?  I'm wondering if the method signature (like before) still invites the possibility to provide a non-null `indexfile` without the callback to write the blocklists.

It would be nice if we could inline the body of [indexVolumeFinished](https://github.com/duplicati/duplicati/commit/04662bace1393b428e39cdc89a276ca38b44b618#diff-59539f1832ed42c3579b81f766e18bc0R301) here, but it seems that we may be missing an instance to a `LocalDatabase`.

seantempleton: If I am understanding your question correctly, I think in the case of the `CompactHandler` that yes we do always want to write the blocklists since blocks have just been moved to a new volume. If there are no blocklists then the foreach that calls `db.GetBlocklists` will just return an empty list and nothing will happen.
Yes a new call could be made to `Put` and they don't provide the callback, but they would have to be in the context of needing to do something so I would hope testing would show that they didn't provide the callback if needed.

For everyone else that calls `Put` they could use the callback function to do something else, they don't have to write blocklists, or they can ignore the callback completely by not passing one in.
This `Put` function is used by the `CompactHandler`, `PurgeFilesHandler`, and the `RepairHandler`. The `RepairHandler` does create an [index file](https://github.com/duplicati/duplicati/blob/837a787cbf233c07f739e31de3025ebbf94f1229/Duplicati/Library/Main/Operation/RepairHandler.cs#L312) and calls `Put` but it uses the `VolumeWriterBase item` parameter to upload the file and does not use the `indexfile` parameter.

I agree, since the necessary code is really just a foreach with a function call it would be nice to inline here. (Really it would be nice to have it centralized in one place and have it done before calling `Put`, but that is obviously way more involved.)

I'm not sure I actually answered your question, and then I rambled on.

warwickmm: That makes sense.  This might be a case where having different overloads as opposed to one method with optional parameters could help clarify the code.  But we can always refactor this later as needed.

PR #4058, opened 2020-01-20
Build true multi-arch Docker images and add arm64

dnrce:
Docker has native support for multi-arch images and it should no longer be necessary for users to use architecture-specific tags to select images manually.

See https://docs.docker.com/buildx/working-with-buildx/#build-multi-platform-images -- though please note that at this time experimental features must be enabled on the build host in order to use `buildx`.

This PR also adds builds for the `arm64v8` architecture now that 64-bit Linux distributions for the Raspberry Pi are becoming more common.

Supersedes #4057.

### DIFF
--- a/Installer/Docker/README.md
+++ b/Installer/Docker/README.md
@@ -17,6 +17,7 @@ Images for the following OS/architecture combinations are available using Docker
 
   * `linux/amd64`
   * `linux/arm/v7` - 32-bit ARMv7 devices like the Raspberry Pi 2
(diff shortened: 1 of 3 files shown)


kenkendk:
Looks good, great contribution!

ts678:
[https://hub.docker.com/r/duplicati/duplicati](https://hub.docker.com/r/duplicati/duplicati) did not receive [Release: 2.0.5.104 (canary) 2020-03-25](https://forum.duplicati.com/t/release-2-0-5-104-canary-2020-03-25/9468)
Wondering whether this is related. Last Canary was 2.0.5.103_canary_2020-02-18 if I read that right.

PR #4061, opened 2020-01-24
Build RecoveryTool index in memory by default

warwickmm:
This modifies the `RecoveryTool` so that the index is created in memory by default.  This should yield a dramatic increase in performance.  For low-resource situations, we allow one to use the original file-based index creation via the new `--build-index-with-files` option.  We also modified the `RecoveryTool` test to verify the behavior for both cases.

The work to build the index in memory was done by @tfriedel (see revision 8848ad4a74ab7dde300a8363d646a9af5362962b and pull request #4040).

### DIFF
--- a/Duplicati/CommandLine/RecoveryTool/FileIndex.cs
+++ b/Duplicati/CommandLine/RecoveryTool/FileIndex.cs
@@ -1,0 +1,246 @@
+﻿//  Copyright (C) 2015, The Duplicati Team
+//  http://www.duplicati.com, info@duplicati.com
+//
+//  This library is free software; you can redistribute it and/or modify
+//  it under the terms of the GNU Lesser General Public License as
+//  published by the Free Software Foundation; either version 2.1 of the
+//  License, or (at your option) any later version.
+//
+//  This library is distributed in the hope that it will be useful, but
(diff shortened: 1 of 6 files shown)


Duplicati/CommandLine/RecoveryTool/FileIndex.cs
```diff
@@ -0,0 +1,246 @@
+﻿//  Copyright (C) 2015, The Duplicati Team
```
warwickmm: This is a rename of the original file-based `Index` class.  The only modification was to move the call to `SortFile` [from the](https://github.com/duplicati/duplicati/pull/4061/files#diff-b389042e9144a8f9c477f231f51c2427L61) `Restore` class [to the end](https://github.com/duplicati/duplicati/pull/4061/files#diff-c02ab5ee0afabad37142d2d16e7851bfR135) of the `FileIndex.Run` method.

warwickmm:
The `Index` and `FileIndex` classes contain a bit of duplication, but I figure we can refactor this later.

kenkendk:
Great work!

kenkendk:
Merged items from @tfriedel and @warwickmm such that the recovery tool runs much faster by default, but can switch back to the low-memory mode if required.

PR #4080, opened 2020-02-01
Change config hash calculation (USN improvement)

drwtsn32x:
Duplicati calculates a configuration hash based on source folder list and exclude/include filters, the purpose of which is to detect changes that could cause problems with USN journal usage.  If a config change is detected, it forces a full volume scan.

The proposed change adds File Attribute and Max File Size filtering into this config hash calculation.

Since I needed access to m_options, the most straight forward approach I could think of was to move the hash calculation to BackupHandler.cs.

### DIFF
--- a/Duplicati/Library/Main/Operation/BackupHandler.cs
+++ b/Duplicati/Library/Main/Operation/BackupHandler.cs
@@ -104,17 +104,9 @@ namespace Duplicati.Library.Main.Operation
         {
             if (m_options.UsnStrategy == Options.OptimizationStrategy.Off) return null;
 
-            // get hash identifying current source filter / sources configuration
-            // ReSharper disable once PossibleMultipleEnumeration
-            var configHash = Library.Utility.Utility.ByteArrayAsHexString(MD5HashHelper.GetHash(
-                (filter == null ? string.Empty : filter.ToString()) +
-                string.Join("; ", sources) +
-                m_options.FileAttributeFilter.ToString() +
-                m_options.SkipFilesLargerThan.ToString()
-            ));
-
             var journalData = m_database.GetChangeJournalData(lastfilesetid);
-            var service = new UsnJournalService(sources, snapshot, configHash, journalData, cancellationTokenSource.Token);
+            var service = new UsnJournalService(sources, snapshot, filter, m_options.FileAttributeFilter, m_options.SkipFilesLargerThan,
+                journalData, cancellationTokenSource.Token);
 
             foreach (var volumeData in service.VolumeDataList)
             {
--- a/Duplicati/Library/Main/Operation/BackupHandler.cs
+++ b/Duplicati/Library/Main/Operation/BackupHandler.cs
@@ -103,8 +103,18 @@ namespace Duplicati.Library.Main.Operation
         private UsnJournalService GetJournalService(IEnumerable<string> sources, ISnapshotService snapshot, IFilter filter, long lastfilesetid)
         {
             if (m_options.UsnStrategy == Options.OptimizationStrategy.Off) return null;
+
+            // get hash identifying current source filter / sources configuration
+            // ReSharper disable once PossibleMultipleEnumeration
+            var configHash = Library.Utility.Utility.ByteArrayAsHexString(MD5HashHelper.GetHash(
+                (filter == null ? string.Empty : filter.ToString()) +
+                string.Join("; ", sources) +
+                m_options.FileAttributeFilter.ToString() +
+                m_options.SkipFilesLargerThan.ToString()
+            ));
+
             var journalData = m_database.GetChangeJournalData(lastfilesetid);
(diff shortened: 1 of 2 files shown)


Duplicati/Library/Main/Operation/BackupHandler.cs
```diff
@@ -103,8 +103,18 @@ public static Snapshots.ISnapshotService GetSnapshot(string[] sources, Options o
...
+            var configHash = Library.Utility.Utility.ByteArrayAsHexString(MD5HashHelper.GetHash(
+                (filter == null ? string.Empty : filter.ToString()) +
+                string.Join("; ", sources) +
+                m_options.FileAttributeFilter.ToString() +
+                m_options.SkipFilesLargerThan.ToString()
+            ));
+
             var journalData = m_database.GetChangeJournalData(lastfilesetid);
-            var service = new UsnJournalService(sources, snapshot, filter, journalData, cancellationTokenSource.Token);
+            var service = new UsnJournalService(sources, snapshot, filter, configHash, journalData, cancellationTokenSource.Token);
```
warwickmm: Is there any downside to passing in `m_options` instead?  This would keep the `UsnJournalService` responsible for the hash calculation and would make it a bit easier to create instances of `UsnJournalService` elsewhere in the future if needed.

drwtsn32x: That sounds like a good idea .. I'll adjust. Thanks!

drwtsn32x: @warwickmm - actually, I'm not sure how to accomplish this. I cannot reference the Options class directly in UsnJournalService.cs without adding a reference to Duplicati.Library.Main to Duplicati.Library.Snapshots.  But this causes a circular reference so is not allowed. I'm not super fluent in C# so maybe there's a simple solution I'm overlooking.

warwickmm: I see, that's unfortunate.  Perhaps that's why many classes take a `Dictionary<string, string>` of options, which I would like to eliminate as well.
  
What are your thoughts about simply adding a `FileAttributesFilter fileAttributesFilter` and a `long maxFileSizeFilter` argument to the `UsnJournalService` constructor?  The way you've written this is fine, but it just feels a bit nicer to keep the hash calculation inside the `UsnJournalService` class.

It might make sense to later introduce an object that encapsulates all the possible exclusions.  This would make it easier to manage in the future, especially if we start to introduce usage of Linux `fanotify` filesystem events.

drwtsn32x: > What are your thoughts about simply adding a FileAttributesFilter fileAttributesFilter and a long maxFileSizeFilter argument to the UsnJournalService constructor?

Yep, I can do that.  It's how I was originally going to implement this, but then thought it'd be cleaner this way.  But I agree with your thoughts on keeping the hash calculation in UsnJournalService.  I'll make the change soon.

Duplicati/Library/Snapshots/UsnJournalService.cs
```diff
@@ -60,18 +63,26 @@ public class UsnJournalService
...
 
-            // get filter identifying current source filter / sources configuration
+            // get hash identifying current source filter / sources configuration
             // ReSharper disable once PossibleMultipleEnumeration
-            var configHash = (emitFilter == null ? string.Empty : emitFilter.GetFilterHash()) + Utility.Utility.ByteArrayAsHexString(MD5HashHelper.GetHash(m_sources));
+            var configHash = Library.Utility.Utility.ByteArrayAsHexString(MD5HashHelper.GetHash(
+                (emitFilter == null ? string.Empty : emitFilter.ToString()) +
+                string.Join("; ", m_sources) +
+                fileAttributeFilter.ToString() +
+                skipFilesLargerThan.ToString()
```
warwickmm: Instead of concatenating everything into a single string, can we create a collection of strings
```
List<string> sourcesAndFilters = new List<string>(this.m_sources)
{   
    emitFilter is null ? String.Empty : emitFilter.ToString(),
    fileAttributeFilter.ToString(),
    skipFilesLargerThan.ToString()
};
```
and provide that to `MD5HashHelper.GetHash`?

With this, I think the comment disabling the ReSharper warning is also no longer needed.

drwtsn32x: From what I can tell, GetHash only takes a single string, not an array of strings. Do you want me to adjust it to accept an array of strings?

warwickmm: Did the previous implementation call [this overload](https://github.com/duplicati/duplicati/blob/v2.0.5.101-2.0.5.101_canary_2020-01-23/Duplicati/Library/Utility/MD5HashHelper.cs#L49), which accepts an `IEnumerable<string>`?

drwtsn32x: Sorry, I just saw the overload! Will submit an update.

PR #4085, opened 2020-02-04
Improve USN and Snapshot error handling

drwtsn32x:
If snapshot fails and policy is set to Required, show "Failed to create a snapshot" message instead of cryptic exception message like "access denied."

If snapshot policy set to Auto, record the failure in the Information log.

If USN policy set to On, write to Warning log instead of Error log.  (This is more consistent with how snapshot policy is handled.)

### DIFF
--- a/Duplicati/Library/Main/Operation/BackupHandler.cs
+++ b/Duplicati/Library/Main/Operation/BackupHandler.cs
@@ -126,7 +126,7 @@ namespace Duplicati.Library.Main.Operation
                         }
                         else if (m_options.UsnStrategy == Options.OptimizationStrategy.On)
                         {
-                            Logging.Log.WriteErrorMessage(LOGTAG, "FailedToUseChangeJournal", volumeData.Exception,
+                            Logging.Log.WriteWarningMessage(LOGTAG, "FailedToUseChangeJournal", volumeData.Exception,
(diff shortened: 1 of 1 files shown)


Duplicati/Library/Main/Operation/BackupHandler.cs
```diff
@@ -81,9 +81,11 @@ public static Snapshots.ISnapshotService GetSnapshot(string[] sources, Options o
             catch (Exception ex)
             {
                 if (options.SnapShotStrategy == Options.OptimizationStrategy.Required)
-                    throw;
+                    throw new Exception("Failed to create a snapshot");
```
warwickmm: We should try to [avoid throwing](https://docs.microsoft.com/en-us/dotnet/standard/design-guidelines/using-standard-exception-types#exception-and-systemexception) the base `System.Exception` type.  Doing so can make the exception more difficult to handle downstream.  If none of the standard exception types are applicable, we can add a custom type in the `CustomExceptions` class.  We can then throw the custom exception and provide `ex` as the [inner exception](https://docs.microsoft.com/en-us/dotnet/api/system.exception.-ctor?view=netframework-4.8#System_Exception__ctor_System_String_System_Exception_).

drwtsn32x: Ok, updated to use `UserInformationException` class.  Also changed it to use the proper string from `Strings.cs`.

PR #4088, opened 2020-02-06
RPM package - change minimum required mono-runtime version

drwtsn32x:
Duplicati now requires mono >= 5.0 to operate correctly. Changing the dependency in the rpm package will flag old mono version issues prior to Duplicati install. Hopefully this will reduce the reports of strange errors that turn out to be caused by old mono versions.

Can someone test this please?  I don't know how to actually build an rpm package, so I haven't tested it myself.  But I believe it should work.


warwickmm:
Looks good to me, but I don't have a way to test.  Perhaps the best way to test this is to create a canary release.

PR #4105, opened 2020-02-17
Fix path to SQLite.Interop.dll

### LINKED ISSUE #4093: SQLiteLoader uses wrong assembly path?
Duplicati\Library\SQLiteHelper\SQLiteLoader.cs : 192
`PInvoke.LoadLibraryEx(Path.Combine(basePath, "SQLite.Interop.dll"), IntPtr.Zero, 0);`

When "SQLite.Interop.dll" is loaded via PInvoke, it looks in basePath, but I think it should look in assemblyPath. On the file system, the .dll is not located in basePath, but in assemblyPath.

warwickmm:
This assembly lives in the `SQLite/win32` and `SQLite/win64` subdirectories, and not in the base path.

This fixes #4093.


warwickmm:
I don't have a good way to test this, but looking at the code and the contents of the release I believe this is correct.

kenkendk:
Yes, this is correct. The interop library lives in the subfolder with the real library.

Issue #4090, opened 2020-02-08
Filter "Exclude directories whose name contains" doesn't exclude the given directory


JackHewson:
- [ x] I have searched open and closed issues for duplicates.

----------------------------------------

## Environment info

- **Duplicati version**: 2.0.5.101_canary_2020-01-23
- **Operating system**: Windows 10
- **Backend**: Local folder or drive

## Description

Filter "Exclude directories whose name contains" doesn't exclude the given directory.

## Steps to reproduce

1. Set source directory 
`C:\temp\DuplicatiTestSource\`

2. Insert entry for filter "Exclude directories whose name contains"
`C:\temp\DuplicatiTestSource\Folder2\`

3. Run the backup

- **Actual result**:
The given directory isn't exluded (see debug log)

- **Expected result**:
The given directory is excluded.

## Debug log

Filter setting "Edit as list" shows: 
`-*C:\temp\DuplicatiTestSource\Folder2\*\\`

Logfile entry:
```
[Verbose-Duplicati.Library.Main.Operation.Backup.FileEnumerationProcess-IncludingPath]: Including path as no filters matched: D:\temp\DuplicatiTestSource\Folder1\
[Verbose-Duplicati.Library.Main.Operation.Backup.FileEnumerationProcess-IncludingPath]: Including path as no filters matched: D:\temp\DuplicatiTestSource\Folder2\
[Verbose-Duplicati.Library.Main.Operation.Backup.FileEnumerationProcess-IncludingPath]: Including path as no filters matched: D:\temp\DuplicatiTestSource\Folder3\
```

drwtsn32x:
What happens if you just choose "Exclude Folder" and then specify the full folder name without final backslash?

JackHewson:
I changed the filter to "Exclude Folder" and the entry to `C:\temp\DuplicatiTestSource\Folder` .
(This setting results in the "edit as list" view as `-C:\temp\DuplicatiTestSource\Folder2\\` )

When running this backup, the given directory isn't exluded and is part of the backup.
Please see the debug log below:

```
[Verbose-Duplicati.Library.Main.Operation.Backup.FileEnumerationProcess-IncludingPath]: Including path as no filters matched: C:\temp\DuplicatiTestSource\Folder1\
[Verbose-Duplicati.Library.Main.Operation.Backup.FileEnumerationProcess-IncludingPath]: Including path as no filters matched: C:\temp\DuplicatiTestSource\Folder2\
[...]

PR #4112, opened 2020-02-23
Fix issue 4090 - Filter builder issues in web UI

drwtsn32x:
PR https://github.com/duplicati/duplicati/pull/3967 introduced a bug on the Windows platform where backslashes would sometimes be incorrectly added to the end of some filters, causing them to not function properly.

Example - add these four filters:
Capture
Note that after you finish typing C:\Test.txt in that fourth filter and you click on something else, the dropdown will auto change from "Exclude File" to "Exclude Folder".

Click the triple dot menu and then "Edit as text":
Capture2

Line 1 - has an extra backslash
Line 2 - this is ok, backslash needs to be escaped inside regexp
Line 3 - has an extra backslash
Line 4 - has two backslashes, should have none since we picked Exclude File (but as noted above, it automatically changed)

The cause is related to `dirsep` being escaped in all cases when on Windows:

https://github.com/duplicati/duplicati/blob/1f1826bf7a8b2a97956d001677ce0ecf917df047/Duplicati/Server/webroot/ngax/scripts/services/AppUtils.js#L548-L550

https://github.com/duplicati/duplicati/blob/1f1826bf7a8b2a97956d001677ce0ecf917df047/Duplicati/Server/webroot/ngax/scripts/services/AppUtils.js#L590-L592

Also, use of the escaped `dirsep` here is contributing to the weird behavior:

https://github.com/duplicati/duplicati/blob/1f1826bf7a8b2a97956d001677ce0ecf917df047/Duplicati/Server/webroot/ngax/scripts/services/AppUtils.js#L568

I believe the correct solution is to only escape the directory separator when it is used inside a regular expression.  That fixes the issue that the original PR https://github.com/duplicati/duplicati/pull/3967  was trying to fix in Issue https://github.com/duplicati/duplicati/issues/2220 without any of the side effects.

I tested on both Windows and Linux and all filter types seem to be working correctly now.

After fix repeat the above test:
Capture3
Capture4

Can someone else test? Since it's just a change to a javascript text file it's easy to test. You may have to do a hard refresh in your browser to get it to read the updated js.

Fixes issue https://github.com/duplicati/duplicati/issues/4090

### DIFF
--- a/Duplicati/Server/webroot/ngax/scripts/services/AppUtils.js
+++ b/Duplicati/Server/webroot/ngax/scripts/services/AppUtils.js
@@ -131,7 +131,7 @@ backupApp.service('AppUtils', function($rootScope, $timeout, $cookies, DialogSer
             name: gettextCatalog.getString('Exclude files whose names contain'),
             key: '-file*',
             prefix: '-[.*',
-            suffix: '[^!]*]',
+            suffix: '[^\\!]*]',    // Escape dirsep inside regexp. Required on Windows, no effect on other platforms.
             rx: '\\-\\[\\.\\*[\\^\\!\\]\\*\\]'
         }, {
             name: gettextCatalog.getString('Exclude folder'),
@@ -545,10 +545,6 @@ backupApp.service('AppUtils', function($rootScope, $timeout, $cookies, DialogSer
         }
 
         function matches(txt, n) {
-            // We need to escape dirsep if it is the backslash character
-            if (dirsep == '\\')
-                dirsep = '\\\\';
-
             var pre = apputils.replace_all(n.prefix || '', '!', dirsep);
             var suf = apputils.replace_all(n.suffix || '', '!', dirsep);
 
@@ -586,10 +582,6 @@ backupApp.service('AppUtils', function($rootScope, $timeout, $cookies, DialogSer
             return body;
 
         body = body || '';
-
-        // We need to escape dirsep if it is the backslash character
-        if (dirsep == '\\')
-            dirsep = '\\\\';
 
         var f = this.filterTypeMap[type];
         var pre = this.replace_all(f.prefix || '', '!', dirsep);


warwickmm:
On linux, when I enter `Test` for "Exclude files whose name contains", the resulting regex is `[.*Test[^\/]*]`.  On the `master` branch, this yields `[.*Test[^/]*]`.  Is that expected?

Also, on linux when enter `Test` for "Exclude folder" and then click on something else, the dropdown changes to "Exclude file whose name contains."  This happens on both `master` and this branch, and is slightly different than what you described.  Can you reproduce this on linux?

drwtsn32x:
@warwickmm - yes, `[.*Test[^\/]*]` on Linux is expected with this change. Escaping characters that don't normally need escaping has no net effect - it functions as desired.

And yes I can see the behavior you describe when I enter "test" on Linux - filter type changes.  If you change it back, it will stick from what I can tell.  I am not sure how the code works that causes the filter types to change automatically like that.

warwickmm:
> I am not sure how the code works that causes the filter types to change automatically like that.

Yes, this is strange but it sounds like a separate issue.

The changes you illustrate make sense to me.  Let's wait to see if someone else with Windows can confirm.

drwtsn32x:
Boy, that is really annoying how the filter dropdown will change automatically.  I am trying to add an "Exclude regular expression" set to `.*\\~[^\\]*` and it keeps changing to something else!  Both master and this branch are doing it, so at least it didn't get worse with this PR fix.

kenkendk:
> Boy, that is really annoying how the filter dropdown will change automatically. I am trying to add an "Exclude regular expression" set to .*\\~[^\\]* and it keeps changing to something else! Both master and this branch are doing it, so at least it didn't get worse with this PR fix.

I think the only real way to fix this is to store the item the user has picked in the dropdown. Right now, it attempts to "unparse" the filter to figure out what element to display in the dropdown.

The setup is here:
https://github.com/duplicati/duplicati/blob/master/Duplicati/Server/webroot/ngax/scripts/services/AppUtils.js#L124

And it is used here: 
https://github.com/duplicati/duplicati/blob/master/Duplicati/Server/webroot/ngax/scripts/services/AppUtils.js#L539

The idea behind the current setup was that we could reliably guess the simple forms, based on prefix+suffix. Anything that is not a simple form would then be a custom expression.

PR #4128, opened 2020-03-05
Throw TimeoutException upon HTTP timeout for GET operations

warwickmm:
The `HttpClient.GetAsync` method throws an `OperationCanceledException` when the timeout is exceeded.  In order to provide a more informative exception, we will detect this case and throw a `TimeoutException` instead.  Since there is no `CancellationToken`, we will assume that the `OperationCanceledException` is due to a timeout.

This is related to issue #3772.  Pull request #3778 handled a similar issue, but only for PUT operations.

See the following for more information:

https://stackoverflow.com/questions/10547895/how-can-i-tell-when-httpclient-has-timed-out/32230327#32230327
https://github.com/dotnet/runtime/issues/21965

It appears that .NET Core will be handling this a bit better:

https://github.com/dotnet/runtime/pull/2281

### DIFF
--- a/Duplicati/Library/Backend/OAuthHelper/OAuthHttpClient.cs
+++ b/Duplicati/Library/Backend/OAuthHelper/OAuthHttpClient.cs
@@ -65,6 +65,26 @@ namespace Duplicati.Library
         }
 
         /// <summary>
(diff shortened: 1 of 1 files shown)


warwickmm:
Note that while the error popup and job summary might still present a generic "a task was cancelled" or "failed to process file" message, the logs will at least reveal the HTTP timeout.

Issue #4141, opened 2020-03-16
All files modified


Squarebro:
- [ ] I have searched open and closed issues for duplicates.

----------------------------------------

## Environment info

- **Duplicati version**: 2.0.5.103
- **Operating system**: Windows

## Description

I have installed new version of Duplicati (2.0.5.103) from (2.0.5.1) and when i run all my Tasks they all got new backup sets even tho most of them didnt had any files added/deleted/modified. When i checked what was going on i saw this:
![changeno](https://user-images.githubusercontent.com/62223861/76713839-cda6d300-6723-11ea-965a-cbc905230df2.png)

So all files got modified but (0 bytes) added. Also there were new dblocks and one new dlist (i thought dblocks were only created if there are some actuall file changes?)

## Steps to reproduce

- Install 2.0.5.103 after using earlier version
- Run any backup

- **Actual result**:
All files get modified
- **Expected result**:
No files modified, no new backup set

drwtsn32x:
It could be due to improved metadata checks.  There was a change recently (I forget if it made it into 2.0.5.1 or if it was later) where NTFS permission breaks were not backed up properly.  Adding this feature did cause all files to be reprocessed.

How did the next backup perform?  It hopefully only "opened" files that were truly changed since the previous backup.

Squarebro:
@drwtsn32x I see two changes between this two versions (both by you - thanks for all the work)

- Fixed issue with USN returning too many results
- Improved USN changes by looking at size and attributes

So this could be the case, but, I dont use usn-policy and i dont even run Duplicati as admin. Would it be still possible for this to be the reason?

drwtsn32x:
No, if you don't use USN policy then these changes wouldn't have affected you. Plus at most those changes may just trigger a full filesystem scan, not a reprocessing of all file contents.

When you ran your next backup did it still reprocess all file contents?

Squarebro:
Next backup didn't modify any files and didnt create new backup set, so no changes were made if this is what you mean by reprocess.

I guess its metadata as you said, but i dont know why would it be triggered if nothing changed and i dont see any other changes between two versions that could trigger it.

Is there some way you know that i could check what really changed?
I could run some querry in SQLli or somehow check in backup files?

Thanks

drwtsn32x:
I did a quick test and confirmed what you saw. Did a backup of some files with 2.0.5.1, upgraded to 2.0.5.103, and ran another backup.  A new backup version was created.

[...]

PR #4142, opened 2020-03-17
Only store inheritance breaks in metadata when True

drwtsn32x:
See issue https://github.com/duplicati/duplicati/issues/4141

Previous pull request https://github.com/duplicati/duplicati/pull/4031 stores inheritance break information in file metadata, regardless of it being true or false.

This has a side effect of Duplicati detecting a metadata change for EVERY file the first time a backup is run with the updated version.  That first backup will take longer to complete.

I recommend the following refinement so that inheritance break information is only stored when it's actually True.  Duplicati will only detect metadata change (and therefore reprocess file contents) for files where inheritance is actually broken.  (Again, only applies to first backup after upgrading.)

There is no change in net functionality of PR 4031 because if the inheritance break information is not present in the metadata, it is assumed to be False.

It'd be best if this were merged before the next beta release so the larger Beta audience is never affected.  (Beta 2.0.5.1 does not include PR 4031.)

Canary users running 2.0.5.101 or later will experience another "longer than normal" backup as Duplicati will again see metadata changes when doing the first backup after upgrading.

### DIFF
--- a/Duplicati/Library/Common/IO/SystemIOWindows.cs
+++ b/Duplicati/Library/Common/IO/SystemIOWindows.cs
@@ -490,7 +490,7 @@ namespace Duplicati.Library.Common.IO
 
             // Only include the following key when its value is True.
             // This prevents unnecessary 'metadata change' detections when upgrading from
-            //   older versions (pre-2.0.5.101) that didn't store this value at all.
+            // older versions (pre-2.0.5.101) that didn't store this value at all.
             // When key is not present, its value is presumed False by the restore code.
             if (rules.AreAccessRulesProtected)
             {
--- a/Duplicati/Library/Common/IO/SystemIOWindows.cs
+++ b/Duplicati/Library/Common/IO/SystemIOWindows.cs
@@ -487,7 +487,11 @@ namespace Duplicati.Library.Common.IO
                 objs.Add(new FileSystemAccess((FileSystemAccessRule)f));
 
             dict["win-ext:accessrules"] = SerializeObject(objs);
-            dict["win-ext:accessrulesprotected"] = rules.AreAccessRulesProtected.ToString();
+
+            if (rules.AreAccessRulesProtected)
+            {
+                dict["win-ext:accessrulesprotected"] = rules.AreAccessRulesProtected.ToString();
+            }
 
             return dict;
         }
--- a/Duplicati/Library/Common/IO/SystemIOWindows.cs
+++ b/Duplicati/Library/Common/IO/SystemIOWindows.cs
@@ -488,9 +488,13 @@ namespace Duplicati.Library.Common.IO
 
             dict["win-ext:accessrules"] = SerializeObject(objs);
 
+            // Only include the following key when its value is True.
+            // This prevents unnecessary 'metadata change' detections when upgrading from
+            //   older versions (pre-2.0.5.101) that didn't store this value at all.
+            // When key is not present, its value is presumed False by the restore code.
             if (rules.AreAccessRulesProtected)
             {
-                dict["win-ext:accessrulesprotected"] = rules.AreAccessRulesProtected.ToString();
+                dict["win-ext:accessrulesprotected"] = "True";
             }
 
             return dict;


Duplicati/Library/Common/IO/SystemIOWindows.cs
```diff
@@ -487,7 +487,11 @@ IFileEntry FileEntryAlphaFS(string p)
                 objs.Add(new FileSystemAccess((FileSystemAccessRule)f));
 
             dict["win-ext:accessrules"] = SerializeObject(objs);
-            dict["win-ext:accessrulesprotected"] = rules.AreAccessRulesProtected.ToString();
+
+            if (rules.AreAccessRulesProtected)
+            {
+                dict["win-ext:accessrulesprotected"] = rules.AreAccessRulesProtected.ToString();
+            }
```
warwickmm: Can this be simplified to 
```
if (rules.AreAccessRulesProtected)
{
    dict["win-ext:accessrulesprotected"] = "True";
}
```
or alternatively without the `if`
```
dict["win-ext:accessrulesprotected"] = rules.AreAccessRulesProtected.ToString();
```

drwtsn32x: Your first suggestion is fine with me if you like that better.

Second one is how code is right now... we want to avoid listing this property in metadata when the value is False

warwickmm: > Second one is how code is right now... we want to avoid listing this property in metadata when the value is False

Oops, not sure what I was thinking here. 

> Second one is how code is right now... we want to avoid listing this property in metadata when the value is False

I think I understand the issue now.  Is this simply a workaround that can be removed once the next beta is released?  If so, can we add a comment in the code explaining why we're doing this, and also create a ticket to clean this up after the next beta is released?

drwtsn32x: Sorry, I'm probably not explaining the issue very well.

This change should stay indefinitely.  Versions prior to 2.0.5.101 did not store `win-ext:accessrulesprotected` at all - it was added very recently.  But when I did add support for it, I had Duplicati store it for every single file with a value of True or False.  This had a negative side effect of making Duplicati think the metadata changed for every single file (the first time you ran a backup after upgrading), and caused it to reprocess all file contents.  An unnecessary - and potentially lengthy - operation.

If we change it so that `win-ext:accessrulesprotected` is only stored when True, then metadata will appear the same as how Duplicati 2.0.5.1 (and older) calculated it for files where inheritance is NOT broken (which is the vast majority of files). 

Not sure if that explanation is any better...

warwickmm: I see.  This is still probably worth explaining in the code with a comment, since at a quick glance it's not clear why "False" is different than the option being unset.

drwtsn32x: Sure, I'll put in a comment and also implement your other suggestion....thanks

Duplicati/Library/Common/IO/SystemIOWindows.cs
```diff
@@ -487,7 +487,15 @@ IFileEntry FileEntryAlphaFS(string p)
                 objs.Add(new FileSystemAccess((FileSystemAccessRule)f));
 
             dict["win-ext:accessrules"] = SerializeObject(objs);
-            dict["win-ext:accessrulesprotected"] = rules.AreAccessRulesProtected.ToString();
+
+            // Only include the following key when its value is True.
+            // This prevents unnecessary 'metadata change' detections when upgrading from
+            //   older versions (pre-2.0.5.101) that didn't store this value at all.
```
warwickmm: Sorry for being overly fastidious, but can we remove these extra spaces before the word `older`?

drwtsn32x: sure thing :)

warwickmm: Thanks!

Issue #4108, opened 2020-02-18
Suspected memory leak


fred-cardoso:
- [x] I have searched open and closed issues for duplicates.

----------------------------------------

## Environment info
- **Duplicati version**: 2.0.5.101
- **Operating system**: Linux
- **Backend**: OneDrive v2
## Description
Hello guys,

I wanted to message someone before I opened this ticket so I could point more directly to the supposed bug. So the issue is the following. I was doing a backup whenit failed with this error:
`2020-02-18 12:51:18 +00 - [Error-Duplicati.Library.Main.Operation.TestHandler-FailedToProcessFile]: Failed to process file duplicati-....`

A more detailed analysis returned the following information:
`"Key": "Error", "Value": "Insufficient memory to continue the execution of the program."`

At this point I went on I checked the RAM usage:
![image](https://user-images.githubusercontent.com/32043667/74738153-38acd900-524e-11ea-8fd8-bf5419379206.png)

At the end of the graph, you can see that there are "two steps". This steps were two tries of backing up some data, where you can see that the memory keeps going up and so forth.

This is not the first time I see this error, actually it's the 3rd. I know that is related to duplicati because it only happens when I run duplicati, where the memory usage keeps increasing. Also, after restarting duplicati docker instance, the free memory increases in more than 1GB.
![image](https://user-images.githubusercontent.com/32043667/74738515-fc2dad00-524e-11ea-93a3-10a71e5d27d3.png)

How can I try to debug this even further to find the culprit?

## Debug log
`{
  "DeletedFiles": 0,
  "DeletedFolders": 0,
  "ModifiedFiles": 0,
  "ExaminedFiles": 640,
  "OpenedFiles": 0,
  "AddedFiles": 0,
  "SizeOfModifiedFiles": 0,
  "SizeOfAddedFiles": 0,
  "SizeOfExaminedFiles": 665451250,
  "SizeOfOpenedFiles": 0,
  "NotProcessedFiles": 0,
  "AddedFolders": 0,
  "TooLargeFiles": 0,
  "FilesWithError": 0,
  "ModifiedFolders": 0,
  "ModifiedSymlinks": 0,
  "AddedSymlinks": 0,
  "DeletedSymlinks": 0,
  "PartialBackup": false,
  "Dryrun": false,
  "MainOperation": "Backup",
  "CompactResults": null,
  "VacuumResults": null,
  "DeleteResults": {
    "DeletedSetsActualLength": 0,
    "DeletedSets": [],
[...]

PR #4145, opened 2020-03-23
Allow MS Graph backends (e.g., OneDrive v2) to use OAuthHelper instead of OAuthHttpClient

tygill:
When I originally wrote the Microsoft Graph backends (OneDrive v2, SharePoint v2, etc.), I also wrote a new OAuthHttpClient class based on the standard HttpClient class, rather than using the existing OAuthHelper class. I did this partially out of having more familiarity with the HttpClient based code, but also because that seemed to be the newer API, as opposed to the HttpWebRequest pattern most other backends used.

This change adds support to these backends for using the 'standard' OAuthHelper instead of the OAuthHttpClient by way of the new `--use-http-client=false` parameter. This logic lives in parallel to the original implementation that uses the OAuthHttpClient class,

This is intended primarily to see if this helps resolve the memory leaks seen in some versions of Mono (see #4108), since the main difference between these backends and others that don't seem to have any problems is the use of the HttpClient APIs. For this reason, the new behavior (using OAuthHelper) is the default on Mono, while the original behavior (using OAuthHttpClient) is the default otherwise.

I've tested both version of this using the OneDrive v2 backend on Windows, but don't have a good way to test on Mono. At any rate, the behavior for both should be the same (barring any bugs in either the backend or the framework). I'm not currently sure what performance differences there might be between the two either.

### DIFF
--- a/Duplicati/CommandLine/BackendTester/Program.cs
+++ b/Duplicati/CommandLine/BackendTester/Program.cs
@@ -394,6 +394,30 @@ namespace Duplicati.CommandLine.BackendTester
                             Console.WriteLine("*** Remote folder contains {0} after cleanup", fe.Name);
                         }
 
+                    // Test some error cases
+                    Console.WriteLine("Checking retrieval of non-existent file...");
+                    bool caughtExpectedException = false;
+                    try
+                    {
+                        using (Duplicati.Library.Utility.TempFile tempFile = new Duplicati.Library.Utility.TempFile())
+                        {
+                            backend.Get(string.Format("NonExistentFile-{0}", Guid.NewGuid()), tempFile.Name);
+                        }
+                    }
+                    catch (FileMissingException ex)
+                    {
+                        Console.WriteLine("Caught expected FileMissingException");
+                        caughtExpectedException = true;
+                    }
+                    catch (Exception ex)
+                    {
+                        Console.WriteLine("*** Retrieval of non-existent file failed: {0}", ex);
+                    }
+
+                    if (!caughtExpectedException)
+                    {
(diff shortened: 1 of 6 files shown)


warwickmm:
My OneDrive tests passed on Linux with mono 6.4.0.

ts678:
One aspect that possibly differs is timeouts. The current code, I believe, left [HttpClient.Timeout Property](https://docs.microsoft.com/en-us/dotnet/api/system.net.http.httpclient.timeout?view=netframework-4.8) at its 100 second default which is sometimes too short, so I have people raise --http-operation-timeout. Duplicati's default output is too short to diagnose the issue, so it's some manual work to guide the user.

The problem is that the best timeout varies with situation, but I was thinking of asking for longer, maybe 5 minutes. With --use-http-client=false, I guess OneDrive v2 will work like the others -- whatever that is.

This is just a side comment on the PR, given work is underway. Timeouts are a general area of concern...
The 100 seconds could be a separate issue if need be, but I thought I'd start on the timeouts topic  here.

Thanks!

tygill:
@ts678, timeouts are a valid concern, and I don't think I can take are a reason to not have a default of 5 minutes or something higher. Maybe I look at that as a separate PR.

warwickmm:
How can we test if this fixes the potential memory leak issues?  Are you able to replicate the situations you encountered in issue #4108?

tygill:
I have a fairly persistent repro with Duplicati running on my Raspberry Pi (enabling backup testing or compaction both fairly quickly cause out of memory exceptions for me). However, I'm using Docker there and don't currently have anything set up to test an unofficial/prerelease build. (For previous attempts to fix this, I've waited until a canary with my changes was ready and then pulled that.)

PR #4150, opened 2020-03-28
Sort file and folder listing in web UI

drwtsn32x:
When selecting source data (or selecting destination in case of local backups), the file/folder listing is not sorted and is presented in the raw order as returned by the API.

On my Debian Linux system, the ordering is completely random:
linux-before

On my Windows system, it does appear to be alphabetical but folders and files are mixed together:
windows-before

I suggest grouping and showing folders first before files. Both groups will be sorted alphabetically.
The following change achieves that goal but not sure if it's the most efficient. Thoughts?

### DIFF
--- a/Duplicati/Server/WebServer/RESTMethods/Backup.cs
+++ b/Duplicati/Server/WebServer/RESTMethods/Backup.cs
@@ -65,7 +65,9 @@ namespace Duplicati.Server.WebServer.RESTMethods
                 result[n.Name] = n.Value;
 
             result["Filesets"] = r.Filesets;
-            result["Files"] = r.Files;
+            result["Files"] = r.Files
+                .OrderByDescending(f => f.Path.EndsWith(Path.DirectorySeparatorChar.ToString(), StringComparison.Ordinal))
+                .ThenBy(f => f.Path);
 
             info.OutputOK(result);
 
--- a/Duplicati/Server/WebServer/RESTMethods/Backup.cs
+++ b/Duplicati/Server/WebServer/RESTMethods/Backup.cs
@@ -66,7 +66,9 @@ namespace Duplicati.Server.WebServer.RESTMethods
 
             result["Filesets"] = r.Filesets;
             result["Files"] = r.Files
-                .OrderByDescending(f => f.Path.EndsWith("/", StringComparison.Ordinal) || f.Path.EndsWith("\\", StringComparison.Ordinal))
+                // Group directories first - support either directory separator here as we may be restoring data from an alternate platform
+                .OrderByDescending(f => (f.Path.StartsWith("/", StringComparison.Ordinal) && f.Path.EndsWith("/", StringComparison.Ordinal)) || (!f.Path.StartsWith("/", StringComparison.Ordinal) && f.Path.EndsWith("\\", StringComparison.Ordinal)))
+                // Sort both groups (directories and files) alphabetically
                 .ThenBy(f => f.Path);
 
             info.OutputOK(result);
--- a/Duplicati/Server/WebServer/RESTMethods/Backup.cs
+++ b/Duplicati/Server/WebServer/RESTMethods/Backup.cs
@@ -66,7 +66,7 @@ namespace Duplicati.Server.WebServer.RESTMethods
 
             result["Filesets"] = r.Filesets;
             result["Files"] = r.Files
-                .OrderByDescending(f => f.Path.EndsWith(Path.DirectorySeparatorChar.ToString(), StringComparison.Ordinal))
+                .OrderByDescending(f => f.Path.EndsWith("/", StringComparison.Ordinal) || f.Path.EndsWith("\\", StringComparison.Ordinal))
                 .ThenBy(f => f.Path);
 
             info.OutputOK(result);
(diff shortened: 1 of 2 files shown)


Duplicati/Server/WebServer/RESTMethods/Filesystem.cs
```diff
@@ -208,7 +208,9 @@ private static IEnumerable<Serializable.TreeNode> ListFolderAsNodes(string entry
                 return false;
             };
 
-            foreach (var s in SystemIO.IO_OS.EnumerateFileSystemEntries(entrypath))
+            foreach (var s in SystemIO.IO_OS.EnumerateFileSystemEntries(entrypath)
+                .OrderByDescending(f => SystemIO.IO_OS.GetFileAttributes(f) & FileAttributes.Directory)
+                .ThenBy(f => SystemIO.IO_OS.PathGetFileName(f)))
```
warwickmm: Is `SystemIO.IO_OS.PathGetFileName` necessary, or does `.ThenBy(f => f)` achieve the same result?

drwtsn32x: I'll test it and let you know.  If it does still work I'll update the pull request.

drwtsn32x: Simplification works

Duplicati/Server/WebServer/RESTMethods/Backup.cs
```diff
@@ -65,7 +65,9 @@ private void SearchFiles(IBackup backup, string filterstring, RequestInfo info)
                 result[n.Name] = n.Value;
 
             result["Filesets"] = r.Filesets;
-            result["Files"] = r.Files;
+            result["Files"] = r.Files
+                .OrderByDescending(f => f.Path.EndsWith(Path.DirectorySeparatorChar.ToString(), StringComparison.Ordinal))
```
warwickmm: How does this affect cross-system restores, e.g., a Windows backup being restored to a Linux machine?

Also, there's a [cached version](https://github.com/duplicati/duplicati/blob/v2.0.5.104-2.0.5.104_canary_2020-03-25/Duplicati/Library/Common/IO/Util.cs#L27) of the directory separator to avoid repeated conversions from `char` to `string`.  I'm not sure how much it saves in terms of performance, but it is being used elsewhere.

drwtsn32x: I tested restoring Linux data on a Windows machine, and basically the first sort doesn't work because of the wrong slash.  The end result was the files and folders were sorted alphabetically but mixed.

One possible solution is to check if the string EndsWith either forward slash OR backslash, regardless of platform.  What do you think?

drwtsn32x: Went ahead and pushed a commit with that change.  Tested and it works even when restoring from an alternate platform.

warwickmm: There is the case where several Linux filesystems (e.g., ext4) allow `\` in filenames, so technically there could be a file whose name ends with `\`.  For example, using revision 7a4d90f4af329409e47718bdb25d21a24853693f, I have a file called `weird\` that is sorted as if it was a folder:

image

Somehow, the restore tree knows that this is a file.  Can you test backing up a file that ends with `\` on Linux, and see how the restore dialog appears in Windows?  Is it rendered as a file or as a folder?

drwtsn32x: Very interesting. I'll do some testing tomorrow.

drwtsn32x: Ok, backed up some test data on Linux that includes a filename with a trailing backslash.  Doesn't surprise me that on Linux this has the regular file icon, since a backslash is not the platform's path separator.

But I was surprised to see that on Windows this Linux backup data is displayed the same way.  The file is shown with a file icon.  It's also NOT grouped with the folders:
Capture

(Out of curiosity I tried restoring it on Windows. The restore was successful and the file was named afile.txt_, which I think is a good compromise.)

This file shows up with two backslashes when debugging, which makes sense to me.
Capture2

I would have expected this to cause the file to be grouped with folders though.

How about this idea - instead of going by the trailing slash/backslash, we go by the Size property. Looks like Duplicati sets this to `-1` for folders.  Do you think that'd be a more reliable method?

drwtsn32x: Sorry, I made a mistake in my first screen shot. That was on Windows but Duplicati didn't have this change applied.  When I have this pull request applied it definitely sorts `afile.txt\` up with the folder group.

I'd like to test going by size = -1 unless you think that's a bad idea.

warwickmm: The use of `-1` as the size is littered throughout the code, and its indication of a path being a folder seems more like a side effect rather than an enforced standard.  I would prefer not to code against this.

The UI code attempts to [guess](https://github.com/duplicati/duplicati/blob/v2.0.5.104-2.0.5.104_canary_2020-03-25/Duplicati/Server/webroot/ngax/scripts/controllers/RestoreController.js#L175) the path separator by inspecting the first character.  If it's `/`, then the files came from a non-Windows backup, and hence a path ending with `/` should indicate a folder.  If the first character is not `/`, then the files came from a Windows backup  and hence a path ending with `\` should indicate a folder.

Do you think we can apply the same logic here?

drwtsn32x: Perhaps... give me some time to mess around with this.  Thanks!

PR #4151, opened 2020-03-30
Remove unused BackendHandler members

warwickmm:
This removes unused members from the `BackendHandler` class.  Most of the members were unused, leaving only the [FileEntryItem](https://github.com/duplicati/duplicati/blob/v2.0.5.104-2.0.5.104_canary_2020-03-25/Duplicati/Library/Main/Operation/Common/BackendHandler.cs#L47) class, which looks quite similar to the [BackendManager.FileEntryItem](https://github.com/duplicati/duplicati/blob/v2.0.5.104-2.0.5.104_canary_2020-03-25/Duplicati/Library/Main/BackendManager.cs#L66) class.


warwickmm:
There is a comment below (from revision 537fe208385bc4b5bfd5cfef03f614491457d2c6 on Feb 25, 2016) that alludes to plans for the `BackendHandler` class that apparently never materialized.  @kenkendk, do you recall the reason for using the `BackendHandler` here?
https://github.com/duplicati/duplicati/blob/8bb7b27c1f3c3ac0947c279692b2aa08a2be22be/Duplicati/Library/Main/Operation/BackupHandler.cs#L475-L480

kenkendk:
The idea was to remove the `BackendManager.cs` file completely, and only use the `BackendHandler` class. This class was written to use `async` logic, but it requires that all the other operations (cleanup, list, verify, etc) are rewritten and I simply did not have the time. I tried then to get `BackendManager` to just be an interface to `BackendHandler` but apparently I did not manage to get that done either :/

warwickmm:
No problem.  Do you think that's still a worthwhile endeavor?  Is it worth keep this around?

kenkendk:
I still think we should change it to have a single shared implementation of the backend logic, but I don't see a reason to keep the unused code in the source. Currently it is just confusing, and if anyone needs it, they can pull it from the old revisions.

PR #4153, opened 2020-03-30
Add missing dependency for Mono certificates on Debian

borfast:
Without this package, Duplicati can't connect to the Backblaze B2 storage.

It always says:
> Failed to connect: Error: TrustFailure (The authentication or decryption has failed.)


drwtsn32x:
Just checked a Debian testing (bullseye) vm...  If I try to install `mono-runtime`, it will install `ca-certificates-mono` automatically (among many other packages).

Capture

What Debian version are you using?

That being said I doubt this change would cause any problems...

borfast:
I'm on Linux Mint 19, which is based on Ubuntu 18.04, which in turn is based on buster/sid.

I just checked `mono-runtime` and it does not depend on `ca-certificates-mono`:

```
$ apt-cache depends mono-runtime                                                                                                                                                                 
mono-runtime
  Depends: libc6
  Depends: mono-runtime-sgen
```

Doing a reverse check on `ca-certificates-mono` I get these packages:
```
$ apt-cache rdepends ca-certificates-mono                                                                                                                                                        
ca-certificates-mono
Reverse Depends:
  libmono-security4.0-cil
  mono-complete
  libmono-system4.0-cil
  libmono-security4.0-cil
  mono-devel
  mono-complete
  libmono-system4.0-cil
```

borfast:
I think I figured it out.

`ca-certificates-mono` is recommended by [libmono-system4.0-cil](https://packages.debian.org/bullseye/libmono-system4.0-cil) - but it's just recommended, it's not a requirement.

The thing is, in Debian, the `apt::install-recommends` setting is set to `true` by default, which explains why it was installed on your VM.

But like you said, adding the explicit dependency won't hurt regular Debian and will make the package work without a problem in other Debian derivatives that don't set `apt::install-recommends=true`.

drwtsn32x:
Agreed, I think this is a good change to merge.  Thanks!

borfast:
My pleasure. Thank you for this great software! :)

PR #4161, opened 2020-04-05
Remove boilerplate code in calls to VerifyRemoteList

warwickmm:
This adds a new overload for the `FilelistProcessor.VerifyRemoteList` method that allows us to remove some boilerplate code that appeared with each of the invocations.

### DIFF
--- a/Duplicati/Library/Main/Operation/CompactHandler.cs
+++ b/Duplicati/Library/Main/Operation/CompactHandler.cs
@@ -98,11 +98,9 @@ namespace Duplicati.Library.Main.Operation
 				using(var bk = sharedBackend == null ? new BackendManager(m_backendurl, m_options, m_result.BackendWriter, db) : null)
                 {
                     var backend = bk ?? sharedBackend;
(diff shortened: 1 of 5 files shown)


Duplicati/Library/Main/Operation/CompactHandler.cs
```diff
@@ -98,11 +98,9 @@ internal bool DoCompact(LocalDeleteDatabase db, bool hasVerifiedBackend, ref Sys
 				using(var bk = sharedBackend == null ? new BackendManager(m_backendurl, m_options, m_result.BackendWriter, db) : null)
                 {
                     var backend = bk ?? sharedBackend;
-                    if (!hasVerifiedBackend && !m_options.NoBackendverification)
+                    if (!hasVerifiedBackend)
```
kenkendk, line 101: Does this change the logic? Or does the calling code now set the flag according to the user options?

warwickmm: I moved the check for `!options.NoBackendverification` into the new `FileListProcessor.VerifyRemoteList` overload.

PR #4166, opened 2020-04-06
Add Edge and Chrome cache folders to exclusion list

drwtsn32x:
Newer Edge is based on Chromium so uses similar paths as Chrome.  Also noticed that the Chrome cache wasn't in the exclusion list either, so added that too.


warwickmm:
Looks good to me.  I find the [documentation](https://chromium.googlesource.com/chromium/src/+/HEAD/docs/user_data_dir.md) slightly misleading, but maybe I'm interpreting things incorrectly.

Do you know if the Chrome cache filter [here](https://github.com/duplicati/duplicati/blob/v2.0.5.104-2.0.5.104_canary_2020-03-25/Duplicati/Library/Utility/FilterGroups.cs#L294) matches anything?

drwtsn32x:
> Do you know if the Chrome cache filter here matches anything?

It may be valid on other platforms. I can check Linux tomorrow, but I don't have the resources to check MacOS.

PR #4178, opened 2020-04-21
Do not remove Alpha VSS dependencies in installer packages

### LINKED ISSUE #4149: VSS warning on non-Windows platforms
The following error is sometimes generated on non-Windows platforms:

```
LimitedWarnings: [
    2020-03-27 13:59:48 -07 - [Warning-Duplicati.Library.DynamicLoader.DynamicLoader`1-SoftError]: Failed to load process type Duplicati.Library.Common.IO.VssBackupComponents assembly /usr/lib/duplicati/Duplicati.Library.IO.dll, error message: Could not load type of field 'Duplicati.Library.Common.IO.VssBackupComponents:_vssBackupComponents' (1) due to: Could not load file or assembly 'AlphaVSS.Common, Version=1.4.0.0, Culture=neutral, PublicKeyToken=959d3993561034e3' or one of its dependencies.
]
```

Obviously VSS is Windows specific.  Not sure of the best way to fix this.  Ideas:

~~Exclude any VSS dlls from being collected here, when running on non-Windows platforms:
https://github.com/duplicati/duplicati/blob/master/Duplicati/Library/DynamicLoader/DynamicLoader.cs#L99~~

~~Switch from Warning to Information logging here:
https://github.com/duplicati/duplicati/blob/master/Duplicati/Library/DynamicLoader/DynamicLoader.cs#L138~~

Thoughts?

warwickmm:
Although these should not be needed for the Linux and OS X packages, it appears that there are times when this dependency is woken up, possibly due to issues regarding caching and static initialization of the `VssBackupComponents._vssBackupComponents` field.

This fixes #4149.

### DIFF
--- a/Installer/Makefile/Makefile
+++ b/Installer/Makefile/Makefile
@@ -37,12 +37,9 @@ package: build
 	mkdir "$(PACKAGE_DIR)/usr/share/doc/duplicati"	
 	cp -R "$(SOURCE_DIR)/Duplicati/GUI/Duplicati.GUI.TrayIcon/bin/Release" "$(PACKAGE_DIR)/usr/lib/duplicati"
 
-	rm -rf "$(PKG_TARGET)/alphavss"
 	rm -rf "$(PKG_TARGET)/SQLite"
 	rm -rf "$(PKG_TARGET)/win-tools"
 	rm -rf "$(PKG_TARGET)/"*.mdb
-	rm -rf "$(PKG_TARGET)/AlphaFS.dll"
-	rm -rf "$(PKG_TARGET)/AlphaVSS.Common.dll"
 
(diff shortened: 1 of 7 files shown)


warwickmm:
I don't know if there are any side effects other than larger packages due to the inclusion of the dlls.  If anyone has any insight into why these were removed from certain installation packages, it would be helpful to know.

An alternative approach is the patch proposed [here](https://github.com/duplicati/duplicati/commit/2e32833e8e6a67d3aa29147dfdf7beeabbd54e95), which appears to fix the known cases that trigger the dependency.  However, it's unclear if the dependency will reveal itself in other ways in the future.

kenkendk:
This should not have any adverse effects. If the libraries are loaded, they will not work anyway, so same situation as them not being there in the first place.

Issue #2249, opened 2017-01-11
Missed jobs queued multiple times?


autumnwalker:
I have:
- [x] searched open and closed issues for duplicates

----------------------------------------

### Version info
**Duplicati Version:** 2.0.1.39_canary_2017-01-08
**Operating System:** Ubuntu 16.04
**Backend:** ACD

### Bug description
I'm not sure how to confirm this, but going on a hunch ... I'm seeding a large(ish) backup right now which is causing my other jobs to miss their scheduled time, the large seed failed (that'll be another issue), and immediately Duplicati started running the two missed jobs (which are on a 15 minute schedule) repeatedly one after the other. I watched it cycle through the same two jobs twice.

Does Duplicati create an entry in the job queue each time a particular job is missed (i.e. if one job is missed three times there will be three entries in the queue to run it)?

canove:
I have the same issue on my side, my backup took some days to be done, and when it finally finished, the other scheduled daily backups were run one time for each day missed.

agelbukh:
I have the same issue. 

I backup the Outlook PST files on my laptop, which are some 30 GB (in total) database files being changed very frequently. So with each job, Duplicati needs to analyze the entire file to find the few changed blocks. This takes hours, heats the CPU a lot, and drains the battery. 

When the laptop is not used for some days (or when I am traveling and don't have Internet access for some days), Duplicati repeats this multi-hour analysis over and over, dozens of times, for all the scheduled instances missed while the laptop was not used.

duplicatibot:
This issue has been mentioned on **Duplicati**. There might be relevant details there:

https://forum.duplicati.com/t/when-duplicati-is-launched-the-backup-is-duplicated/9630/32

ts678:
> There might be relevant details there

There are definitely details now, primarily test results on behavior.

PR #4180, opened 2020-04-26
Fix issue where multiple backups could occur after pause

warwickmm:
Previously, the value of the `LastRun` field in the `Schedule` table was updated to be (approximately) the time that the job was added to the worker's list of tasks.  However, if the backup was missed (via pause or system suspend), this value would be incorrect.  If the next scheduled backup was also missed, then after the backup was finally run, the computed next start time would be based on the (incorrect) last run time of the first missed backup.  This would then immediately trigger an unnecessary second backup since the repetition internal has been exceeded.

Now, we add a handler for the `WorkerThread.StartingWork` event to set a more accurate value for the `LastRun` field in the `Schedule` table.

This fixes issue #2249.

### DIFF
--- a/Duplicati/Server/Scheduler.cs
+++ b/Duplicati/Server/Scheduler.cs
@@ -215,6 +215,11 @@ namespace Duplicati.Server
 
         private void OnStartingWork(WorkerThread<Runner.IRunnerData> worker, Runner.IRunnerData task)
         {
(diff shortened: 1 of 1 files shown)


drwtsn32x:
This worked for me. I am no longer able to reproduce the problem.

PR #4197, opened 2020-05-11
Add Tardigrade as new Backend

TopperDEL:
This PR adds Tardigrade as a new Backend using the native uplink.NET-Library. I've tested it with the BackendTester and did some manual tests. Would love to get feedback if I've missed anything.

Technically it should also work on Linux and MacOs, but I did not yet test this.


kenkendk:
This looks very thorough.

One addition could be to implement `IStreamingBackend` which adds methods that stream the data (preventing the use of temporary files). Looking at the code, it should be easy to add them.

It is not a requirement that this interface is added, but it performs better, if the backend supports it. Is there a reason it is not added? Such as the support library making a temporary file anyway.

warwickmm:
Nice work.  It appears that Tardigrade might be S3-compatible.  Has anyone tried Duplicati's existing S3 support with a Tardigrade account?

kenkendk:
@warwickmm Yes, this already works: https://documentation.tardigrade.io/how-tos/backup-with-duplicati
But with this PR, it is not required that you run the S3 gateway.

TopperDEL:
Yes, this PR implements a native integration where no special gateway is necessary. Currently one needs to setup a local gateway with Minio - this is no more needed with my PR.

Regarding `IStreamingBackend` @kenkendk: I simply forgot this one. Will add it later. :)

TopperDEL:
@kenkendk `IStreamingBackend `is implemented now, too.

Duplicati.Library.Backend.Tardigrade/TardigradeBackend.cs
```diff
@@ -0,0 +1,330 @@
...
+using System.Text;
+using System.Threading;
+using System.Threading.Tasks;
+using uplink.NET.Interfaces;
+using uplink.NET.Models;
+using uplink.NET.Services;
+
+namespace Duplicati.Library.Backend.Tardigrade
+{
+    public class Tardigrade : IBackend, IStreamingBackend
```
warwickmm, line 16: `IStreamingBackend` implements `IBackend`.  Is implementing `IBackend` here redundant?

Duplicati.Library.Backend.Tardigrade/TardigradeBackend.cs
```diff
@@ -0,0 +1,330 @@
...
+
+        public string ProtocolKey => PROTOCOL_KEY;
+
+        public IList<ICommandLineArgument> SupportedCommands
+        {
+            get
+            {
+                return new List<ICommandLineArgument>(new ICommandLineArgument[] {
+                    new CommandLineArgument(TARDIGRADE_AUTH_METHOD, CommandLineArgument.ArgumentType.String, Strings.Tardigrade.TardigradeAuthMethodDescriptionShort, Strings.Tardigrade.TardigradeAuthMethodDescriptionLong, "API key"),
+                    new CommandLineArgument(TARDIGRADE_SATELLITE, CommandLineArgument.ArgumentType.String, Strings.Tardigrade.TardigradeSatelliteDescriptionShort, Strings.Tardigrade.TardigradeSatelliteDescriptionShort, "us-central-1.tardigrade.io:7777"),
```
warwickmm, line 116: Should the 4th argument be `Strings.Tardigrade.TardigradeSatelliteDescriptionLong`?

Duplicati.Library.Backend.Tardigrade/TardigradeBackend.cs
TopperDEL: You're Right, I've educed it to IStreamingBackend

Duplicati.Library.Backend.Tardigrade/TardigradeBackend.cs
TopperDEL: True, Thanks. Resolved!

Duplicati.Library.Backend.Tardigrade/packages.config
```diff
@@ -0,0 +1,4 @@
+﻿<?xml version="1.0" encoding="utf-8"?>
+<packages>
+  <package id="uplink.NET" version="2.1.3" targetFramework="net462" />
```
warwickmm, line 3: We should add the license information, etc. to the `Duplicati.License` project.

TopperDEL: I've Added the License-info

warwickmm:
Great work.  Thanks!

PR #4202, opened 2020-05-19
Switch to uplink.Net v2.1.5 to correctly support Tardigrade-Backend on Linux

TopperDEL:
The uplink.Net-Library had a small error with the handling of the listorj_uplink.so on NetFramework. It deployed the windows-DLL instead of the linux-so. This got fixed in uplink.NET v2.1.5 and therefore I'll update the reference here.

I could successfully Test the Tardigrade-Backend and Windows and Linux now.


TopperDEL:
This one got fixed with #4203 already, sorry for the confusion.

Issue #3656, opened 2019-02-09
`__MonoCS__` can no longer be relied upon as of Mono 5.0


Ajedi32:
Various parts of Duplicati currently rely on the `__MonoCS__` symbol being defined during preprocessing. For example:

* [`Duplicati/Library/Modules/Builtin/CheckMonoSSL.cs#L58-L62`](https://github.com/duplicati/duplicati/blob/fd37a1384c2cb4eede8907ceb863286d4a205dd8/Duplicati/Library/Modules/Builtin/CheckMonoSSL.cs#L58-L62)
* [`Duplicati/GUI/Duplicati.GUI.TrayIcon/Program.cs#L46-L55`](https://github.com/duplicati/duplicati/blob/fd37a1384c2cb4eede8907ceb863286d4a205dd8/Duplicati/GUI/Duplicati.GUI.TrayIcon/Program.cs#L46-L55)

This is an artifact from the mcs compiler, which defined this symbol automatically and was used in Mono prior to version 5.0. As of 5.0 however, Mono uses the csc compiler [which does not define this symbol](https://www.mono-project.com/docs/about-mono/releases/5.0.0/#csc):

> Note: `mcs` defined the `__MonoCS__` symbol and some users incorrectly used it to conditionally compile code specific to Mono. This is considered bad practice and won’t work anymore now that `csc` is the default compiler. You can detect [whether you’re running on Mono at runtime](https://www.mono-project.com/docs/faq/technical/#how-can-i-detect-if-am-running-in-mono) instead.

It was previously discovered that this update broke the build on MonoDevelop (see #3648), but given the reliance on this symbol in other parts of the code I'm concerned the change might have had other unanticipated effects.

It seems like [`build-release.sh#L141`](https://github.com/duplicati/duplicati/blob/fd37a1384c2cb4eede8907ceb863286d4a205dd8/build-release.sh#L141) does manually define the symbol as of fc5b2097e1d45fbb376a00d38b47b17963b3b687 so it may be that this change isn't relevant outside MonoDevelop, but I'm not familiar enough with Duplicati's codebase to make that determination for myself, so I'm raising this issue just to be sure.

Do any parts of Duplicati need to be modified to account for the fact that the `__MonoCS__` preprocessor symbol is no longer defined automatically during compilation on Mono?

verhoek:
[...]

PR #4213, opened 2020-05-28
Removed references to `__MonoCS__`

kenkendk:
The compiler directive `__MonoCS__` was being set by the Mono compiler, but removed some years ago.
The Duplicati code "fixed" the removal by setting the directive during builds, but this causes the debugging and released versions to differ a bit. Also, the use of the compiler directive suggests we are using a functionality that does not exist.

This commit removes all references to `__MonoCS__` (except for one, which is removed in #4212).


warwickmm:
I believe this fixes #3656.

PR #4214, opened 2020-05-28
Add Retina Icons (High Resolution)

baumerdev:
I added images with douple or triple the dimensions for devices that have a higher resolution/ppi like modern smartphones and notebooks (like retina displays). Otherwise normal images look kind of blurred.

For every image there are two images with the same basename and an added suffix "_2x.png" and "_3x.png". Examples:

| restore.png | restore_2x.png | restore_3x.png |
|-------------|----------------|----------------|
|  |  |  |

| treeicons.png | treeicons_2x.png | treeicons_3x.png |
|---------------|------------------|------------------|
|  |  |  |

The images will be displayed in the original size in the web frontend but much cleaner/more crisp on higher resolution displays. If you opened this Github-Issue on a device that has a higher resolution, you should see the effect with the following images:

| restore.png | restore_2x.png | restore_3x.png |
|-------------|----------------|----------------|
|  |  |  |

| treeicons.png | treeicons_2x.png | treeicons_3x.png |
|---------------|------------------|------------------|
|  | |  |

Those new images are referenced from HTML templates with an additional srcset attribute:
``.

From Stylesheets (LESS and CSS) the images are added within selectors:
`@media only screen and (-webkit-min-device-pixel-ratio: 1.25), only screen and (min-resolution: 192dpi), only screen and (min-resolution: 2.25dppx)`
and
`@media only screen and (-webkit-min-device-pixel-ratio: 1.25), only screen and (min-resolution: 288dpi), only screen and (min-resolution: 2.25dppx)`

That way the images are only loaded if necessary and these methods guarantee backward compatibility with older web browsers.

The vector images used to create these images can be found in a folder "_svg". If there is any need to update the PNG at a later time.


kenkendk:
Wow, that looks fantastic!

PR #4231, opened 2020-06-18
Tardigrade-Backend: Resolve two issues

TopperDEL:
This PR bumps the Tardigrade-Backend to v2.3.1 and solves two issues:

- we had random crashes with the uplink.Net-Library that got solved with v2.3.1. The garbage collection moved memory that the underlying uplink-c expected at a certain address.
- listing files without a folder did not work due to a small error

### DIFF
--- a/Duplicati/Library/Backend/Tardigrade/TardigradeBackend.cs
+++ b/Duplicati/Library/Backend/Tardigrade/TardigradeBackend.cs
@@ -46,7 +46,7 @@ namespace Duplicati.Library.Backend.Tardigrade
 
         static Tardigrade()
         {
-            Access.SetTempDirectory(System.IO.Path.GetTempPath());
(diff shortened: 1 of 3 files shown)


Duplicati/Library/Backend/Tardigrade/TardigradeBackend.cs
```diff
@@ -44,6 +44,11 @@ public class Tardigrade : IStreamingBackend
             { "Access grant", "Access grant" },
         };
 
+        static Tardigrade()
+        {
+            Access.SetTempDirectory(System.IO.Path.GetTempPath());
```
warwickmm, line 49: The temporary directory can be set by the user.  Should we reference `Library.Utility.TempFolder.SystemTempPath` instead?

TopperDEL: Ah, didn't know that one. Thanks @warwickmm - resolved it!

PR #4232, opened 2020-06-18
Tardigrade: Correctly deploy binaries with the installer

TopperDEL:
This PR is a first try to deploy all the necessary binaries for Windows, Linux and MacOS with the installers.

Basically I've added all binaries as "Copy to output"-assets in the Tardigrade-Backend. From my understanding they would be copied into the "Installer-Zip-File". From there I remove all binaries that are not necessary for the current Platform.

On windows we have a 64 and a 32 bit version of the library, so I preload that DLL in the Tardigrade-Backend (only on windows).

I would be happy if someone who knows it could create an installer with it. I and others from Storj Labs could then test the different versions. I did and could not do any further testing with the scripts, but I've tried to do it exactly the same like the script already do it with SQLite and other assets, that get removed from platform-installers. The changes are quite straight-forward.

Thank you!

### DIFF
--- a/Duplicati/Library/Backend/Tardigrade/TardigradeBackend.cs
+++ b/Duplicati/Library/Backend/Tardigrade/TardigradeBackend.cs
@@ -267,7 +267,9 @@ namespace Duplicati.Library.Backend.Tardigrade
             foreach (var obj in objects.Items)
             {
                 TardigradeFile file = new TardigradeFile(obj);
-                file.Name = file.Name.Replace(GetBasePath(), "");
+                var basePath = GetBasePath();
+                if (basePath != "")
+                    file.Name = file.Name.Replace(basePath, "");
                 files.Add(file);
             }
 
--- a/Duplicati/Library/Backend/Tardigrade/TardigradeBackend.cs
+++ b/Duplicati/Library/Backend/Tardigrade/TardigradeBackend.cs
@@ -54,20 +54,19 @@ namespace Duplicati.Library.Backend.Tardigrade
             if (_libraryLoaded)
                 return;
 
-            if(System.Environment.Is64BitProcess)
+            if (Duplicati.Library.Common.Platform.IsClientWindows) //We need to init only on Windows to distinguish between x64 and x86
             {
-               var res = LoadLibrary("win-x64/storj_uplink.dll");
+                if (System.Environment.Is64BitProcess)
+                {
+                    var res = LoadLibrary("win-x64/storj_uplink.dll");
+                }
+                else
+                {
+                    var res = LoadLibrary("win-x86/storj_uplink.dll");
+                }
             }
-            else
-            {
-                var res = LoadLibrary("win-x86/storj_uplink.dll");
-            }
-            _libraryLoaded = true;
-        }
-
-        static Tardigrade()
-        {
             Access.SetTempDirectory(Library.Utility.TempFolder.SystemTempPath);
+            _libraryLoaded = true;
         }
 
         // ReSharper disable once UnusedMember.Global
@@ -80,8 +79,7 @@ namespace Duplicati.Library.Backend.Tardigrade
         // This constructor is needed by the BackendLoader.
         public Tardigrade(string url, Dictionary<string, string> options)
         {
-            if (Duplicati.Library.Common.Platform.IsClientWindows)
-                InitStorjLibrary(); //We need to init only on Windows to distinguish between x64 and x86
+            InitStorjLibrary();
 
             var auth_method = options[TARDIGRADE_AUTH_METHOD];
             if (auth_method == "Access grant")
--- a/Duplicati/Library/Backend/Tardigrade/TardigradeBackend.cs
+++ b/Duplicati/Library/Backend/Tardigrade/TardigradeBackend.cs
@@ -80,7 +80,7 @@ namespace Duplicati.Library.Backend.Tardigrade
         // This constructor is needed by the BackendLoader.
         public Tardigrade(string url, Dictionary<string, string> options)
         {
-            if (RuntimeInformation.IsOSPlatform(OSPlatform.Windows))
+            if (Duplicati.Library.Common.Platform.IsClientWindows)
                 InitStorjLibrary(); //We need to init only on Windows to distinguish between x64 and x86
 
             var auth_method = options[TARDIGRADE_AUTH_METHOD];
--- a/Duplicati/Library/Backend/Tardigrade/TardigradeBackend.cs
+++ b/Duplicati/Library/Backend/Tardigrade/TardigradeBackend.cs
@@ -4,6 +4,7 @@ using System.Collections.Generic;
 using System.IO;
 using System.Linq;
 using System.Linq.Expressions;
+using System.Runtime.InteropServices;
 using System.Text;
 using System.Threading;
 using System.Threading.Tasks;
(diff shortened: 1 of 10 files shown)


warwickmm:
There appear to be some merge conflicts.  @TopperDEL, would you mind resolving them?

I'm not very familiar with the installer code.  We may need to wait for @kenkendk's input here.

TopperDEL:
@warwickmm I've resolved the merge conflict.

littleskunk:
I would love to run the docker image on my server and test out the new tardigrade integration. Is there something I can do to help getting the PR over the finish line?

Duplicati/Library/Backend/Tardigrade/TardigradeBackend.cs
```diff
@@ -59,6 +80,9 @@ public Tardigrade()
         // This constructor is needed by the BackendLoader.
         public Tardigrade(string url, Dictionary<string, string> options)
         {
+            if (RuntimeInformation.IsOSPlatform(OSPlatform.Windows))
```
warwickmm, line 83: I like the use of `RuntimeInformation.IsOSPlatform`, but we should probably be consistent and use `Duplicati.Library.Common.Platform.IsClientWindows`.  We can consider changing the implementation of `Platform.IsClientWindows` in the future.

warwickmm:
The changes look reasonable to me and I think we can merge after the trivial suggestion [above](https://github.com/duplicati/duplicati/pull/4232#discussion_r446553804).  We may need @kenkendk's help with creating the installers for testing purposes, although it might be easier to just create another canary release.

Duplicati/Library/Backend/Tardigrade/TardigradeBackend.cs
TopperDEL: No problem, I've changed it.

Duplicati/Library/Backend/Tardigrade/TardigradeBackend.cs
```diff
@@ -44,6 +45,26 @@ public class Tardigrade : IStreamingBackend
...
+
+            if(System.Environment.Is64BitProcess)
+            {
+               var res = LoadLibrary("win-x64/storj_uplink.dll");
+            }
+            else
+            {
+                var res = LoadLibrary("win-x86/storj_uplink.dll");
+            }
+            _libraryLoaded = true;
```
warwickmm, line 65: Instead of keeping track of `_libraryLoaded`, what are your thoughts about doing the above in a static constructor?

TopperDEL: I wanted to be sure that the DLL gets loaded before everythin else. But I assume this now leads to the situation that the DLL gets loaded whenever the BackendLoader collects the available backends, right? Didn't think of that.

warwickmm: Good point.  We should probably avoid loading the library if it's not necessary.

Any other changes you want to make before I merge?

TopperDEL: Resolved this, too. I'm fine for merging! Thank you! @warwickmm

warwickmm: Thanks!  One last question.  I noticed when I build the solution that a file `libstorj_uplink.so` appears in the output directory.  How is this file appearing without the treatment being applied in this pull request?

TopperDEL: You should also see a libstorj_uplink.dylib. The idea is to place all binaries for the different platforms in the output dir and include it in the installer. But I've also adjusted the build scripts for the platforms to remove all files that are not necessary there. So the *.so and *.dylib would be removed on Windows, the dylib and dll would be removed on linux and so on. You can see this in the other files I've change with this PR.

warwickmm: Hmmm...I don't see `libstorj_uplink.dylib`.  This is after a rebuild of the solution:
```
$ ls -1 Duplicati/GUI/Duplicati.GUI.TrayIcon/bin/Debug/*.{so,dylib}
Duplicati/GUI/Duplicati.GUI.TrayIcon/bin/Debug/libstorj_uplink.so
Duplicati/GUI/Duplicati.GUI.TrayIcon/bin/Debug/storj_uplink.dylib
Duplicati/GUI/Duplicati.GUI.TrayIcon/bin/Debug/storj_uplink.so
```
How does `libstorj_uplink.so` appear in the output directory without the explicit `CopyToOutputDirectory` entries in the `.csproj` file that the other files need?

TopperDEL: There's a "storj_uplink.dylib" - that's ok.

Are you building duplicati on Linux?

warwickmm: Yes, this is on Linux.

TopperDEL: So then the file comes from the "normal" build on linux. You won't have the storj_uplink.dll then.
The installer is build on windows, as I've understood? there you would habe the correct files then - the linux-so, the mac-dyli and the Window-dll.

I could change storj_uplink.so to libstorj_uplink.so so we only have one linux-binary in all cases.

warwickmm: It looks like `libstorj_uplink.so` was added to the project as a link to the nuget package folder, while `storj_uplink.dylib` and `storj_uplink.so` were added separately.  Is there a reason for the difference?

image

warwickmm: > The installer is build on windows, as I've understood?

I don't build the installers (@kenkendk creates the releases), but my impression is that it's built on either Mac or Linux via the `build-release.sh` script.

TopperDEL: I've just commited a change. You should now only see one file.

Yes, the uplink.net deploys the binaries for the system you build on. But for the installer, we need all - so I've added them as assets to the Tardigrad-Backend-Project so they always get deployed to the output-dir. the installer-script should remove it from the platform-installers afterwards.

TopperDEL: Well, now it should work on all platforms as all three binaries (4 if you count win-x64 and win-x86 seperately) get deployed as assets. We should give it a try.

warwickmm: > Yes, the uplink.net deploys the binaries for the system you build on. But for the installer, we need all - so I've added them as assets to the Tardigrad-Backend-Project so they always get deployed to the output-dir. the installer-script should remove it from the platform-installers afterwards.

Ah, ok.  That makes sense.  Thanks.

TopperDEL: One for Linux, one for Mac. Plus there should be win-DLLs in subfolders. storj_uplink.dll.dll

warwickmm: Yup, I see them.  Looks good, I'll merge.

PR #4242, opened 2020-07-01
Bump the Tardigrade-Backend to NetFramework 4.7.1

TopperDEL:
The Tardigrade-Backend uses the uplink.NET-library which is on NetStandard 2.0. If this one gets included in a project targeting NetFramework prio to 4.7.1, many DLL-files get added to the output-dir and clutter the installer (see issue #4234).

This PR simply changes the Tardigrade-Backend to 4.7.1 as this seems to solve that issue and does not break anything else. It might be that the Tardigrade-Backend won't work for some users that don't have NetFramework 4.7.1.


warwickmm:
Thanks @TopperDEL.

Issue #4148, opened 2020-03-25
Cannot restore empty file if target path does not exist


Champ-Goblem:
- [x] I have searched open and closed issues for duplicates.

----------------------------------------

## Environment info

- **Duplicati version**: 2.0.5.103_canary_2020-02-18
- **Operating system**: Manjaro 5.4.24
- **Backend**: SFTP on Raspberry Pi running Raspbian Lite

## Description

I successfully backup my laptop running Linux Mint to the remote server without any problems, but as soon as I try to restore from the remote server to the Manjaro machine, we get through some of the downloading process but hit the error:
```
System.IO.DirectoryNotFoundException: Could not find a part of the path "/home/administrator/BTHH5a-fw-mod/build/rootfs-openwrt/squashfs-root/etc/dropbear/dropbear_dss_host_key"
```

Whenever I try and restore, no matter how many different fresh backups I've made this error still persists on the restoration side. The file itself is empty and im trying to restore the files to /, I had also created the folder /home/administrator beforehand which doesnt have any child items and is owned by the main user account (not root) if that raises some problems?
## Steps to reproduce

Just making a backup of that one file and restoring:
1. Backup that single empty file at that path
2. Attempt to restore to another computer
3. Fails with error above

- **Actual result**:
```
System.IO.DirectoryNotFoundException: Could not find a part of the path "/home/administrator/BTHH5a-fw-mod/build/rootfs-openwrt/squashfs-root/etc/dropbear/dropbear_dss_host_key"
```
- **Expected result**:
It should restore the file fine

## Debug log

```
[...]

PR #4245, opened 2020-07-01
Ensure that paths exist when restoring empty files

warwickmm:
This fixes an issue where the folders leading to an empty file weren't created properly during the restore process.  This also adds a test that would previously fail without this fix.

This fixes issue #4148.


warwickmm:
Looks like the AppVeyor test failed.  I'll have to investigate.

warwickmm:
The AppVeyor failure should be fixed now.  On Windows, the drive letter is included as part of the path when the `dont-compress-restore-paths` option is used.

PR #4249, opened 2020-07-08
Catch compression errors and report them

jpschewe:
When the file list is over 4GB one needs to have the zip64 option
enabled. This commit catches the exception and reports it as an error so
that the user knows what the problem is.

Fixes duplicati/duplicati/#4206

### DIFF
--- a/Duplicati/Library/Main/Volumes/FilesetVolumeWriter.cs
+++ b/Duplicati/Library/Main/Volumes/FilesetVolumeWriter.cs
@@ -8,6 +8,11 @@ namespace Duplicati.Library.Main.Volumes
 {
     public class FilesetVolumeWriter : VolumeWriterBase
     {
+        /// <summary>
+        /// The tag used for logging
+        /// </summary>
+        private static readonly string LOGTAG = Logging.Log.LogTagFromType<FilesetVolumeWriter>();
+
         private readonly Library.Utility.TempFile m_tempFile;
         private readonly Stream m_tempStream;
         private StreamWriter m_streamwriter;
@@ -159,11 +164,18 @@ namespace Duplicati.Library.Main.Volumes
             m_writer.Flush();
             m_streamwriter.Flush();
 
-            using (Stream sr = m_compression.CreateFile(FILELIST, CompressionHint.Compressible, DateTime.UtcNow))
+            try
             {
-                m_tempStream.Seek(0, SeekOrigin.Begin);
-                m_tempStream.CopyTo(sr);
-                sr.Flush();
+                using (Stream sr = m_compression.CreateFile(FILELIST, CompressionHint.Compressible, DateTime.UtcNow))
+                {
+                    m_tempStream.Seek(0, SeekOrigin.Begin);
+                    m_tempStream.CopyTo(sr);
+                    sr.Flush();
+                }
+            }
+            catch (System.NotSupportedException e)
+            {
+                Logging.Log.WriteErrorMessage(LOGTAG, "CompressionError", e, "Compression Error: {0}", e.Message);
             }
         }
 
--- a/Duplicati/Library/Main/Volumes/FilesetVolumeWriter.cs
+++ b/Duplicati/Library/Main/Volumes/FilesetVolumeWriter.cs
@@ -8,11 +8,6 @@ namespace Duplicati.Library.Main.Volumes
 {
     public class FilesetVolumeWriter : VolumeWriterBase
     {
-        /// <summary>
-        /// The tag used for logging
-        /// </summary>
-        private static readonly string LOGTAG = Logging.Log.LogTagFromType<FilesetVolumeWriter>();
-
         private readonly Library.Utility.TempFile m_tempFile;
         private readonly Stream m_tempStream;
         private StreamWriter m_streamwriter;
--- a/Duplicati/Library/Main/Volumes/FilesetVolumeWriter.cs
+++ b/Duplicati/Library/Main/Volumes/FilesetVolumeWriter.cs
@@ -170,11 +170,6 @@ namespace Duplicati.Library.Main.Volumes
                     sr.Flush();
                 }
             }
-            catch (System.NotSupportedException e)
-            {
-                Logging.Log.WriteErrorMessage(LOGTAG, "CompressionError", e, "Compression Error: {0}", e.Message);
-                throw e;
-            }
             finally
             {
                 m_writer.Close();
--- a/Duplicati/Library/Main/Volumes/FilesetVolumeWriter.cs
+++ b/Duplicati/Library/Main/Volumes/FilesetVolumeWriter.cs
@@ -140,9 +140,6 @@ namespace Duplicati.Library.Main.Volumes
             if (m_streamwriter != null)
             {
                 this.AddFilelistFile();
-                m_writer.Close();
-                m_streamwriter.Dispose();
-                m_streamwriter = null;
             }
 
             if (m_tempStream != null)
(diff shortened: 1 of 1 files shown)


warwickmm:
Thanks for your help in debugging this!

Duplicati/Library/Main/Volumes/FilesetVolumeWriter.cs
```diff
@@ -159,11 +164,18 @@ private void AddFilelistFile()
...
+                    m_tempStream.CopyTo(sr);
+                    sr.Flush();
+                }
+            }
+            catch (System.NotSupportedException e)
             {
-                m_tempStream.Seek(0, SeekOrigin.Begin);
-                m_tempStream.CopyTo(sr);
-                sr.Flush();
+                Logging.Log.WriteErrorMessage(LOGTAG, "CompressionError", e, "Compression Error: {0}", e.Message);
```
warwickmm, line 178: Does this really work?  How does logging the error prevent `Dispose` from calling `Close` too soon?

Sorry for asking you to test so many different scenarios, but I'm wondering if the following patch yields the desired results (that is, it reveals the "larger than 4GiB" exception instead of the "No token to close. path " one):
```
diff --git a/Duplicati/Library/Main/Volumes/FilesetVolumeWriter.cs b/Duplicati/Library/Main/Volumes/FilesetVolumeWriter.cs
index ab0f505dc..df57b4ebd 100644
--- a/Duplicati/Library/Main/Volumes/FilesetVolumeWriter.cs
+++ b/Duplicati/Library/Main/Volumes/FilesetVolumeWriter.cs
@@ -135,9 +135,6 @@ namespace Duplicati.Library.Main.Volumes
             if (m_streamwriter != null)
             {
                 this.AddFilelistFile();
-                m_writer.Close();
-                m_streamwriter.Dispose();
-                m_streamwriter = null;
             }
 
             if (m_tempStream != null)
@@ -159,11 +156,20 @@ namespace Duplicati.Library.Main.Volumes
             m_writer.Flush();
             m_streamwriter.Flush();
 
-            using (Stream sr = m_compression.CreateFile(FILELIST, CompressionHint.Compressible, DateTime.UtcNow))
+            try
+            {
+                using (Stream sr = m_compression.CreateFile(FILELIST, CompressionHint.Compressible, DateTime.UtcNow))
+                {
+                    m_tempStream.Seek(0, SeekOrigin.Begin);
+                    m_tempStream.CopyTo(sr);
+                    sr.Flush();
+                }
+            }
+            finally
             {
-                m_tempStream.Seek(0, SeekOrigin.Begin);
-                m_tempStream.CopyTo(sr);
-                sr.Flush();
+                m_writer.Close();
+                m_streamwriter.Dispose();
+                m_streamwriter = null;
             }
         }
```

Basically, if creation of the zip file succeeds _or_ fails, immediately close/null out the underlying writers so that future calls to `Close` don't attempt to create the filelist again.

jpschewe: The logging was enough for me to get the log entries that I reported on the ticket. I can give your patch a try the next chance I get.

warwickmm: My suspicion with the logging approach is that while the logs and UI indicate an error, the backup isn't actually halted due to the exception.  This leaves you with a (possibly broken) backup that is missing the filelist.

kenkendk: I agree, the code should have a `throw` statement, so we stop the backup, because the written zip file is incomplete.

jpschewe: I can rethrow the exception, however something was swallowing that exception previously causing us to get to the point of the array error. I haven't been able to figure out where the `NotSupportedException` is going.

warwickmm: My guess is that the `FilesetVolumeWriter` was being disposed before the exception from the `async` call was actually raised.  The call to `Dispose` threw a new exception, and that was hiding the actual exception.

jpschewe: I've pushed an updated change that when I execute without the zip64 flag I get the following exception in the log.

```
System.NotSupportedException: Attempted to write a stream that is larger than 4GiB without setting the zip64 option
  at Duplicati.Library.Main.Volumes.FilesetVolumeWriter.AddFilelistFile () [0x00099] in <115666528dc74e7e8bb3080680ba47bc>:0 
  at Duplicati.Library.Main.Volumes.FilesetVolumeWriter.Close () [0x0000f] in <115666528dc74e7e8bb3080680ba47bc>:0 
  at Duplicati.Library.Main.Operation.Backup.UploadRealFilelist+<>c__DisplayClass1_0.<Run>b__0 (<>f__AnonymousType6`1[<Output>j__TPar] self) [0x0047a] in <115666528dc74e7e8bb3080680ba47bc>:0 
  at CoCoL.AutomationExtensions.RunTask[T] (T channels, System.Func`2[T,TResult] method, System.Boolean catchRetiredExceptions) [0x000d5] in <9a758ff4db6c48d6b3d4d0e5c2adf6d1>:0 
  at Duplicati.Library.Main.Operation.BackupHandler.RunAsync (System.String[] sources, Duplicati.Library.Utility.IFilter filter, System.Threading.CancellationToken token) [0x014f4] in <115666528dc74e7e8bb3080680ba47bc>:0 
  at CoCoL.ChannelExtensions.WaitForTaskOrThrow (System.Threading.Tasks.Task task) [0x00050] in <9a758ff4db6c48d6b3d4d0e5c2adf6d1>:0 
  at Duplicati.Library.Main.Operation.BackupHandler.Run (System.String[] sources, Duplicati.Library.Utility.IFilter filter, System.Threading.CancellationToken token) [0x0000a] in <115666528dc74e7e8bb3080680ba47bc>:0 
  at Duplicati.Library.Main.Controller+<>c__DisplayClass14_0.<Backup>b__0 (Duplicati.Library.Main.BackupResults result) [0x0004d] in <115666528dc74e7e8bb3080680ba47bc>:0 
  at Duplicati.Library.Main.Controller.RunAction[T] (T result, System.String[]& paths, Duplicati.Library.Utility.IFilter& filter, System.Action`1[T] method) [0x0029a] in <115666528dc74e7e8bb3080680ba47bc>:0 
  at Duplicati.Library.Main.Controller.Backup (System.String[] inputsources, Duplicati.Library.Utility.IFilter filter) [0x0007a] in <115666528dc74e7e8bb3080680ba47bc>:0 
  at Duplicati.Server.Runner.Run (Duplicati.Server.Runner+IRunnerData data, System.Boolean fromQueue) [0x003a8] in <05c676d9d79840da886d514feedc4e80>:0
```

I think this resolves the question here.

I have noticed that when a backup fails, the log messages don't always show up on the log for the backup, but rather I need to look in "About" and "Show Log".

warwickmm: Thanks @jpschewe.  Does it work without the catch (that is, using only a try-finally)?  The exception should automatically show up in the log, so catching the exception and rethrowing it feels redundant.

jpschewe: I'll give it a try. I'm concerned that something else up the stack is swallowing the exception and no error will be logged.

jpschewe: OK, I just finished a test with the catch removed and I got the proper error in the log.

warwickmm: Great!

Duplicati/Library/Main/Volumes/FilesetVolumeWriter.cs
```diff
@@ -8,6 +8,11 @@ namespace Duplicati.Library.Main.Volumes
 {
     public class FilesetVolumeWriter : VolumeWriterBase
     {
+        /// <summary>
+        /// The tag used for logging
+        /// </summary>
+        private static readonly string LOGTAG = Logging.Log.LogTagFromType<FilesetVolumeWriter>();
```
warwickmm, line 14: One last thing.  Can you remove this field since it is no longer used?

warwickmm:
Thanks for your help @jpschewe!

Issue #4225, opened 2020-06-14
S3 backend adds double slashes when verifying backend files


jpschewe:
I’ve recently tried using duplicati with minio (S3 compatible storage). I want to backup multiple systems to the same bucket, so need to use a folder per system. I’ve configured my first system with a folder path and started a backup. This correctly creates the folder in my bucket and backed up the files. However when duplicati got to the verification step it is adding an extra slash to the request for the file to read. This results in the following error from minio and the verify step fails.

```
minio[3812]: Error: Object name invalid: jon-backup#jon-2019_2020-06-08//duplicati-20200609T000849Z.dlist.zip.aes
minio[3812]:        6: /home/jpschewe/.gocode/pkg/mod/github.com/minio/minio@v0.0.0-20200608193042-c1382c09d994/cmd/object-api-input-checks.go:74:cmd.checkListObjsArgs()
minio[3812]:        5: /home/jpschewe/.gocode/pkg/mod/github.com/minio/minio@v0.0.0-20200608193042-c1382c09d994/cmd/object-api-common.go:279:cmd.listObjects()
minio[3812]:        4: /home/jpschewe/.gocode/pkg/mod/github.com/minio/minio@v0.0.0-20200608193042-c1382c09d994/cmd/fs-v1.go:1268:cmd.(*FSObjects).ListObjects()
minio[3812]:        3: /home/jpschewe/.gocode/pkg/mod/github.com/minio/minio@v0.0.0-20200608193042-c1382c09d994/cmd/fs-v1.go:1379:cmd.(*FSObjects).ListObjectsV2()
minio[3812]:        2: /home/jpschewe/.gocode/pkg/mod/github.com/minio/minio@v0.0.0-20200608193042-c1382c09d994/cmd/bucket-listobjects-handlers.go:240:cmd.objectAPIHandlers.ListObjectsV2Handler()
minio[3812]:        1: net/http/server.go:2012:http.HandlerFunc.ServeHTTP()
```

I have tested this under duplicati 2.0.5.1_beta_2020-01-18 and 2.0.5.107_canary_2020-05-26.

```
>mono --version
Mono JIT compiler version 6.8.0.123 (tarball Tue May 12 15:11:57 UTC 2020)
Copyright (C) 2002-2014 Novell, Inc, Xamarin Inc and Contributors. www.mono-project.com
        TLS:           __thread
        SIGSEGV:       altstack
        Notifications: epoll
        Architecture:  amd64
        Disabled:      none
[...]

PR #4250, opened 2020-07-08
Remove try/catch that is hiding errors

jpschewe:
The try/catch here was hiding timeouts fetching the test files. This
logic appears to be for handling a very old version of S3. They should
not be needed anymore.

Fixes duplicati/duplicati#4225


warwickmm:
Thanks @jpschewe.

Issue #4229, opened 2020-06-15
"UNIQUE constraint failed" error trying to back up Firefox profile in Windows


dferreyra:
- [x] I have searched open and closed issues for duplicates.

----------------------------------------

## Environment info

- **Duplicati version**: 2.0.5.1
- **Operating system**: Windows 10
- **Backend**: Amazon S3

## Description

My backups started failing with the following error:
```
UNIQUE constraint failed: FilesetEntry.FilesetID, FilesetEntry.FileID
```
I've narrowed it down to directories that end with a dot; e.g., "https+++www.example.com." -- notice the dot after the "com".

The directories are being created by Firefox in the "storage" directory in my Firefox profile.  Here's what a simplified reproduction I have looks like:
```
    C:\Temp>dir storage /b /s
    C:\Temp\storage\default
    C:\Temp\storage\default\https+++www.example.com
    C:\Temp\storage\default\https+++www.example.com\cache
    C:\Temp\storage\default\https+++www.example.com\cache\caches.sqlite
    C:\Temp\storage\default\https+++www.example.com.
    C:\Temp\storage\default\https+++www.example.com.\cache
    C:\Temp\storage\default\https+++www.example.com.\cache\caches.sqlite
```

Maybe it's related to #3642 or #3839.

My workaround is to exclude the "storage" directory in my Firefox profile.

## Steps to reproduce

1. Create a tree like the one described above; i.e., two directories where they only differ in that one ends in a dot, each containing a file with the same name.  To create directories that end in a dot, you can use the following syntax, `mkdir \\?\C:\Temp\storage\default\https+++www.example.com.`.
2. Configure a backup of the tree in Duplicati.  The backend you choose probably doesn't matter.
3. Run the backup.

- **Actual result**:
The backup fails with this error:
```
UNIQUE constraint failed: FilesetEntry.FilesetID, FilesetEntry.FileID
```
- **Expected result**:
A good backup with no errors.

## Debug log

```
2020-06-14 21:23:45 -07 - [Information-Duplicati.Library.Main.Controller-StartingOperation]: The operation Backup has started
2020-06-14 21:23:45 -07 - [Verbose-Duplicati.Library.Main.Operation.Backup.FileEnumerationProcess-IncludingSourcePath]: Including source path: C:\Temp\storage\
[...]

Issue #3518, opened 2018-11-27
Unique constraints violation in 2.0.4.4


Kasheen:
- [x] I have searched open and closed issues for duplicates.
(Found other unique constraints violation bugs which apparently have been fixed in prior versions).

----------------------------------------

## Environment info

- **Duplicati version**: 2.0.4.4_canary_2018-11-14
- **Operating system**: Windows 10 Pro Version	10.0.17134 Build 17134
- **Backend**: Stack (WebDAV)

## Description

Running a backup causes the following error since updating to 2.0.4.4:

> constraint failed UNIQUE constraint failed: FilesetEntry.FilesetID, FilesetEntry.FileID

## Steps to reproduce

Updated duplicati to 2.0.4.4 from a previous version, I'm not sure what I was running before - I don't update every version but rather every few versions.

Attempted to recreate database which completes successfully but backups still result in the same error.

## Debug log

```System.Data.SQLite.SQLiteException (0x80004005): constraint failed
UNIQUE constraint failed: FilesetEntry.FilesetID, FilesetEntry.FileID
   at System.Data.SQLite.SQLite3.Reset(SQLiteStatement stmt)
   at System.Data.SQLite.SQLite3.Step(SQLiteStatement stmt)
   at System.Data.SQLite.SQLiteDataReader.NextResult()
   at System.Data.SQLite.SQLiteDataReader..ctor(SQLiteCommand cmd, CommandBehavior behave)
   at System.Data.SQLite.SQLiteCommand.ExecuteReader(CommandBehavior behavior)
   at System.Data.SQLite.SQLiteCommand.ExecuteNonQuery(CommandBehavior behavior)
   at Duplicati.Library.Main.Database.ExtensionMethods.ExecuteNonQuery(IDbCommand self, Boolean writeLog, String cmd, Object[] values)
   at Duplicati.Library.Main.Database.LocalBackupDatabase.AddFile(String filename, DateTime lastmodified, Int64 blocksetID, Int64 metadataID, IDbTransaction transaction)
   at Duplicati.Library.Main.Operation.Backup.BackupDatabase.<>c__DisplayClass8_0.<AddDirectoryEntryAsync>b__0()
   at Duplicati.Library.Main.Operation.Common.SingleRunner.<>c__DisplayClass3_0.<RunOnMain>b__0()
   at Duplicati.Library.Main.Operation.Common.SingleRunner.<DoRunOnMain>d__2`1.MoveNext()
[...]

Issue #3839, opened 2019-07-29
Backup (or exclude) Files with only .dot Extention


Photographix:
- [x] I have searched open and closed issues for duplicates.

----------------------------------------

## Environment info
- **Duplicati version**: 2.0.4.22
- **Operating system**: Windows Server 2019
- **Backend**: Jotta

## Description
Backup or omit files that have **only** one dot (.) extension -> MyFile.

## Steps to reproduce
1. create a file with only one point as extention
2. try a backup of this file

- **Actual result**:
System.IO.FileNotFoundException
- **Expected result**:
Backup from file 
or or the possibility to skip them via filter

## Screenshots
![2019-07-29 09_15_20-Window](https://user-images.githubusercontent.com/16647382/62029609-5bf53580-b1e3-11e9-8e8b-677ea09199cc.png)

## Debug log
`2019-07-29 01:45:39 +02 - [Information-Duplicati.Library.Main.Controller-StartingOperation]: Die Operation Backup wurde gestartet
2019-07-29 01:54:01 +02 - [Information-Duplicati.Library.Main.BasicResults-BackendEvent]: Backend event: List - Started:  ()
2019-07-29 01:54:05 +02 - [Information-Duplicati.Library.Main.BasicResults-BackendEvent]: Backend event: List - Completed:  (13,45 KB)
2019-07-29 02:28:09 +02 - [Warning-Duplicati.Library.Main.Operation.Backup.FileEnumerationProcess-FileAccessError]: Error reported while accessing file: \\aio-pod\Anika\BKP-WhatsApp\Media\WhatsApp Documents\Sent\DOC-20180611-WA0009.
System.IO.FileNotFoundException: Die Datei "\\aio-pod\Anika\BKP-WhatsApp\Media\WhatsApp Documents\Sent\DOC-20180611-WA0009" konnte nicht gefunden werden.
Dateiname: "\\aio-pod\Anika\BKP-WhatsApp\Media\WhatsApp Documents\Sent\DOC-20180611-WA0009"
   bei System.IO.__Error.WinIOError(Int32 errorCode, String maybeFullPath)
   bei System.IO.File.GetAttributes(String path)
   bei Duplicati.Library.Common.IO.SystemIOWindows.PathTooLongFuncWrapper[T](Func`2 nativeIOFunc, Func`2 alternativeIOFunc, String path, Boolean prefixWithUnc)
[...]

PR #4251, opened 2020-07-09
Exclude invalid Windows paths

dferreyra:
This change explicity excludes paths ending in a dot or a space from
being considered for back up in Windows and generates a clear
warning to the user.

Without this change, under Windows, directories ending in a dot or a
space can trigger a constraint violation in Duplicati when updating
the SQLite database and in that case will abort the entire backup
(see, for example, duplicati/duplicati#3518 and
duplicati/duplicati#4229).  Also, files ending in a dot or a space can
silently fail to be backed up at all.

In Windows, files and directories ending in a dot (".") or a space (" ")
are problematic.

For example,
https://docs.microsoft.com/en-gb/windows/win32/fileio/naming-a-file
says: "Do not end a file or directory name with a space or a
period. Although the underlying file system may support such names,
the Windows shell and user interface does not."

Also, some .NET path manipulation functions do not support trailing
dots and spaces.  For example, under Windows,
Path.GetFullPath(@"C:\dir \file.") will return @"C:\dir\file"; i.e.,
the trailing dot and space have been stripped.

Though they are problematic, directories and files ending in a dot or
a space can be created on Windows in a number of ways; e.g., by
various programs (like Firefox), by using the Windows command line,
and by using environments like Cygwin, and so can trip up Duplicati if
they are not explicitly excluded.


Duplicati/Library/Main/Operation/Backup/FileEnumerationProcess.cs
```diff
@@ -386,6 +392,45 @@ private static bool AttributeFilter(string path, FileAttributes attributes, Snap
...
+        /// For example, calling Path.GetFullPath() in Windows on files or
+        /// directories ending in a dot or a space will return a path without
+        /// the dot or the space.
+        /// </remarks>
+        /// <param name="path">The current path.</param>
+        /// <param name="attributes">The file or folder attributes.</param>
+        /// <returns>Returns true if file or directory ends in an invalid character for Windows.</returns>
+        private static bool IsInvalidWindowsPath(string path, FileAttributes attributes)
+        {
+            if ((attributes & FileAttributes.Directory) != 0)
```
warwickmm, line 425: Can we simplify this by using `attributes.HasFlag(FileAttributes.Directory)`?

ts678:
I'd note that such paths are legal from an NTFS point of view, and Duplicati can currently back them up -- sometimes.
The trick I used here was to use VSS, which I think works because it uses `\\?\` prefix which stops [path normalization](https://docs.microsoft.com/en-us/archive/blogs/jeremykuhne/path-normalization).
Without that, I "think" the pattern I found is that files in the oddly named folders work, but subfolders break Duplicati.
This is a very messy area, and has also had possible [behavioral changes](https://docs.microsoft.com/en-us/dotnet/framework/migration-guide/retargeting/4.6.1-4.6.2#changes-in-path-normalization) if one targets .NET Framework 4.6.2 as we did.

```
Listing contents 0 (7/11/2020 4:38:47 PM):
C:\tmp\sub1\ 
C:\tmp\sub1\subend \ 
C:\tmp\sub1\subend \dir_with_space (0 bytes)
C:\tmp\sub1\subend.\ 
C:\tmp\sub1\subend.\dir_without_space (0 bytes)
C:\tmp\sub1\subend\dir_withoutspace_subdir\ 
Return code: 0
```

warwickmm:
This also relates to issue #3839.

dferreyra:
@ts678, thanks for the information.  So this is much more complicated than I imagined!

So, reading those links, what I implemented goes too far, stopping Duplicati from backing up some paths that it can actually handle correctly.

I was originally running my tests using LINQPad, but it must be using something earlier than .NET Framework 4.6.2 because I was seeing the old behavior.

I just ran some tests with files and directories ending in dots and spaces, calling `Path.GetFullPath()`, `Path.GetDirectoryName()`, and `Path.GetFileName()`.  Compiled under .NET Framework 4.6.2, `Path.GetDirectoryName()` and `Path.GetFileName()` do fine with those endings.

Here are some of paths that don't work with `Path.GetFullPath()`.

Files ending in space or dots:
- `Path.GetFullPath(@"C:\dir\file ")`             returns `@"C:\dir\file"`
- `Path.GetFullPath(@"C:\dir\file.")`             returns `@"C:\dir\file"`
- `Path.GetFullPath(@"C:\dir\file..")`            returns `@"C:\dir\file"`
- `Path.GetFullPath(@"C:\dir\file...")`           returns `@"C:\dir\file"`

Directories ending in a _single_ dot:
- `Path.GetFullPath(@"C:\dir.\file")`             returns `@"C:\dir\file"`

On the other hand, a directory ending in two or three dots was fine:
- `Path.GetFullPath(@"C:\dir..\file")`            returns `@"C:\dir..\file"`
- `Path.GetFullPath(@"C:\dir...\file")`           returns `@"C:\dir...\file"`

Also, prefixing with "`\\?\`" gave good results for all three Path functions.  (For .NET Framework 4.5.2, on the other hand, calling `Path.GetFullPath()` with the "`\\?\`" prefix always threw an exception.) 

Would it be okay for me to enhance the `IsInvalidWindowsPath()` test to handle the cases where `Path.GetFullPath()` breaks?  For example, `IsInvalidWindowsPath()` would look for the "`\\?\`" prefix to avoid flagging those, it would flag _directories_ ending in a _single_ dot, etc.?

@ts678, I also saw the pattern of needing a subfolder to break Duplicati.  I assume that if I enhanced the `IsInvalidWindowsPath()` test to handle the cases where `Path.GetFullPath()` breaks, it would still stop Duplicati from backing up some files that didn't quite fall into the subfolder pattern.  I'm new to the Duplicati codebase, but I'd be happy to investigate this case further.

BTW, my impetus for wanting to resolve this is that I ran into the contraints violation problem and for a several days I had no backups.  I kept thinking it was a problem with the database, and kept doing a "Delete and Repair", but of course that didn't help (but my AWS bill was larger that month).  In playing with it, I also noticed that under some circumstances, files would not get backed up with a trailing dot, but that happened silently, so you wouldn't know it until you tried to restore.  These seem like bad outcomes, so it would be nice to at least give a warning for most of these cases.

Here's a Gist of the test code I wrote: https://gist.github.com/dferreyra/dfa8fcc404b3cc5321c39c67b2dfe665

dferreyra:
Trying out more test cases in Duplicati itself, while the native .NET functions allow directories with trailing spaces and two or more trailing dots, `AlphaFS.Directory.GetDirectories()` itself strips trailing dots and spaces from directories.  This is true even if you pass a UNC path to `AlphaFS.Directory.GetDirectories()`.

For example, on my system, calling `Directory.GetDirectories()` on the following three directories:
* `@"\\?\C:\Temp\storage\default\https+++www.example.com \"`
* `@"\\?\C:\Temp\storage\default\https+++www.example.com.\"`
* `@"\\?\C:\Temp\storage\default\https+++www.example.com..\"`

gives back the correct answers:
* `@"\\?\C:\Temp\storage\default\https+++www.example.com \cache"`
* `@"\\?\C:\Temp\storage\default\https+++www.example.com.\cache"`
* `@"\\?\C:\Temp\storage\default\https+++www.example.com..\cache"`

Calling `AlphaFS.Directory.GetDirectories()` returns this instead:
* `@"C:\Temp\storage\default\https+++www.example.com\cache"`
* `@"C:\Temp\storage\default\https+++www.example.com\cache"`
* `@"C:\Temp\storage\default\https+++www.example.com\cache"`

When gathering the list of sub-directories via `ListFolders()` in Windows, it looks like paths are unconditionally prefixed with the UNC string.  As long as the path is prefixed with the UNC string, `PathTooLongFuncWrapper()` will call `AlphaFS.Directory.GetDirectories()`.  With the trailing dots and spaces being stripped out by `AlphaFS.Directory.GetDirectories()`, we trigger the duplicate constraint violation.

I think that means that as long as AlphaFS behaves this way, the test for invalid files under Windows needs to say these are invalid, even if they contain the UNC prefix.  Also, it can't allow trailing spaces, and can't allow two or more trailing dots.

Given all this, I don't know how to improve on the validity test.

Duplicati/Library/Main/Operation/Backup/FileEnumerationProcess.cs
dferreyra: Yes.  I've pushed a change to use `HasFlag()`, thanks.

dferreyra:
One more thing.  I tried backing up the following files (without my changes in place):
* `@"C:\Temp\storage2\file.txt"`
* `@"C:\Temp\storage2\file.txt "`
* `@"C:\Temp\storage2\file.txt."`
* `@"C:\Temp\storage2\file.txt.."`

The contents of each of the files were different.  The backup completed without any warnings, but when I restored each of them in turn, I always got a file called `@"file.txt"`, and it always contained the contents of the basic `@"C:\Temp\storage2\file.txt"`.  So, though the other files appear in the list of files that can be restored, you can't actually restore them.

dferreyra:
In my experiment backing up files with trailing dots, looking at the filelist.json, the hashes are all the same, so the contents of the other files were not backed up:
```json
[
  {
    "type": "Folder",
    "path": "C:\\Temp\\storage2\\",
    "metahash": "GGJ1eLqHD1aigLvzUTsUMgOVs2KHgc0JvDSqXc9IvrA=",
    "metasize": 2111,
    "metablockhash": "GGJ1eLqHD1aigLvzUTsUMgOVs2KHgc0JvDSqXc9IvrA="
  },
  {
    "type": "File",
    "path": "C:\\Temp\\storage2\\file.txt",
    "hash": "X11YTFhX2Fr5Ea3hsq58tZPBdlQoIJHzrOMe/Z6VE2A=",
    "size": 7,
    "time": "20200705T135321Z",
    "metahash": "/qgYcfVp8wqaKwaL6modbSAdjtnxfscOgkCKvEHnuXU=",
    "metasize": 1776
  },
  {
    "type": "File",
    "path": "C:\\Temp\\storage2\\file.txt ",
    "hash": "X11YTFhX2Fr5Ea3hsq58tZPBdlQoIJHzrOMe/Z6VE2A=",
    "size": 7,
    "time": "20200705T135321Z",
    "metahash": "/qgYcfVp8wqaKwaL6modbSAdjtnxfscOgkCKvEHnuXU=",
    "metasize": 1776
  },
  {
    "type": "File",
    "path": "C:\\Temp\\storage2\\file.txt.",
    "hash": "X11YTFhX2Fr5Ea3hsq58tZPBdlQoIJHzrOMe/Z6VE2A=",
    "size": 7,
    "time": "20200705T135321Z",
    "metahash": "/qgYcfVp8wqaKwaL6modbSAdjtnxfscOgkCKvEHnuXU=",
    "metasize": 1776
  },
  {
    "type": "File",
    "path": "C:\\Temp\\storage2\\file.txt..",
    "hash": "X11YTFhX2Fr5Ea3hsq58tZPBdlQoIJHzrOMe/Z6VE2A=",
    "size": 7,
    "time": "20200705T135321Z",
    "metahash": "/qgYcfVp8wqaKwaL6modbSAdjtnxfscOgkCKvEHnuXU=",
    "metasize": 1776
  }
]
```

ts678:
Good point on restoring. Even though I got what looked like a good backup using VSS, probably due to the seeming use of the question mark form of the DOS device path (which technically isn't the same as a UNC path), the restore attempt got confused:

> Could not find a part of the path 'c:\backup restore\subend \dir_with_space'.

What actually got created was a `subend` folder without the trailing space. I'm trying to restore the whole tree, not bottom files, meaning I don't "think" this is the recently fixed bug where empty files might not always have parent paths created as required.

I took a look at what FileEnumerationProcess.cs has to say at Verbose level in its "IncludingPath" lines, and some later loggings:

IncludingPath might be a good initial view of what I think is System.IO.EnumerateFileSytemEntries (and if so, is hard to change).
CheckFileForChanges gives a somewhat better view of trailing spaces because there is further text to the right on the same line.
AddDirectory is there because these seem to be the problem somehow.
637300966260864791 is a timestamp to help locate the double-INSERT in profiling log with --profile-all-data-base-queries set.

```
2020-07-12 16:02:04 -04 - [Verbose-Duplicati.Library.Main.Operation.Backup.FileEnumerationProcess-IncludingPath]: Including path as no filters matched: C:\tmp\sub1\sub2\
2020-07-12 16:02:04 -04 - [Verbose-Duplicati.Library.Main.Operation.Backup.FileEnumerationProcess-IncludingPath]: Including path as no filters matched: C:\tmp\sub1\subend\
2020-07-12 16:02:04 -04 - [Verbose-Duplicati.Library.Main.Operation.Backup.FileEnumerationProcess-IncludingPath]: Including path as no filters matched: C:\tmp\sub1\subend \
2020-07-12 16:02:04 -04 - [Verbose-Duplicati.Library.Main.Operation.Backup.FileEnumerationProcess-IncludingPath]: Including path as no filters matched: C:\tmp\sub1\subend.\
2020-07-12 16:02:04 -04 - [Verbose-Duplicati.Library.Main.Operation.Backup.FileEnumerationProcess-IncludingPath]: Including path as no filters matched: C:\tmp\sub1\subend\dir_withoutspace_subdir\
2020-07-12 16:02:04 -04 - [Verbose-Duplicati.Library.Main.Operation.Backup.FileEnumerationProcess-IncludingPath]: Including path as no filters matched: C:\tmp\sub1\subend.\dir_without_space
2020-07-12 16:02:04 -04 - [Verbose-Duplicati.Library.Main.Operation.Backup.FileEnumerationProcess-IncludingPath]: Including path as no filters matched: C:\tmp\sub1\subend\dir_withoutspace_subdir\
2020-07-12 16:02:04 -04 - [Verbose-Duplicati.Library.Main.Operation.Backup.FileEnumerationProcess-IncludingPath]: Including path as no filters matched: C:\tmp\sub1\subend \dir_with_space
2020-07-12 16:02:04 -04 - [Verbose-Duplicati.Library.Main.Operation.Backup.FileEnumerationProcess-IncludingPath]: Including path as no filters matched: C:\tmp\sub1\subend\dir_withoutspace_subdir\
2020-07-12 16:02:04 -04 - [Verbose-Duplicati.Library.Main.Operation.Backup.FileEnumerationProcess-IncludingPath]: Including path as no filters matched: C:\tmp\sub1\subend\dir_without_space
2020-07-12 16:02:04 -04 - [Verbose-Duplicati.Library.Main.Operation.Backup.FileEnumerationProcess-IncludingPath]: Including path as no filters matched: C:\tmp\sub1\sub2\dir.txt
2020-07-12 16:02:04 -04 - [Verbose-Duplicati.Library.Main.Operation.Backup.FileEnumerationProcess-IncludingPath]: Including path as no filters matched: C:\tmp\sub1\sub2\ends_with_space.txt
2020-07-12 16:02:04 -04 - [Verbose-Duplicati.Library.Main.Operation.Backup.FileEnumerationProcess-IncludingPath]: Including path as no filters matched: C:\tmp\sub1\sub2\ends_with_space.txt 
2020-07-12 16:02:04 -04 - [Verbose-Duplicati.Library.Main.Operation.Backup.FileEnumerationProcess-IncludingPath]: Including path as no filters matched: C:\tmp\sub1\sub2\tree.txt
2020-07-12 16:02:04 -04 - [Verbose-Duplicati.Library.Main.Operation.Backup.FileEnumerationProcess-IncludingPath]: Including path as no filters matched: C:\tmp\sub1\sub2\
2020-07-12 16:02:04 -04 - [Verbose-Duplicati.Library.Main.Operation.Backup.FileEnumerationProcess-IncludingPath]: Including path as no filters matched: C:\tmp\sub1\subend\
2020-07-12 16:02:04 -04 - [Verbose-Duplicati.Library.Main.Operation.Backup.FileEnumerationProcess-IncludingPath]: Including path as no filters matched: C:\tmp\sub1\subend \
2020-07-12 16:02:04 -04 - [Verbose-Duplicati.Library.Main.Operation.Backup.MetadataPreProcess.FileEntry-AddDirectory]: Adding directory C:\tmp\sub1\
2020-07-12 16:02:04 -04 - [Verbose-Duplicati.Library.Main.Operation.Backup.FileEnumerationProcess-IncludingPath]: Including path as no filters matched: C:\tmp\sub1\subend.\
2020-07-12 16:02:04 -04 - [Verbose-Duplicati.Library.Main.Operation.Backup.FileEnumerationProcess-IncludingPath]: Including path as no filters matched: C:\tmp\sub1\subend\dir_withoutspace_subdir\
2020-07-12 16:02:04 -04 - [Verbose-Duplicati.Library.Main.Operation.Backup.FileEnumerationProcess-IncludingPath]: Including path as no filters matched: C:\tmp\sub1\subend.\dir_without_space
2020-07-12 16:02:04 -04 - [Verbose-Duplicati.Library.Main.Operation.Backup.MetadataPreProcess.FileEntry-AddDirectory]: Adding directory C:\tmp\sub1\subend.\
2020-07-12 16:02:04 -04 - [Profiling-Timer.Begin-Duplicati.Library.Main.Database.ExtensionMethods-ExecuteNonQuery]: Starting - ExecuteNonQuery: INSERT INTO "FilesetEntry" ("FilesetID", "FileID", "Lastmodified") VALUES (1, 2, 637300966260864791)
2020-07-12 16:02:04 -04 - [Profiling-Timer.Finished-Duplicati.Library.Main.Database.ExtensionMethods-ExecuteNonQuery]: ExecuteNonQuery: INSERT INTO "FilesetEntry" ("FilesetID", "FileID", "Lastmodified") VALUES (1, 2, 637300966260864791) took 0:00:00:00.000
2020-07-12 16:02:04 -04 - [Verbose-Duplicati.Library.Main.Operation.Backup.FilePreFilterProcess.FileEntry-CheckFileForChanges]: Checking file for changes C:\tmp\sub1\subend.\dir_without_space, new: True, timestamp changed: True, size changed: True, metadatachanged: True, 7/11/2020 6:31:48 PM vs 1/1/0001 12:00:00 AM
2020-07-12 16:02:04 -04 - [Verbose-Duplicati.Library.Main.Operation.Backup.MetadataPreProcess.FileEntry-AddDirectory]: Adding directory C:\tmp\sub1\subend\dir_withoutspace_subdir\
2020-07-12 16:02:04 -04 - [Profiling-Timer.Begin-Duplicati.Library.Main.Database.ExtensionMethods-ExecuteNonQuery]: Starting - ExecuteNonQuery: INSERT INTO "FilesetEntry" ("FilesetID", "FileID", "Lastmodified") VALUES (1, 4, 637300966260864791)
2020-07-12 16:02:04 -04 - [Profiling-Timer.Finished-Duplicati.Library.Main.Database.ExtensionMethods-ExecuteNonQuery]: ExecuteNonQuery: INSERT INTO "FilesetEntry" ("FilesetID", "FileID", "Lastmodified") VALUES (1, 4, 637300966260864791) took 0:00:00:00.000
2020-07-12 16:02:04 -04 - [Verbose-Duplicati.Library.Main.Operation.Backup.FileEnumerationProcess-IncludingPath]: Including path as no filters matched: C:\tmp\sub1\subend\dir_withoutspace_subdir\
2020-07-12 16:02:04 -04 - [Verbose-Duplicati.Library.Main.Operation.Backup.FileEnumerationProcess-IncludingPath]: Including path as no filters matched: C:\tmp\sub1\subend \dir_with_space
2020-07-12 16:02:04 -04 - [Verbose-Duplicati.Library.Main.Operation.Backup.MetadataPreProcess.FileEntry-AddDirectory]: Adding directory C:\tmp\sub1\subend \
2020-07-12 16:02:04 -04 - [Verbose-Duplicati.Library.Main.Operation.Backup.FilePreFilterProcess.FileEntry-CheckFileForChanges]: Checking file for changes C:\tmp\sub1\subend \dir_with_space, new: True, timestamp changed: True, size changed: True, metadatachanged: True, 7/11/2020 6:31:25 PM vs 1/1/0001 12:00:00 AM
2020-07-12 16:02:04 -04 - [Verbose-Duplicati.Library.Main.Operation.Backup.MetadataPreProcess.FileEntry-AddDirectory]: Adding directory C:\tmp\sub1\subend\dir_withoutspace_subdir\
2020-07-12 16:02:04 -04 - [Profiling-Timer.Begin-Duplicati.Library.Main.Database.ExtensionMethods-ExecuteNonQuery]: Starting - ExecuteNonQuery: INSERT INTO "FilesetEntry" ("FilesetID", "FileID", "Lastmodified") VALUES (1, 4, 637300966260864791)
2020-07-12 16:02:04 -04 - [Profiling-Timer.Finished-Duplicati.Library.Main.Database.ExtensionMethods-ExecuteNonQuery]: ExecuteNonQuery: INSERT INTO "FilesetEntry" ("FilesetID", "FileID", "Lastmodified") VALUES (1, 4, 637300966260864791) took 0:00:00:00.000
...
2020-07-12 16:02:04 -04 - [Error-Duplicati.Library.Main.Operation.BackupHandler-FatalError]: Fatal error
code = Constraint (19), message = System.Data.SQLite.SQLiteException (0x800027AF): constraint failed
UNIQUE constraint failed: FilesetEntry.FilesetID, FilesetEntry.FileID
   at System.Data.SQLite.SQLite3.Reset(SQLiteStatement stmt)
   at System.Data.SQLite.SQLite3.Step(SQLiteStatement stmt)
   at System.Data.SQLite.SQLiteDataReader.NextResult()
   at System.Data.SQLite.SQLiteDataReader..ctor(SQLiteCommand cmd, CommandBehavior behave)
   at System.Data.SQLite.SQLiteCommand.ExecuteReader(CommandBehavior behavior)
   at System.Data.SQLite.SQLiteCommand.ExecuteNonQuery(CommandBehavior behavior)
   at Duplicati.Library.Main.Database.ExtensionMethods.ExecuteNonQuery(IDbCommand self, Boolean writeLog, String cmd, Object[] values)
   at Duplicati.Library.Main.Database.ExtensionMethods.ExecuteNonQuery(IDbCommand self, Boolean writeLog)
   at Duplicati.Library.Main.Database.LocalBackupDatabase.AddFile(Int64 pathprefixid, String filename, DateTime lastmodified, Int64 blocksetID, Int64 metadataID, IDbTransaction transaction)
   at Duplicati.Library.Main.Database.LocalBackupDatabase.AddFile(String filename, DateTime lastmodified, Int64 blocksetID, Int64 metadataID, IDbTransaction transaction)
   at Duplicati.Library.Main.Database.LocalBackupDatabase.AddDirectoryEntry(String path, Int64 metadataID, DateTime lastmodified, IDbTransaction transaction)
   at Duplicati.Library.Main.Operation.Backup.BackupDatabase.<>c__DisplayClass8_0.<AddDirectoryEntryAsync>b__0()
   at Duplicati.Library.Main.Operation.Common.SingleRunner.<>c__DisplayClass3_0.<RunOnMain>b__0()
   at Duplicati.Library.Main.Operation.Common.SingleRunner.<DoRunOnMain>d__2`1.MoveNext()
```

`C:\tmp\sub1\subend\dir_withoutspace_subdir\` shows up way too many times, right from the start, and possibly is fatal when added twice. I'm not sure exactly how a directory walk is done, but if stripping happened before descending, it'd cause duplicates.

In my research to look into the detailed behavior of this, I might have heard that the MAX_PATH check is also suppressed by the `\\?\` syntax. If that is true, then it might be newly possible in .NET Framework 4.6.2 to skip the decision split where AlphaFS will handle long paths. Restores to local files could possibly take the same approach to enjoy long paths, without path normalization. Genuine SMB-remote backup sources do get used (but are not encouraged -- ideally Duplicati is on the system with source files)
and whatever is done needs to handle those. There are all kinds of other odd things theoretically possible. Some good resources:

[File path formats on Windows systems](https://docs.microsoft.com/en-us/dotnet/standard/io/file-path-formats)
[Jeremy Kuhne's Blog](https://docs.microsoft.com/en-us/archive/blogs/jeremykuhne/)

I'm not quite sure what to suggest. Other input would be helpful, though some key developers might not have Windows to try.
My own role is not so much a developer as a forum and GitHub Issues helper. I do try to fathom the Duplicati code sometimes.

dferreyra:
My thought with this pull request was to stop Duplicati from trying to back up files we know it has a problem with, and surface them as warnings to the user.

Getting better support in Duplicati for these files and directories would be ideal!  Maybe that could be a separate ticket (or an appropriate existing ticket)?  I'd be glad to help test and investigate various scenarios for this under Windows, including changing the code some clear ways to help out (e.g., "stop using AlphfaFS for handling long paths", etc.).

warwickmm:
Thanks @dferreyra, that would be great.  I think issue #3839 would be an appropriate place to continue the work on supporting these strange paths.

ts678:
I think the phased proposal makes sense, however I wonder if an option should be provided to allow currently maybe-bad behavior.

This could help end users who (maybe) have good backups using [snapshot-policy](https://duplicati.readthedocs.io/en/latest/06-advanced-options/#snapshot-policy) to use VSS, thus maybe get question-mark device path to GLOBALROOT. Boolean options are by convention false by default. There are people who know how to add them (not me...).

For developers and technical people, allowing current behavior would allow testing what works (and doesn't) without a custom build.

I poked a little, attempting to restore my VSS backup to a manually typed `\\?\` version of my usual destination.  Duplicati asked for:

```
2020-07-13 07:29:39 -04 - [Verbose-Duplicati.Library.Main.Operation.RestoreHandler-CreateFolder]: Creating folder: \\?\C:\backup restore\subend.\
2020-07-13 07:29:39 -04 - [Verbose-Duplicati.Library.Main.Operation.RestoreHandler-CreateFolder]: Creating folder: \\?\C:\backup restore\subend.\dir_withdot_subdir\
2020-07-13 07:29:39 -04 - [Verbose-Duplicati.Library.Main.Operation.RestoreHandler-CreateFolder]: Creating folder: \\?\C:\backup restore\subend \dir_withspace_subdir\
2020-07-13 07:29:39 -04 - [Verbose-Duplicati.Library.Main.Operation.RestoreHandler-CreateFolder]: Creating folder: \\?\C:\backup restore\subend\dir_withoutspace_subdir\
2020-07-13 07:29:39 -04 - [Verbose-Duplicati.Library.Main.Operation.RestoreHandler-CreateFolder]: Creating folder: \\?\C:\backup restore\sub2\
```

but watching with [Sysinternals Process Monitor](https://docs.microsoft.com/en-us/sysinternals/downloads/procmon) showed stripped versions going into the driver (I think). At the end of a restore, file hashes are checked, and I think what happened is it tried to open a trailing-space version of filename, actually got the non-trailing-space-version, and complained about the hash. It announces what it found and expected, and the DB has the expected one on the trailing-space version, thus my tentative conclusion...

[This week, we speak with Jeremy Kuhne about long path support in .NET.](https://channel9.msdn.com/Shows/On-NET/Jeremy-Kuhne-long-path-support) at the five-minute mark talked about (I think) .NET Core converting to the question-mark syntax before calling the Win32 API, so things just work. @mnaiman has been working on a [pull request](https://github.com/duplicati/duplicati/pull/3124) for this possible future version. and might have some thoughts on long paths and trailing spaces and dots. @verhoek has done some previous I/O changes, including some VSS and long path work. Not sure if you like help, but the names came to mind.

Thanks for helping fix this!

dferreyra:
@ts678, thanks again for all that!  As far as pitching in with helping get actual support for these funny paths, I would definitely need and accept guidance on what I might do to help.

As far as my proposed pull request, having an option sounds good.  Based on what you said, I've added this option:
> allow-possibly-invalid-windows-paths
> 
> Use this option to allow possibly invalid Windows paths to be considered for back up on Windows. This option may allow files with historically problematic paths (e.g., directories or files that end in a dot) to be backed up, but may also trigger problems in Duplicati; e.g., this may generate errors that abort a backup, or may result in a \"successful\" backup where files with the problematic path have not actually been backed up.
> 
> Default value: "false"

If this seems okay, I can test this out and push a commit with this addition.

ts678:
The approach sounds good. Should message also mention space? I think there's a short message too, but this is hard to explain.

[Constraint failed UNIQUE constraint failed: FilesetEntry.FilesetID, FilesetEntry.FileID](https://forum.duplicati.com/t/constraint-failed-unique-constraint-failed-filesetentry-filesetid-filesetentry-fileid/10123) was a recent forum case with a trailing space.
Fortunately the user recognized this issue, having run into it elsewhere. This particular Windows path-mangling was news to me.

Thanks!

dferreyra:
I've updated the long description to mention dot and space.  The short description is not very informative.

Short description:
> Allows possibly invalid Windows paths (Windows only)

Long description:
> Use this option to allow possibly invalid Windows paths to be considered for back up on Windows. This option may allow files with historically problematic paths (i.e., files or directories that end in a dot or a space) to be backed up, but may also trigger problems in Duplicati; e.g., this may generate errors that abort a backup, or may result in a \"successful\" backup where files with the problematic path have not actually been backed up.

dferreyra:
I've submitted pull request #4256 that attempts to fix the funny file handling.  If that works out, this pull request would not be needed.

Thanks @ts678 for all the pointers.  I didn't know about all the improvements with .NET 4.6.2.  Reading up on it and watching the Jeremy Kuhne interview was very informative.  I also didn't know about the special VSS handling in Duplicati.  Basically, there was a lot I didn't know about coming into this!

I checked out the current changes in #3124 to see if there were useful patterns established along the lines of what I was thinking of changing for #4256, but I didn't see any, so I forged ahead.

warwickmm:
Closing this in favor of pull request #4256, which adds support for problematic paths.

PR #4256, opened 2020-07-18
Support problematic Windows paths

### LINKED ISSUE #3839: Backup (or exclude) Files with only .dot Extention

dferreyra:
Allow historically problematic Windows paths (i.e., files or
directories that end in a dot or a space) to be backed up with
Duplicati in Windows.

This fixes #3839.

I've taken a stab at changing Duplicati to support files and directories that end in a dot or a space.

Just a quick recap of some things that weren't working in Duplicati under Windows as I tested out changes for #4251:
1. Paths that include directories that end in one or more spaces ("` `") or one or more dots ("`.`") can trigger a UNIQUE CONSTRAINT violation in Duplicati (and from reading #3839, other kinds of exceptions)
1. Files that end in one or more spaces ("` `") or one or more dots ("`.`") could appear to have been successfully backed up and you could even select the files during a restore, but the contents of those files never actually made it into the back up so they could not actually be restored

These kinds of files and directories trip up the AlphaFS library (and, historically, .NET).  For example, this call
```cs
AlphaFS.Directory.GetDirectories(@"\\?\C:\Temp\dir.")
```
can trigger an exception about the path not being found, or, if a path without the trailing dot exists (e.g., plain "`C:\Temp\dir`"), it can silently send AlphaFS down into the wrong directory.

.NET 4.6.2 has added support for using the "`\\?\`" prefix, so long paths can now be handled with the native .NET functions.  Plus the .NET 4.6.2 functions do not have problems with trailing dots and spaces.

I changed SystemIOWindows.cs to rely on the new behavior of .NET 4.6.2.

At this point, unit tests pass, I can back up combinations of directories and files that end in dots or spaces, I can continue to back up files with long paths, and everything successfully restores.  I've also tried snapshot (VSS) backups against locked files, and I've tried backups of network shares (e.g., "`\\example.com\share\file.txt`").

I would like to know what else I should look at to make sure things are working -- I don't have a feel for how comprehensive the unit tests are in that regard.  Also, I'm wondering if I should try creating unit tests for any of this.

As far as the source code changes themselves, I removed the uses of the `PathTooLong...()` wrappers.  For calls where the "`prefixWithUnc`" parameter was true, I replaced them with calls to the native .NET functions where the path parameters are passed through `PrefixWithUNC()`.  For example, this
```cs
public bool DirectoryExists(string path)
{
    return PathTooLongFuncWrapper(System.IO.Directory.Exists,
                                  Alphaleonis.Win32.Filesystem.Directory.Exists, path, true);
}
```
was replaced with
```cs
public bool DirectoryExists(string path)
{
    return System.IO.Directory.Exists(PrefixWithUNC(path));
}
```

For calls where the "`prefixWithUnc`" parameter was false, I still want to protect each path with a UNC prefix.  At the same time I want the return values to be UNC prefixed or not depending whether the given path was prefixed.  So, for example, this
```cs
public string GetPathRoot(string path)
{
    return PathTooLongFuncWrapper(Path.GetPathRoot, AlphaFS.Path.GetPathRoot, path, false);
}
```
was replaced with
```cs
public string GetPathRoot(string path)
{
    if (IsPrefixedWithUNC(path))
    {
        return Path.GetPathRoot(path);
    }
    else
    {
        return StripUNCPrefix(Path.GetPathRoot(PrefixWithUNC(path)));
    }
}
```

I also changed `StripUNCPrefix()` to correctly handle the "`\\?\UNC\`" case.

### DIFF
--- a/Duplicati/Library/Common/IO/SystemIOWindows.cs
+++ b/Duplicati/Library/Common/IO/SystemIOWindows.cs
@@ -53,12 +53,12 @@ namespace Duplicati.Library.Common.IO
 
         public static string StripUNCPrefix(string path)
         {
-            if (path.StartsWith(UNCPREFIX_SERVER))
+            if (path.StartsWith(UNCPREFIX_SERVER, StringComparison.Ordinal))
             {
                 // @"\\?\UNC\example.com\share\file.txt" to @"\\example.com\share\file.txt"
                 return PATHPREFIX_SERVER + path.Substring(UNCPREFIX_SERVER.Length);
             }
-            else if (path.StartsWith(UNCPREFIX))
+            else if (path.StartsWith(UNCPREFIX, StringComparison.Ordinal))
             {
                 // @"\\?\C:\file.txt" to @"C:\file.txt"
                 return path.Substring(UNCPREFIX.Length);
--- a/Duplicati/Library/Common/IO/SystemIOWindows.cs
+++ b/Duplicati/Library/Common/IO/SystemIOWindows.cs
@@ -34,28 +34,39 @@ namespace Duplicati.Library.Common.IO
         private const string PATHPREFIX_SERVER = @"\\";
         private static readonly string DIRSEP = Util.DirectorySeparatorString;
 
-        private static bool IsPathTooLong(string path)
-        {
-            // Use 258 for length check instead of 260 (MAX_PATH) - we need to leave room for the 16-bit (wide) null terminator
-            return path.StartsWith(UNCPREFIX, StringComparison.Ordinal) || path.StartsWith(UNCPREFIX_SERVER, StringComparison.Ordinal) || path.Length > 258;
-        }
-
         public static string PrefixWithUNC(string path)
         {
-            if (path.StartsWith(UNCPREFIX_SERVER, StringComparison.Ordinal))
+            if (IsPrefixedWithUNC(path))
+            {
                 return path;
-
-            if (path.StartsWith(UNCPREFIX, StringComparison.Ordinal))
-                return path;
-
+            }
             return path.StartsWith(PATHPREFIX_SERVER, StringComparison.Ordinal)
-                ? UNCPREFIX_SERVER + path.Remove(0, PATHPREFIX_SERVER.Length)
+                ? UNCPREFIX_SERVER + path.Substring(PATHPREFIX_SERVER.Length)
                 : UNCPREFIX + path;
+        }
+
+        private static bool IsPrefixedWithUNC(string path)
+        {
+            return path.StartsWith(UNCPREFIX_SERVER, StringComparison.Ordinal) ||
+                path.StartsWith(UNCPREFIX, StringComparison.Ordinal);
         }
 
         public static string StripUNCPrefix(string path)
         {
-            return path.StartsWith(UNCPREFIX, StringComparison.Ordinal) ? path.Substring(UNCPREFIX.Length) : path;
+            if (path.StartsWith(UNCPREFIX_SERVER))
+            {
+                // @"\\?\UNC\example.com\share\file.txt" to @"\\example.com\share\file.txt"
+                return PATHPREFIX_SERVER + path.Substring(UNCPREFIX_SERVER.Length);
+            }
+            else if (path.StartsWith(UNCPREFIX))
+            {
+                // @"\\?\C:\file.txt" to @"C:\file.txt"
+                return path.Substring(UNCPREFIX.Length);
+            }
+            else
+            {
+                return path;
+            }
         }
 
         private class FileSystemAccess
@@ -139,135 +150,90 @@ namespace Duplicati.Library.Common.IO
 
         private System.Security.AccessControl.FileSystemSecurity GetAccessControlDir(string path)
         {
-            return PathTooLongFuncWrapper(System.IO.Directory.GetAccessControl,
-                                          Alphaleonis.Win32.Filesystem.Directory.GetAccessControl, path, true);
+            return System.IO.Directory.GetAccessControl(PrefixWithUNC(path));
         }
 
         private System.Security.AccessControl.FileSystemSecurity GetAccessControlFile(string path)
         {
-            return PathTooLongFuncWrapper(System.IO.File.GetAccessControl,
-                                          Alphaleonis.Win32.Filesystem.File.GetAccessControl, path, true);
+            return System.IO.File.GetAccessControl(PrefixWithUNC(path));
         }
 
         private void SetAccessControlFile(string path, FileSecurity rules)
         {
-            PathTooLongActionWrapper(p => System.IO.File.SetAccessControl(p, rules),
-                                     p => Alphaleonis.Win32.Filesystem.File.SetAccessControl(p, rules, AccessControlSections.All),
-                                     path, true);
+            System.IO.File.SetAccessControl(PrefixWithUNC(path), rules);
         }
 
         private void SetAccessControlDir(string path, DirectorySecurity rules)
         {
-            PathTooLongActionWrapper(p => System.IO.Directory.SetAccessControl(p, rules),
-                                     p => Alphaleonis.Win32.Filesystem.Directory.SetAccessControl(p, rules, AccessControlSections.All),
-                                     path, true);
-        }
-
-        private static void PathTooLongVoidFuncWrapper<U, T>(Func<string, T> nativeIOFunc,
-                                                    Func<string, U> alternativeIOFunc,
-                                                    string path, bool prefixWithUnc = false)
-        {
-            // Wrap void into bool return type to avoid code duplication. Code at anytime available for replacement.
-            PathTooLongFuncWrapper(p => { nativeIOFunc(p); return true; }, p => { alternativeIOFunc(p); return true; } , path, prefixWithUnc);
-        }
-
-        private static void PathTooLongActionWrapper(Action<string> nativeIOFunc,
-                                                     Action<string> alternativeIOFunc,
-                                                     string path, bool prefixWithUnc = false)
-        {
-            // Wrap void into bool return type to avoid code duplication. Code at anytime available for replacement.
-            PathTooLongFuncWrapper(p => { nativeIOFunc(p); return true; }, p => { alternativeIOFunc(p); return true; }, path, prefixWithUnc);
-        }
-
-        private static T PathTooLongFuncWrapper<T>(Func<string, T> nativeIOFunc,
-                                                   Func<string, T> alternativeIOFunc,
-                                                   string path, bool prefixWithUnc = false)
-        {
-            if (!IsPathTooLong(path))
-                try { return nativeIOFunc(path); }
-                catch (System.IO.PathTooLongException) { }
-                catch (System.ArgumentException) { }
-
-            return !prefixWithUnc ? alternativeIOFunc(path) : alternativeIOFunc(PrefixWithUNC(path));
+            System.IO.Directory.SetAccessControl(PrefixWithUNC(path), rules);
         }
 
         #region ISystemIO implementation
         public void DirectoryCreate(string path)
         {
-            PathTooLongVoidFuncWrapper(System.IO.Directory.CreateDirectory,
-                                       Alphaleonis.Win32.Filesystem.Directory.CreateDirectory, path, true);
+            System.IO.Directory.CreateDirectory(PrefixWithUNC(path));
         }
 
         public bool DirectoryExists(string path)
         {
-            return PathTooLongFuncWrapper(System.IO.Directory.Exists,
-                                          Alphaleonis.Win32.Filesystem.Directory.Exists, path, true);
+            return System.IO.Directory.Exists(PrefixWithUNC(path));
         }
 
         public void FileDelete(string path)
         {
-            PathTooLongActionWrapper(System.IO.File.Delete,
-                                     Alphaleonis.Win32.Filesystem.File.Delete, path, true);
+            System.IO.File.Delete(PrefixWithUNC(path));
         }
 
         public void FileSetLastWriteTimeUtc(string path, DateTime time)
         {
-            PathTooLongActionWrapper(p => System.IO.File.SetLastWriteTimeUtc(p, time),
-                                     p => Alphaleonis.Win32.Filesystem.File.SetLastWriteTimeUtc(p, time), path, true);
+            System.IO.File.SetLastWriteTimeUtc(PrefixWithUNC(path), time);
         }
 
         public void FileSetCreationTimeUtc(string path, DateTime time)
         {
-            PathTooLongActionWrapper(p => System.IO.File.SetCreationTimeUtc(p, time),
-                                     p => Alphaleonis.Win32.Filesystem.File.SetCreationTimeUtc(p, time), path, true);
+            System.IO.File.SetCreationTimeUtc(PrefixWithUNC(path), time);
         }
 
         public DateTime FileGetLastWriteTimeUtc(string path)
         {
-            return PathTooLongFuncWrapper(System.IO.File.GetLastWriteTimeUtc,
-                                          Alphaleonis.Win32.Filesystem.File.GetLastWriteTimeUtc, path, true);
+            return System.IO.File.GetLastWriteTimeUtc(PrefixWithUNC(path));
         }
 
         public DateTime FileGetCreationTimeUtc(string path)
         {
-            return PathTooLongFuncWrapper(System.IO.File.GetCreationTimeUtc,
-                                          Alphaleonis.Win32.Filesystem.File.GetCreationTimeUtc, path, true);
+            return System.IO.File.GetCreationTimeUtc(PrefixWithUNC(path));
         }
 
         public bool FileExists(string path)
         {
-            return PathTooLongFuncWrapper(System.IO.File.Exists,
-                                          Alphaleonis.Win32.Filesystem.File.Exists, path, true);
+            return System.IO.File.Exists(PrefixWithUNC(path));
         }
 
         public System.IO.FileStream FileOpenRead(string path)
         {
-            return PathTooLongFuncWrapper(p => System.IO.File.Open(p, System.IO.FileMode.Open, System.IO.FileAccess.Read, System.IO.FileShare.ReadWrite),
-                                   p => Alphaleonis.Win32.Filesystem.File.Open(p, System.IO.FileMode.Open, System.IO.FileAccess.Read, System.IO.FileShare.ReadWrite),
-                                   path, true);
+            return System.IO.File.Open(PrefixWithUNC(path), System.IO.FileMode.Open, System.IO.FileAccess.Read, System.IO.FileShare.ReadWrite);
         }
 
         public System.IO.FileStream FileOpenWrite(string path)
         {
             return !FileExists(path)
                 ? FileCreate(path)
-                : PathTooLongFuncWrapper(System.IO.File.OpenWrite, Alphaleonis.Win32.Filesystem.File.OpenWrite, path, true);
+                : System.IO.File.OpenWrite(PrefixWithUNC(path));
         }
 
         public System.IO.FileStream FileCreate(string path)
         {
-            return PathTooLongFuncWrapper(System.IO.File.Create, Alphaleonis.Win32.Filesystem.File.Create, path, true);
+            return System.IO.File.Create(PrefixWithUNC(path));
         }
 
         public System.IO.FileAttributes GetFileAttributes(string path)
         {
-            return PathTooLongFuncWrapper(System.IO.File.GetAttributes, AlphaFS.File.GetAttributes, path, true);
+            return System.IO.File.GetAttributes(PrefixWithUNC(path));
         }
 
         public void SetFileAttributes(string path, System.IO.FileAttributes attributes)
         {
-            PathTooLongActionWrapper(p => System.IO.File.SetAttributes(p, attributes),
-                                      p => Alphaleonis.Win32.Filesystem.File.SetAttributes(p, attributes), path, true);
+            System.IO.File.SetAttributes(PrefixWithUNC(path), attributes);
         }
 
         /// <summary>
@@ -279,7 +245,7 @@ namespace Duplicati.Library.Common.IO
         {
             try
             {
-                return PathTooLongFuncWrapper(AlphaFS.File.GetLinkTargetInfo, AlphaFS.File.GetLinkTargetInfo, file, true).PrintName;
+                return AlphaFS.File.GetLinkTargetInfo(PrefixWithUNC(file)).PrintName;
             }
             catch (AlphaFS.NotAReparsePointException) { }
             catch (AlphaFS.UnrecognizedReparsePointException) { }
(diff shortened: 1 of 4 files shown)


ts678:
This is wonderful to see. Thanks! There's long been a wish to reduce reliance on libraries such as AlphaFS, and the move to net462 even had [long path support in mind](https://github.com/duplicati/duplicati/pull/3844#issuecomment-525843997), as did the current [Net 5 migration](https://github.com/duplicati/duplicati/pull/3124) PR (that previously had .NET Core or Net Standard plans).

Someone else will need to look at the code details, but I can give some ideas on making sure that things are still working with this.

[Duplicati.CommandLine.RecoveryTool.exe](https://duplicati.readthedocs.io/en/latest/07-other-command-line-utilities/#duplicaticommandlinerecoverytoolexe) [targetfile](https://github.com/duplicati/duplicati/blob/45da5a4e2bb2a6ee964e7c6ceb19a0e3b8b7a550/Duplicati/CommandLine/RecoveryTool/Restore.cs#L160) uses may need changing. I'm not sure how well they handle long or odd paths. This tool is occasionally needed for [disaster recovery](https://duplicati.readthedocs.io/en/latest/08-disaster-recovery/#restoring-files-using-the-recovery-tool) for backups that are too broken for usual restore, so should handle all paths.

[restore_from_python.py](https://github.com/duplicati/duplicati/blob/master/Tools/Commandline/RestoreFromPython/README) gets little to no usage, but is sort of comforting to know exists because it has so few dependencies. I think there's a report it doesn't work (due to library drift or something), but I'm not sure how [this](https://github.com/duplicati/duplicati/blob/45da5a4e2bb2a6ee964e7c6ceb19a0e3b8b7a550/Tools/Commandline/RestoreFromPython/restore_from_python.py#L89) handles long or odd paths in backups.

Presumably the Web UI for both backup and restore selection does OK even if actually seeing trailing space presence isn't possible.
[The FIND command](https://duplicati.readthedocs.io/en/latest/04-using-duplicati-from-the-command-line/#the-find-command) is in a similar position, but can at least more easily write into a file where one can look at it with better editors.

Restore should work with or without setting [no-local-blocks](https://duplicati.readthedocs.io/en/latest/06-advanced-options/#no-local-blocks) true. True stops an optimization which get blocks from original source. Verbose log looks like it will announce [here](https://github.com/duplicati/duplicati/blob/45da5a4e2bb2a6ee964e7c6ceb19a0e3b8b7a550/Duplicati/Library/Main/Operation/RestoreHandler.cs#L643-L646) whether Duplicati was able to find some local file blocks to patch into that restored file.

Compatibility with previous backups may be hard to test. Old backups might have recorded the same actual tree in a different way. New code potentially makes old files "disappear" and new ones "appear" in their true places. I guess it's an improvement if it works.

Although it might be a temporary situation if above "fixup-on-backup" happens, ideally old backups should restore with new code, winding up no worse than trying to restore old backup with the old code. I consider it less important than continuing old backups.

warwickmm:
Thanks @dferreyra!  I added some tests in #4257 to help us verify the behavior.  Currently, this tests backup and restore operations for the following paths:
* `C:\projects\duplicati\testdata\backup-data\xxxxxxxxxx\yyyyyyyyyyyyyyyyyyyyyyyyyyyyyyyyyyyyyyyyyyyyyyyyyyyyyyyyyyyyyyyyyyyyyyyyyyyyyyyyyyyyyyyyyyyyyyyyyyyyyyyyyyyyyyyyyyyyyyyyyyyyyyyyyyyyyyyyyyyyyyyyyyyyyyyyyyyyyyyyyyyyyyyyyyyyyyyyyyyyyyyyyyyyyyyyyyyyyyyyyyyyyyyyyyyyyyyyyyyyyyyyyyyyyyyyyyyyyyyyyyyyyyyyyyyyyyy`
* `C:\projects\duplicati\testdata\backup-data\ends_with_dot.\ends_with_dot.`
* `C:\projects\duplicati\testdata\backup-data\ends_with_dots..\ends_with_dots..`
* `C:\projects\duplicati\testdata\backup-data\ends_with_space \ends_with_space ` (this has a space at the end of the folders)
* `C:\projects\duplicati\testdata\backup-data\ends_with_spaces  \ends_with_spaces  ` (this has two spaces at the end of the folders)

I have the same concern that @ts678 mentions above.  I don't have access to a Windows machine, so can someone try the following to see what happens?
1. Checkout the `master` branch and run a backup.
2. Apply the changes in this pull request and run the backup again.

Are the paths to the source directory and individual files represented differently?  If so, what are the consequences?

dferreyra:
@warwickmm, I don't know if your intention is to have someone besides me try this (i.e., another pair of eyes), but I'll try the experiment and report back.  I'll also apply your unit tests and see how they fare.

@ts678, thanks for that list of things to look into.  I'll start to take a look at them after I try the experiment suggested by @warwickmm,

warwickmm:
Thanks. The new tests are passing in the automated AppVeyor builds, but it would be good to see if it passes in a different Windows environment as well.

dferreyra:
@warwickmm, your unit tests pass for me, too.

As far as your `master` vs. pull request experiment, here's what I tried:
1. Set up some funny paths and some normal paths
1. Backed it all up with `master` (I'll call it the "master backup")
1. Backed it up again with the pull request applied (I'll call it the "pr backup")
1. Still with the pull request applied, restored the "master backup"
1. Still with the pull request applied, restored the "pr backup"

I had to craft the funny directories to avoid the unique constraint violation because that will abort the backup.  Here's what my problematic directories looked like (I'm using Cygwin here):
```
$ find problematic-directories2 -type f -print0 | xargs -0 ls -1
'problematic-directories2/default/https+++www.example.com /cache/caches.sqlite'
```

Here's what my problematic files looked like:
```
$ find problematic-files -type f -print0 | xargs -0 ls -1
problematic-files/file.txt
'problematic-files/file.txt '
problematic-files/file.txt.
problematic-files/file.txt..
```

The "master backup" gave warnings for the problematic paths:
```
2020-07-18 21:34:10 -07 - [Warning-Duplicati.Library.Main.Operation.Backup.FileEnumerationProcess-FileAccessError]: Error reported while accessing file: C:\Temp\duplicati-tests\3839\problematic-directories2\default\https+++www.example.com \
2020-07-18 21:34:10 -07 - [Warning-Duplicati.Library.Main.Operation.Backup.MetadataGenerator.Metadata-MetadataProcessFailed]: Failed to process metadata for "C:\Temp\duplicati-tests\3839\problematic-directories2\default\https+++www.example.com \", storing empty metadata
```
Looking in the generated "filelist.json", I can confirm that nothing below that problematic directory was stored, but there was an entry for the directory:
```json
  {
    "type": "Folder",
    "path": "C:\\Temp\\duplicati-tests\\3839\\problematic-directories2\\default\\https+++www.example.com \\",
    "metahash": "qiXpeARtaA74dA2Dfm3lvB4qLcYInb2hASVEtTjVP2U=",
    "metasize": 5,
    "metablockhash": "qiXpeARtaA74dA2Dfm3lvB4qLcYInb2hASVEtTjVP2U="
  },
```

As usual, the "master backup" got no warnings for the problematic files.  Looking in the generated "filelist.json", I can see that, though each file has different content, only the content of "file.txt" was stored -- note the identical hashes and metadata:
```json
  {
    "type": "File",
    "path": "C:\\Temp\\duplicati-tests\\3839\\problematic-files\\file.txt",
    "hash": "GuTNUyFUCcc4uQFJNIM2inZSSmK4hGuB5AiIK6nwD7s=",
    "size": 38,
    "time": "20200718T043251Z",
    "metahash": "XS0XhoaxbMfiLAboYOO+BRh9r3QrPm1jfRkSYHKzk4I=",
    "metasize": 1776
  },
  {
    "type": "File",
    "path": "C:\\Temp\\duplicati-tests\\3839\\problematic-files\\file.txt ",
    "hash": "GuTNUyFUCcc4uQFJNIM2inZSSmK4hGuB5AiIK6nwD7s=",
    "size": 38,
    "time": "20200718T043251Z",
    "metahash": "XS0XhoaxbMfiLAboYOO+BRh9r3QrPm1jfRkSYHKzk4I=",
    "metasize": 1776
  },
  {
    "type": "File",
    "path": "C:\\Temp\\duplicati-tests\\3839\\problematic-files\\file.txt.",
    "hash": "GuTNUyFUCcc4uQFJNIM2inZSSmK4hGuB5AiIK6nwD7s=",
    "size": 38,
    "time": "20200718T043251Z",
    "metahash": "XS0XhoaxbMfiLAboYOO+BRh9r3QrPm1jfRkSYHKzk4I=",
    "metasize": 1776
  },
  {
    "type": "File",
    "path": "C:\\Temp\\duplicati-tests\\3839\\problematic-files\\file.txt..",
    "hash": "GuTNUyFUCcc4uQFJNIM2inZSSmK4hGuB5AiIK6nwD7s=",
    "size": 38,
    "time": "20200718T043251Z",
    "metahash": "XS0XhoaxbMfiLAboYOO+BRh9r3QrPm1jfRkSYHKzk4I=",
    "metasize": 1776
  }
```
The code is stripping the trailing characters and incorrectly picking up the plain "file.txt" file.  (If the plain "file.txt" file is not there, the "master backup" gives warnings for each of the problematic files and in that case, none of the problematic files appear in the backup data.)

In the "pr backup", the directories and files under the problematic directory now appear, and the entry for the problematic directory itself has changed:
```json
  {
    "type": "Folder",
    "path": "C:\\Temp\\duplicati-tests\\3839\\problematic-directories2\\default\\https+++www.example.com \\",
    "metahash": "5YRk2A8XIbNtMUag8wXE4FadH7rYos2barALy1sf6jg=",
    "metasize": 2223,
    "metablockhash": "5YRk2A8XIbNtMUag8wXE4FadH7rYos2barALy1sf6jg="
  },
```

For the problematic files, the contents of the files are now backed up correctly, and you can see the different hashes and different metadata:
```json
  {
    "type": "File",
    "path": "C:\\Temp\\duplicati-tests\\3839\\problematic-files\\file.txt",
    "hash": "GuTNUyFUCcc4uQFJNIM2inZSSmK4hGuB5AiIK6nwD7s=",
    "size": 38,
    "time": "20200718T043251Z",
    "metahash": "XS0XhoaxbMfiLAboYOO+BRh9r3QrPm1jfRkSYHKzk4I=",
    "metasize": 1776
  },
  {
    "type": "File",
    "path": "C:\\Temp\\duplicati-tests\\3839\\problematic-files\\file.txt ",
    "hash": "zJ5DcOls2VSz7Qo+NSKAfjf1GzFPwiPDrbyEOEePqhU=",
    "size": 21,
    "time": "20200715T204404Z",
    "metahash": "iDqH7hEcFa6PiICKJSch7hWdGDTSko3DydPpXdXSEVQ=",
    "metasize": 1776
  },
  {
    "type": "File",
    "path": "C:\\Temp\\duplicati-tests\\3839\\problematic-files\\file.txt.",
    "hash": "8yFYYvG7hbYzMy2k0lxlOhIjNUMHnZ6o/2kigV4n108=",
    "size": 24,
    "time": "20200718T043113Z",
    "metahash": "xAF7EW8aSEfSch43n6MjxXPDD2D6OYqb+lePr7akzuw=",
    "metasize": 1776
  },
  {
    "type": "File",
    "path": "C:\\Temp\\duplicati-tests\\3839\\problematic-files\\file.txt..",
    "hash": "OMhs+iI6aNr2UtBQtedgw0QBidmHgEMhlTVflE9CdWA=",
    "size": 28,
    "time": "20200715T204439Z",
    "metahash": "MOo6fYt+TNSU2eVT94WDEQc8/t/n22Z6GYTwYVe6TNo=",
    "metasize": 1776
  }
```

Given all this, my restores were somewhat predictable.  Using the pull request build, the restore of the "master backup" gave me the problematic directory, with the trailing space, but nothing below it.  All the problematic files were restored, but they were incorrect: The files with the trailing dots and spaces were all just identical copies of "file.txt".

Still using the pull request build, the restore of the "pr backup" restored everything correctly.

As far as the normal, non-problematic directories and files I included, looking in the filelist.json for the two backups, everything was identical.

I hope this is along the lines of what you were looking for.  Let me know if you want me to try something different.

dferreyra:
@warwickmm, I meant to add that I also had several files with long paths in the backups.  The contents of "filelist.json" was identical for them in both backups, and both restores worked and gave identical results.

dferreyra:
@ts678, I looked at Duplicati.CommandLine.RecoveryTool.exe and I've pushed a change to support restoring problematic paths.  Since it used the native .NET functions directly, it must never have worked with files with long paths (or, of course, paths with trailing dots or spaces).

I went through the Duplicati.CommandLine.RecoveryTool.exe `restore` code and changed it to use ISystemIO, instead.  That allowed me to restore files with long paths and files and directories with trailing dots and spaces.  You can even pass in a long path for the `--targetpath` parameter itself (with or without the "`\\?\`" prefix) and the restore will work.

I didn't change any of the path handling dealing with the Duplicati backup files themselves or temporary files, just code that deals directly with restoring backed up files.  So you can't, for example, pass in a "`<localpath>`" parameter that is itself a long path (or a path with trailing dots or spaces).  Supporting this would take more work.  For example, `Path.SetCurrentDirectory()` does not support long paths even in .NET 4.6.2 as far as I can tell, and you might need to add more functions to ISystemIO to cover the APIs that are being used.

dferreyra:
@ts678, I tried restores with "`--no-local-blocks`" false and true and confirmed in the logs that it was or wasn't using local data as expected.  In both cases, my restores of problematic and non-problematic files were identical, so I think we're good there.

warwickmm:
Thanks @dferreyra.  This is looking very promising!

Duplicati/Library/Common/IO/SystemIOWindows.cs
```diff
@@ -34,28 +34,39 @@ public struct SystemIOWindows : ISystemIO
...
+        private static bool IsPrefixedWithUNC(string path)
+        {
+            return path.StartsWith(UNCPREFIX_SERVER, StringComparison.Ordinal) ||
+                path.StartsWith(UNCPREFIX, StringComparison.Ordinal);
+        }
+
         public static string StripUNCPrefix(string path)
         {
-            return path.StartsWith(UNCPREFIX, StringComparison.Ordinal) ? path.Substring(UNCPREFIX.Length) : path;
+            if (path.StartsWith(UNCPREFIX_SERVER))
```
warwickmm, line 56: I think it would be better to specify `StringComparison.Ordinal` as the comparison type.

Duplicati/Library/Common/IO/SystemIOWindows.cs
```diff
@@ -34,28 +34,39 @@ public struct SystemIOWindows : ISystemIO
...
+            {
+                // @"\\?\UNC\example.com\share\file.txt" to @"\\example.com\share\file.txt"
+                return PATHPREFIX_SERVER + path.Substring(UNCPREFIX_SERVER.Length);
+            }
+            else if (path.StartsWith(UNCPREFIX))
```
warwickmm, line 61: I think it would be better to specify `StringComparison.Ordinal` as the comparison type.

Duplicati/Library/Snapshots/WindowsSnapshot.cs
```diff
@@ -285,15 +265,8 @@ public override string ConvertToSnapshotPath(string localPath)
...
-
-            // Note: Do NOT use Path.Combine as it strips the UNC path prefix
             var subPath = localPath.Substring(root.Length);
-            if (!subPath.StartsWith(Util.DirectorySeparatorString, StringComparison.Ordinal))
-            {
-                volumePath = Util.AppendDirSeparator(volumePath, Util.DirectorySeparatorString);
-            }
-
-            var mappedPath = volumePath + subPath;
+            var mappedPath = Path.Combine(volumePath, subPath);
```
warwickmm, line 269: Would it be safer to use `SystemIO.IO_WIN.PathCombine` instead of `Path.Combine` here?

Duplicati/Library/Common/IO/SystemIOWindows.cs
dferreyra: Thanks.  I've pushed a fix.

Duplicati/Library/Common/IO/SystemIOWindows.cs
dferreyra: Thanks.  I've pushed a fix.

Duplicati/Library/Snapshots/WindowsSnapshot.cs
dferreyra: Thanks.  I've pushed a fix.

warwickmm:
Thanks.  The code looks reasonable to me.  My ability to test this is limited, so I'll give others a chance to review and suggest additional tests to perform before we merge.  If there are any suggestions for the automated tests in #4257, let me know.

/cc @drwtsn32x

dferreyra:
I tried running the Python script, [restore_from_python.py](https://github.com/duplicati/duplicati/blob/master/Tools/Commandline/RestoreFromPython/README).  As long as I feed it a destination path that contains the UNC prefix, it is able to restore long paths and paths that end in dots and spaces.

I did make the Windows Registry change to support long paths; i.e., ["Removing the MAX_PATH Limitation"](https://docs.python.org/3/using/windows.html#removing-the-max-path-limitation).

An example run, note the UNC prefix:
```
C:\Temp>python restore_from_python.py
Welcome to Python Duplicati recovery.
Please type the full path to a directory with Duplicati's .aes or .zip files:C:\Temp\duplicati-tests\3839\backup-files
Please type * to restore all files, or a pattern like /path/to/files/* to restore the files in a certain directory)*
Please enter the path to an empty destination directory:\\?\C:\Temp\duplicati-tests\3839\restores\python3
note: paths on windows have limited length, you might want to consider a shorter output path.
using duplicati-20200718T051302Z.dlist.zip which looks like the most recent dlist.
Creating index, this may take some time...
......Restoring files...


0 warnings/errors seen.
Complete.
```

dferreyra:
I have one more thought.  On a `master` build, if I manually enter a backup path that has the "`\\?\`" prefix, a couple of things aren't quite right.

I have an additional small commit I can make against this pull request that fixes the quirks.  I've tested it against snapshot and non-snapshot backups, using the GUI and the Duplicati.CommandLine, and I've tested restoring with the Duplicati.CommandLine.RecoveryTool.

I was wondering if I should push this change as well?  I ask because it might be considered a slight tangent.

As far as what my commit fixes, let's say I've specified that I want to back up "`\\?\C:\Temp\`":
1. The resulting Duplicati backup data will contain an entry for the path "`\\?\C:\Temp\`", but the entries for subdirectories and files will not include the "`\\?\`" prefix; e.g., you might see "`C:\Temp\file.txt`".
1. When you go to restore via the GUI, you see entries for "`\\?\C:\Temp\`" and "`C:\Temp\`" in the file picker
1. Again in the GUI, if try to expand one of the entries in the file picker with the "`\\?\`" prefix, you get this error:
```
```

The commit changes `ListFiles()` and `ListFolders()` in NoSnapshotWindows.cs to not unconditionally strip UNC prefixes from the results.  It also changes the `FilterExpression` class in FilterExpression.cs to stop identifying the question mark in the "`\\?\`" prefix as a wildcard in Windows.

warwickmm:
Thanks @dferreyra for finding and fixing more issues!  I'd prefer that we work on supporting source paths that begin with `\\?\` in a separate pull request.  That feels like a different issue than what is being covered here, and may require different testing and other considerations.

dferreyra:
@warwickmm, sounds good, thanks.

warwickmm:
@ts678, can you think of any additional tests that we should run here?

dferreyra:
@warwickmm, @ts678, I have not tested symbolic links under Windows with these changes.  I wasn't sure if that might be affected.  I can take a look at that tonight.

ts678:
> can you think of any additional tests that we should run here?

Can I assume that unit tests have some filter (include/exclude) testing? I sort of assumed so, and also sort of assumed that question mark device paths would get converted back to standard form that starts with drive letter (but minus annoying path normalization).

I'd also like to know that non-NTFS things that I guess are accessed as paths that start in the regular drive filesystem still work, but what to test may depend on what's available. Personal tools like Google Backup and Sync actually make offline copies, but business variants such as Google's [Drive File Stream](https://support.google.com/a/answer/7491144?hl=en) or Microsoft [OneDrive Files On-Demand](https://support.microsoft.com/en-us/office/save-disk-space-with-onedrive-files-on-demand-for-windows-10-0e6860d3-d9f3-4971-b321-7092438fb38e) may be a better test because real-time remote access may occur. I _think_ SMB was already tested, but it uses a special path syntax whereas these other filesystems use normal one.

I think some people have used [FUSE](https://en.wikipedia.org/wiki/Filesystem_in_Userspace) software to feed Duplicati source files... We can't test all of these, but it'd be nice to smoke-test something special that's stitched into usual path space. I'm not familiar enough with Windows' internals to know for sure that we're only bypassing the annoying edits by going in a different way. The .NET Core plan of doing it invisibly does make me think it'll work.

dferreyra:
Just FYI, I did some basic symbolic link tests -- regular NTFS "SYMLINK" and "SYMLINKD" links, not the fancy ones @ts678 mentioned.

Things seem to work correctly.  I used the UI to back up, and the UI and Duplicati.CommandLine.exe to restore.  I tried snapshot and no snapshot, and `--no-local-blocks` true and false.  Even symbolic links that have the problematic names are restored, but I do get these warnings because AlphaFS is still involved:
```
[Warning-Duplicati.Library.Main.Operation.RestoreHandler-MetadataWriteFailed]: Failed to apply metadata to file: "C:\Temp\duplicati-tests\3839\restores\symlinks\WinMetadata-symlink-relative.\", message: (2) The system cannot find the file specified: [C:\Temp\duplicati-tests\3839\restores\symlinks\WinMetadata-symlink-relative.]
System.IO.FileNotFoundException: (2) The system cannot find the file specified: [C:\Temp\duplicati-tests\3839\restores\symlinks\WinMetadata-symlink-relative.]
```

warwickmm:
> Can I assume that unit tests have some filter (include/exclude) testing? I sort of assumed so, and also sort of assumed that question mark device paths would get converted back to standard form that starts with drive letter (but minus annoying path normalization).

There are some simple unit tests, but the coverage isn't great.  Let us know if you have a specific test in mind and perhaps someone can try running it.  If it's something that can be easily automated, I can look into adding a unit test.

warwickmm:
I added a test to pull request #4257 for excluding problematic paths (long paths, file and folders that end with dots and spaces) and it passes in both AppVeyor (Windows) and Travis (Linux).

warwickmm:
Anything else can test here?  One way to get better test coverage is to put this out in a canary release.

ts678:
> Anything else we can test here?

Although I don't think it was tested, I'm less worried now about potential oddities of special things existing near the end of the path, and a little more worried about potential variations found at start of path, such as using a relative path or leaving off the drive letter.  

[The Definitive Guide on Win32 to NT Path Conversion](https://googleprojectzero.blogspot.com/2016/02/the-definitive-guide-on-win32-to-nt.html) gives a good taxonomy. To avoid possible limitations of the GUI, I tested from [Export as Command-line](https://duplicati.readthedocs.io/en/latest/03-using-the-graphical-user-interface/#exporting-a-backup-job-configuration), changing the requested backup to things like relative-to-current-directory-and-drive (basically just a folder in current directory). Although I think it'd be less usual, I also tested `..\` at start. All got what I wanted, verified by `find` command.

There are other varieties of paths described in that document, but some seem highly unlikely to be used for Duplicati backup sources.

> One way to get better test coverage is to put this out in a canary release.

That's what it's for. Unfortunately I'm not sure how much control there is on the timing. When this is in `master` there's a chance I'll be able to do a build myself and play with it (still wishing for a nightly) before the release, but I'd have to install a build environment.

warwickmm:
> When this is in master there's a chance I'll be able to do a build myself and play with it (still wishing for a nightly) before the release, but I'd have to install a build environment.

If you're curious, it's possible to [checkout the code prior to merge](https://docs.github.com/en/github/collaborating-with-issues-and-pull-requests/checking-out-pull-requests-locally) if you want to test things locally.

Since the changes looks reasonable to me, I'm going to go ahead and merge this.  We can always follow-up with additional fixes as we discover any issues.  Thanks @dferreyra!

ts678:
Getting two variations of odd access errors from the combination of a Windows drive root plus VSS snapshot.
[Warning with 2.0.5.109](https://forum.duplicati.com/t/warning-with-2-0-5-109/10448). I know there are some newer PRs in the queue, but this is report from new 2.0.5.109.

ts678:
The HResult”:-2147024773 (0x8007007B) from forum report possibly decodes per [2.1 HRESULT](https://docs.microsoft.com/en-us/openspecs/windows_protocols/ms-erref/0642cb2f-2075-4469-918c-4441e69c548a) and [2.2 Win32 Error Codes](https://docs.microsoft.com/en-us/openspecs/windows_protocols/ms-erref/18d8fbe8-a967-4f1c-ae50-99ca8e491d2d) to an ERROR_INVALID_NAME 0x0000007B  "The file name, directory name, or volume label syntax is incorrect." from FACILITY_WIN32 7.

It looks like the CLI error I got is just abbreviated for some reason. When I run GUI live log testing trying to backup C:\\ using VSS:

`{"ClassName":"System.IO.IOException","Message":"The filename, directory name, or volume label syntax is incorrect.\r\n","Data":null,"InnerException":null,"HelpURL":null,"StackTraceString":"   at System.IO.__Error.WinIOError(Int32 errorCode, String maybeFullPath)\r\n   at System.IO.File.GetAttributes(String path)\r\n   at Duplicati.Library.Utility.Utility.<EnumerateFileSystemEntries>d__20.MoveNext()","RemoteStackTraceString":null,"RemoteStackIndex":0,"ExceptionMethod":"8\nWinIOError\nmscorlib, Version=4.0.0.0, Culture=neutral, PublicKeyToken=b77a5c561934e089\nSystem.IO.__Error\nVoid WinIOError(Int32, System.String)","HResult":-2147024773,"Source":"mscorlib","WatsonBuckets":null}`

dferreyra:
I've submitted #4278 to fix the problem reported in [Warning with 2.0.5.109](https://forum.duplicati.com/t/warning-with-2-0-5-109/10448).

PR #4258, opened 2020-07-20
Fix Recovery Tool handling of paths with dashes

dferreyra:
In Duplicati.CommandLine.RecoveryTool, the `index` and `list` commands
do not recognize Duplicati backup files when those files are in paths
containing dashes.

For example, without this change, if you call
```
Duplicati.CommandLine.RecoveryTool index C:\Temp\duplicati-tests\recovery-tool\backup-files
```
the Duplicati backup files are displayed, but the recovery tool
reports "Not a Duplicati file, ignoring" for each of them.

The problem was that `Library.Main.Volumes.VolumeBase.ParseFilename()` was being called with the entire path to the Duplicati backup files; e.g., "`C:\Temp\duplicati-tests\recovery-tool\backup-files\duplicati-20200720T005143Z.dlist.zip`", but `Library.Main.Volumes.VolumeBase.ParseFilename()` only expects the file name portion.  This change uses `Path.GetFileName()` to pass the file name to `Library.Main.Volumes.VolumeBase.ParseFilename()` where appropriate.

### DIFF
--- a/Duplicati/CommandLine/RecoveryTool/Index.cs
+++ b/Duplicati/CommandLine/RecoveryTool/Index.cs
@@ -64,7 +64,7 @@ namespace Duplicati.CommandLine.RecoveryTool
 
                 try
                 {
-                    var p = Library.Main.Volumes.VolumeBase.ParseFilename(file);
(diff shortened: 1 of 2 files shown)


warwickmm:
Thanks @dferreyra.

PR #4267, opened 2020-07-26
Support Windows source paths starting with "\\?\"

dferreyra:
Under Windows, when a backup path has been specified that uses the
"`\\?\`" prefix, change Duplicati to consistently record all files and
folders with that prefix.  Change the GUI so that it does not mistake
the question mark in the "`\\?\`" prefix with a wildcard.

Without this change, if you manually enter, for example, "`\\?\C:\Temp\`" in the source data:
1. The resulting Duplicati backup data will contain an entry for the path "`\\?\C:\Temp\`", but the entries for subdirectories and files will not include the "`\\?\`" prefix; e.g., you might see "`C:\Temp\file.txt`".
1. When you go to restore via the GUI, you see entries for both "`\\?\C:\Temp\`" and "`C:\Temp\`" in the file picker
1. In the file pickter, if you try to expand one of the entries with the "`\\?\`" prefix, you get this error:
```
```

To fix these problems, this commit changes `ListFiles()` and `ListFolders()` in NoSnapshotWindows.cs to not unconditionally strip UNC prefixes from the results.

It also changes the `FilterExpression` class in FilterExpression.cs to stop identifying the question mark in the "`\\?\`" prefix as a wildcard.  Note that the question mark in the "`\\?\`" prefix is ignored on Linux, too, to allow for the possibility of restoring a Windows backup using Duplicati running on Linux.

Things I've tested:
1. In Windows, used the GUI and Duplicati.CommandLine to perform backups with the "`\\?\`" prefix
1. In Windows and Ubuntu, used the GUI, Duplicati.CommandLine and Duplicati.CommandLine.Recoverytool to perform the restores of Windows backups that used the "`\\?\`" prefix
1. Performed snapshot and non-snapshot backups with the "`\\?\`" prefix

### DIFF
--- a/Duplicati/Library/Snapshots/NoSnapshotWindows.cs
+++ b/Duplicati/Library/Snapshots/NoSnapshotWindows.cs
@@ -93,12 +93,7 @@ namespace Duplicati.Library.Snapshots
         /// <param name='localFolderPath'>The folder to examinate</param>
         protected override string[] ListFiles (string localFolderPath)
         {
-            string[] tmp = SystemIO.IO_WIN.GetFiles(localFolderPath);
-            string[] res = new string[tmp.Length];
-            for(int i = 0; i < tmp.Length; i++)
-                res[i] = SystemIOWindows.StripUNCPrefix(tmp[i]);
-
-            return res;
+            return SystemIO.IO_WIN.GetFiles(localFolderPath);
         }
 
         
@@ -109,12 +104,7 @@ namespace Duplicati.Library.Snapshots
         /// <param name='localFolderPath'>The folder to examinate</param>
         protected override string[] ListFolders (string localFolderPath)
         {
-            string[] tmp = SystemIO.IO_WIN.GetDirectories(SystemIOWindows.PrefixWithUNC(localFolderPath));
-            string[] res = new string[tmp.Length];
-            for (int i = 0; i < tmp.Length; i++)
-                res[i] = SystemIOWindows.StripUNCPrefix(tmp[i]);
-
-            return res;
+            return SystemIO.IO_WIN.GetDirectories(localFolderPath);
         }
 
         /// <summary>
--- a/Duplicati/Library/Utility/FilterExpression.cs
+++ b/Duplicati/Library/Utility/FilterExpression.cs
@@ -20,6 +20,8 @@ using System.Linq;
 using System.Collections.Generic;
 using System.Text.RegularExpressions;
 using System.Globalization;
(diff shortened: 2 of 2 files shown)


warwickmm:
> Note that the question mark in the "`\\?\`" prefix is ignored on Linux, too, to allow for the possibility of restoring a Windows backup using Duplicati running on Linux.

I can create a file named `\\?\` on Linux (ext4 fileystem).  Can ignoring this pattern cause unexpected behavior with the filters on Linux?

dferreyra:
@warwickmm, that's a good question.  I think my change in FilterExpression.cs to unconditionally strip the prefix is asking for trouble.

I have a different approach that I've coded up that might be a better fit.  It does the following:
1. Adds a new `--list-filter-literal` option to suppress searching the filter for regex, group, and wildcard characters.  It essentially says to take the filter literally
1. Changes restoreFilePicker.js to add a "`filter-literal=true`" query parameter that gets mapped to the new "list-filter-literal" option
1. The "list-filter-literal" value makes it's way into FilterExpression.cs to suppress the identification of special characters for groups, regexes, and wildcards

That fixes the restore picker in the GUI, the idea being that the picker is dealing with literal paths anyway, so there's no need to try to interpret the filter strings as anything else.

Adding the `--list-filter-literal` option also allows you to use the "find" command to list the contents of paths with funny characters; e.g.,
```
Duplicati.CommandLine.exe --list-filter-literal --list-folder-contents --dbpath=UKCZMQQVAL.sqlite find C:\Temp\duplicati-unc\unc-prefix \\?\C:\Temp\duplicati-tests\3839\WinMetadata\
```

In your example above, if you'd successfully backed up your Linux path starting with "`\\?\`", and you tried to list the contents with the "find" command, without this change I think you would get the same error that the GUI reports:
```
System.ArgumentException: Filter for list-folder-contents must be a path prefix with no wildcards
Parameter name: filter
```

I'd like to present this new commit, but I'm not sure what the right approach is.  I could push the new commit here, or I could submit it as a separate pull request for folks to consider.

dferreyra:
> I'd like to present this new commit, but I'm not sure what the right approach is. I could push the new commit here, or I could submit it as a separate pull request for folks to consider.

I went ahead and submitted pull request #4275 for the approach I outlined in my previous comment.

warwickmm:
It sounds like this approach has some undesirable drawbacks.  @dferreyra, should we close this pull request?

dferreyra:
@warwickmm, yes I agree on closing this request, thanks.

PR #4269, opened 2020-07-27
Target .NET Framework 4.7.1

warwickmm:
This updates all projects to target .NET Framework 4.7.1.  The TencentCOS and Tardigrade backends depend on .NET Standard 2.0.  When a .NET Framework prior to 4.7.1 is targeted, the system cannot be sure that all the dependencies exist, so it copies all dependent assemblies to the output directory.  This causes many assemblies from the `System` namespace to become bundled in the release.

https://stackoverflow.com/a/48875007

We had previously attempted to make individual projects target 4.7.1 (see pull request #4242), but this can cause compatibility issues when 4.6.2 projects depend on 4.7.1. projects.

This will require Mono 5.10.0 or greater (previously, we required 5.0.0 or greater).

https://www.mono-project.com/docs/about-mono/releases/5.10.0/#class-libraries

This fixes issue #4234.


warwickmm:
I'll have to investigate the Travis failures.

drwtsn32x:
Before this merges we should also update the deb and rpm package requirements to flag mono 5.10.x as minimum version.

warwickmm:
Thanks @drwtsn32x.  Revision 70e8c93d26dd8b55986ca86bc8d8ad59a94c1eb8 updates the package requirements.

warwickmm:
Looks like the selenium docker image we're using is [based on ubuntu bionic](https://github.com/SeleniumHQ/docker-selenium/blob/trunk/Base/Dockerfile#L1), which whose version of `mono-complete` is [4.6.2.7](https://packages.ubuntu.com/bionic/mono-complete).  I'll see if there's a way to bump this to version 5.10.0 or greater.  Hopefully that fixes the Travis tests.

drwtsn32x:
Interesting... I run the docker image on my NAS and it reports as containing mono 5.20.1.34

warwickmm:
The Travis build uses the [selenium docker image](https://github.com/SeleniumHQ/docker-selenium/blob/trunk/NodeFirefox/Dockerfile) to programmatically click around the UI for some simple tests.  I think I got it to work by updating to the latest version of mono inside of it.  Just running a few more tests before I push the fix to this branch.

warwickmm:
Revision c3c80f58ef14b67bd3a57f63ee1ba9303cb07721 updates the version of mono in the docker container running the GUI tests.

kenkendk:
Excellent work!

ts678:
One small step forward. Thanks! 5.10 is a good choice because I think FreeBSD and its NAS variants are stuck there for now.
[Bug 238089 - lang/mono: Update to 5.20 (5.10.0 dates to 2018-02) ](https://bugs.freebsd.org/bugzilla/show_bug.cgi?id=238089)
[lang/mono: take maintainership, update to 6.8.0.105](https://reviews.freebsd.org/D23300).

I put a note on the [Release: 2.0.5.109 (canary) 2020-08-07](https://forum.duplicati.com/t/release-2-0-5-109-canary-2020-08-07/10447/4?u=ts678) forum release note about this, but I think it'll mostly just work...

Maybe the time to update all documentation should be Beta. [Manual](https://duplicati.readthedocs.io/en/latest/02-installation/) 4.6.2 should update. [Duplicati home page](https://www.duplicati.com/) still has 4.5.

PR #4270, opened 2020-07-28
Change tests to work with long paths in Windows

### LINKED ISSUE #3863: Shorten paths in data.zip file used for BulkData tests
Our AppVeyor configuration currently ignores the `BulkData` tests.  This was intended originally to avoid running into AppVeyor's time limit restriction (currently at 60 minutes).  While this limit now appears to provide ample time for us to run all the tests (including `BulkData`), there are paths inside the `https://s3.amazonaws.com/duplicati-test-file-hosting/data.zip` file that result in test paths longer than what is supported by Windows.  For example:
```
C:\projects\duplicati\testdata\data\Icons\077434-black-inlay-crystal-clear-bubble-icon-business-full-set\077434-black-inlay-crystal-clear-bubble-icon-business-full-set\077238-black-inlay-crystal-clear-bubble-icon-business-currency-british-pound-sc35.png
```

We should modify the data in this file so that the resulting paths satisfy the length requirement.  Once we do so, we should be able to enable the `BulkData` tests in AppVeyor.

dferreyra:
Even in .NET 4.6.2
`System.IO.Compression.ZipFile.ExtractToDirectory()` cannot handle
long paths.  Replace call to
`System.IO.Compression.ZipFile.ExtractToDirectory()` with an
equivalent that extracts files to a temporary location and uses I/O
functions that support long paths to move them to their final
location.

In CommandLineOperationsTests.cs, use ISystemIO functions to handle
potentially long paths.

Add a fix to RecoveryTool for long paths that was missed by #4258.

This fixes #3863.

Note that I'm not currently set up to run unit tests in Linux, so I haven't tried my changes there.

### DIFF
--- a/Duplicati/UnitTest/BasicSetupHelper.cs
+++ b/Duplicati/UnitTest/BasicSetupHelper.cs
@@ -176,9 +176,8 @@ namespace Duplicati.UnitTest
                     {
                         // By the ZIP spec, directories end in a forward slash
                         var isDirectory = entry.FullName.EndsWith("/");
-                        // Convert forward slashes to the Windows separator
-                        var fullNamePath = entry.FullName.Replace("/", Util.DirectorySeparatorString);
-                        var destination = systemIO.PathCombine(destinationDirectoryName, fullNamePath);
+                        var destination =
+                            systemIO.PathGetFullPath(systemIO.PathCombine(destinationDirectoryName, entry.FullName));
                         if (isDirectory)
                         {
                             systemIO.DirectoryCreate(destination);
--- a/Duplicati/UnitTest/BasicSetupHelper.cs
+++ b/Duplicati/UnitTest/BasicSetupHelper.cs
@@ -18,6 +18,9 @@ using System;
 using NUnit.Framework;
 using System.IO;
 using System.Collections.Generic;
+using System.IO.Compression;
+using Duplicati.Library.Common;
+using Duplicati.Library.Common.IO;
 
 namespace Duplicati.UnitTest
 {
@@ -61,6 +64,8 @@ namespace Duplicati.UnitTest
         /// </summary>
         public static readonly bool DEBUG_OUTPUT = Library.Utility.Utility.ParseBool(Environment.GetEnvironmentVariable("DEBUG_OUTPUT"), false);
 
+        protected static readonly ISystemIO systemIO = SystemIO.IO_OS;
+
         /// <summary>
         /// Writes a message to TestContext.Progress and Console.Out
         /// </summary>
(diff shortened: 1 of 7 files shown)


warwickmm:
Thanks @dferreyra.  The new `ISystemIO` methods will allow me to clean up some hacks I needed in pull request #4257.

Duplicati/UnitTest/BasicSetupHelper.cs
```diff
@@ -153,6 +158,60 @@ public virtual void TearDown()
...
+                // long paths.
+                using (var archive = ZipFile.OpenRead(sourceArchiveFileName))
+                {
+                    foreach (var entry in archive.Entries)
+                    {
+                        // By the ZIP spec, directories end in a forward slash
+                        var isDirectory = entry.FullName.EndsWith("/");
+                        // Convert forward slashes to the Windows separator
+                        var fullNamePath = entry.FullName.Replace("/", Util.DirectorySeparatorString);
+                        var destination = systemIO.PathCombine(destinationDirectoryName, fullNamePath);
```
warwickmm, line 181: Should we use `PathGetFullPath` to convert the destination path to an absolute path?  The current working directory is unclear if `destinationDirectoryName` is a relative path.

If that works, then it might be possible to remove the conversion of `/` to `\`, since I think `PathGetFullPath` should convert all the separators to the standard one.

dferreyra: I agree it needs a call to `PathGetFullPath()` and I can do that.

As far as the slashes, the problem I encountered is that `Path.GetFullPath()` will strip trailing dots and spaces unless it's prefixed with "`\\?\`".  But once you add the prefix, that disables the slash fix-up.  `SystemIOWindows.PathGetFullPath()` adds the "`\\?\`" to handle the trailing dots and spaces, so the forward slashes remain and the resulting path then trips up `DirectoryCreate()`.

One alternative would be to add the slash fix-up code right into `SystemIOWindows.PathGetFullPath()`, or even into `PrefixWithUNC()` itself.  Then the last two lines in the snippet above can turn into what you were proposing:
```
var destination =
    systemIO.PathGetFullPath(systemIO.PathCombine(destinationDirectoryName, entry.FullName));
```

warwickmm: > One alternative would be to add the slash fix-up code right into SystemIOWindows.PathGetFullPath()

That would be nice so that `SystemIOWindows.PathGetFullPath` is closer in behavior to the built-in `Path.GetFullPath`.

dferreyra: I've pushed a change that adds the slash fix-up code to `SystemIOWindows.PathGetFullPath()`.

PR #4275, opened 2020-08-02
Support Windows source paths starting with "\\?\"

dferreyra:
Under Windows, when a backup path has been specified that uses the
"`\\?\`" prefix, change Duplicati to consistently record all files and
folders with that prefix.

Add option `--list-filter-literal` to the Duplicati.CommandLine.exe
"find" command to disable the check for regex, group, and wildcard
characters in the "`<filename>`" argument.

Change the GUI so that it does not mistake the question mark in the
"`\\?\`" prefix with a wildcard by changing the restore picker to pass
`--list-filter-literal=true` when obtaining folder contents.

### DIFF
--- a/Duplicati/Library/Snapshots/NoSnapshotWindows.cs
+++ b/Duplicati/Library/Snapshots/NoSnapshotWindows.cs
@@ -93,12 +93,7 @@ namespace Duplicati.Library.Snapshots
         /// <param name='localFolderPath'>The folder to examinate</param>
         protected override string[] ListFiles (string localFolderPath)
         {
-            string[] tmp = SystemIO.IO_WIN.GetFiles(localFolderPath);
-            string[] res = new string[tmp.Length];
-            for(int i = 0; i < tmp.Length; i++)
-                res[i] = SystemIOWindows.StripUNCPrefix(tmp[i]);
-
-            return res;
+            return SystemIO.IO_WIN.GetFiles(localFolderPath);
         }
 
         
@@ -109,12 +104,7 @@ namespace Duplicati.Library.Snapshots
         /// <param name='localFolderPath'>The folder to examinate</param>
         protected override string[] ListFolders (string localFolderPath)
         {
-            string[] tmp = SystemIO.IO_WIN.GetDirectories(SystemIOWindows.PrefixWithUNC(localFolderPath));
-            string[] res = new string[tmp.Length];
-            for (int i = 0; i < tmp.Length; i++)
-                res[i] = SystemIOWindows.StripUNCPrefix(tmp[i]);
-
-            return res;
+            return SystemIO.IO_WIN.GetDirectories(localFolderPath);
         }
 
         /// <summary>
--- a/Duplicati/Library/Main/Operation/ListFilesHandler.cs
+++ b/Duplicati/Library/Main/Operation/ListFilesHandler.cs
@@ -29,7 +29,7 @@ namespace Duplicati.Library.Main.Operation
 
         public void Run(IEnumerable<string> filterstrings = null, Library.Utility.IFilter compositefilter = null)
         {
-            var parsedfilter = new Library.Utility.FilterExpression(filterstrings);
+            var parsedfilter = new Library.Utility.FilterExpression(filterstrings, true, m_options.ListFilterLiteral);
             var filter = Library.Utility.JoinedFilterExpression.Join(parsedfilter, compositefilter);
             var simpleList = !((filter is FilterExpression expression && expression.Type == Library.Utility.FilterType.Simple) || m_options.AllVersions);
 
--- a/Duplicati/Library/Main/Options.cs
+++ b/Duplicati/Library/Main/Options.cs
@@ -263,6 +263,7 @@ namespace Duplicati.Library.Main
                     new CommandLineArgument("list-prefix-only", CommandLineArgument.ArgumentType.Boolean, Strings.Options.ListprefixonlyShort, Strings.Options.ListprefixonlyLong, "false"),
                     new CommandLineArgument("list-folder-contents", CommandLineArgument.ArgumentType.Boolean, Strings.Options.ListfoldercontentsShort, Strings.Options.ListfoldercontentsLong, "false"),
                     new CommandLineArgument("list-sets-only", CommandLineArgument.ArgumentType.Boolean, Strings.Options.ListsetsonlyShort, Strings.Options.ListsetsonlyLong, "false"),
+                    new CommandLineArgument("list-filter-literal", CommandLineArgument.ArgumentType.Boolean, Strings.Options.ListfilterliteralShort, Strings.Options.ListfilterliteralLong, "false"),
                     new CommandLineArgument("disable-autocreate-folder", CommandLineArgument.ArgumentType.Boolean, Strings.Options.DisableautocreatefolderShort, Strings.Options.DisableautocreatefolderLong, "false"),
                     new CommandLineArgument("allow-missing-source", CommandLineArgument.ArgumentType.Boolean, Strings.Options.AllowmissingsourceShort, Strings.Options.AllowmissingsourceLong, "false"),
 
@@ -550,6 +551,11 @@ namespace Duplicati.Library.Main
         /// A value indicating that only filesets are returned
         /// </summary>
         public bool ListSetsOnly { get { return GetBool("list-sets-only"); } }
+
+        /// <summary>
+        /// A value indicating whether list filters should be considered literally
+        /// </summary>
+        public bool ListFilterLiteral { get { return GetBool("list-filter-literal"); } }
 
         /// <summary>
         /// A value indicating if file time checks are skipped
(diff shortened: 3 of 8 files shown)


Duplicati/Library/Snapshots/NoSnapshotWindows.cs
```diff
@@ -93,12 +93,7 @@ public override System.IO.Stream OpenRead (string localPath)
         /// <param name='localFolderPath'>The folder to examinate</param>
         protected override string[] ListFiles (string localFolderPath)
         {
-            string[] tmp = SystemIO.IO_WIN.GetFiles(localFolderPath);
-            string[] res = new string[tmp.Length];
-            for(int i = 0; i < tmp.Length; i++)
-                res[i] = SystemIOWindows.StripUNCPrefix(tmp[i]);
```
kenkendk, line 99: This seems to have implications beyond the fix. Is this required for the the folder listing functionality to work?
The methods are used throughout the code, especially in the backup code. I would think this fix would start recording files with UNC paths instead of their normalized form?

dferreyra: Regarding this question:
> Is this required for the the folder listing functionality to work?

The change you highlight is intended to make the backup itself consistent.  Without this change, if you manually enter, for example, "`\\?\C:\Temp\`" in the source data, the resulting Duplicati backup data will contain an entry for the path "`\\?\C:\Temp\`", but the entries for subdirectories and files will not include the "`\\?\`" prefix; e.g., you might see "`C:\Temp\file.txt`".  The inconsistency in the backup then leaks through to the restore GUI that will show entries for both "`\\?\C:\Temp\`" and "`C:\Temp\`" in the file picker.  With this change, if someone manually enters "`\\?\C:\Temp\`", all the files and subdirectories will also be recorded with the prefix.  And the restore GUI then displays directory "`\\?\C:\Temp\`" and the appropriate files and subdirectories beneath it.

Regarding this question:
> The methods are used throughout the code, especially in the backup code. I would think this fix would start recording files with UNC paths instead of their normalized form?

Pull request #4256 changed `SystemIO.IO_WIN.GetFiles()` to essentially honor the form of the `path` argument in the results it returns.  If the `path` passed into `SystemIO.IO_WIN.GetFiles()` is itself _not_ prefixed with "`\\?\`", the code behaves like it did before #4256 and it strips the prefix from the results.  On the other hand, if `path` is already prefixed, `SystemIO.IO_WIN.GetFiles()` passes it straight to `Directory.GetFiles()` and the results of `SystemIO.IO_WIN.GetFiles()` will likewise be prefixed.

The code you highlight here changes `NoSnapshotWindows.ListFiles()` to _also_ honor the form of its path argument, `localFolderPath`.  If `localFolderPath` is not prefixed, the results will also not be prefixed.

For places in Duplicati that end up calling `NoSnapshotWindows.ListFiles()` with a non-prefixed path, Duplicati would continue to behave like it did before #4256 and before this pull request.  On the other hand, if there are places that call `NoSnapshotWindows.ListFiles()` with prefixed paths but were expecting non-prefixed results, that would be a problem.  And the same goes for the changes in #4256 and `SystemIO.IO_WIN.GetFiles()`.

I've done backups with non-prefixed paths with these changes; e.g., using the source data picker, and the entries stored in the backup are not prefixed, just like before these changes.  Also, the unit tests pass, though I don't know if the tests would be sensitive to this.

I'll take a look at the call tree for these functions and see how paths are obtained and how they are used to see if I can identify areas that might pose a problem.  If you have any specific things you can think of for me to test, I can do that, too.

dferreyra:
I've updated the original pull request comment to include information I had originally included  with #4267.

This pull request is similar to #4267, but, after a question from @warwickmm, attempts to improve on the technique used to fix the error in the GUI:
```
```

From what I know now, I don't think the solution to this error in #4267 is a good one.  The change in this pull request, that adds `--list-filter-literal` and changes to GUI to use it to get lists of files, reduces the impact of the change and I think follows existing patterns in the code.

Sorry if there's any confusion with the two pull requests.

dferreyra:
@kenkendk, I probably misread your questions in the review:
> This seems to have implications beyond the fix. Is this required for the the folder listing functionality to work?
The methods are used throughout the code, especially in the backup code. I would think this fix would start recording files with UNC paths instead of their normalized form?

Yes, with this pull request, if a user manually specifies a source path using the `\\?\` notation, either through the GUI or through the command line, Duplicati will begin to store directories and files using the prefix, in a sense "honoring" the users's choice of notation.  Maybe this is a stretch, but it reminds me of how I might have a drive `F:` mapped to `\\server.example.com\files`, and I can ask Duplicati to back up either `\\server.example.com\files\foo` or `F:\foo` and Duplicati preserves my choice of notation.  Users might expect Duplicati to use the `\\?\` notation in cases where they've specified it themselves, but I can't think of a strong technical reason to preserve the notation since we apply the prefix internally anyway to support long paths and funny characters.

Perhaps you'd prefer that Duplicati _never_ record entries using the `\\?\` notation, even if the user specified their source paths that way?  In that case, a different fix could be to unconditionally strip the `\\?\` prefixes for any source paths that are provided by the user under Windows.  If we did that instead, we probably don't need the "--list-filter-literal" change.  I think that would be enough to fix the three problems I wrote in my main comment:
>Without this change, if you manually enter, for example, `\\?\C:\Temp\` in the source data:
>1. The resulting Duplicati backup data will contain an entry for the path `\\?\C:\Temp\`, but the entries for subdirectories and files will not include the `\\?\` prefix; e.g., you might see `C:\Temp\file.txt`.
>1. When you go to restore via the GUI, you see entries for both `\\?\C:\Temp\` and `C:\Temp\` in the file picker
>1. In the file pickter, if you try to expand one of the entries with the `\\?\` prefix, you get this error:
> ```
> Filter for list-folder-contents must be a path prefix with no wildcards Parameter name: filter
> ```

Duplicati/Library/Snapshots/NoSnapshotWindows.cs
dferreyra: Just to follow up, I did look at the call tree for these functions and ran some basic experiments.  As far as I can tell, `\\?\` prefixed paths will only be recorded if the user provides `\\?\` prefixed paths, either through manually typing in paths this way via the GUI or by providing them on the command line.  That was the intent of these changes.  I have added another comment on this pull request for an alternative that would change Duplicati to _never_ store `\\?\` prefixed paths, if that would be preferred.

warwickmm:
> Perhaps you'd prefer that Duplicati never record entries using the \\?\ notation, even if the user specified their source paths that way? In that case, a different fix could be to unconditionally strip the \\?\ prefixes for any source paths that are provided by the user under Windows. If we did that instead, we probably don't need the "--list-filter-literal" change. I think that would be enough to fix the three problems I wrote in my main comment:

This feels a bit cleaner to me.  The user entering a source path with `\\?\` is likely a rare scenario, so my preference would be for a solution that touches less code.

dferreyra:
@warwickmm, @kenkendk, I've submitted a pull request so you can see what the alternative might look like: #4285.

I have a more general question about pull requests, is someone can help me.  Let's pretend for a moment that it's decided to try an approach like #4285.  I see at least three ways I might approach that:
1. Add a single commit to this pull request that simultaneously undoes what I did in the previous commit, and implements the new approach.
1. Add two commits to this pull request, the first that reverts the changes of the previous commit for an explicit clean slate, and the second that implements the new approach.
1. Create a new pull request.

What's the right thing to do when one would essentially like to start over?  The first two approaches keep everything under a single pull request along with the context that led to the changes.  The last approach is clean, but cut off from the original context.

warwickmm:
My preference would be approaches 1 and 2.  It's easier to follow the evolution of the work within a single pull request, and it's nice to have the revisions corresponding to the alternative approaches available in the repository after the branch is merged.

dferreyra:
@warwickmm, thanks for answering my question about pull requests.

@kenkendk and @warwickmm, I was wondering if there's any interest in a pull request that implements the restore-related GUI change with the `--list-filter-literal` option.

I don't think it's needed for Windows since it looks like Windows won't let you include `*` or `?` in folder names even if you use the `\\?\` prefix.  But Linux will allow `*` and `?` in folder names, and if you do happen to back up a folder with a `*` or a `?`, the restore GUI won't let you expand the folder because you get the error mentioned in my original comment:
```
```

I would describe the change something like this:

> Fix restore GUI handling of folders with `*` or `?`
>
> Add option `--list-filter-literal` to the Duplicati.CommandLine "find" command that disables the check for regex, group, and wildcard characters in the `<filename>` argument.  Since the restore GUI is only dealing with actual folder and file names, change the restore file picker to pass in `--list-filter-literal=true` to avoid mistaking literal `*` or `?` characters in folder and file names for wildcards and triggering the following error:
> ```
> Filter for list-folder-contents must be a path prefix with no wildcards Parameter name: filter
> ```

warwickmm:
Sorry @dferreyra, this question got lost in my emails.  Can you create a new issue for the below problem?  We can continue the discussion there.

> But Linux will allow * and ? in folder names, and if you do happen to back up a folder with a * or a ?, the restore GUI won't let you expand the folder because you get the error mentioned in my original comment:
> 
> Filter for list-folder-contents must be a path prefix with no wildcards Parameter name: filter

PR #4277, opened 2020-08-06
Enabled support for \$Extend\$Deleted special MTF folder when using USN

dgehri:
See discussion in https://forum.duplicati.com/t/usn-option-working-inconsistently/9224: instead of deleting files, Windows 10 moves them to a special MTF folder named _\$Extend\$Deleted_, which doesn't show up in the USN journal.

I did not manage to find a programmatic way to look up the special folder name above from it's FRN. However, when files are "deleted" that way, they will be renamed to a sequence of hex digits, where the first 16 digits are the file's FRN in hex, and the remaining digits are random. The code checks this, in addition to the fact that the parent FRN cannot be looked up, as well as that this special FRN is the same for all such files. This is sufficient to be 100% sure that the file was, in fact, deleted. 

In any case, the USN code doesn't establish what was deleted, renamed or modified. It just provides a list of files & folders that were "touched" since the last update. So the relevant entry, isn't the new name of the "deleted" file, but it's old name and location. That latter entry will be present in the USN (there are always two entries for a rename, the old and the new name), and hence the file scanner will establish the fact that it was, indeed, deleted.

### DIFF
--- a/Duplicati/Library/Main/Operation/Backup/CountFilesHandler.cs
+++ b/Duplicati/Library/Main/Operation/Backup/CountFilesHandler.cs
@@ -20,20 +20,21 @@
 using System.Collections.Generic;
 using System.Threading.Tasks;
 using Duplicati.Library.Utility;
+using Duplicati.Library.Snapshots;
 using CoCoL;
 
 namespace Duplicati.Library.Main.Operation.Backup
 {
     internal static class CountFilesHandler
     {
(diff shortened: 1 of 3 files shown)


dgehri:
Note: this PR postpones the Backup.CountFilesHandler execution, so it can use the USN filtered file set, instead of attempting to scanning everything. This isn't strictly related to this fix, so feel free to reject that part if needed.

dgehri:
I removed the scheduling change to CountsFileHandler, as it caused weird behavior in the GUI. Still, it doesn't make sense for this thread to extend beyond the actual backup thread. The count thread takes ~15 minutes on my machine, whereas the USN backup thread is done after ~1-2 minutes.

kenkendk:
It would be awesome if the `CountFilesHandler` could use the USN as well, so we can avoid the delays.

It thougt that the counter task was aborted if the backup task completes?

PR #4278, opened 2020-08-08
Fix snapshots broken by #4256

dferreyra:
Revert the change made to `WindowsSnapshot.ConvertToSnapshotPath()`
by pull request #4256.

The change in #4256 replaced some existing manual path construction
code with a call to `SystemIO.IO_WIN.PathCombine()`, but calling
`SystemIO.IO_WIN.PathCombine()` this way did not always result in
valid VSS paths.  For example, trying to back up a root volume like
"`C:\`" would result in the following error:
```
The filename, directory name, or volume label syntax is incorrect.
```

This commit addresses the problem reported in the forum, [Warning with 2.0.5.109](https://forum.duplicati.com/t/warning-with-2-0-5-109/10448).  That specific problem was that trying to back up the root of a drive, like "`D:\`", would result in a VSS path that looked something like this:
```
\\?\GLOBALROOT\Device\HarddiskVolumeShadowCopy5
```
This lacks a directory separator at the end.  The valid VSS path would look like this:
```
\\?\GLOBALROOT\Device\HarddiskVolumeShadowCopy5\
```
This might not be the only thing that would not work with the change to `WindowsSnapshot.ConvertToSnapshotPath()`.  This revert restores the original behavior of `WindowsSnapshot.ConvertToSnapshotPath()`.

Note that in the revert, I have left out this comment since, as of #4256, it is no longer true:
```csharp
// Note: Do NOT use Path.Combine as it strips the UNC path prefix
```

### DIFF
--- a/Duplicati/Library/Snapshots/WindowsSnapshot.cs
+++ b/Duplicati/Library/Snapshots/WindowsSnapshot.cs
@@ -266,6 +266,10 @@ namespace Duplicati.Library.Snapshots
             var root = SystemIO.IO_WIN.GetPathRoot(localPath);
             var volumePath = _vssBackupComponents.GetVolumeFromCache(root);
 
+            // Note that using a simple Path.Combine() for the following code
+            // can result in invalid snapshot paths; e.g., if localPath is
+            // @"C:\", mappedPath would not have the required trailing
+            // directory separator.
             var subPath = localPath.Substring(root.Length);
             if (!subPath.StartsWith(Util.DirectorySeparatorString, StringComparison.Ordinal))
             {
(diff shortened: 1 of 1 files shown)


ts678:
> This might not be the only thing that would not work with the change

Maybe we'll hear. Now that release notes in the forum are working again, I put a [notice](https://forum.duplicati.com/t/release-2-0-5-109-canary-2020-08-07/10447/2?u=ts678) on it to potential Canary users,
however I only mentioned the root path case, and tried to steer people to the original report topic if they saw an issue.
Can't rely on everyone reading release note posting. Thanks very much for the fast turnaround on this analysis and PR.

warwickmm:
Thanks @dferreyra for the quick fix.  Can we add a comment in the code explaining why we can't use `SystemIO.IO_WIN.PathCombine`, which would feel like a natural thing to do?

dferreyra:
@warwickmm, yes, good idea.  I've pushed a commit that adds a comment.

PR #4287, opened 2020-08-17
Feature/improve usn 2

dgehri:
This is an update to my previous improvement to USN. It replays the file history for each file, and removes any change sequence(s) that don't affect the backup (e.g. creation of a file, followed by deletion).

This further increases the number of times the USN can be used.

This patch also restores use of the USN by the CountFilesHandler, if available.

### DIFF
--- a/Duplicati/Library/Main/Operation/Backup/CountFilesHandler.cs
+++ b/Duplicati/Library/Main/Operation/Backup/CountFilesHandler.cs
@@ -20,20 +20,21 @@
 using System.Collections.Generic;
 using System.Threading.Tasks;
 using Duplicati.Library.Utility;
+using Duplicati.Library.Snapshots;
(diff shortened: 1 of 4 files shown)


dgehri:
I'll combine PR 4288 into this one.

dgehri:
This PR also fixes the issue raised in https://forum.duplicati.com/t/untraceable-long-operation-query-during-backup/9007/24.

dgehri:
**This PR is ready to merge.** I tested it extensively, and it brings down backup times (in my case) from ~25 minutes to 1'30". 

This is mainly due to three improvements:

1. Smarter parsing of USN events, excluding non relevant events
2. Better resolving algorithm for full paths from USN 
2. Performance improvement of `AppendFilesFromPreviousSetWithPredicate()`

PR #4302, opened 2020-08-31
Fix restore GUI handling of folders with * or ?

### LINKED ISSUE #4300: The restore GUI gives an error when trying to expand a directory containing a literal wildcard character

## Environment info

- **Operating system**: Ubuntu 18.04.5 LTS
- **Backend**: S3

## Description

If you back up a directory containing literal wildcard characters (`*` or `?`), if you go into the restore GUI and try to expand the directory to see its contents, you get the following error:
```
```

## Steps to reproduce

1. Create a directory with literal wildcard characters; e.g., 
```
mkdir 'does-question-mark-work?'
```
2. Add some files in the directory
1. Back up the directory
1. In the restore GUI, try to expand the directory with the wildcard characters to view the contents

- **Actual result**:
You get the following error trying to expand the directory, and you cannot view the contents:
```
```
- **Expected result**:
To be able to view the contents of the directory.

## Debug log

```
Aug 29, 2020 10:47 AM: Request for http://example.com:8200/api/v1/backup/1/files//home/example/duplicati-data/does-question-mark-work?/?prefix-only=false&folder-contents=true&time=2020-08-29T10:00:10%2B00:00&filter=%2Fhome%2Fexample%2Fduplicati-data%2Fdoes-question-mark-work%3F%2F gave error
System.ArgumentException: Filter for list-folder-contents must be a path prefix with no wildcards
Parameter name: filter
  at Duplicati.Library.Main.Database.LocalListDatabase+FileSets+<SelectFolderContents>d__10.MoveNext () [0x000a7] in <8f1de655bd1240739a78684d845cecc8>:0 
  at System.Linq.Enumerable+SelectEnumerableIterator`2[TSource,TResult].ToArray () [0x00030] in <a63e4875bcd94dc2960d31c0cf2bbd30>:0 
  at System.Linq.Enumerable.ToArray[TSource] (System.Collections.Generic.IEnumerable`1[T] source) [0x0001f] in <a63e4875bcd94dc2960d31c0cf2bbd30>:0 
  at Duplicati.Library.Main.Operation.ListFilesHandler.Run (System.Collections.Generic.IEnumerable`1[T] filterstrings, Duplicati.Library.Utility.IFilter compositefilter) [0x001e4] in <8f1de655bd1240739a78684d845cecc8>:0 
  at Duplicati.Library.Main.Controller+<>c__DisplayClass21_0.<List>b__0 (Duplicati.Library.Main.ListResults result) [0x0001c] in <8f1de655bd1240739a78684d845cecc8>:0 
  at Duplicati.Library.Main.Controller.RunAction[T] (T result, System.String[]& paths, Duplicati.Library.Utility.IFilter& filter, System.Action`1[T] method) [0x0026f] in <8f1de655bd1240739a78684d845cecc8>:0 
  at Duplicati.Library.Main.Controller.RunAction[T] (T result, Duplicati.Library.Utility.IFilter& filter, System.Action`1[T] method) [0x00007] in <8f1de655bd1240739a78684d845cecc8>:0 
  at Duplicati.Library.Main.Controller.List (System.Collections.Generic.IEnumerable`1[T] filterstrings, Duplicati.Library.Utility.IFilter filter) [0x00021] in <8f1de655bd1240739a78684d845cecc8>:0 
  at Duplicati.Server.Runner.Run (Duplicati.Server.Runner+IRunnerData data, System.Boolean fromQueue) [0x00668] in <c5f097a49c0a4f1fb0f93cf3f5f218b1>:0 
  at Duplicati.Server.WebServer.RESTMethods.Backup.SearchFiles (Duplicati.Server.Serialization.Interface.IBackup backup, System.String filterstring, Duplicati.Server.WebServer.RESTMethods.RequestInfo info) [0x000cb] in <c5f097a49c0a4f1fb0f93cf3f5f218b1>:0 
  at Duplicati.Server.WebServer.RESTMethods.Backup.GET (System.String key, Duplicati.Server.WebServer.RESTMethods.RequestInfo info) [0x001e0] in <c5f097a49c0a4f1fb0f93cf3f5f218b1>:0 
  at Duplicati.Server.WebServer.RESTHandler.DoProcess (Duplicati.Server.WebServer.RESTMethods.RequestInfo info, System.String method, System.String module, System.String key) [0x00146] in <c5f097a49c0a4f1fb0f93cf3f5f218b1>:0 
```

dferreyra:
Add new filter syntax, prefixing with an "at sign"; i.e., `@...`, that
disables glob expansion, treating `*` and `?` wildcard characters as
literal characters.

In the GUI, change the restore file picker to prefix paths with `@`
when expanding a node to avoid mistaking literal `*` or `?` characters
for wildcards characters and triggering the following error:
```
```

This fixes #4300.

### DIFF
--- a/Duplicati/CommandLine/Commands.cs
+++ b/Duplicati/CommandLine/Commands.cs
@@ -21,6 +21,7 @@ using System.Collections.Generic;
 using System.IO;
 using System.Text.RegularExpressions;
 using Duplicati.Library.Common.IO;
+using Duplicati.Library.Utility;
 
 namespace Duplicati.CommandLine
 {
@@ -254,7 +255,7 @@ namespace Duplicati.CommandLine
             if (!containsSeparators && arg.StartsWith("@", StringComparison.Ordinal))
             {
                 // Convert to Regexp filter and prefix with ".*/"
-                return $"[.*{Regex.Escape(Util.DirectorySeparatorString + arg.Substring(1))}]";
+                return $"[.*{Utility.ConvertLiteralToRegExp(Util.DirectorySeparatorString + arg.Substring(1))}]";
             }
             else if (!containsSeparators && !containsWildcards && !arg.StartsWith("[", StringComparison.Ordinal))
             {
@@ -281,12 +282,12 @@ namespace Duplicati.CommandLine
         private static string SuffixArgWithAsterisk(string arg)
         {
             var containsWildcards = ContainsWildcards(arg);
-            var endsWithSeparator = arg.EndsWith(Util.DirectorySeparatorString);
+            var endsWithSeparator = arg.EndsWith(Util.DirectorySeparatorString, StringComparison.Ordinal);
 
             if (endsWithSeparator && arg.StartsWith("@", StringComparison.Ordinal))
             {
                 // Convert to Regexp filter and suffix with ".*"
-                return $"[{Regex.Escape(arg.Substring(1))}.*]";
+                return $"[{Utility.ConvertLiteralToRegExp(arg.Substring(1))}.*]";
             }
             else if (endsWithSeparator && !containsWildcards && !arg.StartsWith("[", StringComparison.Ordinal))
             {
@@ -304,16 +305,14 @@ namespace Duplicati.CommandLine
         /// <summary>
         /// Returns true if <paramref name="s"/> contains directory separator characters.
         /// </summary>
-        public static bool ContainsDirectorySeparators(string s) =>
-            s.IndexOfAny(pathSeparatorsCharacters) >= 0;
+        public static bool ContainsDirectorySeparators(string s) => s.IndexOfAny(pathSeparatorsCharacters) >= 0;
 
         private static readonly char[] wildcardCharacters = new[] { '*', '?' };
 
         /// <summary>
         /// Returns true if <paramref name="s"/> contains wildcard characters.
         /// </summary>
-        public static bool ContainsWildcards(string s) =>
-            s.IndexOfAny(wildcardCharacters) >= 0;
+        public static bool ContainsWildcards(string s) => s.IndexOfAny(wildcardCharacters) >= 0;
 
         public static int List(TextWriter outwriter, Action<Duplicati.Library.Main.Controller> setup, List<string> args, Dictionary<string, string> options, Library.Utility.IFilter filter)
         {
--- a/Duplicati/CommandLine/Commands.cs
+++ b/Duplicati/CommandLine/Commands.cs
@@ -239,7 +239,8 @@ namespace Duplicati.CommandLine
             // Prefix all filenames with "*/" so we search all folders
             var specialArgChars = new[] { '*', '?', Path.DirectorySeparatorChar, Path.AltDirectorySeparatorChar };
             return argList.Select(arg => arg.IndexOfAny(specialArgChars) < 0
-                 && !arg.StartsWith("[", StringComparison.Ordinal) ? "*" + Util.DirectorySeparatorString + arg : arg).ToList();
+                 && !arg.StartsWith("[", StringComparison.Ordinal)
+                 && !arg.StartsWith("@", StringComparison.Ordinal) ? "*" + Util.DirectorySeparatorString + arg : arg).ToList();
         }
 
         public static int List(TextWriter outwriter, Action<Duplicati.Library.Main.Controller> setup, List<string> args, Dictionary<string, string> options, Library.Utility.IFilter filter)
@@ -267,7 +268,7 @@ namespace Duplicati.CommandLine
                             options["version"] = v.ToString();
                         }
                     }
-                    else if (args[0].IndexOfAny(new char[] { '*', '?' }) < 0 && !args[0].StartsWith("[", StringComparison.Ordinal))
+                    else if (args[0].IndexOfAny(new char[] { '*', '?' }) < 0 && !args[0].StartsWith("[", StringComparison.Ordinal) && !args[0].StartsWith("@", StringComparison.Ordinal))
                     {
                         try
                         {
@@ -452,7 +453,7 @@ namespace Duplicati.CommandLine
 
             // suffix all folders with "*" so we restore all contents in the folder
             for (var ix = 0; ix < args.Count; ix++)
-                if (args[ix].IndexOfAny(new char[] { '*', '?' }) < 0 && !args[ix].StartsWith("[", StringComparison.Ordinal) && args[ix].EndsWith(Util.DirectorySeparatorString, StringComparison.Ordinal))
+                if (args[ix].IndexOfAny(new char[] { '*', '?' }) < 0 && !args[ix].StartsWith("[", StringComparison.Ordinal) && args[ix].EndsWith(Util.DirectorySeparatorString, StringComparison.Ordinal) && !args[ix].StartsWith("@", StringComparison.Ordinal))
                     args[ix] += "*";
 
             using(var output = new ConsoleOutput(outwriter, options))
--- a/Duplicati/CommandLine/Commands.cs
+++ b/Duplicati/CommandLine/Commands.cs
@@ -19,6 +19,7 @@ using System;
 using System.Linq;
 using System.Collections.Generic;
 using System.IO;
+using System.Text.RegularExpressions;
 using Duplicati.Library.Common.IO;
 
 namespace Duplicati.CommandLine
@@ -234,14 +235,85 @@ namespace Duplicati.CommandLine
             }
         }
 
-        public static List<string> PrefixArgsWithAsterisk(IList<string> argList)
+        /// <summary>
+        /// For bare file names with no wildcards, replace with the
+        /// equivalent of prefixing with "*/" so we search all folders.
+        /// </summary>
+        public static IEnumerable<string> PrefixArgsWithAsterisk(IEnumerable<string> argList) =>
+            argList.Select(PrefixArgWithAsterisk).ToList();
+
+        /// <summary>
+        /// For bare file names with no wildcards, replace with the
+        /// equivalent of prefixing with "*/" so we search all folders.
+        /// </summary>
+        private static string PrefixArgWithAsterisk(string arg)
         {
-            // Prefix all filenames with "*/" so we search all folders
-            var specialArgChars = new[] { '*', '?', Path.DirectorySeparatorChar, Path.AltDirectorySeparatorChar };
-            return argList.Select(arg => arg.IndexOfAny(specialArgChars) < 0
-                 && !arg.StartsWith("[", StringComparison.Ordinal)
-                 && !arg.StartsWith("@", StringComparison.Ordinal) ? "*" + Util.DirectorySeparatorString + arg : arg).ToList();
+            var containsSeparators = ContainsDirectorySeparators(arg);
+            var containsWildcards = ContainsWildcards(arg);
+
+            if (!containsSeparators && arg.StartsWith("@", StringComparison.Ordinal))
+            {
+                // Convert to Regexp filter and prefix with ".*/"
+                return $"[.*{Regex.Escape(Util.DirectorySeparatorString + arg.Substring(1))}]";
+            }
+            else if (!containsSeparators && !containsWildcards && !arg.StartsWith("[", StringComparison.Ordinal))
+            {
+                // Prefix files with "*/"
+                return "*" + Util.DirectorySeparatorString + arg;
+            }
+            else
+            {
+                return arg;
+            }
         }
+
+        /// <summary>
+        /// For folders, replace with the equivalent of suffixing with "/*"
+        /// so we restore contents in the folder.
+        /// </summary>
+        public static IEnumerable<string> SuffixArgsWithAsterisk(IEnumerable<string> argList) =>
+            argList.Select(SuffixArgWithAsterisk).ToList();
+
+        /// <summary>
+        /// For folders, replace with the equivalent of suffixing with "/*"
+        /// so we restore contents in the folder.
+        /// </summary>
+        private static string SuffixArgWithAsterisk(string arg)
+        {
+            var containsWildcards = ContainsWildcards(arg);
+            var endsWithSeparator = arg.EndsWith(Util.DirectorySeparatorString);
+
+            if (endsWithSeparator && arg.StartsWith("@", StringComparison.Ordinal))
+            {
+                // Convert to Regexp filter and suffix with "/.*"
+                return $"[{Regex.Escape(arg.Substring(1))}.*]";
+            }
+            else if (endsWithSeparator && !containsWildcards && !arg.StartsWith("[", StringComparison.Ordinal))
+            {
+                // Suffix with "/*"
+                return arg + "*";
+            }
+            else
+            {
+                return arg;
+            }
+        }
+
+        private static readonly char[] pathSeparatorsCharacters = new[] { Path.DirectorySeparatorChar, Path.AltDirectorySeparatorChar };
+
+        /// <summary>
+        /// Returns true if <paramref name="s"/> contains directory separator characters.
+        /// </summary>
+        public static bool ContainsDirectorySeparators(string s) =>
+            s.IndexOfAny(pathSeparatorsCharacters) >= 0;
+
+        private static readonly char[] wildcardCharacters = new[] { '*', '?' };
+
+        /// <summary>
+        /// Returns true if <paramref name="s"/> contains wildcard characters.
+        /// </summary>
+        public static bool ContainsWildcards(string s) =>
+            s.IndexOfAny(wildcardCharacters) >= 0;
 
         public static int List(TextWriter outwriter, Action<Duplicati.Library.Main.Controller> setup, List<string> args, Dictionary<string, string> options, Library.Utility.IFilter filter)
         {
@@ -283,7 +355,7 @@ namespace Duplicati.CommandLine
                     }
                 }
 
-                args = PrefixArgsWithAsterisk(args);
+                args = PrefixArgsWithAsterisk(args).ToList();
 
                 // Support for not adding the --auth-username if possible
                 string dbpath;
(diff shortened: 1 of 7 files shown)


warwickmm:
Thanks @dferreyra!  I was able to backup a directory named `*` and expand it in the restore tree.

However, I wasn't able to successfully use the `@` character in any filter (UI or command-line) and get it to behave as I expected.  I know there's a few bugs in the UI filter handling code, but I don't think it's related.

Can you think of a way to fix the restore tree without exposing the functionality to the user in any way?  I wouldn't want to add to existing frustrations about how the filters work (or don't work).

dferreyra:
@warwickmm, thanks for giving it a try.

As far as a fix that doesn't expose the functionality to users, the closest I've come to that was to add the extra command line option to the "find" function to disable globbing.  The restore picker uses "find", so the fix would have the restore picker provide the new option.  Here's how I described that proposed change:

> Fix restore GUI handling of folders with `*` or `?`
>
> Add option `--list-filter-literal` to the Duplicati.CommandLine "find" command that disables the check for regex, group, and wildcard characters in the `<filename>` argument.  Since the restore GUI is only dealing with actual folder and file names, change the restore file picker to pass in `--list-filter-literal=true` to avoid mistaking literal `*` or `?` characters in folder and file names for wildcards and triggering the following error:
> ```
> Filter for list-folder-contents must be a path prefix with no wildcards Parameter name: filter
> ```

The extra parameter has the advantage that unless someone asks for it, they'll get the existing behavior.

I already have the code for it, but in https://github.com/duplicati/duplicati/issues/4300#issuecomment-683444960, you said you preferred a solution that didn't add another command line option.  Also, it does touch more code, at least the way I implemented it.  You can look at most of the changes here, https://github.com/duplicati/duplicati/pull/4275/files; just ignore the NoSnapshotWindows.cs changes as they don't apply.

In the meantime, I'll think about it some more.

dferreyra:
And @warwickmm, sorry the `@` change didn't work.  I only tried it against the "find" command, but never actually tried to use it as an actual backup filter!

warwickmm:
Interesting, I tested it with `find` and didn't think it worked, but perhaps I was doing something wrong.  I was backing up
```
test_source/*/hello.txt
test_source/dir/hello.txt
```
and couldn't get `find` to only display the first entry.  Admittedly, I didn't try very hard.

I think it would be nice to get this working, although I'm guessing having `*` or `?` in a path is probably uncommon.

warwickmm:
I didn't get a chance to debug, but I wonder if this was interfering with my tests:

https://github.com/duplicati/duplicati/blob/145ef49b91fc43a77dabfb66cd8ae3c8b7ca93c2/Duplicati/CommandLine/Commands.cs#L285

dferreyra:
@warwickmm, I'll run some tests and see if I can see what's going on.  Thanks for the hints.

dferreyra:
@warwickmm, I've pushed a fix.

I tried your "find" experiment and with this fix it works now for me.  (When I originally tested "find", I did it without providing a `--dbpath` option and that seemed to give me the correct result.  Once I provided a `--dbpath`, I saw the same behavior as you reported.)

I also tried a command line backup with an `@` filter using `--include='@/test_source/*/'` to only include the `*` directory and not the `dir` directory, and that worked.

I also tried the "test-filters" command and it seems good, too.  (I didn't know about it until Duplicati itself told me about it after a backup.)  Anyway, here's the output:
```
$ mono Duplicati.CommandLine.exe test-filters --include='@/test_source/*/' --exclude='*/' /test_source/
Including source path: /test_source/
Excluding path due to filter: /test_source/dir/ => (*/)
Including path due to filter: /test_source/*/ => ([/test_source/\*/])
Excluding path due to filter: /test_source/does-question-mark-work?/ => (*/)
Including path as no filters matched: /test_source/*/hello.txt
Matched 1 files (10 bytes)
```

I tried testing filters through the GUI, but I couldn't figure out how to enter filters manually.  I tried using the "Show advanced editor" option and entering paths prefixed with `+` and `-`, but Duplicati gave an errors about the paths with `+` and `-` not existing.

Oh, and at least for me, the call to `PrefixArgsWithAsterisk()` had no effect on the filters I entered.

warwickmm:
Thanks @dferreyra.  That fixed `find` for me, but now the restore tree gives me the `Filter for list-folder-contents must be a path prefix with no wildcards Parameter name: filter` error.  In fact, now this error shows up before attempting to expand the `*` folder.  Does the restore tree still work for you?

dferreyra:
@warwickmm, yes, I get the same error.  I changed the code to convert `@` filters to "Regex" filters to get the correct behavior for "find" and "backup", but the filters need to be "Simple" to avoid the error.  I'll take a closer look.

dferreyra:
@warwickmm, with my last commit of considering the `@` filter a `FilterType.Regexp`, besides undoing the restore picker fix, I could also see places in the code that benefit when the filter can be identified as being of type `FilterType.Simple` and the `FilterType.Regexp` type thwarts those.  And I couldn't think of a nice way to fix the restore picker while using `FilterType.Regexp`.

So I went back to using `FilterType.Simple`, like in the original commit, and started investigating the things you found that weren't working.

The switch back to `FilterType.Simple` makes it so the restore picker is working again.

The problem with the "find" command when using a database was that there was some database code that was inspecting the fiter string directly, looking for wildcard characters.  I changed that code to trigger off of `FilterType.Wildcard` instead, so that fixed the "find" command for me.

I tried doing command line backups with the `@` and they worked for me, though it's a little awkward working with the `@` since once you prefix with an `@`, you can't take advantage of globbing.  Here's the command I ran.  It correctly backed up `/test_source/*/hello.txt` and not `/test_source/dir/hello.txt`:
```
mono Duplicati.CommandLine.exe backup destination /test_source/ --no-encryption=true --include='@/test_source/*/hello.txt'
```

I see now that the `PrefixArgsWithAsterisk()` function you mentioned earlier can get in the way of things in the "find" command and the "restore" command if you specify a filter that looks like a file with no path separators and no wildcard characters.  For example, `@hello.txt` gets turned into `*/@hello.txt`.  I'm still planning on fixing that, so the commit I pushed does not address this.

It's obvious to me now, but this change is making it so that you can create `FilterType.Simple` filters explicitly, even when they contain wildcard characters.  I saw that `FilterEntry.ToString()` is relied on in some places, so I changed it to unconditionally prefix all `FilterType.Simple` filters with the `@` so that they can make a round trip through a string and back to a `FilterEntry` without turning into `FilterType.Wildcard` when the filter contains wildcard characters.  I still want to take a closer look at this and run some tests to exercise the code.

I'm planning on doing more test.  I still haven't figured out where or how to exercise the `+` and `-` style of filters, so I haven't been able to test that yet.

dferreyra:
I've pushed another commit that addresses the issue of the "find" and "restore" command prefixing (and suffixing) some filters with `*`.  The commit disables the prefixing and suffixing of `*` for `@` filters since wildcards cannot be supported with `@` filters.

warwickmm:
I tested a backup that contains
```
test_source/*/hello.txt
test_source/dir/hello.txt
test_source/dir/bye.txt
```
In the UI, if I ask to restore `test_source/*/hello.txt` to a new location, it also restores the `dir` directory and its files as well.  Are you able to reproduce this?

dferreyra:
@warwickmm, thanks again for looking at this!  Yes I get the same behavior.

I'll take a look.  I assume that any path selected via the picker needs to be presented to the server as an `@` style filter, too.

dferreyra:
@warwickmm, I've pushed a fix for the restore GUI.

Now that I've implemented this solution, I can see that I can apply the same approach to the "find" and "restore" commands and how they want to add a `*` prefix to bare files (i.e., where no path is included) and a `*` suffix to all directories.  With b8501dfd77f20400df827b20381d94f5e6c205bc , I just disabled the `*` prefixing and suffixing if the filter provided is an `@` filter.  Instead, I can change "find" and "restore" to convert `@` filters to regular expression filters in those specific cases so we can preserve the literal wildcard characters and still perform the expected globbing.  That would restore the existing behavior even with `@` filters.  When I get a chance, I'll try coding it up and pushing another commit for it.

Circling back to some of my previous comments, I was able to test code paths related to the FilterEntry.ToString() with `FilterExpression.Serialize()` and `FilterExpression.Deserialize()`, and it looked good to me.  I was also able to test creating `+` and `-` style filters by using `--parameters-file`; e.g.,
```
--target=/target
--source=/test_source/
--no-encryption=true
--append-filter=+@/test_source/*/hello.txt
```
I've also performed various command line backups and restores, providing various styles of filters, both `@` and regular filters, include and exclude, separately and together and I haven't encountered any problems yet.

If you can think of some specific areas I should test, let me know.

warwickmm:
Thanks @dferreyra.  I ran a few quick manual tests and didn't encounter anything unexpected.  Since this does touch some core code paths, I think we should spend some extra time testing and studying this.

I added a test here (7d13d9ed81d3d6bb375ba2d08a2d5690b175e83d) to try and help us verify the behavior.  Can you take a look and if it seems helpful, cherry pick it into your branch to help us test any further changes?

Thanks for your help and patience.

dferreyra:
@warwickmm, thanks for the tests -- very helpful!

Even though Windows won't let you create paths containing `*` or `?`, I thought your tests would be helpful in Windows to verify the basic functionality of `@` filters.  For example, one thing that could trip up the code if not handled properly is that the Windows directory separator, `\`, needs to be escaped in regular expressions.  So I changed the tests to also run under Windows.  I hope that's okay and that you don't feel it obscures things.

Duplicati/CommandLine/Commands.cs
```diff
@@ -234,14 +235,86 @@ public static int Affected(TextWriter outwriter, Action<Duplicati.Library.Main.C
...
+            argList.Select(SuffixArgWithAsterisk);
+
+        /// <summary>
+        /// For folders, return argument with the equivalent of suffixing
+        /// with "*" so we restore contents in the folder.
+        /// </summary>
+        private static string SuffixArgWithAsterisk(string arg)
+        {
+            var containsWildcards = ContainsWildcards(arg);
+            var endsWithSeparator = arg.EndsWith(Util.DirectorySeparatorString);
```
warwickmm, line 284: Can we use the `StringComparison.Ordinal` comparison type here?

Duplicati/CommandLine/Commands.cs
```diff
@@ -234,14 +235,86 @@ public static int Affected(TextWriter outwriter, Action<Duplicati.Library.Main.C
...
+            }
         }
 
+        private static readonly char[] pathSeparatorsCharacters = new[] { Path.DirectorySeparatorChar, Path.AltDirectorySeparatorChar };
+
+        /// <summary>
+        /// Returns true if <paramref name="s"/> contains directory separator characters.
+        /// </summary>
+        public static bool ContainsDirectorySeparators(string s) =>
+            s.IndexOfAny(pathSeparatorsCharacters) >= 0;
```
warwickmm, line 308: I think this is short enough where we don't need the line break.

Duplicati/CommandLine/Commands.cs
```diff
@@ -234,14 +235,86 @@ public static int Affected(TextWriter outwriter, Action<Duplicati.Library.Main.C
...
+
+        private static readonly char[] wildcardCharacters = new[] { '*', '?' };
+
+        /// <summary>
+        /// Returns true if <paramref name="s"/> contains wildcard characters.
+        /// </summary>
+        public static bool ContainsWildcards(string s) =>
+            s.IndexOfAny(wildcardCharacters) >= 0;
```
warwickmm, line 316: I think this is short enough where we don't need the line break.

Duplicati/CommandLine/Commands.cs
```diff
@@ -234,14 +235,86 @@ public static int Affected(TextWriter outwriter, Action<Duplicati.Library.Main.C
...
-            var specialArgChars = new[] { '*', '?', Path.DirectorySeparatorChar, Path.AltDirectorySeparatorChar };
-            return argList.Select(arg => arg.IndexOfAny(specialArgChars) < 0
-                 && !arg.StartsWith("[", StringComparison.Ordinal) ? "*" + Util.DirectorySeparatorString + arg : arg).ToList();
+            var containsSeparators = ContainsDirectorySeparators(arg);
+            var containsWildcards = ContainsWildcards(arg);
+
+            if (!containsSeparators && arg.StartsWith("@", StringComparison.Ordinal))
+            {
+                // Convert to Regexp filter and prefix with ".*/"
+                return $"[.*{Regex.Escape(Util.DirectorySeparatorString + arg.Substring(1))}]";
```
warwickmm, line 257: Instead of `Regex.Escape` here, can we use the `Utility.ConvertLiteralToRegExp` helper method to isolate the conversion logic in one place?

Duplicati/CommandLine/Commands.cs
dferreyra: Fixed in latest commit.

Duplicati/CommandLine/Commands.cs
dferreyra: Fixed in latest commit.

Duplicati/CommandLine/Commands.cs
dferreyra: Fixed in latest commit.

Duplicati/CommandLine/Commands.cs
dferreyra: Fixed in latest commit.

dferreyra:
I see a failure in the AppVeyor with this last commit.  I saw it once myself, though I couldn't reproduce it and didn't know what to make of it:
```
1) Error : Duplicati.UnitTest.ProblematicPathTests.DirectoriesWithWildcards
System.Exception : The previous backup has time 9/7/2020 4:52:54 AM, but this backup has time 9/7/2020 4:52:53 AM. Something is wrong with the clock.
   at Duplicati.Library.Main.Operation.BackupHandler.<RunAsync>d__20.MoveNext() in C:\projects\duplicati\Duplicati\Library\Main\Operation\BackupHandler.cs:line 592
--- End of stack trace from previous location where exception was thrown ---
   at System.Runtime.ExceptionServices.ExceptionDispatchInfo.Throw()
   at CoCoL.ChannelExtensions.WaitForTaskOrThrow(Task task)
   at Duplicati.Library.Main.Operation.BackupHandler.Run(String[] sources, IFilter filter, CancellationToken token) in C:\projects\duplicati\Duplicati\Library\Main\Operation\BackupHandler.cs:line 349
   at Duplicati.Library.Main.Controller.<>c__DisplayClass14_0.<Backup>b__0(BackupResults result) in C:\projects\duplicati\Duplicati\Library\Main\Controller.cs:line 128
   at Duplicati.Library.Main.Controller.RunAction[T](T result, String[]& paths, IFilter& filter, Action`1 method) in C:\projects\duplicati\Duplicati\Library\Main\Controller.cs:line 468
   at Duplicati.Library.Main.Controller.Backup(String[] inputsources, IFilter filter) in C:\projects\duplicati\Duplicati\Library\Main\Controller.cs:line 124
   at Duplicati.UnitTest.ProblematicPathTests.DirectoriesWithWildcards() in C:\projects\duplicati\Duplicati\UnitTest\ProblematicPathTests.cs:line 165
```

warwickmm:
I've seen that failure occasionally when running the tests locally, but don't know what causes it.  My guess is that when multiple backups are run one after another that take very little time to complete, something is off.

I started another AppVeyor build to see if it happens again.

Duplicati/CommandLine/Commands.cs
```diff
@@ -267,7 +339,7 @@ public static int List(TextWriter outwriter, Action<Duplicati.Library.Main.Contr
                             options["version"] = v.ToString();
                         }
                     }
-                    else if (args[0].IndexOfAny(new char[] { '*', '?' }) < 0 && !args[0].StartsWith("[", StringComparison.Ordinal))
+                    else if (args[0].IndexOfAny(new char[] { '*', '?' }) < 0 && !args[0].StartsWith("[", StringComparison.Ordinal) && !args[0].StartsWith("@", StringComparison.Ordinal))
```
warwickmm, line 342: Can we use `!ContainsWildcards` instead of `IndexOfAny` here?

Duplicati/CommandLine/Commands.cs
```diff
@@ -234,14 +236,84 @@ public static int Affected(TextWriter outwriter, Action<Duplicati.Library.Main.C
             }
         }
 
-        public static List<string> PrefixArgsWithAsterisk(IList<string> argList)
+        /// <summary>
+        /// For bare file names with no wildcards, replace argument with
+        /// the equivalent of prefixing with "*/" so we search all folders.
+        /// </summary>
+        public static IEnumerable<string> PrefixArgsWithAsterisk(IEnumerable<string> argList) =>
+            argList.Select(PrefixArgWithAsterisk);
```
warwickmm, line 244: This is simple enough to be on one line.

Duplicati/CommandLine/Commands.cs
```diff
@@ -234,14 +236,84 @@ public static int Affected(TextWriter outwriter, Action<Duplicati.Library.Main.C
...
+                return arg;
+            }
+        }
+
+        /// <summary>
+        /// For folders, replace argument with the equivalent of suffixing
+        /// with "*" so we restore contents in the folder.
+        /// </summary>
+        public static IEnumerable<string> SuffixArgsWithAsterisk(IEnumerable<string> argList) =>
+            argList.Select(SuffixArgWithAsterisk);
```
warwickmm, line 276: This is simple enough to be on one line.

Duplicati/CommandLine/Commands.cs
dferreyra: Fixed in latest commit.

Duplicati/CommandLine/Commands.cs
dferreyra: Fixed in latest commit.

Duplicati/CommandLine/Commands.cs
dferreyra: Fixed in latest commit.

Duplicati/UnitTest/ProblematicPathTests.cs
```diff
@@ -21,6 +22,153 @@ private static void WriteFile(string path, byte[] contents)
...
+                directories.Add(dirWithQuestionMark);
+                questionMarkWildcardShouldMatchCount++;
+            }
+
+            // Include at least one single character directory in Windows
+            // for a '?' wildcard can match on
+            const string singleCharacterDir = "X";
+            string dirWithSingleCharacter = Path.Combine(this.DATAFOLDER, singleCharacterDir);
+            SystemIO.IO_OS.DirectoryCreate(dirWithSingleCharacter);
+            WriteFile(SystemIO.IO_OS.PathCombine(dirWithSingleCharacter, file), new byte[] { 1 });
```
warwickmm, line 65: Can we make the file contents here different than the others?  This way, we can hopefully detect a case where the restore operation restores the wrong file.

dferreyra: Fixed in latest commit.

warwickmm:
Thanks @dferreyra.  Looks good to me.

PR #4311, opened 2020-09-08
Add symlink policy tests

dferreyra:
This adds a basic symlink policy test that performs backups using each
symlink policy.

If you don't have the required privilege in Windows, you can get an
exception trying to create a symbolic link.  If an exception is thrown
trying to create a symbolic link, the exception is trapped and the
test is marked as "Ignored"; e.g.,
```
1) Ignored : Duplicati.UnitTest.SymLinkTests.SymLinkPolicy
Client could not create a symbolic link.  Error reported: (1314) A required privilege is not held by the client. | Read: [C:\td\backup-data\target] | Write: [C:\td\backup-data\symlink]
```

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -57,7 +57,7 @@ jobs:
         directories:
           - $BUILD_DIR
       script:
-        - ${ROOT_DIR}/pipeline/jobs/unittest_job.sh --testcategories SVNDataLong,SVNData,RecoveryTool,ProblematicPath
+        - ${ROOT_DIR}/pipeline/jobs/unittest_job.sh --testcategories SVNDataLong,SVNData,RecoveryTool,ProblematicPath,SymLink
     - stage: tests
       cache:
         directories:
--- a/.travis.yml
+++ b/.travis.yml
@@ -63,7 +63,7 @@ jobs:
         directories:
           - $BUILD_DIR
       script:
-        - ${ROOT_DIR}/pipeline/jobs/unittest_job.sh --testcategories Border
+        - travis_wait ${ROOT_DIR}/pipeline/jobs/unittest_job.sh --testcategories Border
     - stage: tests
       cache:
         directories:
--- a/Duplicati/UnitTest/BasicSetupHelper.cs
+++ b/Duplicati/UnitTest/BasicSetupHelper.cs
@@ -21,6 +21,7 @@ using System.Collections.Generic;
 using System.IO.Compression;
 using Duplicati.Library.Common;
 using Duplicati.Library.Common.IO;
+using Duplicati.Library.Utility;
 
 namespace Duplicati.UnitTest
 {
@@ -209,6 +210,17 @@ namespace Duplicati.UnitTest
             else
             {
                 ZipFile.ExtractToDirectory(sourceArchiveFileName, destinationDirectoryName);
+            }
+        }
+
+        /// <summary>
+        /// Write file <paramref name="path"/> with <paramref name="contents"/>.
+        /// </summary>
+        protected static void WriteFile(string path, byte[] contents)
+        {
+            using (FileStream fileStream = SystemIO.IO_OS.FileOpenWrite(path))
+            {
+                Utility.CopyStream(new MemoryStream(contents), fileStream);
             }
         }
     }
--- a/Duplicati/UnitTest/BasicSetupHelper.cs
+++ b/Duplicati/UnitTest/BasicSetupHelper.cs
@@ -212,17 +212,6 @@ namespace Duplicati.UnitTest
                 ZipFile.ExtractToDirectory(sourceArchiveFileName, destinationDirectoryName);
             }
         }
-
-        /// <summary>
-        /// Write file <paramref name="path"/> with <paramref name="contents"/>.
-        /// </summary>
-        protected static void WriteFile(string path, byte[] contents)
-        {
-            using (FileStream fileStream = SystemIO.IO_OS.FileOpenWrite(path))
-            {
-                Utility.CopyStream(new MemoryStream(contents), fileStream);
-            }
-        }
     }
 }
 
--- a/Duplicati/UnitTest/CommandLineOperationsTests.cs
+++ b/Duplicati/UnitTest/CommandLineOperationsTests.cs
@@ -216,7 +216,7 @@ namespace Duplicati.UnitTest
 
             ProgressWriteLine("Verifying partial restore ...");
             using (new Library.Logging.Timer(LOGTAG, "VerifiationOfPartialRestore", "Verification of partial restored files"))
-                TestUtils.VerifyDir(rf, RESTOREFOLDER, true);
+                TestUtils.AssertDirectoryTreesAreEquivalent(rf, RESTOREFOLDER, true, "VerifiationOfPartialRestore");
 
             systemIO.DirectoryDelete(RESTOREFOLDER, true);
 
@@ -226,7 +226,7 @@ namespace Duplicati.UnitTest
 
             ProgressWriteLine("Verifying partial restore ...");
             using (new Library.Logging.Timer(LOGTAG, "VerificationOfPartialRestore", "Verification of partial restored files"))
-                TestUtils.VerifyDir(rf, RESTOREFOLDER, true);
+                TestUtils.AssertDirectoryTreesAreEquivalent(rf, RESTOREFOLDER, true, "VerificationOfPartialRestore");
 
             systemIO.DirectoryDelete(RESTOREFOLDER, true);
 
@@ -237,7 +237,7 @@ namespace Duplicati.UnitTest
             ProgressWriteLine("Verifying full restore ...");
             using (new Library.Logging.Timer(LOGTAG, "VerificationOfFullRestore", "Verification of restored files"))
                 foreach (var s in systemIO.EnumerateDirectories(DATAFOLDER))
-                    TestUtils.VerifyDir(s, Path.Combine(RESTOREFOLDER, Path.GetFileName(s)), true);
+                    TestUtils.AssertDirectoryTreesAreEquivalent(s, Path.Combine(RESTOREFOLDER, Path.GetFileName(s)), true, "VerificationOfFullRestore");
 
             systemIO.DirectoryDelete(RESTOREFOLDER, true);
 
@@ -248,7 +248,7 @@ namespace Duplicati.UnitTest
             ProgressWriteLine("Verifying full restore ...");
             using (new Library.Logging.Timer(LOGTAG, "VerificationOfFullRestoreWithoutDb", "Verification of restored files"))
                 foreach (var s in systemIO.EnumerateDirectories(DATAFOLDER))
-                    TestUtils.VerifyDir(s, Path.Combine(RESTOREFOLDER, Path.GetFileName(s)), true);
+                    TestUtils.AssertDirectoryTreesAreEquivalent(s, Path.Combine(RESTOREFOLDER, Path.GetFileName(s)), true, "VerificationOfFullRestoreWithoutDb");
 
             ProgressWriteLine("Testing data ...");
             using (new Library.Logging.Timer(LOGTAG, "TestRemoteData", "Test remote data"))
--- a/Duplicati/UnitTest/CommandLineOperationsTests.cs
+++ b/Duplicati/UnitTest/CommandLineOperationsTests.cs
@@ -215,8 +215,8 @@ namespace Duplicati.UnitTest
                 Duplicati.CommandLine.Program.RealMain((new string[] { "restore", target, rf + "*", "--restore-path=\"" + RESTOREFOLDER + "\"" }.Union(opts)).ToArray());
 
             ProgressWriteLine("Verifying partial restore ...");
-            using (new Library.Logging.Timer(LOGTAG, "VerifiationOfPartialRestore", "Verification of partial restored files"))
-                TestUtils.AssertDirectoryTreesAreEquivalent(rf, RESTOREFOLDER, true, "VerifiationOfPartialRestore");
+            using (new Library.Logging.Timer(LOGTAG, "VerificationOfPartialRestore", "Verification of partial restored files"))
+                TestUtils.AssertDirectoryTreesAreEquivalent(rf, RESTOREFOLDER, true, "VerificationOfPartialRestore");
 
             systemIO.DirectoryDelete(RESTOREFOLDER, true);
 
--- a/Duplicati/UnitTest/ProblematicPathTests.cs
+++ b/Duplicati/UnitTest/ProblematicPathTests.cs
@@ -32,7 +32,7 @@ namespace Duplicati.UnitTest
             if (!Platform.IsClientWindows)
             {
                 SystemIO.IO_OS.DirectoryCreate(dirWithAsterisk);
-                WriteFile(SystemIO.IO_OS.PathCombine(dirWithAsterisk, file), new byte[] {0});
+                TestUtils.WriteFile(SystemIO.IO_OS.PathCombine(dirWithAsterisk, file), new byte[] {0});
                 directories.Add(dirWithAsterisk);
                 questionMarkWildcardShouldMatchCount++;
                 verbatimAsteriskShouldMatchCount++;
@@ -44,7 +44,7 @@ namespace Duplicati.UnitTest
             if (!Platform.IsClientWindows)
             {
                 SystemIO.IO_OS.DirectoryCreate(dirWithQuestionMark);
-                WriteFile(SystemIO.IO_OS.PathCombine(dirWithQuestionMark, file), new byte[] { 1 });
+                TestUtils.WriteFile(SystemIO.IO_OS.PathCombine(dirWithQuestionMark, file), new byte[] { 1 });
                 directories.Add(dirWithQuestionMark);
                 questionMarkWildcardShouldMatchCount++;
             }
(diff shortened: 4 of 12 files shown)


.travis.yml
```diff
@@ -57,7 +57,7 @@ jobs:
         directories:
           - $BUILD_DIR
       script:
-        - ${ROOT_DIR}/pipeline/jobs/unittest_job.sh --testcategories SVNDataLong,SVNData,RecoveryTool,ProblematicPath
+        - ${ROOT_DIR}/pipeline/jobs/unittest_job.sh --testcategories SVNDataLong,SVNData,RecoveryTool,ProblematicPath,SymLink
```
warwickmm, line 60: Can we add the `SymLink` category to `pipeline/start.sh`?  This file is used to run the Docker-based tests locally.

Duplicati/UnitTest/BasicSetupHelper.cs
```diff
@@ -211,6 +212,17 @@ protected static void ZipFileExtractToDirectory(string sourceArchiveFileName, st
                 ZipFile.ExtractToDirectory(sourceArchiveFileName, destinationDirectoryName);
             }
         }
+
+        /// <summary>
+        /// Write file <paramref name="path"/> with <paramref name="contents"/>.
+        /// </summary>
+        protected static void WriteFile(string path, byte[] contents)
```
warwickmm, line 219: This could be useful in other tests.  Can we move this to the `TestUtils` class?

Duplicati/UnitTest/SymLinkTests.cs
```diff
@@ -0,0 +1,157 @@
...
+                        systemIO.FileDelete($"{this.DBFILE}-journal");
+                    }
+                }
+            }
+        }
+
+        /// <summary>
+        /// Asserts that the two directories are equivalent; i.e., they they contain the same subdirectories and files, recursively.
+        /// </summary>
+        private static void AssertDirectoryTreesAreEquivalent(string expectedDir, string actualDir, string contextMessage)
```
warwickmm, line 110: The `TestUtils.VerifyDir` method seems to have a similar intent.  I do like how this recursive implementation is simpler.  Is there any reason to prefer one over the other?

Duplicati/UnitTest/SymLinkTests.cs
```diff
@@ -0,0 +1,157 @@
...
+            while (expectedFilesEnumerator.MoveNext() && actualFilesEnumerator.MoveNext())
+            {
+                AssertFilesAreEqual(expectedFilesEnumerator.Current, actualFilesEnumerator.Current, contextMessage);
+            }
+        }
+
+        /// <summary>
+        /// Asserts that two files are the same by comparing their size and their contents.
+        /// </summary>
+        private static void AssertFilesAreEqual(string expectedFile, string actualFile, string contextMessage)
```
warwickmm, line 136: The `TestUtils.CompareFiles` method seems to have a similar intent.  Is there any reason to prefer one over the other?

.travis.yml
dferreyra: Will do.  I see that the `ProblematicPath` category does not appear in `pipeline/start.sh`.  Should it?  (I'm not familiar with any of this -- I was just trying to follow patterns.)

warwickmm: `ProblematicPath` is added in pull request #4279.

warwickmm: I just merged #4279 so that we can avoid potential merge conflicts later.

Duplicati/UnitTest/SymLinkTests.cs
dferreyra: `TestUtils.VerifyDir` throws an exception for some problems (e.g., for missing folders or files), but only writes log messages for others (e.g., if the contents of two files are different).  On the other hand, `AssertDirectoryTreesAreEquivalent` uses the NUnit Assert mechanism for any problems it encounters, including if the contents of two files are different.  Also, `TestUtils.VerifyDir` (via `TestUtils.CompareFiles`) verifies some file metadata, while `AssertDirectoryTreesAreEquivalent` and `AssertFilesAreEqual` don't, though that would be easy to add.

I generally like using the NUnit Assert messages to capture context that can make the source of a failure more aparent.  It keeps the failure reasons localized in the NUnit failure report.  For example, here's what NUnit reports when `AssertDirectoryTreesAreEquivalent` fails due to file contents not being equal:
```
1) Failed : Duplicati.UnitTest.SymLinkTests.SymLinkPolicy
  Symlink policy: Store, file contents mismatch at position 0 for C:\D\td\backup-data\target\b.txt and C:\D\td\restored\symlink\b.txt
  Expected: 98
  But was:  65
```

I'll take this in any direction you'd like.  Just let me know how you'd like to proceed.

.travis.yml
dferreyra: Sounds good, thanks!

Duplicati/UnitTest/SymLinkTests.cs
warwickmm: I took a look at `TestUtils.VerifyDir`, and you're correct.  This implementation is much better.  Can we:
1. Move `AssertDirectoryTreesAreEquivalent` to `TestUtils`
2. Replace usages of `VerifyDir` with `AssertDirectoryTreesAreEquivalent`
3. Remove `VerifyDir`.
4. Move `AssertFilesAreEqual` to `TestUtils`
5. Add the ability for `AssertFilesAreEqual` to compare metadata
6. Replace usages of `CompareFiles` with `AssertFilesAreEqual`
7. Remove `CompareFiles`

dferreyra: Will do, thanks!

dferreyra: I've pushed a commit that includes these changes.

Duplicati/UnitTest/BasicSetupHelper.cs
dferreyra: Addressed in latest commit.

.travis.yml
dferreyra: Addressed in latest commit.

dferreyra:
I see the new symlink unit test is failing in the Travis CI run now that it's checking metadata.  I've since tried it on my Ubuntu machine and it fails the same way.  It always passes for me on Windows.

I've run it a few times now in Ubuntu and I see that it seems to be that the original file time has more precision than the restored file.  Some failures for the last write time:
```
  Expected: 2020-09-12 05:18:14.6245344
  But was:  2020-09-12 05:18:14.624534

  Expected: 2020-09-12 05:34:53.8480951
  But was:  2020-09-12 05:34:53.848095

  Expected: 2020-09-12 05:35:45.5730138
  But was:  2020-09-12 05:35:45.573013

  Expected: 2020-09-12 05:40:46.5941515
  But was:  2020-09-12 05:40:46.594151
```
I can look into it.

If this is expected, one way to deal with it is to change the DateTime comparison to require less precision; e.g., check if they are within, say, 1 ms or 100 ms of each other.

I also saw that one of the tests timed out in the Travis CI run.  At one point I did experience a hang myself running the full unit tests under Windows.  After a couple of hours, I broke in using the debugger and it appeared to be stuck in a call to a compression library:
https://github.com/duplicati/duplicati/blob/fcc44eacafe2f575e71c28670ec18995d10c1469/Duplicati/UnitTest/CompressionTests.cs#L44-L46
That's a different test category than what timed out in the Travis CI run, but I thought I'd mention it.  I've also gotten out of memory exceptions running the full unit tests under Windows, but I've never actually been that low in memory; e.g., at least 4 GB free, sometimes more.  I don't know if the out of memory exceptions and the hang I saw are related, but after the hang, I closed a bunch of applications and ran the full unit tests a few more times and they always passed.

dferreyra:
In the mean time, I've gone ahead and pushed a change that adds a 1 ms tolerance to the file time comparisons.

Duplicati/UnitTest/CommandLineOperationsTests.cs
```diff
@@ -216,7 +216,7 @@ private void DoRunCommands(string target)
 
             ProgressWriteLine("Verifying partial restore ...");
             using (new Library.Logging.Timer(LOGTAG, "VerifiationOfPartialRestore", "Verification of partial restored files"))
-                TestUtils.VerifyDir(rf, RESTOREFOLDER, true);
+                TestUtils.AssertDirectoryTreesAreEquivalent(rf, RESTOREFOLDER, true, "VerifiationOfPartialRestore");
```
warwickmm, line 219: Typo in "Verifiation"?

Duplicati/UnitTest/ProblematicPathTests.cs
```diff
@@ -96,7 +88,7 @@ public void DirectoriesWithWildcards()
                     {
                         string fileName = SystemIO.IO_OS.PathGetFileName(expectedFilePath);
                         string restoredFilePath = SystemIO.IO_OS.PathCombine(this.RESTOREFOLDER, directoryName, fileName);
-                        Assert.IsTrue(TestUtils.CompareFiles(expectedFilePath, restoredFilePath, expectedFilePath, false));
+                        TestUtils.AssertFilesAreEqual(expectedFilePath, restoredFilePath, false, expectedFilePath);
                     }
```
warwickmm, line 92: Is it possible to replace this `foreach` loop with a call to `AssertDirectoryTreesAreEquivalent`?

Duplicati/UnitTest/CommandLineOperationsTests.cs
dferreyra: Typos fixed in the latest commit.

Duplicati/UnitTest/TestUtils.cs
```diff
@@ -169,129 +170,88 @@ private static int IndexOf(List<string> lst, string m)
...
+            using (var actualFileStream = SystemIO.IO_OS.FileOpenRead(actualFile))
+            {
+                // Compare file lengths
+                var expectedFileStreamLength = expectedFileStream.Length;
+                var actualFileStreamLength = actualFileStream.Length;
+                Assert.That(actualFileStreamLength, Is.EqualTo(expectedFileStreamLength), $"{contextMessage}, file size mismatch for {expectedFile} and {actualFile}");
+                // Compare file contents
+                // The byte-by-byte compare is dog-slow, so we use a fast(-er) check, and then report the first byte diff if required
+                if (!Utility.CompareStreams(expectedFileStream, actualFileStream, true))
                 {
```
warwickmm, line 226: I tried testing this with two files of the same length containing different bytes, and this method didn't raise a failure.  After the call to `Utility.CompareStreams`, do we need to rewind the streams by setting their `Positions` to `0`?

dferreyra: Yes, I missed that.  I've added a commit for this.

Duplicati/UnitTest/TestUtils.cs
```diff
@@ -169,129 +170,88 @@ private static int IndexOf(List<string> lst, string m)
...
-                                Console.WriteLine("Mismatch in byte " + l.ToString() + " in file " + display);
-                                return false;
-                            }
+                        var expectedByte = expectedFileStream.ReadByte();
+                        var actualByte = actualFileStream.ReadByte();
+                        // For performance reasons, only exercise Assert mechanism and generate message if byte comparison fails
+                        if (expectedByte != actualByte)
+                        {
+                            var message =
+                                $"{contextMessage}, file contents mismatch at position {i} for {expectedFile} and {actualFile}";
```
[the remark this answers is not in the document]
dferreyra: Addressed in the most recent commit.

Duplicati/UnitTest/ProblematicPathTests.cs
```diff
@@ -96,7 +88,7 @@ public void DirectoriesWithWildcards()
...
                 }
```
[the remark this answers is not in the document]
dferreyra: Yes, that worked.  It's in the most recent commit.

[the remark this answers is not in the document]
warwickmm: Thanks.  Can we actually remove the outer `foreach` as well by comparing `this.DATAFOLDER` to `this.RESTOREFOLDER`?

Duplicati/UnitTest/SymLinkTests.cs
```diff
@@ -0,0 +1,107 @@
...
+using NUnit.Framework;
+
+namespace Duplicati.UnitTest
+{
+    [TestFixture]
+    public class SymLinkTests : BasicSetupHelper
+    {
+        [Test]
+        [Category("SymLink")]
+        public void SymLinkPolicy()
```
warwickmm, line 17: What are your thoughts about simplifying this by parameterizing the test with
```
[Test]
[Category("SymLink")]
[TestCase(Options.SymlinkStrategy.Store)]
[TestCase(Options.SymlinkStrategy.Follow)]
[TestCase(Options.SymlinkStrategy.Ignore)]
public void SymLinkPolicy(Options.SymlinkStrategy symlinkPolicy)
```
We can then remove the 
```
foreach (var symlinkPolicy in symlinkPolicies)
```
and remove all the file and directory deletion statements as well.

This also improves the test reporting by treating each parameter as a separate test.

dferreyra: That's very nice!  I've pushed a commit with this change.

Duplicati/UnitTest/ProblematicPathTests.cs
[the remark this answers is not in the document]
dferreyra: Sorry you had to spell it out!  I've pushed a commit with this change.

dferreyra:
@warwickmm, currently, the pull request includes this behavior described in the description:
> If you don't have the required privilege in Windows, you can get an
> exception trying to create a symbolic link.  If an exception is thrown
> trying to create a symbolic link, the exception is trapped and the
> test is marked as "Ignored"; e.g.,
> ```
> 1) Ignored : Duplicati.UnitTest.SymLinkTests.SymLinkPolicy
> Client could not create a symbolic link.  Error reported: (1314) A required privilege is not held by the client. | Read: [C:\td\backup-data\target] | Write: [C:\td\backup-data\symlink]
> ```

I originally did that because I didn't know what kinds of privilege the Windows CI builds ran under, but the tests run fine in the Windows CI.  I also wasn't sure if this would be a pain for people running the unit tests manually under Windows and encountering this failure because they weren't running "elevated".

Anyway, do you think it should remain like this, or maybe this behavior should be removed and let the test fail outright if it's not run "elevated" in Windows?

dferreyra:
I'm wondering about the "Border" category timeouts.  They seem to be persistent.  In my testing in Windows and Ubuntu, the tests pass, but they do seem to run a bit slower, maybe 1% to 4% slower under my recent commits.  I don't know if the slowdown is real -- I've only run the "Border" tests a couple of time in each environment.  But if it's real, could that be enough to push it over the edge and into a timeout?

warwickmm:
The timeout happens often, even before these changes.  Do you want to try increasing the timeout using `travis_wait` in `.travis.yml` to see if that helps?

https://docs.travis-ci.com/user/common-build-problems/#build-times-out-because-no-output-was-received

warwickmm:
> Anyway, do you think it should remain like this, or maybe this behavior should be removed and let the test fail outright if it's not run "elevated" in Windows?

I think we can keep the `Assert.Ignore`.  My IDE (Rider) shows "ignored tests", so it's not completely invisible.  Does yours show something similar?

Duplicati/UnitTest/SymLinkTests.cs
```diff
@@ -0,0 +1,87 @@
...
+                        Assert.That(restoredSymlinkFullPath, Is.EqualTo(symlinkTargetFullPath));
+                        break;
+                    case Options.SymlinkStrategy.Follow:
+                        // Restore should contain a regular directory with copies of the files in the symlink target
+                        Assert.That(systemIO.IsSymlink(restoreSymlinkDir), Is.False);
+                        TestUtils.AssertDirectoryTreesAreEquivalent(targetDir, restoreSymlinkDir, true, $"Symlink policy: {Options.SymlinkStrategy.Store}");
+                        break;
+                    case Options.SymlinkStrategy.Ignore:
+                        // Restore should not contain the symlink at all
+                        Assert.That(systemIO.DirectoryExists(restoreSymlinkDir), Is.False);
```
warwickmm, line 81: Would it hurt to check `systemIO.FileExists` as well just in case?

dferreyra: That sounds good.  I've pushed that change.

dferreyra:
Regarding the `Assert.Ignore`, leaving it in is good.  I've been running the unit tests via the command line.  That always prominently displays the ignore message.  I've used the Visual Studio Test Explorer before, but I haven't been able to get it working for the Duplicati project.

dferreyra:
Well, it looks like adding `travis_wait` has solved the timeout problem, thanks!

I think I understand the "Border" timeouts and why they're so consistent now.  The original `TestUtils.CompareFiles` would complain about file times not being exactly equal and would generate output, but not interrupt the test.  For example, from a Travis CI run before this pull request:
```
Sep 08 02:17:54 => Duplicati.UnitTest.BorderTests.RunMixedBlockFile_1
Sep 08 02:17:54 Mismatch in lastmodified for /root/duplicati_testdata/restored/b-p500, 09/08/2020 02:17:29 vs. 09/08/2020 02:17:29
Sep 08 02:17:54 Mismatch in create-time for /root/duplicati_testdata/restored/b-p500, 09/08/2020 02:17:29 vs. 09/08/2020 02:17:29
Sep 08 02:17:54 Mismatch in lastmodified for /root/duplicati_testdata/restored/b-b, 09/08/2020 02:17:29 vs. 09/08/2020 02:17:29
Sep 08 02:17:54 Mismatch in create-time for /root/duplicati_testdata/restored/b-b, 09/08/2020 02:17:29 vs. 09/08/2020 02:17:29
Sep 08 02:17:54 Mismatch in lastmodified for /root/duplicati_testdata/restored/a-b, 09/08/2020 02:17:27 vs. 09/08/2020 02:17:27
Sep 08 02:17:54 Mismatch in create-time for /root/duplicati_testdata/restored/a-b, 09/08/2020 02:17:27 vs. 09/08/2020 02:17:27
...
```
These messages would stave off the Travis CI timeout.

Since that was causing `TestUtils.AssertFilesAreEqual` to fail, I added the 1 ms tolerance.  But that means that the "Border" tests now run absolutely silently, so without `travis_wait`, the Travis CI times out after 10 minutes.

warwickmm:
Thanks @dferreyra.  Excellent work!

Issue #4295, opened 2020-08-25
Unable to start up after upgrade from 2.0.5.108 - to 2.0.5.110 -  System.ArgumentException: Illegal characters in path.


Athep:
Duplicati-crashlog:

> System.Reflection.TargetInvocationException: Exception has been thrown by the target of an invocation. ---> System.Exception: Ett allvarligt fel inträffade i Duplicati: System.Exception: Unable to start up, perhaps another process is already running?
Error message: System.ArgumentException: Illegal characters in path.
   at System.Security.Permissions.FileIOPermission.EmulateFileIOPermissionChecks(String fullPath)
   at System.IO.FileSystemEnumerableIterator`1..ctor(String path, String originalUserPath, String searchPattern, SearchOption searchOption, SearchResultHandler`1 resultHandler, Boolean checkHost)
   at System.IO.Directory.GetFiles(String path)
   at Duplicati.Library.Common.IO.SystemIOWindows.GetFiles(String path)
   at Duplicati.Server.SingleInstance..ctor(String basefolder)
   at Duplicati.Server.Program.CreateApplicationInstance(Boolean writeConsole)
   at Duplicati.Server.Program.CreateApplicationInstance(Boolean writeConsole)
   at Duplicati.Server.Program.RealMain(String[] _args) ---> System.Exception: Unable to start up, perhaps another process is already running?
Error message: System.ArgumentException: Illegal characters in path.
   at System.Security.Permissions.FileIOPermission.EmulateFileIOPermissionChecks(String fullPath)
   at System.IO.FileSystemEnumerableIterator`1..ctor(String path, String originalUserPath, String searchPattern, SearchOption searchOption, SearchResultHandler`1 resultHandler, Boolean checkHost)
   at System.IO.Directory.GetFiles(String path)
   at Duplicati.Library.Common.IO.SystemIOWindows.GetFiles(String path)
   at Duplicati.Server.SingleInstance..ctor(String basefolder)
   at Duplicati.Server.Program.CreateApplicationInstance(Boolean writeConsole)
   at Duplicati.Server.Program.CreateApplicationInstance(Boolean writeConsole)
   at Duplicati.Server.Program.RealMain(String[] _args)
   --- End of inner exception stack trace ---
   at Duplicati.GUI.TrayIcon.HostedInstanceKeeper..ctor(String[] args)
   at Duplicati.GUI.TrayIcon.Program.RealMain(String[] _args)
   --- End of inner exception stack trace ---
[...]

PR #4321, opened 2020-09-19
Disable "legacy path handling" for executables

dferreyra:
This is a defensive measure in case legacy path handling has been
enabled at the system level.  Duplicati requires that legacy path
handling be _disabled_ to function properly.

Issue #4295 appears to be a result of legacy path handling being
enabled globally.

This pull request changes the shipped app.config files for Duplicati EXEs to explicitly disable legacy path handling.  I did some very light testing that consisted of _enabling_ legacy path handling via the Windows registry (i.e., putting my system in a state that would trigger the exception reported in #4295), then confirming that each EXE worked at a very basic level.  Also, specifically for `Duplicati.CommandLine.exe` and `Duplicati.GUI.TrayIcon.exe`, I did confirm that without the app.config setting, both get the exception reported in #4295 with legacy path handling enabled globally, but do not get the exception with the app.config settings in this pull request.

At this point, I don't think we know if a change like this would actually fix the issue mentioned, but I thought I'd provide this in case it helps with investigating it.

### DIFF
--- a/Duplicati/CommandLine/BackendTester/app.config
+++ b/Duplicati/CommandLine/BackendTester/app.config
@@ -4,6 +4,7 @@
     <supportedRuntime version="v4.0" sku=".NETFramework,Version=v4.7.1" />
   </startup>
   <runtime>
+    <AppContextSwitchOverrides value="Switch.System.IO.UseLegacyPathHandling=false" />
     <assemblyBinding xmlns="urn:schemas-microsoft-com:asm.v1">
       <dependentAssembly>
         <assemblyIdentity name="Newtonsoft.Json" publicKeyToken="30ad4fe6b2a6aeed" culture="neutral" />
--- a/Duplicati/CommandLine/BackendTool/app.config
+++ b/Duplicati/CommandLine/BackendTool/app.config
@@ -4,6 +4,7 @@
     <supportedRuntime version="v4.0" sku=".NETFramework,Version=v4.7.1" />
   </startup>  
   <runtime>  
+    <AppContextSwitchOverrides value="Switch.System.IO.UseLegacyPathHandling=false" />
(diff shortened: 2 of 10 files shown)


warwickmm:
I am not familiar with this legacy path issue. The article [here](https://docs.microsoft.com/en-us/dotnet/framework/migration-guide/mitigation-path-normalization#what-is-path-normalization) seems to indicate that `Switch.System.IO.UseLegacyPathHandling=false` is the default behavior since we target .NET 4.7.1.  However, if it's possible for the OS registry to override this, I think it's reasonable to make our reliance on the new path handling behavior explicit.

How did you decide which `app.config` files to modify?  My local repository seems to have a lot of `app.config` files that I don't see in the set of changed files.

dferreyra:
Right, it is supposed to be disabled by default.  In trying to investigate #4295, I was able to enable the setting globally with the Windows Registry and break Duplicati.

My intent was to only make the change for Duplicati EXEs:
1. `Duplicati.CommandLine.BackendTester.exe`
1. `Duplicati.CommandLine.BackendTool.exe`
1. `Duplicati.CommandLine.ConfigurationImporter.exe`
1. `Duplicati.CommandLine.exe`
1. `Duplicati.CommandLine.RecoveryTool.exe`
1. `Duplicati.GUI.TrayIcon.exe`
1. `Duplicati.Library.AutoUpdater.exe`
1. `Duplicati.Library.Snapshots.exe`
1. `Duplicati.Server.exe`
1. `Duplicati.Service.exe`
1. `Duplicati.WindowsService.exe`

I was assuming the app.config for the "entry assembly" would be in control of this, so I thought that would just be EXEs.  I can't say that they all actually require it.  Most functions in `SystemIOWindows.cs` prefix with `\\?\`, so anything that calls those functions would probably break with legacy path handling enabled.

warwickmm:
Sounds good.  Thanks!

PR #4324, opened 2020-09-20
Telegram backend implementation

martikyan:
@warwickmm I reopened the PR because of the force pushes, sorry for the inconvenience. Fixed current comments. Thanks


Duplicati/Library/Backend/Telegram/TelegramBackend.cs
```diff
@@ -0,0 +1,473 @@
...
+using TLSharp.Core.Network.Exceptions;
+using TLSharp.Core.Utils;
+using TLRequestDeleteMessages = TeleSharp.TL.Channels.TLRequestDeleteMessages;
+
+namespace Duplicati.Library.Backend
+{
+    public class Telegram : IBackend
+    {
+        private const int MEBIBYTE_IN_BYTES = 1048576;
+        private static readonly string LOG_TAG = Log.LogTagFromType(typeof(Telegram));
```
martikyan, line 23: Keep the convention of names

Duplicati/Library/Backend/Telegram/TelegramBackend.cs
```diff
@@ -0,0 +1,473 @@
...
+
+        private void EnsureConnected()
+        {
+            var isConnected = false;
+
+            while (isConnected == false)
+            {
+                try
+                {
+                    isConnected = m_telegramClient.ConnectAsync().Wait(TimeSpan.FromSeconds(5));
```
martikyan, line 387: Make sure Wait() will not wrap FloodException, if so handle the specific case, or use another timeout logic

Duplicati/Library/Backend/Telegram/ChannelFileInfo.cs
```diff
@@ -0,0 +1,102 @@
...
+            {
+                return false;
+            }
+
+            if (ReferenceEquals(this, obj))
+            {
+                return true;
+            }
+
+            if (obj.GetType() != this.GetType())
```
martikyan, line 79: Do not use reflection while comparing the objects

Duplicati/Library/Backend/Telegram/Strings.cs
```diff
@@ -0,0 +1,63 @@
...
+        #region General
+
+        public static string DisplayName { get; } = "Telegram";
+        public static string Description => LC.L("This backend can read and write data to a Telegram backend");
+
+        #endregion
+
+        #region Errors
+
+        public static string NoOrWrongAuthCodeError => LC.L("The auth code is incorrect or missing");
```
martikyan, line 27: Separate wrong and missing auth code messages

Duplicati/Library/Backend/Telegram/TelegramBackend.cs
```diff
@@ -0,0 +1,473 @@
...
+            new CommandLineArgument(Strings.CHANNEL_NAME, CommandLineArgument.ArgumentType.String, Strings.ChannelName, Strings.ChannelName)
+        };
+
+        public string Description { get; } = Strings.Description;
+
+        public string[] DNSName { get; }
+
+        public void Test()
+        {
+            // No need to use our method
```
martikyan, line 277: Please clarify

Duplicati/Library/Backend/Telegram/TelegramBackend.cs
```diff
@@ -0,0 +1,475 @@
...
+using TeleSharp.TL.Messages;
+using TLSharp.Core;
+using TLSharp.Core.Exceptions;
+using TLSharp.Core.Network.Exceptions;
+using TLSharp.Core.Utils;
+using TLRequestDeleteMessages = TeleSharp.TL.Channels.TLRequestDeleteMessages;
+
+namespace Duplicati.Library.Backend
+{
+    public class Telegram : IBackend
```
warwickmm, line 20: Does this backend support streaming?  If so, we can implement `IStreamingBackend` instead of `IBackend` to get better performance.

Duplicati/Library/Backend/Telegram/TelegramBackend.cs
```diff
@@ -0,0 +1,475 @@
...
+                        cancelToken.ThrowIfCancellationRequested();
+                        var inputPeerChannel = new TLInputPeerChannel {ChannelId = channel.Id, AccessHash = (long)channel.AccessHash};
+                        var fileNameAttribute = new TLDocumentAttributeFilename
+                        {
+                            FileName = remotename
+                        };
+
+                        EnsureConnected();
+                        m_telegramClient.SendUploadedDocument(inputPeerChannel, file, remotename, "application/zip", new TLVector<TLAbsDocumentAttribute> {fileNameAttribute}, cancelToken).GetAwaiter().GetResult();
+                        fs.Close();
```
warwickmm, line 145: I don't think it's necessary to explicitly call `Close`.  The call to `Dispose` should take care of it.

Duplicati/Library/Backend/Telegram/TelegramBackend.cs
```diff
@@ -0,0 +1,475 @@
...
+                            catch (InvalidOperationException e)
+                            {
+                                if (e.Message.Contains("Couldn't read the packet length") == false)
+                                {
+                                    throw;
+                                }
+                            }
+                        }
+
+                        fs.Close();
```
warwickmm, line 183: I don't think it's necessary to explicitly call `Close`.  The call to `Dispose` should take care of it.

Duplicati/Library/Backend/Telegram/TelegramBackend.cs
martikyan: Fixed, thanks

Duplicati/Library/Backend/Telegram/TelegramBackend.cs
martikyan: Fixed, thanks

Duplicati/Library/Backend/Telegram/TelegramBackend.cs
martikyan: Fixed, thanks

Duplicati/Library/Backend/Telegram/TelegramBackend.cs
```diff
@@ -0,0 +1,504 @@
...
+
+            if (string.IsNullOrEmpty(m_phoneNumber))
+            {
+                throw new UserInformationException(Strings.NoPhoneNumberError, nameof(Strings.NoPhoneNumberError));
+            }
+
+            if (string.IsNullOrEmpty(m_channelName))
+            {
+                throw new UserInformationException(Strings.NoChannelNameError, nameof(Strings.NoChannelNameError));
+            }
```
warwickmm, line 90: If all these fields are required, we should display them by default in the UI.  For example, see the following examples for the Tardigrade backend:

* https://github.com/duplicati/duplicati/blob/v2.0.5.110-2.0.5.110_canary_2020-08-10/Duplicati/Library/Backend/Tardigrade/TardigradeConfig.cs
* https://github.com/duplicati/duplicati/pull/4197/files#diff-09aad22515a671afd5defa7c0a14c108
* https://github.com/duplicati/duplicati/blob/v2.0.5.110-2.0.5.110_canary_2020-08-10/Duplicati/Server/webroot/ngax/templates/backends/tardigrade.html

image

Duplicati/Library/Backend/Telegram/TelegramBackend.cs
```diff
@@ -0,0 +1,504 @@
...
+namespace Duplicati.Library.Backend
+{
+    public class Telegram : IStreamingBackend, IBackend
+    {
+        private const int MEBIBYTE_IN_BYTES = 1048576;
+
+        private static readonly InMemorySessionStore m_sessionStore = new InMemorySessionStore();
+        private static readonly string m_logTag = Log.LogTagFromType(typeof(Telegram));
+        private static readonly object m_lockObj = new object();
+        private static TelegramClient m_telegramClient;
```
martikyan, line 27: Remove telegram client staticness, think of better solution

Duplicati/Library/Backend/Telegram/TelegramBackend.cs
```diff
@@ -0,0 +1,504 @@
...
+                    return result;
+                },
+                nameof(List));
+        }
+
+        public Task PutAsync(string remotename, Stream stream, CancellationToken cancelToken)
+        {
+            SafeExecute(tc =>
+                {
+                    m_telegramClient = tc;
```
martikyan, line 128: No need for this, fix

Duplicati/Library/Backend/Telegram/TelegramBackend.cs
```diff
@@ -0,0 +1,504 @@
...
+                    var result = fileInfos.Select(fi => fi.ToFileEntry());
+                    return result;
+                },
+                nameof(List));
+        }
+
+        public Task PutAsync(string remotename, Stream stream, CancellationToken cancelToken)
+        {
+            SafeExecute(tc =>
+                {
```
martikyan, line 127: ThrowIfCancellationRequested

warwickmm:
Thanks @martikyan.  I was finally able to create an account and a channel to test this.  My first test was successful.  However, my second attempt resulted in an error "The auth code is incorrect".  Is the authentication code expected to change?

martikyan:
> Thanks @martikyan. I was finally able to create an account and a channel to test this. My first test was successful. However, my second attempt resulted in an error:
> 
> "The numeric code used to authenticate does not match the numeric code sent by SMS/Telegram".
> 
> Is the authentication code expected to change?

Dear @warwickmm 
Thanks for the review. The problem here is connected with the telegram session that is stored in the memory. After restarting the duplicati, It would try to reauthenticate with the same inputs and fails because Telegram sends a new auth code. We can solve this problem by writing the Telegram session into a temp file, but I don't know whether it is an acceptable solution for the project. If you agree I will change the code to write the session wherever you say.

warwickmm:
> After restarting the duplicati, it tries to reauthenticate with the same inputs and fails because Telegram sends a new auth code.

Sorry, I'm not familiar with Telegram so I don't quite understand this.  I also restarted Duplicati and tried to configure a new backup using the same channel and still encountered the error about the wrong authentication code.  I tried using old and new authentication codes that Telegram sent to my iOS app after logging out and logging in.

What situations require a new authentication code?

Also, do the API limits implied [here](https://stackoverflow.com/questions/47389575/telegram-api-development-tools-limits) apply?  Is there a limit to how many backups can be made in a certain time period?

martikyan:
There is a so-called session that contains the authenticated user with some information needed to get online and post files, read messages, etc. Currently, it is stored in the memory in a static field, which is not the best solution and causes problems like the one you mentioned. You restart the duplicati, and it forgets the session and Telegram asks for reauthentication and sends the code. So all we need is a place where I can serialize and store that info (the session).

The API limits:
Telegram has protection for DOS and flooding, so any client which shows too much activity is temporarily being blocked. It doesn't matter what library/app you use so the TLSharp is not the problem. The code in this PR works even Telegram detects flooding. We wait the time Telegram asks us to wait.

One more thing I want to mention about sessions: they contain sensitive information.

warwickmm:
Thanks.  I think I am starting to understand.  This sounds like it could be problematic, for example when users reboot their machine, or if they are using the command-line version of Duplicati.

Unfortunately, authentication is an area that I am not familiar with.  We may need other developers input (@kenkendk) to understand what options we have available.

Are there any OAuth approaches that would work with Telegram, similar to what's described below?

https://www.duplicati.com/articles/OAuth/
https://duplicati-oauth-handler.appspot.com/

martikyan:
Before @kenkendk will joint this conversation I want to mention out that actually there is OAuth available to authenticate the user of Telegram, but that will not help because it will not open the access to the files/channels.


You can check this out, but this won't work for us: [https://core.telegram.org/widgets/login](url)

warwickmm:
It looks like the authentication code is embedded in the [TargetURL](https://github.com/duplicati/duplicati/blob/v2.0.5.111-2.0.5.111_canary_2020-09-26/Duplicati/Server/Duplicati.Server.Serialization/Interface/IBackup.cs#L47).  I'm seeing a query string that looks like
```
telegram://t.me/?api-id=<api id>&api-hash=<api hash>&auth-code=<authentication code>&channel-name=<channel name>&phone-number=<phone number>&auth-password<authentication password>
```

I noticed that the `Telegram` constructor doesn't make use of the `url` parameter.  Could it be retrieved from here?  

Also, when does Telegram determine that a new authentication code is needed?  Would a new code ever need to be reentered?

martikyan:
> It looks like the authentication code is embedded in the [TargetURL](https://github.com/duplicati/duplicati/blob/v2.0.5.111-2.0.5.111_canary_2020-09-26/Duplicati/Server/Duplicati.Server.Serialization/Interface/IBackup.cs#L47). I'm seeing a query string that looks like
> 
> ```
> telegram://t.me/?api-id=<api id>&api-hash=<api hash>&auth-code=<authentication code>&channel-name=<channel name>&phone-number=<phone number>&auth-password<authentication password>
> ```
> 
> I noticed that the `Telegram` constructor doesn't make use of the `url` parameter. Could it be retrieved from here?
> 
> Also, when does Telegram determine that new authentication code is needed? Would a new code ever need to be re-entered?

The can actually be empty, only the protocol `telegram://` should be present. Retrieving needed information from options more simple, why should we make the code complex?

If you send the session information to the Telegram it will not ask for authentication. As soon as after restart we have no session, it asks for reauth. All these problems can be solved if we'll be able to store the session in a file so that after restarting Duplicati no more authentication will be needed.

I will make a separate PR for this. If the problem will still present, we would just revert that PR.

warwickmm:
> All these problems can be solved if we'll be able to store the session in a file so that after restarting Duplicati no more authentication will be needed.

Sounds good.  That's similar to how the SFTP backend refers to a local file for the private key.

martikyan:
> > All these problems can be solved if we'll be able to store the session in a file so that after restarting Duplicati no more authentication will be needed.
> 
> Sounds good. That's similar to how the SFTP backend refers to a local file for the private key.

I hope the checks will complete successfully.

The encrypted sessions implemented. Need to test under OSX and Linux

warwickmm:
Great work @martikyan.  I was only able to run some simple tests, but the session storage seems to be working well.

However, have you experienced extremely slow upload speeds?  I have a local test suite that runs various operations, and it never finishes.  At times, it seems that files are uploaded at a rate of maybe a few Kb every 30 minutes.  During this time, there is no apparent network activity, so I'm not sure what is happening.  Is this the result of hitting some API limit?

martikyan:
> Great work @martikyan. I was only able to run some simple tests, but the session storage seems to be working well.
> 
> However, have you experienced extremely slow upload speeds? I have a local test suite that runs various operations, and it never finishes. At times, it seems that files are uploaded at a rate of maybe a few Kb every 30 minutes. During this time, there is no apparent network activity, so I'm not sure what is happening. Is this the result of hitting some API limit?

Thanks, @warwickmm 
Yes, there are problems connected with reconnecting to Telegram servers. I'll ping you when I solve those.

martikyan:
Sorry for late fix ...

warwickmm:
Thanks @martikyan.  What's the easiest way to get a new authentication code?  I am having a lot of trouble establishing a successful connection.  If I use the "Login code" provided when I log into the iOS app, I get the following error when trying "Test connection"

image

martikyan:
Dear @warwickmm 
You can't use the code from the iOS app to authenticate in Duplicati, you have to test the connection with no auth code filled then you'll get one

warwickmm:
There's a possibility I got blocked by too may attempts.  I started to get something like PHONE_CODE_EXPIRED, and now I see the following when I leave the auth code blank:

image

martikyan:
That's because the backend is waiting for a valid code, and it knows the hash of it. You'd better restart Duplicati and try again.
I just successfully uploaded 30 GB of data to it.

warwickmm:
Thanks, restarting Duplicati worked and I was able to establish a connection.

Running some tests now... :smile:

warwickmm:
All my tests have passed so far.  Nice work.  :+1: 

I'll take a closer look at the code later when I have more time.

Duplicati/Library/Backend/Telegram/EncryptedFileSessionStore.cs
```diff
@@ -0,0 +1,157 @@
...
+
+        public EncryptedFileSessionStore(string password)
+        {
+            if (string.IsNullOrWhiteSpace(password))
+            {
+                throw new ArgumentNullException(nameof(password));
+            }
+
+            var appData = Environment.GetFolderPath(Environment.SpecialFolder.LocalApplicationData);
+            m_teleDataPath = Path.Combine(appData, nameof(Duplicati), nameof(Telegram));
```
warwickmm, line 29: We allow for [rebranding](https://github.com/duplicati/duplicati/wiki/Branding-and-OEM-customization).  Instead of `nameof(Duplicati)`, would it be better to reference [AutoUpdateSettings.AppName](https://github.com/duplicati/duplicati/blob/v2.0.5.111-2.0.5.111_canary_2020-09-26/Duplicati/Library/AutoUpdater/AutoUpdateSettings.cs#L160)?

Duplicati/Library/Backend/Telegram/TelegramBackend.cs
```diff
@@ -0,0 +1,512 @@
...
+using TeleSharp.TL.Messages;
+using TLSharp.Core;
+using TLSharp.Core.Exceptions;
+using TLSharp.Core.Network.Exceptions;
+using TLSharp.Core.Utils;
+using TLRequestDeleteMessages = TeleSharp.TL.Channels.TLRequestDeleteMessages;
+
+namespace Duplicati.Library.Backend
+{
+    public class Telegram : IStreamingBackend, IBackend
```
warwickmm, line 22: `IStreamingBackend` implements `IBackend`, so we can remove `IBackend` here.

Duplicati/Library/Backend/Telegram/TelegramBackend.cs
```diff
@@ -0,0 +1,512 @@
...
+                    var fileInfos = ListChannelFileInfos();
+                    var result = fileInfos.Select(fi => fi.ToFileEntry());
+                    return result;
+                },
+                nameof(List));
+        }
+
+        public Task PutAsync(string remotename, Stream stream, CancellationToken cancelToken)
+        {
+            Test();
```
warwickmm, line 130: What's the purpose of calling `Test()` here?

Duplicati/Library/Backend/Telegram/ChannelFileInfo.cs
```diff
@@ -0,0 +1,87 @@
...
+    public class ChannelFileInfo : IEquatable<ChannelFileInfo>
+    {
+        public int MessageId { get; set; }
+        public long MediaDocAccessHash { get; set; }
+        public long DocumentId { get; set; }
+        public int Version { get; set; }
+        public long Size { get; set; }
+        public string Name { get; set; }
+        public DateTime Date { get; set; }
+        public long AccessHash { get; set; }
```
warwickmm, line 16: The `AccessHash` property appears to be unused.  Can we remove it?

Duplicati/Library/Backend/Telegram/ChannelFileInfo.cs
```diff
@@ -0,0 +1,87 @@
...
+        }
+
+        public override bool Equals(object obj)
+        {
+            return ReferenceEquals(this, obj) || obj is ChannelFileInfo other && Equals(other);
+        }
+
+        public override int GetHashCode()
+        {
+            return MessageId;
```
warwickmm, line 74: It's usually not recommended for `GetHashCode` to depend on mutable fields.  Can `MessageId` be made `readonly`?

Duplicati/Library/Backend/Telegram/TelegramBackend.cs
```diff
@@ -0,0 +1,512 @@
...
+        private int? RetrieveMessages(TLInputPeerChannel inputPeerChannel, List<ChannelFileInfo> result, int maxDate)
+        {
+            EnsureConnected();
+            var absHistory = m_telegramClient.GetHistoryAsync(inputPeerChannel, offsetDate: maxDate).GetAwaiter().GetResult();
+            var history = ((TLChannelMessages)absHistory).Messages.OfType<TLMessage>();
+            var minDate = (int?)null;
+
+            foreach (var msg in history)
+            {
+                minDate = minDate == null || minDate > msg.Id ? msg.Date : minDate;
```
warwickmm, line 270: This looks strange, so I just want to confirm.  Do we really want to compare `minDate > msg.Id`?

Duplicati/Library/Backend/Telegram/Strings.cs
```diff
@@ -0,0 +1,64 @@
...
+        public static string PasswordShort => LC.L("2FA password (if enabled)");
+        public static string PasswordLong => LC.L("The 2 step verification password. Input only if you have set it up");
+        public static string ChannelName => LC.L("The channel name of the backup");
+
+        #endregion
+
+        #region Formats
+
+        public const string TELEGRAM_FLOOD = "It's required to wait {0} seconds before continuing";
+        public const string EXCEPTION_RETRY = "An exception was thrown, retrying";
```
warwickmm, line 57: ` EXCEPTION_RETRY` appears to be unused.  Can we remove it?

Duplicati/Library/Backend/Telegram/TelegramBackend.cs
martikyan: Ensuring that the client is Authenticated was mandatory. So I replaced Test with Authenticate()

Duplicati/Library/Backend/Telegram/TelegramBackend.cs
martikyan: I have refactored the expression, so it is more clear

Duplicati/Library/Backend/Telegram/ChannelFileInfo.cs
martikyan: Surely

Duplicati/Library/Backend/Telegram/TelegramBackend.cs
warwickmm: Thanks.  Sorry, I'm not familiar with the Telegram data model, but I just want double check again.  Should this really be `minDate < msg.Id` instead of `minDate < msg.Date`?

martikyan: You are correct there was a bug there. I refactored the code again. I didn't notice it, sorry :)

martikyan: It was working because the messages were coming in the correct order, which we cannot lean on. Fixed, thank you

Duplicati/Library/Backend/Telegram/TelegramBackend.cs
```diff
@@ -0,0 +1,516 @@
...
+            var inputPeerChannel = new TLInputPeerChannel {ChannelId = channel.Id, AccessHash = channel.AccessHash.Value};
+            var result = new List<ChannelFileInfo>();
+            var oldMinDate = 0;
+            var newMinDate = (int?)null;
+
+            while (oldMinDate != newMinDate)
+            {
+                oldMinDate = newMinDate ?? 0;
+                RetrieveMessages(inputPeerChannel, result, oldMinDate);
+                newMinDate = result.Min(cfi => (int)cfi.Date.GetEpochSeconds());
```
warwickmm, line 256: The `Utility.NormalizeDateTimeToEpochSeconds` method already exists.  Does that work here?  Note that it computes the value in terms of the UTC time.

warwickmm:
I'm also now encountering an error during an initial backup.  Are you able to reproduce this?
```
System.InvalidOperationException: Sequence contains no elements
  at Duplicati.Library.Main.BackendManager.List () [0x0005a] in <71391d08f48d4493bef657a91f8b3bd2>:0 
  at Duplicati.Library.Main.Operation.FilelistProcessor.RemoteListAnalysis (Duplicati.Library.Main.BackendManager backend, Duplicati.Library.Main.Options options, Duplicati.Library.Main.Database.LocalDatabase database, Duplicati.Library.Main.IBackendWriter log, System.Collections.Generic.IEnumerable`1[T] protectedFiles) [0x0000e] in <71391d08f48d4493bef657a91f8b3bd2>:0 
  at Duplicati.Library.Main.Operation.FilelistProcessor.VerifyRemoteList (Duplicati.Library.Main.BackendManager backend, Duplicati.Library.Main.Options options, Duplicati.Library.Main.Database.LocalDatabase database, Duplicati.Library.Main.IBackendWriter log, System.Collections.Generic.IEnumerable`1[T] protectedFiles) [0x00001] in <71391d08f48d4493bef657a91f8b3bd2>:0 
  at Duplicati.Library.Main.Operation.BackupHandler.PreBackupVerify (Duplicati.Library.Main.BackendManager backend, System.String protectedfile) [0x00133] in <71391d08f48d4493bef657a91f8b3bd2>:0 
  at Duplicati.Library.Main.Operation.BackupHandler.RunAsync (System.String[] sources, Duplicati.Library.Utility.IFilter filter, System.Threading.CancellationToken token) [0x014fa] in <71391d08f48d4493bef657a91f8b3bd2>:0 
  at CoCoL.ChannelExtensions.WaitForTaskOrThrow (System.Threading.Tasks.Task task) [0x00050] in <9a758ff4db6c48d6b3d4d0e5c2adf6d1>:0 
  at Duplicati.Library.Main.Operation.BackupHandler.Run (System.String[] sources, Duplicati.Library.Utility.IFilter filter, System.Threading.CancellationToken token) [0x0000a] in <71391d08f48d4493bef657a91f8b3bd2>:0 
  at Duplicati.Library.Main.Controller+<>c__DisplayClass14_0.<Backup>b__0 (Duplicati.Library.Main.BackupResults result) [0x0004d] in <71391d08f48d4493bef657a91f8b3bd2>:0 
  at Duplicati.Library.Main.Controller.RunAction[T] (T result, System.String[]& paths, Duplicati.Library.Utility.IFilter& filter, System.Action`1[T] method) [0x0029a] in <71391d08f48d4493bef657a91f8b3bd2>:0 
  at Duplicati.Library.Main.Controller.Backup (System.String[] inputsources, Duplicati.Library.Utility.IFilter filter) [0x0007a] in <71391d08f48d4493bef657a91f8b3bd2>:0 
  at Duplicati.CommandLine.Commands.Backup (System.IO.TextWriter outwriter, System.Action`1[T] setup, System.Collections.Generic.List`1[T] args, System.Collections.Generic.Dictionary`2[TKey,TValue] options, Duplicati.Library.Utility.IFilter filter) [0x0013b] in <700018d1612946c39533af5802846804>:0 
  at (wrapper delegate-invoke) System.Func`6[System.IO.TextWriter,System.Action`1[Duplicati.Library.Main.Controller],System.Collections.Generic.List`1[System.String],System.Collections.Generic.Dictionary`2[System.String,System.String],Duplicati.Library.Utility.IFilter,System.Int32].invoke_TResult_T1_T2_T3_T4_T5(System.IO.TextWriter,System.Action`1<Duplicati.Library.Main.Controller>,System.Collections.Generic.List`1<string>,System.Collections.Generic.Dictionary`2<string, string>,Duplicati.Library.Utility.IFilter)
  at Duplicati.CommandLine.Program.ParseCommandLine (System.IO.TextWriter outwriter, System.Action`1[T] setup, System.Boolean& verboseErrors, System.String[] args) [0x003df] in <700018d1612946c39533af5802846804>:0 
  at Duplicati.CommandLine.Program.RunCommandLine (System.IO.TextWriter outwriter, System.IO.TextWriter errwriter, System.Action`1[T] setup, System.String[] args) [0x00004] in <700018d1612946c39533af5802846804>:0 
```

martikyan:
@warwickmm Sorry for this too, I am working on the fix, will be ready in few minutes.

Duplicati/Library/Backend/Telegram/TelegramBackend.cs
martikyan: Sorry, I am unable to find the method. Are you sure its' name is correct?

martikyan: Found it, I'll use this, will be ready in few minutes

warwickmm:
Thanks @martikyan, looks good to me.  Since my tests are passing, let's merge.  We can always make improvements later as we find them.  Nice job.

martikyan:
> Thanks @martikyan, looks good to me. Since my tests are passing, let's merge. We can always make improvements later as we find them. Nice job.

@warwickmm Thank you too for review,
I will use the backend, and in case of issues will fix them

PR #4332, opened 2020-09-28
Disable "legacy path handling" for executables

dferreyra:
This disables legacy path handling by changing the
`AssemblyRedirects.xml` that's deployed as the app.config files in the
installation image.

Issue #4295 appears to be a result of legacy path handling being
enabled globally.


warwickmm:
Thanks @dferreyra.  I wasn't aware of `AssemblyRedirects.xml` either.

PR #4351, opened 2020-10-23
Added PLD PLB RW2

dolorosus:
PLD, PLB are picture formats from Photoline
RW2 is Panasonics RAW format

### DIFF
--- a/Duplicati/Library/Main/default_compressed_extensions.txt
+++ b/Duplicati/Library/Main/default_compressed_extensions.txt
@@ -93,11 +93,14 @@
 .jpg #JPEG Image
 .pamp #PowerAmp Image Cache
 .pdn #Paint.NET Image
+.pld # Photoline Image
+.plb # Photoline Image Cache
 .png #Portable Network Graphic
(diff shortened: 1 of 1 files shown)


Duplicati/Library/Main/default_compressed_extensions.txt
```diff
@@ -93,11 +93,14 @@
 .jpg #JPEG Image
 .pamp #PowerAmp Image Cache
 .pdn #Paint.NET Image
+.pld # Photoline Image
+.plb # Photoline Image Cache
```
warwickmm, line 97: It appears that `pld` and `plb` are also used by other applications (Oracle, Adobe Premiere, ...).  I don't know how common these are, but if these extensions aren't primarily associated with Photoline, we might not want to include them here by default if the Oracle and Adobe usages can benefit from compression.

Note that the user can also provide a custom extensions file using the `--compression-extension-file` option if necessary.

warwickmm:
I'm going to close this until we get more clarity on the `.pld` and `.plb` filetypes.

PR #4354, opened 2020-10-29
Upload filesets last during a backup

### LINKED ISSUE #4341: Interrupting backup can leave hidden damage so Recreate will download all dblocks then fail from missing blocks

## Environment info

- **Duplicati version**: 2.0.5.111_canary_2020-09-26 but I think this issue became possible in 2.0.4.16_canary_2019-03-28
- **Operating system**: Windows 10 Version 2004
- **Backend**: Local folder

## Description

Parallel uploads allow the dlist to upload faster than the larger dblocks. An interruption leaves a backup with missing dblocks. 
Testing from the GUI doesn't reveal this because the corrupted version is not in the database, but remains at the destination.
Fixing GUI's Recreate error is easy, at least here, but any issue that causes the entire maybe-large backup to download is bad.
Workaround may be asynchronous-concurrent-upload-limit cut from 4 to 1 so upload is dblock/dindex pairs, then a dlist last.

## Steps to reproduce

1. Add a backup to local folder of a source folder with a Duplicati binary zip file renamed to file.zip, using no encryption, throttle-upload=1MB, and console-log-level=information. The latter two are to slow it down and allow you to see when to interrupt it.
2. Export as command-line.
3. On Windows, set AUTOUPDATER_Duplicati_SKIP_UPDATE=true to ensure later Ctrl-C kills Duplicati (prevents a child process).
4. Start the exported command line, and press Ctrl-C after the dlist file has completed its uploading.
You will see a dlist and partial dblock. The partial dblock will be deleted at the next backup's start.
5. Update file.zip to a different renamed Duplicati binary zip file.
6. Start the exported command line, and let the backup finish.
You now have a dblock, dindex, and dlist from second backup, and dlist from interrupted backup.
7. In the GUI, you can attempt to check backup with test command for all files, list-broken-files, and Database Repair. All OK.
You now see one version on the restore dropdown, despite having two dlist files in the backup and nothing has noticed it.
8. Get About --> Show log --> Live --> Verbose ready to see if it winds up downloading all dblocks.
9. Now try Recreate, and watch it download all dblocks. This would be very slow with a large backup.
Recreate pops up a Red error:
Recreated database has missing blocks and 1 broken filelists. Consider using "list-broken-files" and "purge-broken-files" to purge broken data from the remote store and the database.
You now see two versions on the restore dropdown, although the older one had no dblock uploaded.
list-broken-files shows that older one has a broken file. purge-broken-files purges the empty version.

- **Actual result**:
All looked fine, but a Recreate downloaded all dblocks then failed.
- **Expected result**:
Error prevented or detected, and Recreate downloads no dblocks.

seantempleton:
If parallel uploads are enabled it is possible for the fileset to be uploaded before all dblocks are uploaded.
When the fileset is placed in the queue to be uploaded place it in a separate list and only upload it after all dblocks have been uploaded.

Fixes #4341

### DIFF
--- a/Duplicati/Library/Main/Operation/Backup/BackendUploader.cs
+++ b/Duplicati/Library/Main/Operation/Backup/BackendUploader.cs
@@ -39,12 +39,12 @@ namespace Duplicati.Library.Main.Operation.Backup
     {
         public Task<long> LastWriteSizeAsync { get { return m_tcs.Task; } }
         private readonly TaskCompletionSource<long> m_tcs = new TaskCompletionSource<long>();
-        
+
         public void SetFlushed(long size)
         {
             m_tcs.TrySetResult(size);
         }
-        
+
         public void TrySetCanceled()
         {
             m_tcs.TrySetCanceled();
@@ -108,7 +108,6 @@ namespace Duplicati.Library.Main.Operation.Backup
         private readonly StatsCollector m_stats;
         private readonly ITaskReader m_taskReader;
 
-
         public BackendUploader(Func<IBackend> backendFactory, Options options, DatabaseCommon database, ITaskReader taskReader, StatsCollector stats)
         {
             m_backendFactory = backendFactory;
@@ -135,6 +134,7 @@ namespace Duplicati.Library.Main.Operation.Backup
                 var uploadsInProgress = 0;
                 m_cancelTokenSource = new CancellationTokenSource();
                 m_progressUpdater.Run(m_cancelTokenSource.Token);
+                var filesetsToUpload = new List<FilesetUploadRequest>();
 
                 try
                 {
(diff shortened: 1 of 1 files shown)


Duplicati/Library/Main/Operation/Backup/BackendUploader.cs
```diff
@@ -240,6 +224,49 @@ public Task Run()
...
+
+                Worker finishedWorker = workers.Single(w => w.Task == finishedTask);
+                workers.Remove(finishedWorker);
+                finishedWorker.Dispose();
+            }
+        }
+
+        private void UploadFilesets(List<Worker> workers, List<FilesetUploadRequest> filesets)
+        {
+            if (filesets.Any())
```
warwickmm, line 258: Is this `if` statement necessary given the `foreach` below?

seantempleton: Good catch. You're right, I'll remove it.

warwickmm:
Thanks @seantempleton.  Looks good to me.

Issue #3920, opened 2019-09-21
use-background-io-priority warnings


neekt:
- [x] I have searched open and closed issues for duplicates.

----------------------------------------
## Environment info
- **Duplicati version**: 2.0.4.23_beta_2019-07-14
- **Operating system**: Linux Mint 19.2 Cinnamon, Linux Kernel 5.0.0-29-generic
- **Backend**: Dropbox

## Description
A few days ago, I selected use-background-io-priority in settings, to try to free up system resources a bit while my scheduled backups are running. The option seems to work very well - substantially less slowing down of everything when backups are running - but I now get a warning for every backup that is run, every time each backup is run. However, when I look at the log file, there doesn't appear to be any information about the warnings at all.

## Steps to reproduce
1. Using Duplicati browser GUI, set up a back up of any directory, using Dropbox as the backend, with scheduled hourly backups, default AES-256 encryption, default remote volume size (50 MB)
2. Select use-background-io-priority in settings, leaving all other settings unchanged
3. Run the back up (either manually, or wait for the hourly scheduled backup)

- **Actual result**:
After the backup runs (successfully backing up files), the red warning icon appears in the system tray, and the following dialogue appears in the browser GUI (example):
```
Warning while running Photos
Got 1 warning(s)
Dismiss Show
```
Selecting 'Show' takes me to the relevant log, where it gives me no indication of what the warning is about.

If I deselect the use-background-io-priority option and re-run the backup, no warning appears.

- **Expected result**:
Either there shouldn't be a warning dialogue, or there should be information in the log about the warning.

## Debug log
Example of a backup log of my 'Photos' directory: https://pastebin.com/k7LA8yer

Thank you for maintaining this software, it has worked extremely well for me over the ~6 months I have been using it!

drwtsn32x:
I tried to reproduce your issue but without success. Perhaps the backend choice does matter (I wouldn't have thought so).

But sometimes warnings show up not in the job log but in the main Live log.  After you get the warning, go to the Duplicati Web UI -> About -> Show Log -> Live, and then select Warning from the dropdown.  Does the warning message appear there?

neekt:
Ah, there we go. Here's the warning I got in the live log:

`Oct 15, 2019 12:37 PM: Failed to reset background IO priority, exitcode: 0, stderr: ionice: ignoring given class data for none class`

Does this mean the IO priority is not working, and that I imagined it was benefiting me? :D

drwtsn32x:
[...]

PR #4359, opened 2020-11-05
Bugfix issue #3920

achims311:
only "best-effort" and "realtime" allow a priotity level
"none" and "idle" have no prioity level


warwickmm:
Thanks @achims311.  I can confirm that this fixes the warning.

PR #4371, opened 2020-11-19
better description of the Remote-Volume-Size de translation

apreiml:
Hi,

the description of the remote volume size is misleading. I mistook this for the maximum free space of the remote storage, because it says: "This backup does not relate to your maximum backup or file size." Therefore I thought that it does not relate to the backup volumes.

I'd like to suggest a different description to make clear, that the backup is split into volumes and this setting sets the volume size.


warwickmm:
Thanks @apreiml.  The translations should actually be fixed in our [transifex project](https://www.transifex.com/duplicati/duplicati/).  Can you make the changes there?  The updated translations will be pulled for the next release.

apreiml:
@warwickmm Ok. I just wanted to get feedback, if this makes sense. Should the base string also be changed, so that the other languages also notice the chances? Then I guess it needs a code change also.

warwickmm:
Yes, I think the base string would need to be changed first.  We can have a discussion on how to improve the description there.

PR #4372, opened 2020-11-20
Improve wording of the remote volume size description

apreiml:
Follow up of https://github.com/duplicati/duplicati/pull/4371

The current wording only says what it not is and not what it is. It is also misleading as it hints that it does not relate to the maximum backup or file size. But imho it effects and therefore relates to the sizes of the backup files.

I tried to describe it more cleary in this merge request.


apreiml:
Also the addoredit.html had the execute flag set, which I removed.

warwickmm:
Thanks @apreiml.  This option often confuses users so it would help for the description to be improved.

I think it makes sense for this description to be consistent with the one used in the [advanced options](https://github.com/duplicati/duplicati/blob/v2.0.5.111-2.0.5.111_canary_2020-09-26/Duplicati/Library/Main/Strings.cs#L78):  
image

Feel free to suggest improvements to both descriptions.  The [documentation](https://duplicati.readthedocs.io/en/latest/06-advanced-options/#dblock-size) and [this article](https://www.duplicati.com/articles/Choosing-Sizes/#remote-volume-size) might be helpful.

I don't know how this would affect existing translations.  @ts678, @drwtsn32x, @kenkendk, do you know if changing this would break existing translations?

apreiml:
@warwickmm Ok. For the advanced options this is ok. Maybe a reference would be helpful, that this is the same setting.

For the description above, I would not expect the user to be familiar with how exactly duplicati stores the files and how the files are called. Imho the user just needs to know that duplicati archives the backup into several files of similar size
(besides the index files), and the maximum file size can be changed here. Because this is enough information, to know what the setting is about. If he wants to know more details, he can follow the link to the manual.

ts678:
> do you know if changing this would break existing translations?

I don't do much in translation, but I think @kees-z did the Dutch. My guess is that it would revert to untranslated (English) until translated, however lots of languages are only partially translated. You can figure out the state of specific strings by setting UI to language of your choice then navigating by experience (and maybe guess), or by text-searching the [.po files](https://www.gnu.org/software/gettext/manual/html_node/PO-Files.html) for English snippets.

Usually what I do is search for translated snippets from forum posts to try to get back to the original message to see what it is...

https://github.com/duplicati/duplicati/blob/1c3065aa814d6253ed033d3584c7171708fc9bf0/Localizations/webroot/localization_webroot-sv_SE.po#L271-L273

https://github.com/duplicati/duplicati/blob/1c3065aa814d6253ed033d3584c7171708fc9bf0/Localizations/webroot/localization_webroot-sv_SE.po#L1472-L1474

https://github.com/duplicati/duplicati/blob/1c3065aa814d6253ed033d3584c7171708fc9bf0/Localizations/webroot/localization_webroot-sv_SE.po#L1795-L1797

image

Exactly how the rhythm runs between Duplicati changes and Transifex translators is unknown to me, but @kenkendk gets results. Ideally it seems like we'd like to have message changes done early enough before release to allow time for translation to be done.

warwickmm:
I think the new description is an improvement.  The translations will always take some time to catch up, but hopefully this results in fewer inadvertent bad size settings.

PR #4376, opened 2020-11-21
Fix incorrect or misleading indentations

warwickmm:
This also adds braces to clarify scope where beneficial.


warwickmm:
The diff looks pretty ugly, but if you check "Hide whitespace changes" it should be clearer that the changes mainly involve braces and indentation.

kenkendk:
I guess this means we should have a style-guide?
Also, we should have a CI that enforces the styles?

warwickmm:
Having a style guide and CI  would be nice, but enforcement can be tricky.  One option is to define a style in ReSharper or some other tool and have it fix everything, but that would be quite disruptive when looking at a blame.  For now, I just look for the egregious violations (e.g., no indentation after an `if`) and fix those manually.

PR #4383, opened 2020-12-05
Update localization source to reflect new remote volume size description

warwickmm:
The description of the remote volume size was updated in pull request #4372.  This updates the localization source, and inlines the HTML text to avoid potential whitespace/newline issues in the translation.


warwickmm:
This is probably not necessary, as the `localization_webroot.pot` file appears to be generated automatically:

https://github.com/duplicati/duplicati/blob/7d958adb8a71b334834ce42ecc9571c78e8a259e/Localizations/webroot/extract.sh#L6-L6

Issue #4308, opened 2020-09-04
Duplicati stuck at 0% backuping own SQLITE database file in appdata


Naklarsuper:
Backup seemed to progress until the file xxxxx.sqlite in Duplicati appdata local folder (the database for this very backup) remained stuck at 0%. Had to stop Duplicati process via task manager, remove medium, and restart from scratch, repair database, then it worked again. 

Could not reproduce so far. 

Latest Windows 10, Duplicati - 2.0.5.1_beta_2020-01-18

drwtsn32x:
You should exclude the Duplicati databases from your backup.  There is no benefit to having a backup job back up its own sqlite database.  If the database ever got deleted or damaged, it would need to be recreated from the back end data before any restores could be done.  And once it's recreated you wouldn't need to restore it anyway!

Naklarsuper:
Understood, thank you. So what I have observed was expected behavior that can be prevented by excluding the databases?

drwtsn32x:
Yes, I recall seeing a couple other reports over the years of Duplicati failing when it tried to back up its own database.

IMO Duplicati should probably auto-exclude the job specific sqlite file from the backup.

Naklarsuper:
Agree. Thank you. Closing the issue.

Issue #3416, opened 2018-10-08
Repair command deletes remote files for new backups


Livven:
- [x] I have searched open and closed issues for duplicates.

----------------------------------------

## Environment info

- **Duplicati version**: 2.0.3.3_beta_2018-04-02
- **Operating system**: Windows 10 1803
- **Backend**: OneDrive

## Description

Running the database repair command can cause data loss. This should never happen.

## Steps to reproduce

1. Add backup on PC 1 for an existing backup from another PC 2 so that files from PC 2 can be easily restored to PC 1.
2. Restore some files on PC 1.
3. Run the backup on PC 2.
4. Run the database repair command on PC 1.

- **Actual result**:
All new backup files (from step 3) are deleted from the remote.
- **Expected result**:
The new backup (from step 3) should now be available and selectable from the file restore view on PC 1. At a minimum the files should not be deleted...

## Debug log

```
MainOperation: Repair
RecreateDatabaseResults: null
ParsedResult: Error
EndTime: 09.10.2018 00:19:01 (1539037141)
BeginTime: 09.10.2018 00:16:52 (1539037012)
Duration: 00:02:09.3987071
Messages: []
Warnings: []
Errors: [
    Failed to accept new index file: duplicati-i0c1445caa4b845d882ce766251164713.dindex.zip.aes, message: Volume duplicati-b8ef173135e134164af8b16c6cea4b94b.dblock.zip.aes has local state Deleting => Volume duplicati-b8ef173135e134164af8b16c6cea4b94b.dblock.zip.aes has local state Deleting,
    Failed to accept new index file: duplicati-i0e57eee9039b437c941ffe354faf0a0f.dindex.zip.aes, message: Volume duplicati-b64286173307b4f45bae7e1c35cfbcbf7.dblock.zip.aes has local state Deleting => Volume duplicati-b64286173307b4f45bae7e1c35cfbcbf7.dblock.zip.aes has local state Deleting,
    Failed to accept new index file: duplicati-i1a724a86bf4040a8ba560c543ebd7fe1.dindex.zip.aes, message: Volume duplicati-b7c8ecc9280c54cf8a62750800c5719d8.dblock.zip.aes has local state Deleting => Volume duplicati-b7c8ecc9280c54cf8a62750800c5719d8.dblock.zip.aes has local state Deleting,
[...]

PR #4390, opened 2020-12-11
Always exclude job-specific database files from backup

### LINKED ISSUE #1797: Completing backup endless
Hi,

I have successfully created an initial archive (768.94 GB archived to 50GB files for a total of 303.17GB).
I am now trying to see if incremental runs properly and timely. It seems fast (less than 30 minuets) but get stuck at Completing Backup... for more than 12 hours. This morning I've got this error:

  Failed to process path: /home/CAMPUS2/namlb/.config/Duplicati/UWVLTBPPDV.sqlite-journal
System.Exception: Unexpected result count: 0, expected 1 at Duplicati.Library.Main.Database.LocalBackupDatabase.AddBlockset (System.String filehash, Int64 size, Int32 blocksize, IEnumerable`1 hashes, IEnumerable`1 blocklistHashes, System.Int64& blocksetid, IDbTransaction transaction) [0x00000] in <filename unknown>:0 at Duplicati.Library.Main.Operation.BackupHandler.AddFileToOutput (Duplicati.Library.Main.BackendManager backend, System.String filename, Int64 size, DateTime lastmodified, IMetahash metadata, IEnumerable`1 hashlist, System.String filehash, IEnumerable`1 blocklisthashes) [0x00000] in <filename unknown>:0 at Duplicati.Library.Main.Operation.BackupHandler.HandleFilesystemEntry (ISnapshotService snapshot, Duplicati.Library.Main.BackendManager backend, System.String path, FileAttributes attributes) [0x00000] in <filename unknown>:0 

The failed path is part of the archived folder. Is that a problem?

warwickmm:
This excludes job-specific SQLite files from the backup.  There is no benefit to backing these up, and if the database needs to restored, it should be recreated from the backend files to ensure consistency.

The files being excluded are:

* The job-specific database file (`<database>`)
* The database rollback journals (`<database>-journal`)
* Write-ahead Log (WAL) files (`<database>-wal`)
* Shared-memory files (`<database>-shm`)

According to the [documentation](https://www.sqlite.org/draft/atomiccommit.html#_the_super_journal_file), the super journal file does not have a naming convention that is part of the specification.  As such, we do not exclude it:

> The next step in a multi-file commit is the creation of a "super-journal" file. The name of the super-journal file is the same name as the original database filename (the database that was opened using the sqlite3_open() interface, not one of the ATTACHed auxiliary databases) with the text "-mjHHHHHHHH" appended where HHHHHHHH is a random 32-bit hexadecimal number. The random HHHHHHHH suffix changes for every new super-journal.
> 
> (Nota bene: The formula for computing the super-journal filename given in the previous paragraph corresponds to the implementation as of SQLite version 3.5.0. But this formula is not part of the SQLite specification and is subject to change in future releases.)

We cannot use filters to unconditionally exclude these files, as there exists code paths where filters containing both includes and excludes have a tendency to include unmatched paths, which is probably the safer thing to do if a path is not matched by any (include or exclude) filter.

This fixes #1797 and #4308.

### DIFF
--- a/Duplicati/Library/Main/Operation/Backup/CountFilesHandler.cs
+++ b/Duplicati/Library/Main/Operation/Backup/CountFilesHandler.cs
@@ -27,14 +27,14 @@ namespace Duplicati.Library.Main.Operation.Backup
 {
     internal static class CountFilesHandler
     {
-        public static async Task Run(IEnumerable<string> sources, Snapshots.ISnapshotService snapshot, UsnJournalService journalService, BackupResults result, Options options, IFilter sourcefilter, IFilter filter, Common.ITaskReader taskreader, System.Threading.CancellationToken token)
+        public static async Task Run(IEnumerable<string> sources, Snapshots.ISnapshotService snapshot, UsnJournalService journalService, BackupResults result, Options options, IFilter sourcefilter, IFilter filter, ISet<string> forbiddenPaths, Common.ITaskReader taskreader, System.Threading.CancellationToken token)
         {
             // Make sure we create the enumeration process in a separate scope,
             // but keep the log channel from the parent scope
             using(Logging.Log.StartIsolatingScope(true))
             using (new IsolatedChannelScope())
             {
-                var enumeratorTask = Backup.FileEnumerationProcess.Run(sources, snapshot, journalService, options.FileAttributeFilter, sourcefilter, filter, options.SymlinkPolicy, options.HardlinkPolicy, options.ExcludeEmptyFolders, options.IgnoreFilenames, options.ChangedFilelist, taskreader, token);
+                var enumeratorTask = Backup.FileEnumerationProcess.Run(sources, snapshot, journalService, options.FileAttributeFilter, sourcefilter, filter, options.SymlinkPolicy, options.HardlinkPolicy, options.ExcludeEmptyFolders, options.IgnoreFilenames, options.ChangedFilelist, forbiddenPaths, taskreader, token);
                 var counterTask = AutomationExtensions.RunTask(new
                 {
                     Input = Backup.Channels.SourcePaths.ForRead
--- a/Duplicati/Library/Main/Operation/Backup/FileEnumerationProcess.cs
+++ b/Duplicati/Library/Main/Operation/Backup/FileEnumerationProcess.cs
@@ -42,7 +42,7 @@ namespace Duplicati.Library.Main.Operation.Backup
         /// </summary>
         private static readonly string FILTER_LOGTAG = Logging.Log.LogTagFromType(typeof(FileEnumerationProcess));
 
-        public static Task Run(IEnumerable<string> sources, Snapshots.ISnapshotService snapshot, UsnJournalService journalService, FileAttributes fileAttributes, Duplicati.Library.Utility.IFilter sourcefilter, Duplicati.Library.Utility.IFilter emitfilter, Options.SymlinkStrategy symlinkPolicy, Options.HardlinkStrategy hardlinkPolicy, bool excludeemptyfolders, string[] ignorenames, string[] changedfilelist, ITaskReader taskreader, CancellationToken token)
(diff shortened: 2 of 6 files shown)


warwickmm:
@ts678, @drwtsn32x, any thoughts on this?  Is there any valid case where these shouldn't be excluded?

ts678:
> Is there any valid case where these shouldn't be excluded?

From a performance point of view, any case where adding the exclude slows file scans more than the effort of doing DB backup.
[Slow backup with small files](https://forum.duplicati.com/t/slow-backup-with-small-files/11228/8?u=ts678) got into performance recently in a complex exclude, but how much does a simple (initial?) one hurt?

Other than that, if this excludes only the database for this active job, not all, it'll be stale soon and probably inconsistent anyway.
SQLite is not VSS aware AFAIK. If VSS was used and got a crash-consistent view, it's supposed to work, but won't match backup.
[Repair command deletes remote files for new backups #3416](https://github.com/duplicati/duplicati/issues/3416) then becomes a risk. Recreate is a safer way to get a DB if it works.

There is a chance that disaster recovery would go faster if direct restore from backup files of an in-progress DB avoided Recreate situation where it goes looking for a missing block and reads through all dblocks trying (and maybe failing after months of work).
Even if you lost the backup that was active when the database was backed up, you might be able to get to the backup before it...

There's also a possible Recreate slowness at default 100KB blocksize if backup is big. Someone's [asking](https://forum.duplicati.com/t/200tb-backups-feasible-with-duplicati/11324) right now about a 200 TB.

Having said a couple of cases, I can't recall having heard of use, though I've definitely heard forum complaints on Recreate speed.

There's philosophy to consider too. How far do we want to deviate from doing what is asked, whether or not we think it sensible?
Documenting such deviations would be best, even if's not seen unless somebody asks us. It's nice to have something to point to.

What I'm curious about is the occasional report that Duplicati hangs when trying to backup active database. How does it do that?
Whatever's going on might also cause hangs on some [random file](https://forum.duplicati.com/t/backup-process-stucks-on-random-file/11314) with similar activity (or whatever it takes to get Duplicati hang).
Even though I haven't been able to break it, I still worry about growing and shrinking files as mentioned [here](https://github.com/duplicati/duplicati/pull/3872#issuecomment-525759357) for a bug mostly-fix.

warwickmm:
Thanks @ts678.

Performance-wise, I didn't do much testing but I expect that this should have very little impact on performance.

> Other than that, if this excludes only the database for this active job, not all, it'll be stale soon and probably inconsistent anyway.

This only excludes the SQLite files for the active backup configuration.  One can include SQLite files from a different configuration and those will be backed up.

> There's philosophy to consider too. How far do we want to deviate from doing what is asked, whether or not we think it sensible? Documenting such deviations would be best, even if's not seen unless somebody asks us. It's nice to have something to point to.

This is a good point.  At the very least, the source selection UI should indicate that these files are forbidden.

> What I'm curious about is the occasional report that Duplicati hangs when trying to backup active database. How does it do that?

This is also a good point.  Perhaps the approach proposed here by skipping the SQLite files is just treating the symptom rather than the underlying problem.

Thanks for the feedback.  I'll decline this for now and give it some more thought.

warwickmm:
This also concerns issue #1443.

ts678:
> Perhaps the approach proposed here by skipping the SQLite files is just treating the symptom rather than the underlying problem.

Sometimes one can't get as far down as one might want, so must treat at some possible level while pursuing understanding.

One caveat (maybe not applicable to 1443 find) is that it can be hard to distinguish what appear to be source file hangs from destination hangs which stop the source file reading. One can approximate a destination hang just by setting a slow throttle.
Source file must be large enough that it doesn't all fit into the default 4 volumes that can queue. Reading will appear to stop.

I'm not sure how best to distinguish this. The live log can help some. Looking at the upload volume queue can also give clue.

Another caveat is that general non-Duplicati file locking problems also exist, and I'm surprised there aren't more issues filed...
Windows VSS can help, but it's hard to set up until whenever the installer grows smart enough to offer an easy service setup.

[Duplticati and VSS are bad friends](https://forum.duplicati.com/t/duplticati-and-vss-are-bad-friends/7884/39?u=ts678) shows how trying to open a file in two different ways can increase chance of open success.
Linux might behave differently, but the idea is to find which operations are incompatible and when, to see if relief can be had.
The log file might be an easier one to look at, compared to sorting out the [File Locking And Concurrency In SQLite Version 3](https://sqlite.org/lockingv3.html).

I did run a test for about 10 days, with a CLI job asynchronously backing up its own database and that of a GUI job. Ran fine. Seemingly a collision is hard to hit, but there's nothing like finding an actual test case to look at. Maybe I'll pursue more tests.

PR #4399, opened 2020-12-27
Avoid NullReferenceException if an IAsyncResult or WebRequest doesn't contain a WebResponse property

Gurthurb:
An example of when this happens is when shutting down TrayIcon (using its context menu), at least on Windows. This is quite annoying when debugging with Visual Studio (having the default exception settings).


warwickmm:
Thanks @Gurthurb.

Issue #4102, opened 2020-02-16
Password saver may autofill too many places, getting --auth-password is not supported


ts678:
- [*] I have searched open and closed issues for duplicates.

----------------------------------------

## Environment info

- **Duplicati version**: 2.0.5.1 Beta but this is old...
- **Operating system**: Windows 10 Version 1909
- **Backend**: N/A (mostly)

## Description

Passwords that are saved in password savers get autofill too generously on screens. Any remedies?
[Changing backend type leaves invalid backend options #3082](https://github.com/duplicati/duplicati/issues/3082) fix would help some but leave some.
Though I mostly blame password savers, is there anything Duplicati could do to limit their autofills?

A recurring complaint especially in forum is --auth-password is not supported warning on backups. [Backup to OneDrive v2 triggers UnsupportedOption: --auth-password every time #4099](https://github.com/duplicati/duplicati/issues/4099#issuecomment-586020558) cites a few which got solved to the cause here, but there are quite a few others that didn't resolve the mystery.

## Steps to reproduce

1. At Firefox --> Options --> Privacy & Security --> Logins and Passwords, check at least options:
Ask to save logins and passwords for websites
Autofill logins and passwords
2. At Duplicati --> Settings --> Access to user interface, check password, set `dup_pass` twice, OK.
3. At `http://localhost:8200/login.html?v=2.0.5.1` where you got sent, type `dup_pass`. `Sign In`

![image](https://user-images.githubusercontent.com/40771998/74598324-9c13ec80-503d-11ea-80a8-168c4546bd82.png)

4. Save.
5. Add backup --> Configure a new backup --> Next. Name and Passphrase backgrounds are yellow, and Passphrase is filled (click `Show` to get `dup_pass`). If username got saved, it might be filled too, which is what happened on the test in GitHub issue 4099 that tested with Chrome password saver.
6. Grumble and change `Passphrase` and `Repeat Passphrase` into what you intended, e.g. `password`.
7. `Next` to `Destination` screen finds similar setup. `Username` and `Password` yellow, password filled.
8. Set `Storage Type` away from `Local folder or drive` default, e.g. set it to Microsoft OneDrive v2.
9. Enter dummy Path `path` and AuthID `authid`, just to bypass the check (won't actually run it now).
10. Use three-button menu `Copy Destination URL to Clipboard` and note extraneous auth_password
`onedrivev2://path?authid=authid&auth-password=dup_pass`

- **Actual result**:
As described.
- **Expected result**:
[...]

PR #4402, opened 2020-12-29
Fix unwanted password manager autofill

drwtsn32x:
This attempts to fix an issue where users get this error when restoring:

`Failed to decrypt data (invalid passphrase?): Invalid password or corrupted data`

The root problem seems to be the password manager autofilling the (normally, always?) hidden encryption passphrase field when restoring using the web UI.

This change provides a hint to password managers that they should **not** autofill this field.

Reference:
https://developer.mozilla.org/en-US/docs/Web/Security/Securing_your_site/Turning_off_form_autocompletion#Preventing_autofilling_with_autocompletenew-password

Supported by all modern browsers:
https://developer.mozilla.org/en-US/docs/Web/HTML/Attributes/autocomplete#Browser_compatibility

I have tested this using built in password managers for Edge, Chrome, and Firefox and it resolves the issue.

Note - to fully resolve https://github.com/duplicati/duplicati/issues/4102 ~~we may need to add this fix/workaround to other html files.  Needs more research.  In the meantime I wanted to submit this partial fix for the most common issue we see on the forum.~~  With the third commit issue 4102 should be fully resolved.

### DIFF
--- a/Duplicati/Server/webroot/ngax/templates/addoredit.html
+++ b/Duplicati/Server/webroot/ngax/templates/addoredit.html
@@ -56,11 +56,11 @@
 
                     <div class="input password" ng-hide="checkGpgAsymmetric() || (Options['encryption-module'] || '').length == 0">
                         <label for="passphrase" translate>Passphrase</label>
-                        <input type="{{ShowPassphrase ? 'text' : 'password'}}" name="passphrase" id="passphrase" ng-model="Options['passphrase']" />
+                        <input autocomplete="new-password" type="{{ShowPassphrase ? 'text' : 'password'}}" name="passphrase" id="passphrase" ng-model="Options['passphrase']" />
                     </div>
                     <div class="input password" ng-hide="checkGpgAsymmetric() || (Options['encryption-module'] || '').length == 0">
                         <label for="repeat-passphrase" translate>Repeat Passphrase</label>
-                        <input type="{{ShowPassphrase ? 'text' : 'password'}}" name="repeat-passphrase" id="repeat-passphrase" ng-model="RepeatPasshrase"/>
+                        <input autocomplete="new-password" type="{{ShowPassphrase ? 'text' : 'password'}}" name="repeat-passphrase" id="repeat-passphrase" ng-model="RepeatPasshrase"/>
 
                         <div class="tools">
                             <ul>
--- a/Duplicati/Server/webroot/ngax/templates/backends/azure.html
+++ b/Duplicati/Server/webroot/ngax/templates/backends/azure.html
@@ -9,5 +9,5 @@
 </div>
 <div class="input password">
     <label for="azure_password" translate>Access Key</label>
-    <input type="password" name="azure_password" id="azure_password" ng-model="$parent.Password" placeholder="{{'Enter access key' | translate}}" />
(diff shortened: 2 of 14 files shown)


warwickmm:
Looks good to me.  @drwtsn32x, were you planning on making the same change to `Duplicati/Server/webroot/ngax/templates/addoredit.html` in this pull request as well?

drwtsn32x:
@warwickmm - ok, `addoredit.html` has now been updated as well.

drwtsn32x:
Do we want to update the "Destination" page password fields as well?

warwickmm:
> Do we want to update the "Destination" page password fields as well?

Is it possible that some users don't have a UI password set, but their password manager has the Destination page password saved?  I guess the destination shouldn't be modified often, so there's less impact compared to the Restore page.  @ts678, any thoughts here?

ts678:
Did addoredit.html actually help avoid the `--auth-password is not supported`? I'm now thinking I pointed to the wrong web page.
`https://github.com/duplicati/duplicati/tree/master/Duplicati/Server/webroot/ngax/templates/backends` might be the tree.
https://github.com/duplicati/duplicati/blob/45c4f865e8b5c2e79c96628383d8dca4cdc5317d/Duplicati/Server/webroot/ngax/scripts/services/EditUriBuiltins.js#L14-L14

looks like it's part of page build mechanism that picks which actual backends/*.html file to use. I haven't sorted out some subtleties, but browser password filling still looks like a major mess. Firefox isn't offering to save destination page passwords. Can anyone else? Chromium-based Edge will save them, but on next backup they show up on first screen as `Name` and `Passphrase` (for encryption). however that goof might be stopped by the addoredit.html change. Basically, I can't get my browsers to do any sensible thing here.

We seem to get a surprisingly limited range of forum complaints. The most frightening one is the restore failure. At least the `auth-password` one is fairly harmless and fixable. But why aren't people complaining about "user" and password fills on General screen 1?

drwtsn32x:
@ts678 - very good points, I don't think we've ever seen reports of issues with page 1.  Page 2 is the one that causes the --auth-password issue, right? I will do some testing tomorrow.

drwtsn32x:
The autofill happens on both page 1 and page 2 of a backup job config, so I think we should add the no autofill hint to all those areas.

I'm guessing the reason people don't report problems on page 1 is because it is pretty obvious that it needs manual correction:



On page 2 it's not quite as obvious, but at least it's visible (unlike when you do a restore operation):



I will edit the files in the "backends" folder and then check to see if it solves the auth-password issue.

drwtsn32x:
Ok, after editing the "backends" files the OneDrive URL issue is resolved. I'll add another commit to this PR.

Note that you have to force a full refresh, these pages are aggressively cached.  Should we look at disabling caching on the html files via an HTTP header or something?

warwickmm:
New versions supposedly perform some [cache-busting](https://github.com/duplicati/duplicati/pull/3515), so I'm not sure how important it is to disable caching all the time.

drwtsn32x:
@warwickmm - ok yeah it shouldn't be necessary then. Thanks

warwickmm:
Looks good to me.  Thanks @drwtsn32x!  This should eliminate a really annoying issue that caused users a lot of stress.

PR #4404, opened 2020-12-30
Fix restore issue in Docker installations

### LINKED ISSUE #4403: Query in GetFilesAndSourceBlocksFast does not work in Docker
## Environment info

- **Duplicati version**: 2.0.5.111_canary_2020-09-26
- **Operating system**: Docker
- **Backend**: Local

## Description

Attempts to restore when running the Docker container results in a "SQLite error no such column: S.BlocksetID" error when running the following query

https://github.com/duplicati/duplicati/blob/9b977b7ba04a940868754fd15a37b99fbd77f1bd/Duplicati/Library/Main/Database/LocalRestoreDatabase.cs#L1247

This is likely related to the changes from pull request #4297.

## Steps to reproduce

1. Create a backup configuration, and run a backup.
2. Attempt to restore a file.

- **Actual result**:
An error is encountered: SQLite error no such column: S.BlocksetID

image

warwickmm:
This fixes an issue that caused a "SQLite error no such column: S.BlocksetID" error when attempting to restore in Docker installations.  When creating a temporary table from a `SELECT` statement, we should be providing explicit column names.  If we do not do so, the name of the column in the result set is unspecified and may change depending on the version of SQLite.

https://www.sqlite.org/c3ref/column_name.html

This fixes #4403.


warwickmm:
There appear to be a few other cases of `CREATE TEMP TABLE <table_name> AS SELECT ...` that do not provide explicit column names.  I'm not sure why those are not causing problems (perhaps the query plan leads to a different code path in SQLite), but we should consider fixing those in the future to be safe.

warwickmm:
Without this, the `latestBlockTable` table had column names `File`, `File:1`, and `MAX("Fileset"."Timestamp")`.

ts678:
> I'm not sure why those are not causing problems

[Query in GetFilesAndSourceBlocksFast does not work in Docker #4403](https://github.com/duplicati/duplicati/issues/4403#issuecomment-752753962) proposes it's a view bug, and points to a bug in old SQLite.

> we should consider fixing those in the future to be safe.

If examination can show that Duplicati Docker does ship the buggy version, and newer one works, this might be the only view bug.

Regardless, my independent accidental parallel chase found the same bad column names, and had the same explicit name proposal.

This is also not a Docker-only bug. I got it on a Linux Mint VM that was old enough to have what might be the bad sqlite3 version.
What surprises me is that this isn't being reported from users with regular Linux installs. Maybe everyone else upgrades Linux fast?

ts678:
After finding the Debian 9.11 site and downloading the package listing files, it looks like Docker SQLite "should" be 3.16.2.
Probably best to just set the names as desired, which will also handle old Linux systems if any besides mine are breaking...
I still wonder if it's only a problem with views though, or at least my testing with genuine tables wasn't showing the issue.

warwickmm:
Thanks @ts678.  It doesn't sound like the issue is specific to views.  According to the SQLite documentation, inheriting the name of the column when selecting from a table is not part of the spec and can change based on the implementation.

warwickmm:
I queried `select sqlite_version()` in the Docker container and it returned 3.16.2.

Issue #4389, opened 2020-12-11
Suggestion: Switching to FluentMigrator and Linq2Db for a backup sqlite database operations


SomebodyOdd:
- [x] I have searched open and closed issues for duplicates.

----------------------------------------

Hello there!

I'd like to ask everyone involved in a project on opinion.
IMO, code that deals with local db operations is less than ideal in respect to maintainability and overall readability. Those huge string literals with SQLite commands and low level DB class usage is something that could improved on. How about switching to something like FluentMigrator and Linq2Db (specific tools to use are up for debate, of course, used these as an example).

I'd like to at least try to implement this, but I want to listen for a feedback first.
Will it sit right with core dev team? Is this something that could make dealing with duplicati code easier for everyone and not just myself?

This switch could make optimizing backup operations easier and make possible integration with server DBMS as well (although, this is not final goal of this proposal). Not to mention that this won't hurt possible future migration to .net 5, if this ever arise.

So, what do you think?

SomebodyOdd:
Looks like noone is interested in that. Closing an issue, please reopen if I'm wrong and there is some feedback on that.

ts678:
> Looks like noone is interested

Maybe nobody is familiar with the tools or the philosophy of use. I would be interested in your input on your question:

> Is this something that could make dealing with duplicati code easier for everyone

because Duplicati **very much** needs DB volunteers. I'm not one, but would like to minimize the barriers to new entrants.
This could include an occasional [contributor](https://github.com/duplicati/duplicati/graphs/contributors) who shows up with something, but core team is also not that good at SQL.

These tools (to the extent I could follow them) seemed like they reduced "low level" code but needed high level training.
There are a lot of training resources for generic SQL. I'm not sure about the resources or learning curve for higher tools.

From a non-DB-person view, I find that current code is hard to read and follow. How safely can any migration be done?
There would not be a team of experts available to explain every feature of current code, yet migrated code should work.

From a maintainability point of view, would migrated code be easier to read and follow by those who aren't DB-expert?
Possibly you could post a sample of migrated code, so developers could see whether it seems an improvement to them.

> This switch could make optimizing backup operations easier

Duplicati does has some SQL code that should go faster. Maybe DB design is off, maybe queries are off. Any help there?
The help short-term would probably be in writing equivalent-but-faster SQL. Big schema redesigns might be too much.

There might also be some actual bugs in logic (either C# or the SQL it runs). Tracking them down in field is not possible.
[...]

PR #4297, opened 2020-08-26
Improve query performence used in GetFilesAndSourceBlocksFast

pcmind:
Improve query performance when restoring backups of a lot of files to a new location. 
This fix #3452

Tree changes in query:
* Move inner selection S to a temporary table
* Reorder join field in outer query
* Remove intersection with File view in outer selection (File view is already intersected with in selection S)

With the above-mentioned changes, on a specific testset, query took 9s instead of 6min.

PR #4408, opened 2021-01-02
[WIP] Project structure refactoring and migration to SDK-style project format.

SomebodyOdd:
(this is Work-in-Progress - I haven't really tested resulting binaries yet)

This PR moves CommandLine and Server projects to their own folder and migrates project format to SDK-style. No functional and/or behavioral changes are intended (besides build behaviour changes, that arise from project style migration)

Apart from just ditching old project style, there is a benefit when it comes to using .NET Standard libraries and packages, because there is [known issue](https://github.com/dotnet/sdk/issues/901) with referencing those from old style projects.
This conversion, however, required moving Server and CommandLine projects to their separate folders. While not strictly necessary (you can ignore child projects in main project .csproj), this made transition easier, and would probably be good in a long run. Feel free to correct me if you think otherwise.

This PR is related to #4389 since some of the proposed libraries are .NET Standard.


warwickmm:
Transition to .NET 5 is being done in pull request #3124.  I haven't been able to keep up with the progress there, but perhaps it would be good to coordinate with @mnaiman, since it looks like some of the project format conversion has already been done.

warwickmm:
I'm going to go ahead and close this, since the project format changes are being done in the .NET 5 branch.

SomebodyOdd:
@warwickmm sorry for necroposting, but #3124 seems like a huge pull request with over a 100 commits there already, and it doesn't seem to end anytime soon. Don't get me wrong, there is a progress there, and this would be awesome if duplicati would migrate to .net 5, but project conversion is a thing that could bring some benefit right now. Maybe we could try and merge project format conversion earlier than .net 5 migration?

Thing is, I'd still like to use a FluentMigrator, as discussed in #4389, but old project format really complicates things because of [this](https://github.com/dotnet/sdk/issues/901) known issue.
Could we revisit decision to wait for a .net 5?

warwickmm:
There is some desire to focus more on bug fixes and stabilizing the current code in an effort to ease the transition to .NET 5. 
 Personally, I am a little hesitant to overhaul critical code before we make the transition.

However, I wouldn't want to inhibit your progress on cleaning up the database code and queries.  I'm also not familiar with all the implications of the proposed changes.  Perhaps one way forward could be to keep the necessary infrastructure changes (csproj files, etc.) on your branch, and convert one of the queries as a proof of concept that we can evaluate?  I'd be curious about performance implications, how easy it is to debug a query, etc.  

I'd also be interested to see what the before and after Linq2Db looks like for the changes in pull request #4297.  Is the inefficiency further obfuscated by the "higher-level" query, or is it actually made more obvious somehow?

/cc @kenkendk

SomebodyOdd:
Okay, I just was hesitant that if I do some major stuff now, then it will become merge nightmare after .NET 5 changes roll out.
But proof of concept is fine, I guess. Just need to find some time for it

PR #4409, opened 2021-01-02
Use consistent (and updated) mono image for Docker and Travis

warwickmm:
This makes the Travis builds use a `mono` parent image that is more consistent with our Docker build.  Previously, the Travis build used `mono` (effectively `mono:latest`), while our Dockerfile specified `mono:5-slim`.  This discrepancy caused us to fail to notice the problem described in issue #4403.  By making these consistent, we will be more likely to detect such issues earlier.

We also take this opportunity to update the Docker parent image to `mono:6`.


warwickmm:
Duplicated by #4410.

PR #4410, opened 2021-01-02
Use consistent (and updated) mono versions for Docker and Travis

warwickmm:
This makes the Travis builds use a `mono` parent image that is more consistent with our Docker build. Previously, the Travis build used `mono` (effectively `mono:latest`), while our Dockerfile specified `mono:5-slim`. This discrepancy caused us to fail to notice the problem described in issue #4403. By making these consistent, we will be more likely to detect such issues earlier.

We also take this opportunity to update the Docker parent image to `mono:6`.


warwickmm:
The build for revision 6f41be3 (which uses `mono:5`) fails as desired with the "SQLite error no such column: S.BlocksetID" error described in issue #4403.  While using `mono:6` does avoid this issue, the proper fix for that error is in pull request #4404.

warwickmm:
I wasn't able to test the changes to `build-images.sh`, but I think it should work.

PR #4417, opened 2021-01-10
Abort with error if run-script-before script returns unsupported code

warwickmm:
This makes the code more consistent with the documentation in the example scripts.  Previously, the process would not be aborted if an unsupported code was encountered.

https://github.com/duplicati/duplicati/blob/338f467111f6bdcde971f93b8376daba42827498/Duplicati/Library/Modules/Builtin/run-script-example.sh#L21-L27

This also adds some tests to verify the behavior.

### DIFF
--- a/Duplicati/Library/Modules/Builtin/RunScript.cs
+++ b/Duplicati/Library/Modules/Builtin/RunScript.cs
@@ -286,11 +286,15 @@ namespace Duplicati.Library.Modules.Builtin
                                 {
                                     switch (p.ExitCode)
                                     {
-                                        case 1:
+                                        case 0: // OK, run operation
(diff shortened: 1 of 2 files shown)


warwickmm:
I'm not sure why the new test with exit code `-1` is failing on Travis.  It is passing for me locally, as well as in AppVeyor.

warwickmm:
It appears the behavior of exit code `-1` is not consistent across shells.  Revision c11f7c2e4717db86c7b665198700d54e2320f263 replaces this test case with a few others.

https://unix.stackexchange.com/questions/418784/what-is-the-min-and-max-values-of-exit-codes-in-linux

PR #4418, opened 2021-01-15
Make DUPLICATI__PARSED_RESULT consistent with OperationAbortReason

### LINKED ISSUE #4222: Exit 1 in run-script-before produces fatal operationresult

## Environment info

- **Operating system**: docker container on unRAID host
- **Backend**: webDAV

## Description

I'm using the run-script-before to see if files have changed in the local paths I'm backing up. If no files were changed the script exits with 1 (which is supposed to say, "OK, and don't run the backup").
However entering the run-script-after I'd (of course) want to catch a fatal error and log it and/or send an email based on this result, but because some jobs usually have no changes this would trigger alerts each time.

## Steps to reproduce

1. Deploy docker based duplicati.
2. Create very basic job.
3. Create a very basic script that exits with 1 and use it for run-script-before.
3. Create a very short script that reads the operation result and echos it to a log.

- **Actual result**:
A fatal error is produced.
- **Expected result**:
I would expect success since exiting the BEFORE stage of the script was totally expected and desired, if I wanted to trigger a warning or error I would exit with 3 or 5 respectively.

## Screenshots

Not needed.

## Debug log

Looking into Duplicati\Library\Modules\Builtin\RunScript.cs (Execute() function) it seems like these lines throw an exception on each of those errorcodes (OperationAbortReason enum set to normal though so that seems OK)

    // If this is the start event, we abort the backup 
    if (eventname == "BEFORE")
    {
        switch (p.ExitCode)
        {
            case 1:
                throw new OperationAbortException(OperationAbortReason.Normal, Strings.RunScript.InvalidExitCodeError(scriptpath, p.ExitCode));
            case 3:
                throw new OperationAbortException(OperationAbortReason.Warning, Strings.RunScript.InvalidExitCodeError(scriptpath, p.ExitCode));
            case 5:
                throw new OperationAbortException(OperationAbortReason.Error, Strings.RunScript.InvalidExitCodeError(scriptpath, p.ExitCode));
        }
    }

This causes us to exit the try and enter this catch below:

    catch (OperationAbortException)
    {
        // Do not log this, it is already logged
        throw;
    }

The OnFinish() function from the same class checks the result to see if it is of type Exception, which it is since we got thrown outside of the Execute() function right? This sets the level to ParsedResultType.Fatal (in this case incorrectly).

    ParsedResultType level;
    if (result is Exception)
        level = ParsedResultType.Fatal;
    else if (result != null && result is IBasicResults results)
        level = results.ParsedResult;
    else
        level = ParsedResultType.Error;

I think fixing this should make reporting more reliably (at least in my usecase).
Also I was thinking.. Should this be not as easy as inspecting the object in this catch, to see if the OperationAbortReason enum equal Normal and if so don't throw?

    catch (OperationAbortException)
    {
        // Do not log this, it is already logged
        throw;
    }

I don't have the c# know-how (yet ;) ) to solve these myself or to relate these changes in the overall picture so I hope these suggestions point someone in the right direction and motivate that same person to include this in a next pull-request.

Thanks!

warwickmm:
This makes the `DUPLICATI__PARSED_RESULT` environment variable (available to `run-script-after` scripts) more consistent with the `OperationAbortReason`.  Previously, all `run-script-before` abort exit codes (`1`, `3`, `5`), would result in a `Fatal` parsed result.  Now, `1` (OK, don't run operation) results in `Success`, `3` (Warning, don't run operation) results in `Warning`, and `5` (Error, don't run operation) results in `Fatal`.

This fixes #4222.

### DIFF
--- a/Duplicati/Library/Modules/Builtin/RunScript.cs
+++ b/Duplicati/Library/Modules/Builtin/RunScript.cs
@@ -180,7 +180,25 @@ namespace Duplicati.Library.Modules.Builtin
                 return;
 
             ParsedResultType level;
-            if (result is Exception)
+            if (result is OperationAbortException oae)
+            {
+                switch (oae.AbortReason)
+                {
+                    case OperationAbortReason.Error:
+                        level = ParsedResultType.Error;
+                        break;
+                    case OperationAbortReason.Normal:
+                        level = ParsedResultType.Success;
+                        break;
+                    case OperationAbortReason.Warning:
+                        level = ParsedResultType.Warning;
+                        break;
+                    default:
+                        level = ParsedResultType.Unknown;
+                        break;
+                }
+            }
+            else if (result is Exception)
                 level = ParsedResultType.Fatal;
             else if (result != null && result is IBasicResults results)
                 level = results.ParsedResult;
--- a/Duplicati/UnitTest/RunScriptTests.cs
+++ b/Duplicati/UnitTest/RunScriptTests.cs
@@ -160,10 +160,15 @@ namespace Duplicati.UnitTest
         public void RunScriptParsedResult(int exitCode)
         {
             string parsedResultFile = Path.Combine(this.RESTOREFOLDER, "result.txt");
-            List<string> customCommands = new List<string>
+            List<string> customCommands = new List<string>();
+            if (Platform.IsClientWindows)
             {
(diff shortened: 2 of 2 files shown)


warwickmm:
The failing tests should be fixed by pull request #4417.

PR #4431, opened 2021-02-07
Add GitHub workflow to run tests

warwickmm:
This adds a GitHub workflow that
* Runs the unit tests in Linux, Windows, and macOS environments.
* Runs the Selenium-based user interface tests.

Compared to using AppVeyor and Travis, this has the benefit of consolidating the configuration of our tests into a single YAML file, as opposed to having to maintain separate configurations for AppVeyor and Travis.  One immediate benefit of this was the ability to easily configure the tests to run in a macOS environment.  Also, once we migrate to .NET 5, it should be a simple change to modify the test configuration accordingly.

This did required a few minor changes to the code:
* Dispose an instance of HttpWebResponse to help avoid timeout issues.
* Allow for tolerance when comparing metadata timestamps on macOS.

This does not remove the existing AppVeyor and Travis configurations.  We can remove them if we later decide that those are no longer needed.


warwickmm:
Travis has also adopted a new [pricing model](https://blog.travis-ci.com/2020-11-02-travis-ci-new-billing), where OSS projects are granted an allotment of credits.  Once the credits are used up, then it sounds like we would have to choose one of their pricing plans.

warwickmm:
Some of the failures on macOS are quite strange.  The timestamps in the metadata can be off by quite a bit:
```
1) Failed : Duplicati.UnitTest.BorderTests.Run100k
  Restore, creation time mismatch for /Users/runner/duplicati_testdata/backup-data/a and /Users/runner/duplicati_testdata/restored/a
  Expected: 2021-02-08 18:27:06.3860412 +/- 00:00:00.5000000
  But was:  2021-02-08 18:27:18.3035
  at Duplicati.UnitTest.TestUtils.AssertFilesAreEqual (System.String expectedFile, System.String actualFile, System.Boolean verifymetadata, System.String contextMessage) [0x0016f] in /Users/runner/work/duplicati/duplicati/Duplicati/UnitTest/TestUtils.cs:256 
```

warwickmm:
I've removed the artifacts from the workflow.  We can revisit when the need for a "nightly" release is more apparent.

warwickmm:
Here's an example of a Travis build that was canceled due to an insufficient credits balance:

https://travis-ci.com/github/duplicati/duplicati/jobs/492568705

Issue #4436, opened 2021-02-10
UI - Progress bar is covering the "cancel button"


jfparis:
- [x] I have searched open and closed issues for duplicates.

----------------------------------------

## Environment info
- **Duplicati version**: 2.0.5.112_canary_2021-01-20
- **Operating system**: Ubuntu
- **Backend**: Not relevant

## Description
In the main UI, the progress bar is displayed over the text and the X (circled cross) button. 
As a result, when a task is nearing completion it is impossible to click on the Circled Cross icon to cancel it 

## Steps to reproduce
1. Launch a task and wait for it to reach 95% (or about) completion

- **Actual result**:
Try to click over the X (circled Cross) and you will notice it is impossible 
- **Expected result**:
It should be possible to click on the cross to cancel the task

## Screenshots
![Screenshot from 2021-02-10 10-20-02](https://user-images.githubusercontent.com/48999/107483407-efe5e280-6b78-11eb-9ec6-da04f65b7500.png)

Look at the mouse pointer. It does not allow a click because the bar is displayed above (z-index) the button 
If the task is at 98% then the circle cross is totally covered and it is totally impossible to click

jfparis:
Created a pull request to fix this

warwickmm:
Closed by pull request #4437.

warwickmm:
Reopening because the fix from pull request #4437 caused an issue with the edit backup views.

PR #4437, opened 2021-02-10
Prevent progress bar from obscuring stop button

jfparis:
Brings the Stop button above the progress bar to allow stopping tasks even when they are well advanced

This should address issue #4436


warwickmm:
Thanks @jfparis.  However, we actually generate `style.css` by editing [Duplicati/Server/webroot/ngax/less/style.less](https://github.com/duplicati/duplicati/blob/v2.0.5.112-2.0.5.112_canary_2021-01-20/Duplicati/Server/webroot/ngax/less/style.less) and then compiling it with `lessc`
```
$ lessc Duplicati/Server/webroot/ngax/less/style.less --clean-css > Duplicati/Server/webroot/ngax/styles/style.css
```
Would you be able to make your changes to `style.less`?

jfparis:
This got out of hand quickly.
Lessc seems to have dropped a lot of quotes out of the compiled file. Not sure if this is causing the rest of the trouble. Is there a specific version of lessc? Is there a specific devenv that I can replicate?

jfparis:
The trick is in the "-x" parameter for less. I will update the PR
Thanks

warwickmm:
Thanks @jfparis.  Looks good to me on Brave.  @drwtsn32x, are you able to test this on a few other browsers?

drwtsn32x:
@warwickmm - worked for me on Firefox.  Brave is Chromium-based right, so it should also work with regular Chrome or Edge as well.

warwickmm:
Thanks @drwtsn32x.

I also compiled `style.less` locally and the resulting `style.css` was identical to what was committed.

drwtsn32x:
@warwickmm @jfparis - something may be wrong with this change.  Now I'm seeing this when editing a backup job:

Capture

warwickmm:
Thanks @drwtsn32x, I see it too.  @jfparis, are you able to reproduce this?

I'll revert the merge in `master`.

jfparis:
Yes. I can see it. I need to figure out whether this is a compiler issue or
else
--
Jean-François Paris
+44 7533414619

jfparis:
Figured it out. This was a compiler issue.
The proper way to invoke lessc is 
`lessc -m=always -x Duplicati/Server/webroot/ngax/less/style.less > Duplicati/Server/webroot/ngax/styles/style.css` 

Both options are deprecated :)

WIll submit a new PR

Issue #3748, opened 2019-04-24
Underscores instead of dashes


kees-z:
- [x] I have searched open and closed issues for duplicates.

----------------------------------------

## Environment info

- **Duplicati version**: 2.0.4.17
- **Operating system**: Windows 10 Pro

## Description

For almost all advanced options, words are separated using dashes (`--auth-username`). A few options for specific storage providers use underscores to separate words. I found these options that use underscores:
```
--aws_secret_access_key
--aws_access_key_id
--azure_blob_container_name
--azure_account_name
--azure_access_key
--azure_blob_container_name
```

For these options, underscores (`_`) should be replaced by dashes (`-`). For backward compatibility, the current options with underscores should still be accepted, but preferably be undocumented.

Issue #3082, opened 2018-03-05
Changing backend type leaves invalid backend options


marcusmercier:
I'm getting warnings during every backup for "ignored"  options. Is it possible to suppress them, or atleast, not show them on the dashboard/first page? It is very annoying and misleading.

Image:
![2018-03-05 11_50_56-caesar7 com - duplicati](https://user-images.githubusercontent.com/36671101/36973612-9342d944-206b-11e8-8980-f02686fde713.png)

Pectojin:
Why don't you remove the invalid options?

marcusmercier:
I did not add them, I assume they are integrated in to the "provider", in this case "Hubic". I searched inside the individual backup configurations - there were no extra/additional options associated with any of the backups. They must be integrated. I obviously can be wrong - am I?

Pectojin:
Hmm, that sounds very strange.

I'm not familiar with the Hubic provider, but no providers should be appending invalid options, so if it is, that's a bug.

Did you also check your global options? Just to be safe.

What version of Duplicati are you running?

marcusmercier:
The version is 2.0.2.20_canary_2018-02-27
The "problem" is present for at least 2 months (I update it quote often).
And no, there are no global options set, apart from --thread-priority=idle

I only use Mega.co.nz and Hubic providers, but only Hubic shows this error. Hubic also shows 401 quote often, but it's for they rate limiting.

wjansenw:
Did you move from backend? I had the same issue when moving from a backend
where those options where needed, and another backend where those where not
needed.
I managed removing the options by selecting the old provider, removing the
options (they were not mandatory), saving, and going back to the new
backend.

Op ma 5 mrt. 2018 om 16:08 schreef marcusmercier <notifications@github.com>:

marcusmercier:
I might have copied the configuration from mega.co.nz's configuration, but these messages started occurring no less than one month after I did that. For clarification, I attached a screenshot of what I see in the options section - there is nothing I can remove. Also, I browsed around the Duplicati's .sqlite database, and there were no such options ("auth-username", "auto-password") there. I also checked the individual backup's .sqlite file - no options there either.
![2018-03-05 17_20_51-caesar7 com - duplicati](https://user-images.githubusercontent.com/36671101/36989591-05be15a0-209a-11e8-97f7-fe0ba13e0740.png)

wjansenw:
Can you export to command line and see if the options are there?

[...]

PR #4443, opened 2021-02-15
Use dashes in Azure and S3 options for consistency

julianohubel:
Addressed to issue [#3748](https://github.com/duplicati/duplicati/issues/3748#issue-436683140)


Duplicati/Library/Backend/S3/S3Backend.cs
```diff
@@ -392,7 +392,9 @@ public IList<ICommandLineArgument> SupportedCommands
 
                 var normal = new ICommandLineArgument[] {
                     new CommandLineArgument("aws_secret_access_key", CommandLineArgument.ArgumentType.Password, Strings.S3Backend.AMZKeyDescriptionShort, Strings.S3Backend.AMZKeyDescriptionLong, null, new string[] {"auth-password"}, null),
```
warwickmm, line 394: For the old underscored options, can we use the following `CommandLineArgument` constructor that marks the options as deprecated?

https://github.com/duplicati/duplicati/blob/67aaab15e3a8638806a50cf23b5a085c826a338a/Duplicati/Library/Interface/CommandLineArgument.cs#L258-L263

julianohubel: Shure. I added a new commit change. Could you review?

Duplicati/Library/Backend/AzureBlob/AzureBlobBackend.cs
```diff
@@ -131,15 +131,18 @@ public IList<ICommandLineArgument> SupportedCommands
                     new CommandLineArgument("azure_account_name",
                         CommandLineArgument.ArgumentType.String,
                         Strings.AzureBlobBackend.StorageAccountNameDescriptionShort,
-                        Strings.AzureBlobBackend.StorageAccountNameDescriptionLong),
+                        Strings.AzureBlobBackend.StorageAccountNameDescriptionLong,
+                        null, null, null, "This is deprecated, use azure-account-name insted"),
```
warwickmm, line 135: Thanks.  Can we change the spelling to `instead`?

julianohubel: Right. I Added another commit, thanks for the correction.

warwickmm:
Thanks @julianohubel.  I can see the deprecated messages for azure:
```
$ mono Duplicati/GUI/Duplicati.GUI.TrayIcon/bin/Debug/Duplicati.CommandLine.exe help azure
Azure blob (azure):
 This backend can read and write data to Azure blob storage.  Allowed formats are: "azure://bucketname"
 Supported options:
  --azure_account_name (String): The storage account name
    [DEPRECATED]: This is deprecated, use azure-account-name instead
    The Azure storage account name which can be obtained by clicking the "Manage Access Keys" button on the storage account dashboard
  --azure_access_key (Password): The access key
    [DEPRECATED]: This is deprecated, use azure-access-key instead
    The Azure access key which can be obtained by clicking the "Manage Access Keys" button on the storage account dashboard
  --azure_blob_container_name (String): The name of the storage container
    [DEPRECATED]: This is deprecated, use azure-blob-container-name instead
    All files will be written to the container specified
  --azure-account-name (String): The storage account name
    The Azure storage account name which can be obtained by clicking the "Manage Access Keys" button on the storage account dashboard
  --azure-access-key (Password): The access key
    The Azure access key which can be obtained by clicking the "Manage Access Keys" button on the storage account dashboard
  --azure-blob-container-name (String): The name of the storage container
    All files will be written to the container specified
  --auth-password (Password): Supplies the password used to connect to the server
    The password used to connect to the server. This may also be supplied as the environment variable "AUTH_PASSWORD".
  --auth-username (String): Supplies the username used to connect to the server
    The username used to connect to the server. This may also be supplied as the environment variable "AUTH_USERNAME".
```
However, they don't appear for s3:
```
$ mono Duplicati/GUI/Duplicati.GUI.TrayIcon/bin/Debug/Duplicati.CommandLine.exe help s3

Duplicati can use S3-compatible servers to store backups. The following target URL format is used:
  s3://bucketname/prefix

Options:
  --aws_secret_access_key
    The AWS "Secret Access Key" can be obtained after logging into your AWS account, this can also be supplied through the "auth-password" property
  --aws-secret-access-key
    The AWS "Secret Access Key" can be obtained after logging into your AWS account, this can also be supplied through the "auth-password" property
  --aws_access_key_id
    The AWS "Access Key ID" can be obtained after logging into your AWS account, this can also be supplied through the "auth-username" property
  --aws-access-key-id
    The AWS "Access Key ID" can be obtained after logging into your AWS account, this can also be supplied through the "auth-username" property
...
```
I don't quite understand the difference, but it appears that for `azure`, the following is `false`, while for `s3` it is `true`:
https://github.com/duplicati/duplicati/blob/c47a356d6f2fafbcb990957bcd364df7fe16a80c/Duplicati/CommandLine/Help.cs#L92

For `s3`, this causes the output to be displayed using `PrintArgSimple`, which doesn't respect the deprecated property:
https://github.com/duplicati/duplicati/blob/c47a356d6f2fafbcb990957bcd364df7fe16a80c/Duplicati/CommandLine/Help.cs#L486-L496

I'm not that familiar with this part of the code, so I'm not sure what the best way to proceed is right now.  Any ideas?  I don't know why there are so many different code paths for printing options.

warwickmm:
I think the difference is s3 has an entry here:

https://github.com/duplicati/duplicati/blob/85c9e4394308933d549face3b9cf97ec37c13978/Duplicati/CommandLine/help.txt#L382-L389

Since the following is `true`

https://github.com/duplicati/duplicati/blob/85c9e4394308933d549face3b9cf97ec37c13978/Duplicati/CommandLine/Help.cs#L188

we then replace `%MODULEOPTIONS%` with the result from `PrintArgsSimple`, which doesn't respect the deprecated flag

https://github.com/duplicati/duplicati/blob/85c9e4394308933d549face3b9cf97ec37c13978/Duplicati/CommandLine/Help.cs#L229-L230

One minimally invasive option is to modify `PrintArgsSimple` to call `CommandLineArgument.PrintArgument`, which seems to fix the s3 case:
```
diff --git a/Duplicati/CommandLine/Help.cs b/Duplicati/CommandLine/Help.cs
index 843396fc6..66cf9ba55 100644
--- a/Duplicati/CommandLine/Help.cs
+++ b/Duplicati/CommandLine/Help.cs
@@ -489,8 +489,10 @@ namespace Duplicati.CommandLine
                 return "";
 
             List<string> lines = new List<string>();
-            foreach(Duplicati.Library.Interface.ICommandLineArgument arg in args)
-                lines.Add(PrintArgSimple(arg, arg.Name));
+            foreach (Duplicati.Library.Interface.ICommandLineArgument arg in args)
+            {
+                Duplicati.Library.Interface.CommandLineArgument.PrintArgument(lines, arg);
+            }
 
             return string.Join(Environment.NewLine, lines.ToArray());
         }
```

@julianohubel, what do you think?

kenkendk:
The PR suggested by @julianohubel is correct. The render code is broken.

I think the idea is that we wanted to sometimes show a limited set of options.
I will look an try to fix the help rendering mess.

kenkendk:
It is still a bit of a mess, but the idea is:

1a) If there is an entry in `help.txt` that has been created by a human who decided how it should look.
1b) If the entry in `help.txt` contains "%MODULEOPTIONS%", print the module options in a concise manner

2a) If there are no entries in `help.txt`, auto-create one by using the description from the module.
2b) Print each option with the same code that is used for printing help for a specific option.

It would perhaps make sense to just append the same options list in either case.
I did not change that, but I did fix the output in #4459

ts678:
[AWS Secret Access Key Warning in Duplicati - 2.0.5.113_canary_2021-03-07](https://forum.duplicati.com/t/aws-secret-access-key-warning-in-duplicati-2-0-5-113-canary-2021-03-07/11901) seems related to this. Is issue likely to be widespread?

On a side note, is the UI still using the old option?

https://github.com/duplicati/duplicati/blob/f145b9fcc2f93aaa606bc939373afbb566034647/Duplicati/Server/webroot/ngax/scripts/services/EditUriBuiltins.js#L341-L345

warwickmm:
I can't reproduce the warnings.  There is an outstanding issue (#3082) regarding old options hanging around after changing backends.  Could that be the cause?

I created issue #4461 to fix the UI and will submit a fix soon.  Thanks for noticing that.

warwickmm:
I just noticed that the user in the forum says they're using the Wasabi backend.  I believe this is an S3-compatible backend, so that would make sense why they're seeing warnings about deprecated S3 options.

PR #4445, opened 2021-02-19
Update stylesheet to raise stop button above progress bar

jfparis:
Brings the Stop button above the progress bar to allow stopping tasks even when they are well advanced

This should address issue #4436


jfparis:
@warwickmm any chance the CI can be relaunched? This is the second time this happens to me.

Exact diff between the compiled css code is 
`3372c3372,3374
<     cursor: pointer
---
>     cursor: pointer;
>     z-index: 10;
>     position: relative
5681d5682
< 
`

warwickmm:
I used to have the ability to re-run the Travis jobs, but after we transitioned to travis.com from travis.org I no longer can.  The only way is to make another commit (perhaps force push the same commit with a new message).

warwickmm:
Thanks, that looks better to me.  I merged your branch into one that runs the UI tests via GitHub actions and it passed [here](https://github.com/warwickmm/duplicati/runs/1940060967?check_suite_focus=true).

@drwtsn32x, would you be able to test with Firefox again?

jfparis:
Alright. I pushed again. The diff of the compiled style.css remains above
```
3372,3374c3372
<     cursor: pointer;
<     z-index: 10;
<     position: relative
---
>     cursor: pointer
5682a5681
> 
```

drwtsn32x:
@warwickmm Yep looks good to me now!

warwickmm:
Thanks @jfparis and @drwtsn32x.

Issue #4316, opened 2020-09-13
python-is-python2 is deprecated and can break modern software


HeRo002:
## Description

Actually, not much information is needed to describe this bug. Some may even argue that it's a feature request. 

The thing is that Duplicati (on Ubuntu 20.04) installs the package 'python-is-python2', which creates the file /usr/bin/python which is a symlink to /usr/bin/python2.7. 

This kind of laziness from the developers of Duplicati puts newer software, that need python3 at risk of crashing. I am more or less citing the description of the package - see below.

I am going to develop complex Python3 software on my Ubuntu 20.04, with large amount of dependencies, so Duplicati puts my project at risk of failure or problems that could have been avoided if Duplicati had not installed the /usr/bin/python symlink to /usr/bin/python2.7. I would therefore be very glad if your could correct this as fast as possible. 

Description of python-is-python2:

" In Ubuntu, all python packages use explicit python3 or python2
interpreter and do not use unversioned /usr/bin/python at all. Some
third-party code may still be python2 based, yet may use
/usr/bin/python.

This is a convenience package which ships a symlink to point
/usr/bin/python interpreter at the current default python2. It may
improve compatibility with obsolete 3rd-party software, whilst
breaking some modern software.

This package will be installed upon upgrades to Ubuntu 20.04, if
DEPRECATED python2 was installed.

python2 is DEPRECATED and will not be provided in the future Ubuntu
release. It is recommended to remove python2 and this package after
ensuring that only python3 is in use.

No packages may declare dependencies on this package. "

- [X] I have searched open and closed issues for duplicates.

----------------------------------------

## Environment info

- **Duplicati version**: Duplicati - 2.0.5.1_beta_2020-01-18
- **Operating system**: Ubuntu 20.04.1 LTS
- **Backend**: Backblaze B2

## Debug log

I DID create a 'bug report' zip file in Duplicati. Here is the content of the 'system-info.txt'-file: 

Duplicati: Duplicati.Server, Version=2.0.5.1, Culture=neutral, PublicKeyToken=null (Duplicati.Library.Main, Version=2.0.5.1, Culture=neutral, PublicKeyToken=null)
Autoupdate urls: https://updates.duplicati.com/beta/latest.manifest;https://alt.updates.duplicati.com/beta/latest.manifest
Update folder: /home/henrik/.config/Duplicati/updates
Base install folder: /usr/lib/duplicati
Version name: "2.0.5.1_beta_2020-01-18" (2.0.5.1)
[...]

PR #4448, opened 2021-02-21
Update most of python2 scripts throughout to make them compatible with both python2 and python3

jfparis:
Python2 has reached EoL
Python3 is the only supported version of python. 

This PR update the code of most python2 code in the duplicati repository to make it compatible with python3 (whilst maintaining backward compatibility with python 2 in case this is needed)

1 exception (which I left untouched) is the osX tray icon project as I don't have access to a mac to test this

### DIFF
--- a/Tools/Verification/DuplicatiVerify.py
+++ b/Tools/Verification/DuplicatiVerify.py
@@ -1,13 +1,14 @@
-#!/usr/bin/env python2
+#!/usr/bin/env python
 
 """
-This file is a standalone python script that is tested with python 2.7
+This file is a standalone python script that is tested with python 2.7 & python 3.8
 If Duplicati is producing a backup with the option --upload-verification-file,
 it will produce a *-verification.json file on the backend, which can be verified
 by this script. Simply run this script with the path to the backup
 folder as input, and it will verify all *-verification.json files in the folder.
 """
 
+from __future__ import print_function
 import sys
 import os
 import string
@@ -22,8 +23,9 @@ def bytes_from_file(filename, chunksize=8192):
         while True:
             chunk = f.read(chunksize)
             if chunk:
-                for b in chunk:
-                    yield b
+                #for b in chunk:
+                #    yield b
+                yield chunk
             else:
                 break
 
@@ -33,7 +35,7 @@ def verifyHashes(filename):
     checked = 0
 
     if (not os.path.exists(filename)):
-        print "Specified file does not exist: ", filename
+        print("Specified file does not exist:", filename)
         return -1
     
     folder = os.path.dirname(filename)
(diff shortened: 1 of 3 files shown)


warwickmm:
This is related to issue #4316.

jfparis:
And [this one](https://github.com/kees-z/DuplicatiDocs/issues/65)

warwickmm:
Thanks @jfparis.

Tools/Verification/DuplicatiVerify.py
```diff
@@ -1,98 +1,102 @@
...
+def bytes_from_file(filename, chunksize=8192*1024):
     with open(filename, "rb") as f:
         while True:
             chunk = f.read(chunksize)
             if chunk:
-                for b in chunk:
-                    yield b
+                #for b in chunk:
+                #    yield b
+                yield chunk
```
warwickmm, line 30: What was the purpose of changing the default `chunksize` and removing the `for` loop?

Tools/Verification/DuplicatiVerify.py
```diff
@@ -1,98 +1,102 @@
...
 if __name__ == "__main__":
-    if len(sys.argv) != 1 and len(sys.argv) != 2:
-        print "Usage:", len(sys.argv)
-        print "DuplicatiVerify.py <path to verification file>"
-        print "DuplicatiVerify.py <folder with verification files>"
-        print "DuplicatiVerify.py (no arguments, uses current dir)"
+    parser = argparse.ArgumentParser(description='Verify hashes of backup files.')
+    parser.add_argument("--quiet", action='store_true', help="Be noisy about each file being verified")
+    parser.add_argument("path", type=str, nargs='?',
+                    help="""path to the verification fole or folder containing the verification folder.\
```
warwickmm, line 82: Typo?  Should "fole" be "file"?

warwickmm: Should "verification folder" be "verification file"?

jfparis:
Hi

What the previous code was doing is
1 Read 8 k from the file
2 Feed it byte by byte to the hash function

It is not clear why this was done like that but sounds wrong

Dealing with 2 first.
* Sha256 has a block size of 64bytes so feeding it by byte is inefficient.
Results in a lot of superfluous function call in python which makes the
program dead slow for no reason
* The hash functions of python are meant to be fed a block of arbitrary
size
* As a result the loop can go and we are fine with whatever buffer size we
want to use

Then the size of the buffer
* 8k looks smallish
* 8Mb seems to produce a meaningful performance gain on my desktop machine
whilst maintaining an acceptable mem footprint / execution time on 8 year
old 256MB arm NAS box.

If you think the 8mb could be a problem I can revert or make it a command
line arguments (although I am not sure who uses fully fledged linux machine
smaller than mine 😃)

I shall push a fix for the typos

Thanks for reviewing

warwickmm:
Thanks @jfparis.  I tested `DuplicatiVerify.py` with both python2 and python3 and it seems to be working well.  The other two scripts are simply for maintenance are not deployed.

PR #4455, opened 2021-03-03
Use SI prefixes for sizes

### LINKED ISSUE #3685: Gibibytes vs Gigabytes
It appears, from #3682, that Duplicati is converting byte prefixes incorrectly for any size above 1024 bytes. 

This is AFAIK entirely cosmetical as the method causing the issue is always used for logs or UI displaying. Except for, perhaps, the throttle settings which currently claim to throttle in SI-prefixes but actually throttles in binary prefixes.

The cause is in [the utility class method FormatSizeString](https://github.com/duplicati/duplicati/blob/da64b4447f342199374e6c8c5871b8460d98b6bf/Duplicati/Library/Utility/Utility.cs#L600). It's converting Bytes into the binary prefixes GiB, MiB, and KiB but they are declared as scientific prefixes GB, MB, and KB to the user.
```
        public static string FormatSizeString(double size)
        {
            double sizeAbs = Math.Abs(size);  // Allow formatting of negative sizes
            if (sizeAbs >= 1024 * 1024 * 1024 * 1024L)
                return Strings.Utility.FormatStringTB(size / (1024 * 1024 * 1024 * 1024L));
```
Either the calculations should be changed to be SI - As in divisible by 1000, or the [string utility class](https://github.com/duplicati/duplicati/blob/c3136e31dfcf3c0857ed7f5eecf6cfb362d56df7/Duplicati/Library/Utility/Strings.cs#L27) should be modified to correctly print that they're binary and not SI prefixes.
```
    internal static class Utility {
        public static string FormatStringB(long size) { return LC.L(@"{0} bytes", size); }
        public static string FormatStringGB(double size) { return LC.L(@"{0:N} GB", size); }
```

Additionally, in order to handle the input throttle prefixes consistently, the [size parser utility](https://github.com/duplicati/duplicati/blob/c3136e31dfcf3c0857ed7f5eecf6cfb362d56df7/Duplicati/Library/Utility/Sizeparser.cs#L45) should also be modified to calculate SI prefixes, or to expect binary prefixes (and then the UI also tweaked to provide binary)

I personally believe going with SI is the right way for ideological reasons - and it is a simpler change because we're already calling them by SI names but just converting them wrong two places in the code.

samuel-w:
Fixes #3685

### DIFF
--- a/Duplicati/Library/Utility/Utility.cs
+++ b/Duplicati/Library/Utility/Utility.cs
@@ -643,23 +643,23 @@ namespace Duplicati.Library.Utility
         }
 
         /// <summary>
-        /// Formats a size into a human readable format, eg. 2048 becomes &quot;2 KB&quot; or -2283 becomes &quot;-2.23 KB%quot.
+        /// Formats a size into a human readable format, eg. 2000 becomes &quot;2 KB&quot; or -223 becomes &quot;-2.23 KB%quot.
         /// </summary>
         /// <param name="size">The size to format</param>
         /// <returns>A human readable string representing the size</returns>
         public static string FormatSizeString(double size)
         {
             double sizeAbs = Math.Abs(size);  // Allow formatting of negative sizes
-            if (sizeAbs >= 1024 * 1024 * 1024 * 1024L)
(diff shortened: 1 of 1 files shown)


samuel-w:
Closing pending further discussion in linked issue.

Issue #2833, opened 2017-10-20
Default of snapshot-policy should be auto


Tesla2k:
- [ x] I have searched open and closed issues for duplicates.

----------------------------------------

## Environment info

- **Duplicati version**: 2.0.2.1_beta_2017-08-01
- **Operating system**: Windows 10
- **Backend**: -

## Description

Duplicati should always use snaphosts if possible. This is what I except from a modern backup solution. It also prevents error messages with locked files. So the default for snapshot-policy should be "auto" instead of "off"

kenkendk:
The reason it is "off" is that it will otherwise create log entries in Windows Event Log with "permission denied". My thoughts were that people who know how to run Duplicati as Administrator can also turn on VSS.

Is it less of a problem that you get failure messages in the log?

Tesla2k:
Yes for me it would be better the other way.

When I first used duplicati. I was irritated, to get so many access denied errors (I backup whole C drive). I was thinking it uses snapshots, why all these errors. Then I found out that it is not runnging as Admin so it can't do snapshots. I managed to let it run as service., but still got the error! Till I finally found out, that it is not activated by default.
And honestly I never look at the Windows Event Log so I would never notice this message. :)

Shouldn't it be possible to check if it is runnning with to correct rights an only then try to make a snaphost?
Then we would have not errors ans snapshots would be made if possible.

PR #4456, opened 2021-03-04
Set snapshot-policy to auto by default

samuel-w:
Fixes https://github.com/duplicati/duplicati/issues/2833

### DIFF
--- a/Duplicati/Library/Main/Options.cs
+++ b/Duplicati/Library/Main/Options.cs
@@ -294,7 +294,7 @@ namespace Duplicati.Library.Main
                     
                     new CommandLineArgument("upload-unchanged-backups", CommandLineArgument.ArgumentType.Boolean, Strings.Options.UploadUnchangedBackupsShort, Strings.Options.UploadUnchangedBackupsLong, "false"),
 
-                    new CommandLineArgument("snapshot-policy", CommandLineArgument.ArgumentType.Enumeration, Strings.Options.SnapshotpolicyShort, Strings.Options.SnapshotpolicyLong, "off", null, Enum.GetNames(typeof(OptimizationStrategy))),
+                    new CommandLineArgument("snapshot-policy", CommandLineArgument.ArgumentType.Enumeration, Strings.Options.SnapshotpolicyShort, Strings.Options.SnapshotpolicyLong, "auto", null, Enum.GetNames(typeof(OptimizationStrategy))),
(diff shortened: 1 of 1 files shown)


warwickmm:
I'm not sure this is a good idea.  I believe `snapshot-policy` requires administrator/root privileges, and also doesn't work on MacOS.

samuel-w:
In auto mode, it tries and fails silently, according to the description:
> Setting this to "auto" makes Duplicati attempt to create a snapshot, and fail silently if that was not allowed or supported (note that the OS may still log system warnings).

warwickmm:
There's some discussion [here](https://forum.duplicati.com/t/ignore-locked-files/4326/5) about `snapshot-policy=auto` causing audio/visual notifications in Windows, which would be quite annoying.  What OS are you testing on?

@ts678, @drwtsn32x, any thoughts about this?

drwtsn32x:
I looked at that thread and don't see where `snapshot-policy=auto` itself causes notifications.  I thought the OP was talking about notifications due to locked files, which appears when snapshots aren't enabled.  But I admit I only skimmed the thread.

Intuitively I would think setting it to `auto` would not cause any issues, and could be a better way to go.  That being said, it won't help many people since the default install runs non-elevated anyway.

ts678:
The [explanation](https://github.com/duplicati/duplicati/issues/2833#issuecomment-345243861) for why this isn't the default is in #2833 that this PR means to fix. I haven't found the event yet. Has anyone else?

> The reason it is "off" is that it will otherwise create log entries in Windows Event Log with "permission denied".

warwickmm:
> I looked at that thread and don't see where snapshot-policy=auto itself causes notifications. I thought the OP was talking about notifications due to locked files, which appears when snapshots aren't enabled. But I admit I only skimmed the thread.

You're right, I think I misread the post.  Are windows notification events raised when we attempt to read a locked file?

kenkendk:
IIRC, the errors are shown in the Windows Event Log.
These errors occur if the user does not have permission to create the snapshot, but `auto` mode will not log the error in the Duplicati log.

I personally don't think it is a good idea to set it to auto.
- It only works if the user has permission
- The hard part is setting it up to run with elevated permissions
- It is easy to change the setting
- Snapshots are fairly invasive, with a few caveats of their own

A better approach would be to query if the current user has sufficient permissions, and if that is the case set the default to `On` (I dislike `auto` in general because you either want VSS or not, toggling without error messages is likely to cause issues down the line).

warwickmm:
That makes sense to me.  Thanks @samuel-w for the proposed fix, but I agree with @kenkendk that we should consider a different approach for this.

PR #4459, opened 2021-03-06
Changed the help output to not show deprecated options.

kenkendk:
Commandline help for azure now shows a little cleaner


kenkendk:
The idea here is that users listing the help gets only displayed options that are intended to be used (we can remove deprecated options later).
If they go specifically to get help on an option (deprecated or not) that option will be show, including the deprecation status.
In other words, the option help is still there, it is just not shown unless you look for it.

PR #4462, opened 2021-03-07
S3 configurations result in warnings about deprecated options

### LINKED ISSUE #4461: UI still uses deprecated S3 option names
## Environment info

- **Duplicati version**: 2.0.5.113

## Description
The `aws_access_key_id` and `aws_secret_access_key` S3 options were deprecated in version 2.0.5.113 (see pull request #4443).  However, the deprecated name are still being used in the UI in `EditUriBuiltins.js`

warwickmm:
The old option names (`aws_access_key_id` and `aws_secret_access_key`) were deprecated in favor of their respective hyphen-separated versions (`aws-access-key-id` and `aws-secret-access-key`) in pull request #4443. However, both the old and new options were mapped to the same alias, which resulted in the `Controller` class issuing a warning below:

https://github.com/duplicati/duplicati/blob/f145b9fcc2f93aaa606bc939373afbb566034647/Duplicati/Library/Main/Controller.cs#L774-L789

This pull request reserves the alias for just the new option name, and removes the alias from the deprecated option.

This also replaces some usages of the old option names in the UI code.

This fixes #4461.

### DIFF
--- a/Duplicati/Server/webroot/ngax/scripts/services/EditUriBuiltins.js
+++ b/Duplicati/Server/webroot/ngax/scripts/services/EditUriBuiltins.js
@@ -339,10 +339,10 @@ backupApp.service('EditUriBuiltins', function (AppService, AppUtils, SystemInfo,
     ];
 
     EditUriBackendConfig.parsers['s3'] = function (scope, module, server, port, path, options) {
-        if (options['--aws_access_key_id'])
-            scope.Username = options['--aws_access_key_id'];
-        if (options['--aws_secret_access_key'])
-            scope.Password = options['--aws_secret_access_key'];
+        if (options['--aws-access-key-id'])
+            scope.Username = options['--aws-access-key-id'];
+        if (options['--aws-secret-access-key'])
+            scope.Password = options['--aws-secret-access-key'];
 
         if (options['--s3-use-rrs'] && !options['--s3-storage-class']) {
             delete options['--s3-use-rrs'];
@@ -371,7 +371,7 @@ backupApp.service('EditUriBuiltins', function (AppService, AppUtils, SystemInfo,
         
         scope.s3_storageclass = scope.s3_storageclass_custom = options['--s3-storage-class'];
 
-        var nukeopts = ['--aws_access_key_id', '--aws_secret_access_key', '--s3-use-rrs', '--s3-server-name', '--s3-location-constraint', '--s3-storage-class', '--s3-client'];
+        var nukeopts = ['--aws-access-key-id', '--aws-secret-access-key', '--s3-use-rrs', '--s3-server-name', '--s3-location-constraint', '--s3-storage-class', '--s3-client'];
         for (var x in nukeopts)
             delete options[nukeopts[x]];
     };
--- a/Duplicati/Library/Backend/S3/S3Backend.cs
+++ b/Duplicati/Library/Backend/S3/S3Backend.cs
@@ -391,9 +391,9 @@ namespace Duplicati.Library.Backend
 
 
                 var normal = new ICommandLineArgument[] {
-                    new CommandLineArgument("aws_secret_access_key", CommandLineArgument.ArgumentType.Password, Strings.S3Backend.AMZKeyDescriptionShort, Strings.S3Backend.AMZKeyDescriptionLong, null, new string[] {"auth-password"},null,"This is deprecated, use aws-secret-access-key instead"),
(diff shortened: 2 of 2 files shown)


Duplicati/Server/webroot/ngax/scripts/services/EditUriBuiltins.js
```diff
@@ -339,10 +339,10 @@ backupApp.service('EditUriBuiltins', function (AppService, AppUtils, SystemInfo,
     ];
 
     EditUriBackendConfig.parsers['s3'] = function (scope, module, server, port, path, options) {
-        if (options['--aws_access_key_id'])
-            scope.Username = options['--aws_access_key_id'];
-        if (options['--aws_secret_access_key'])
-            scope.Password = options['--aws_secret_access_key'];
+        if (options['--aws-access-key-id'])
```
warwickmm, line 342: Does anyone have an S3 backend that they can test with?  I'm not sure if `options` here is guaranteed to contain the new name, or if it's possible to contain the old deprecated name.

/cc @kenkendk, @drwtsn32x, @ts678

ts678:
I don't use S3, although I suppose anyone could set up a fake backup. My theory was that the GUI (maybe even the 2.0.5.113 one) attaches the deprecated option to the URL. If so, maybe Duplicati then chokes out a warning. That would be fixable on the four-dot editing menu, but that's not something we want to force everyone to do. If it affects all users on these backends, that's a big deal...

warwickmm:
I don't think it affects all backends (I tested local and SFTP and didn't encounter any warnings).  I was able to setup a new B2 bucket that uses S3, and I get the warnings.  However, I'm not sure where they're coming from yet.

warwickmm:
For my test case where I see the warnings, neither the old nor the new `aws-access-key-id` option appears in the `options` provided to the `S3` constructor.  It must be coming from somewhere else.

warwickmm:
OK, I suspect the problem is the following.  In the `Controller` class, we collect a bunch of options here

https://github.com/duplicati/duplicati/blob/f145b9fcc2f93aaa606bc939373afbb566034647/Duplicati/Library/Main/Controller.cs#L749-L755

In particular, `DynamicLoader.BackendLoader.GetSupportedCommands(m_backend)` returns the collection below, which contains both the deprecated and new versions:

https://github.com/duplicati/duplicati/blob/f145b9fcc2f93aaa606bc939373afbb566034647/Duplicati/Library/Backend/S3/S3Backend.cs#L393-L397

I'm not sure what's supposed to happen here, but if I remove the deprecated versions from the `SupportedCommands` an existing S3 backup seems to work fine in 2.0.5.113.

ts678:
> I don't think it affects all backends

My bad wording. I meant all backends affected by the option change, which maybe is S3 **and** Azure? I'm not following too closely.
I just posted in the forum asking the user to try the three-dot menu to try to suppress the warning from their Wasabi S3 backup.

warwickmm:
I take back my previous comment...I don't think removing an entry from `SupportedCommands` is the correct approach.

It looks like the URL contains `auth-username`, which is an alias for both `aws_access_key_id` and `aws-access-key-id`
https://github.com/duplicati/duplicati/blob/f145b9fcc2f93aaa606bc939373afbb566034647/Duplicati/Library/Backend/S3/S3Backend.cs#L396-L397

When we check all relevant options (from the collection noted above), the deprecated `aws_access_key_id` has alias `auth-username`, which is in the collection of "actually used options" `ropts`:
https://github.com/duplicati/duplicati/blob/f145b9fcc2f93aaa606bc939373afbb566034647/Duplicati/Library/Main/Controller.cs#L774-L789

I believe a better solution is for the deprecated options to not be aliased anymore.  We should have @kenkendk confirm.

warwickmm:
I tested this by created an S3 (B2) backup in 2.0.5.1 and ran it.  I then "upgraded" to 2.0.5.113 and ran a backup and encountered the warnings.  I then "upgraded" to this branch and ran a backup that completed without any warnings.

kenkendk:
I had not considered this effect. But I think you the fix presented here solves it for the S3 backend for most cases.
I think the Azure backend should be fine as well, as it is also using the `auth-username` properties instead of the option names directly.

I think it is actually inconsistent that the Azure backend is not using Alias, as the options are aliased. We could remove some code from the Azure backend if this was updated.

For the web-ui, I think we should check for both versions (`_` and `-`), in case the user somehow manages to import some old configuration. Suggestion:

```js
        if (options['--aws-access-key-id'])
            scope.Username = options['--aws-access-key-id'];
        else if (options['--aws_access_key_id'])
            scope.Username = options['--aws_access_key_id'];
        if (options['--aws-secret-access-key'])
            scope.Password = options['--aws-secret-access-key'];
        else if (options['--aws_secret_access_key'])
            scope.Password = options['--aws_secret_access_key'];
```

And the delete part should take out both variants.

I have added #4464 with the fixes.

warwickmm:
Declining this in favor of pull request #4464.

Issue #3071, opened 2018-02-27
Can't update Duplicati to latest version


thomasyi17:
- [x ] I have searched open and closed issues for duplicates.

https://github.com/duplicati/duplicati/issues/2530
^I could not find any lingers update folders to delete as referenced in this issue, but I doubt there would be any lingering versions considering this is my first update attempt 

----------------------------------------

## Environment info

- **Duplicati version**: 2.0.2.19/2.0.2.20 (Running as service)
- **Operating system**: Windows 7 Enterprise
- **Backend**: FTP

## Description

(New to Duplicati, so apologies for any novice mistakes)
I can't seem to activate the new version of Duplicati from 2.0.2.19 to 2.0.2.20. Following a check for updates, I am presented with text inviting me to activate the new version as it has been installed (see screenshot). However, nothing happens upon clicking "activate now." Multiple clicks in quick succession result in a popup saying "Activate failed: Missing XSRF Token."

For reference, I am running Duplicati as a service to avoid launching tray icons and eating up 8XXX ports for each user when accessing the GUI.

## Steps to reproduce

1. Check for updates
2. Attempt to activate

- **Actual result**:
Inability to activate following download and "installation"
- **Expected result**:
Activates the new version

## Screenshots

![image](https://user-images.githubusercontent.com/8674491/36755987-d1913dce-1bdb-11e8-9e96-446ace0c8521.png)

## Debug log

Sadly, there didn't appear to be any output in the "Show log" tab. Happy to look for logs elsewhere.

thomasyi17:
After restarting the service, the activation appears to go through.

thomasyi17:
Would be cool if the service automatically restarted to allow for activation, or if the client prompted the user to restart the service

Reuenthal767:
I had the same problem today, I was trying to udpate from v2.0.4.23-2.0.4.23_beta_2019-07-14 to v2.0.5.0-2.0.5.0_experimental_2020-01-03 and was getting the same error.

Activate failed: Missing XSRF Token.

Restarting the duplicati service in services.msc did the trick.

Thank you @thomasyi17

ts678:
I looked at this some during 2.0.7.100 Canary rollout when someone reported difficulty getting Windows service update going.
Below is information detail, and there's also information in the PR just above that looks like it was closed before being released.

Sysinternals Process Explorer data on the process tree:
```
PID     Started                         Name                            Environment variable of interest
[...]

PR #4467, opened 2021-03-12
Restart Duplicati Service to apply updates

Germs2004:
fixes the problem (#3071) where the "Apply Updates" button does not work when Duplicati is running as a Windows Service.  This change causes the Duplicati service to restart, which successfully applies the updates.

Note: This may need some tweaks to ensure it only affects Duplicati on Windows, not Linux.

### DIFF
--- a/Duplicati/Server/Program.cs
+++ b/Duplicati/Server/Program.cs
@@ -1,5 +1,6 @@
 ﻿using System;
 using System.Collections.Generic;
+using System.Diagnostics;
 using System.Globalization;
 using System.Linq;
 using Duplicati.Library.Common;
@@ -302,9 +303,53 @@ namespace Duplicati.Server
             }
 
             if (UpdatePoller != null && UpdatePoller.IsUpdateRequested)
+            {
+                if (WindowsServiceExists("Duplicati"))
+                {
+                    RestartWindowsService("Duplicati");
+                }
+                
                 return Library.AutoUpdater.UpdaterManager.MAGIC_EXIT_CODE;
+            }
 
             return 0;
+        }
+        
+        public static bool WindowsServiceExists(string serviceName)
+        {
+            string cmdOutput;
+            using (Process p = new Process())
+            {
+                p.StartInfo.UseShellExecute = false;
+                p.StartInfo.RedirectStandardOutput = true;
+                p.StartInfo.FileName = "cmd.exe";
+                p.StartInfo.Arguments = "/c net start | find \"" + serviceName + "\"";
+                p.Start();
+                cmdOutput = p.StandardOutput.ReadToEnd();
+                p.WaitForExit();
+            }
+
+            return !string.IsNullOrEmpty(cmdOutput);
+        }
+
+        public static void RestartWindowsService(string serviceName)
+        {
+            ProcessStartInfo psi = new ProcessStartInfo()
+            {
+                CreateNoWindow = true,
+                FileName = @"C:\Windows\System32\cmd.exe",
+                Arguments = "/C net stop " + serviceName + " && net start " + serviceName,
+                LoadUserProfile = false,
+                UseShellExecute = false,
+                WindowStyle = ProcessWindowStyle.Hidden
+            };
+
+            Process p = new Process()
+            {
+                StartInfo = psi
+            };
+
+            p.Start();
         }
 
         private static void StartWebServer(Dictionary<string, string> commandlineOptions)


warwickmm:
Thanks @Germs2004.  Unfortunately, I'm not very familiar with this part of the code and I don't have Windows, so I can only offer some limited help.

It looks like an earlier attempt was made in revision 4611ddab2a3050839f88059036b8b4360c33be5c but reverted in 8dd3d1376ca36a13932b8652cc66675c78086aa9, with the [release notes](https://github.com/duplicati/duplicati/releases/tag/v2.0.3.14-2.0.3.14_canary_2018-11-08) saying 

> Removed the WindowsService restart fix as it did not work as expected

Perhaps that revision might provide some additional insight into the issue.

Duplicati/Server/Program.cs
```diff
@@ -302,10 +303,54 @@ public static int RealMain(string[] _args)
             }
 
             if (UpdatePoller != null && UpdatePoller.IsUpdateRequested)
+            {
+                if (Platform.IsClientWindows && WindowsServiceExists("Duplicati"))
```
warwickmm, line 307: Would it be safer to reference `WindowsService.ServiceControl.SERVICE_NAME` instead of hard-coding `"Duplicati"` here?

https://github.com/duplicati/duplicati/blob/beaf03562fdcf4425e962085bdf7175d6a465f49/Duplicati/WindowsService/ServiceControl.cs#L15

Germs2004: Yes, that's a good improvement.  I wasn't aware that was available.

I should also note that .Net has a library called System.ServiceProcess.dll, which has a managed code object called "ServiceController" for working with services.  Even though that code would be better than my solution of invoking a command line program and parsing its output, I didn't know if adding a reference to that new dll would make life more difficult for anyone since it would be the only code in this .Net Project that requires that reference.  If adding that new dll reference would be fine with everyone,  the code could be rewritten to use that instead.

Germs2004:
> Perhaps that revision might provide some additional insight into the issue.

There is a good concept in that revision.  My code is telling the service to restart from the duplicati.server.exe, which works fine for me, but that other revision attempts to restart the service from within the service.  I'm looking into that now to see if I want to rewrite my solution.

_For anyone else investigating how this works, it looks like every time you start your Duplicati Windows Service , it runs the version of "Duplicati.WindowsService.exe" that was originally installed (hopefully it has no major bugs).  That file apparently never gets updated.  Then, that file looks for the most recent subfolder in the "updates" folder, and executes a copy of that newer Duplicati.WindowsService.exe and that newer Duplicati.Server.exe.  That's how updates work.  The updates don't overwrite the main program files - they just add a subfolder, and the next launch of Duplicati looks for the latest update folder and executes that as a child process._

Germs2004:
I've spent a few more hours digging into Duplicati code and I'm ready to say that my solution is good enough for me to present for the testers.

If I wanted to spend many more hours on it, it would be better if Duplicati.Server could be made to communicate upwards to the child Duplicati.WindowsService, and that could communicate upwards to the root Duplicati.WindowsService.  Then the root service could relaunch the child service from the latest update folder, and that child service could relaunch the Duplicati.Server from the latest update folder.  Someone is welcome to make that improvement if they can figure it out, but for now, I think this solution of simply telling the Duplicati service to restart directly from Duplicati.Server as it exits should will work 99% of the time.

I suggest having a few Windows users and Linux users test this before it gets published to the world just in case it affects something unexpected, but I hope that all the Windows users who have until now had to manually restart their service after every update will be pleased.

Duplicati/Server/Program.cs
warwickmm: If using System.ServiceProcess.dll cleans things up, I would be in favor of referencing it.

warwickmm:
Thanks @Germs2004.  Our resources are a bit thin at the moment so it may take some time before someone gets a chance to test this.  Also, we are in the midst of a feature freeze while we work on the next beta release, so to simplify the workflow we probably won't be merging many changes before then.

drwtsn32x:
Spawning cmd.exe and calling net.exe stop/start isn't the prettiest way to do this. Utilizing the ServiceController class would be a lot better.

Duplicati/Server/Program.cs
```diff
@@ -302,10 +303,54 @@ public static int RealMain(string[] _args)
             }
 
             if (UpdatePoller != null && UpdatePoller.IsUpdateRequested)
+            {
+                if (Platform.IsClientWindows && !String.IsNullOrWhiteSpace(WindowsService.ServiceControl.SERVICE_NAME) && WindowsServiceExists(WindowsService.ServiceControl.SERVICE_NAME))
```
sergethedev17, line 307: This checks only if the Duplicati service _is installed_. The proposed check:

1. does not check if the Duplicati service is running;
2. does not check that the process currently executing is in fact the service.

I had a look at `Duplicati\WindowsService\ServiceControl.cs`. It seems that `ServiceControl` creates a new `Duplicati.Service.Runner' instance, which in turn starts a new `Duplicati.Server.exe` process. Perhaps the `Duplicati.Server.exe` process should check:

1. am I a child process of `Duplicati.WindowsService.exe`?
2. is the parent process running as a service and what is its Process Id?

sergethedev17:
Once the `Duplicati.Server.exe` process reliably determines that it is running as a service (see comment above), then it can restart itself this way:

1. Call `Environment.Exit(EXITCODE_RESTART)` where `const int EXITCODE_RESTART = 3` (or some other value), to signal that it wants to be restarted.
2. When the process exits, the loop in `Duplicati.Service.Runner.Run()` method checks the process exit code. If exit code = EXITCODE_RESTART, then Runner restarts the `Duplicati.Server.exe` process.

This avoids potential issues with external utilities (eg `net start`, `sc`) and privileges required to restart a Windows Service.

The downside is that the Service itself is not actually restarted. However, `Duplicati.WindowsService.ServiceControl` method does not have a facility to auto-update, so this limitation may be OK.

ts678:
This is not only a Windows service issue. The Activate button is too unreliable. What's supposed to happen is exit code based.

Duplicati process are usually paired. The parent has a loop that runs most recent update, going again on MAGIC_EXIT_CODE.

https://github.com/duplicati/duplicati/blob/ed86cca3e468f4a3e33c3d0faec1d22cb194bda2/Duplicati/Library/AutoUpdater/UpdaterManager.cs#L1163-L1165

https://github.com/duplicati/duplicati/blob/ed86cca3e468f4a3e33c3d0faec1d22cb194bda2/Duplicati/Server/Program.cs#L304-L305

For a Windows service, the pairing is two Duplicati.WindowsService.exe, and one Duplicati.Server.exe. Why not paired Server?

https://github.com/duplicati/duplicati/blob/ed86cca3e468f4a3e33c3d0faec1d22cb194bda2/Duplicati/Library/AutoUpdater/UpdaterManager.cs#L1088-L1089

so possibly by fiddling with where the parent/child happens, and with exit codes, one could get MAGIC_EXIT_CODE to work.
This could be independent efforts. One could get it to work in the general case, and one could ask if it fits the service model.

I'm not a Windows service expert, and am not sure how a service is supposed to die, ask for a restart if possible, and so on...
If we can keep the restart code within Duplicati Server, leave Service always up, and have everything work, I would be happy.

As a side note, I just installed service on a system that already used an update, and it pointed to the update Service version.
This works until that update is aged off (there is a limit), and at that point the service will break due to no executable to run.

image

Service install would be more long-lasting if it installed the base version instead, e.g. from `C:\Program Files\Duplicati 2`

[Replace auto updater #4542](https://github.com/duplicati/duplicati/issues/4542) is a proposal to do more usual updates, but if it doesn't happen, fixing current one would help.

kenkendk:
I know several users have reported issues with the updater system. For some people it simply will not accept the update.
It is really hard to debug, but I suspect that an Antivirus or similar system will write/touch stuff in the updates folder, which will cause the updater to reject the update (a check done to avoid running unwanted code).

> Service install would be more long-lasting if it installed the base version instead,

Yes!

The executable that starts is considered the "base" version (check via environment variable). If you run from the updates folder, it will consider that version the "base" version. New updates will then be installed in `updates/updates/updates/...` causing unwanted issues.

ts678:
> The executable that starts is considered the "base" version (check via environment variable). If you run from the updates folder

This is a different issue than the vulnerable service install. Here is a transcript of 2.0.5.114 Canary base after update to 2.0.6.100:

```
C:\Program Files\Duplicati 2>Duplicati.WindowsService.exe install

Running a transacted installation.

Beginning the Install phase of the installation.
See the contents of the log file for the C:\ProgramData\Duplicati\updates\2.0.6.100\Duplicati.WindowsService.exe assembly's progress.
The file is located at C:\ProgramData\Duplicati\updates\2.0.6.100\Duplicati.WindowsService.InstallLog.
Installing assembly 'C:\ProgramData\Duplicati\updates\2.0.6.100\Duplicati.WindowsService.exe'.
Affected parameters are:
   logtoconsole =
   assemblypath = C:\ProgramData\Duplicati\updates\2.0.6.100\Duplicati.WindowsService.exe
   logfile = C:\ProgramData\Duplicati\updates\2.0.6.100\Duplicati.WindowsService.InstallLog
   commandline =
Installing service Duplicati...
Service Duplicati has been successfully installed.
Creating EventLog source Duplicati in log Application...

The Install phase completed successfully, and the Commit phase is beginning.
See the contents of the log file for the C:\ProgramData\Duplicati\updates\2.0.6.100\Duplicati.WindowsService.exe assembly's progress.
The file is located at C:\ProgramData\Duplicati\updates\2.0.6.100\Duplicati.WindowsService.InstallLog.
Committing assembly 'C:\ProgramData\Duplicati\updates\2.0.6.100\Duplicati.WindowsService.exe'.
Affected parameters are:
   logtoconsole =
   assemblypath = C:\ProgramData\Duplicati\updates\2.0.6.100\Duplicati.WindowsService.exe
   logfile = C:\ProgramData\Duplicati\updates\2.0.6.100\Duplicati.WindowsService.InstallLog
   commandline =

The Commit phase completed successfully.

The transacted install has completed.

C:\Program Files\Duplicati 2>
```

The installer does the usual [parent/child](https://github.com/duplicati/duplicati/blob/7d71c6a4797cc62a8283c150bda90713fe472474/Duplicati/Library/AutoUpdater/UpdaterManager.cs#L1148), installs from child, which is from updates, and is vulnerable to eventual [version removal](https://github.com/duplicati/duplicati/blob/7d71c6a4797cc62a8283c150bda90713fe472474/Duplicati/Library/AutoUpdater/UpdaterManager.cs#L541-L549).

I'm not familiar with Windows service installs, but I suspect the path is set by:

https://github.com/duplicati/duplicati/blob/7d71c6a4797cc62a8283c150bda90713fe472474/Duplicati/WindowsService/ProjectInstaller.cs#L39-L42

I also notice from the transcript that it wrote log files into the updates directory, thus corrupting it and keeping it from being used.
Sorry if I'm sidetracking this pull request (I'm commenting on existing code), but I thought I would respond to @kenkendk remark, which I might misunderstand. Maybe it's saying that updates\2.0.6.100 won't get removed by the service, but will get other issues.

I'm still not ruling out version removal by some non-service program that started from the C:\Program Files\Duplicati 2 base install.

ts678:
Would anyone be able to find why MAGIC_EXIT_CODE isn't being set? It might help to make the [above](https://github.com/duplicati/duplicati/pull/4467#issuecomment-852593583) `if` at line 304 multi-lines.

Somewhere along the line, Duplicati has decided to return 0 instead of MAGIC_EXIT_CODE when the Activate button gets pressed.
I think the Server (run as a Windows service) and TrayIcon (more widely used -- may need fix even more) use the same mechanism.

I'm not set up for heavy debugging needing code change, but I did one-time fix TrayIcon Activate by changing return value to 126:

https://github.com/duplicati/duplicati/blob/7d71c6a4797cc62a8283c150bda90713fe472474/Duplicati/Library/AutoUpdater/UpdaterManager.cs#L1096-L1100

I haven't tried that debugger edit-the-local-variable on a service yet, but if I did, I'd probably have service run as me (not SYSTEM).

On Windows, [Process Monitor](https://docs.microsoft.com/en-us/sysinternals/downloads/procmon) is an easy way to see exit codes. If exit code is 0 (as it often is on Activate), parent process also exits.

Issue #4438, opened 2021-02-10
OneDrive respect "retry-after" header


basrieter:
- [x] I have searched open and closed issues for duplicates.

----------------------------------------

## Environment info

- **Duplicati version**: 2.0.5.1_beta_2020-01-18
- **Operating system**: Windows 10 Pro
- **Backend**: OneDrive v2

## Description

I am restoring 400 GB of data for verification of backup integrity. After some time I am getting *Error 429: Too Many Requests*. In the response there is a value "retry-after" indictating the number of seconds to wait. Duplicity does not seem to use this and retries after the configured time (10 seocnds for me)

## Steps to reproduce

1. Restore 400 GB (50 MB volume size) from OneDrive

- **Actual result**:
*Error 429: Too Many Requests* errors and failing restores. 
- **Expected result**:
Using the "retry-after" value  and wait for the given time? 

## Debug log

Part of the OneDrive response:
```
"StatusCode": 429,
		"ReasonPhrase": "Too Many Requests",
		"Headers": [{
				"Key": "Retry-After",
				"Value": ["112"]
			}, {
				"Key": "X-MSNSERVER",
				"Value": ["BN4SCH101062818"]
			}, {
				"Key": "Strict-Transport-Security",
				"Value": ["max-age=31536000; includeSubDomains"]
			}, {
				"Key": "MS-CV",
				"Value": ["qdUJs/gP4kq4LJnLXV+K4A.0"]
			}, {
				"Key": "X-QosStats",
				"Value": ["{\"ApiId\":0,\"ResultType\":2,\"SourcePropertyId\":0,\"TargetPropertyId\":42}"]
			}, {
				"Key": "X-ThrowSite",
				"Value": ["3062.afa1"]
			}, {
				"Key": "X-ClientErrorCode",
				"Value": ["ThrottledRequest"]
			}, {
				"Key": "X-ErrorCodeChain",
				"Value": ["AccessDenied,ThrottledRequest"]
			}, {
				"Key": "X-AsmVersion",
				"Value": ["UNKNOWN; 19.619.204.2006"]
			}, {
				"Key": "Date",
				"Value": ["Wed, 10 Feb 2021 12:33:29 GMT"]
			}
		]
```

duplicatibot:
This issue has been mentioned on **Duplicati**. There might be relevant details there:

https://forum.duplicati.com/t/throttle-not-working/11742/7

PR #4469, opened 2021-03-16
Try to fix OneDrive upload throttling

tygill:
This change attempts to fix OneDrive's upload throttling. It does so by hooking the input stream directly up to the request stream during PutAsync() calls instead of reading into a byte[] buffer and then uploading that. The input stream is split into separate fragments by wrapping it in the newly introduced SubStream class, which wraps a stream and exposes only a limited slice of it.

This change also addresses #4438 by adding support for responding to the Retry-After header in the OneDrive backend. It does so by detecting this header when checking each response, and waiting until that point before sending any future requests.

There are also a variety of other minor changes to this backend (changing from Thread.Sleep to Task.Delay, cancelling upload sessions when throwing an exception, etc.)

I've also added some basic buffer caching to Utility.CopyStream so a new buffer isn't allocated for every copy.


Duplicati/Library/Utility/SubStream.cs
```diff
@@ -0,0 +1,118 @@
...
+
+        public override long Position
+        {
+            get => this.m_stream.Position - this.m_start;
+            set => this.m_stream.Position = value + this.m_start;
+        }
+
+        public override void Flush()
+        {
+            this.m_stream.Flush();
```
warwickmm, line 77: According to the [documentation](https://docs.microsoft.com/en-us/dotnet/api/system.io.stream.flush?view=netframework-4.7.1),

> In a class derived from Stream that doesn't support writing, Flush is typically implemented as an empty method to ensure full compatibility with other Stream types since it's valid to flush a read-only stream.

Since `CanWrite` is `false`, should we implement `Flush` to be a no-op?

Duplicati/Library/Utility/Utility.cs
```diff
@@ -18,6 +18,7 @@
 using System.Collections.Generic;
 using System.IO;
 using System.Linq;
+using System.Threading;
```
warwickmm, line 21: It looks like this `using` directive already exists on line 28 below.

Duplicati/Library/Utility/SubStream.cs
tygill: That seems reasonable to me. Updated.

Duplicati/Library/Utility/Utility.cs
tygill: Thanks for pointing this out - I'm not sure how this happened. Maybe Visual Studio automatically added it?

At any rate, it's now cleaned up.

Duplicati/Library/Utility/SubStream.cs
```diff
@@ -0,0 +1,118 @@
...
+using System.Threading.Tasks;
+using System.IO;
+
+namespace Duplicati.Library.Utility
+{
+    /// <summary>
+    /// This class wraps a Stream but only allows reading a limited number of bytes from the underlying stream.
+    /// This can be used to create split a source stream into multiple smaller buffers that can be used as independent streams.
+    /// </summary>
+    public class SubStream : Stream
```
warwickmm, line 34: On the surface, this appears to have a similar purpose to [ReadLimitLengthStream](https://github.com/duplicati/duplicati/blob/v2.0.5.114-2.0.5.114_canary_2021-03-10/Duplicati/Library/Utility/ReadLimitLengthStream.cs).  The implementation is different, probably due to the different use cases.  If these indeed are meant to provide similar functionality, do you see any way to consolidate the two?

tygill: I guess I should have looked harder for something that already did this - I didn't see that class. After a quick look, I think I can add my new features to that class and use it instead.

tygill: Looks like my change is causing some tests to fail. I'll have to look more into it.

tygill: Tests are fixed and some new tests are added to validate the new behavior on ReadLimitLengthStream.

Duplicati/Library/Utility/ReadLimitLengthStream.cs
```diff
@@ -35,12 +66,12 @@ public override void Flush()
             // NOOP
         }
 
-        public override long Length => m_endPosition;
+        public override long Length => m_length;
```
warwickmm, line 69: This appears to change the behavior, but your version makes more sense to me.  Was the previous implementation a bug, or was this property not read?

Duplicati/Library/Utility/ReadLimitLengthStream.cs
```diff
@@ -50,8 +81,24 @@ public override int Read(byte[] buffer, int offset, int count)
...
+                    innerPosition = m_innerStream.Seek(offset, SeekOrigin.Current);
+                    break;
+                case SeekOrigin.End:
+                    innerPosition = m_innerStream.Seek(m_start + m_length + offset, SeekOrigin.Begin);
+                    break;
+                default:
+                    throw new ArgumentException("Unknown SeekOrigin", nameof(origin));
+            }
+
+            return ClampInnerStreamPosition(innerPosition) - m_start;
```
warwickmm, line 101: Since we have seeked `m_innerStream`, can we return `this.Position` to avoid duplicating the calculation?

Duplicati/Library/Utility/ReadLimitLengthStream.cs
```diff
@@ -138,14 +185,38 @@ public override void WriteByte(byte value)
...
                 return (int)maxCount;
             }
 
             return count;
         }
+
+        private long ClampInnerStreamPosition(long position)
+        {
+            // Note that this allows this stream to have positions in the range -1 to m_length.
+            // Reading at -1 or m_length should return nothing.
```
warwickmm, line 208: This feels a bit strange, since other code might perform calculations using the `Position` property with the assumption that the value is nonnegative.  Would you be able to explain the reason for allowing `Position` to be negative?

Duplicati/Library/Utility/ReadLimitLengthStream.cs
```diff
@@ -35,12 +66,12 @@ public override void Flush()
...
 
-        public override long Length => m_endPosition;
+        public override long Length => m_length;
 
         public override long Position
         {
-            get => m_innerStream.Position;
-            set => m_innerStream.Position = value;
+            get => ClampInnerStreamPosition(m_innerStream.Position) - m_start;
+            set => m_innerStream.Position = ClampInnerStreamPosition(value + m_start);
```
warwickmm, line 74: I think most implementations (e.g., `FileStream`, `MemoryStream`, etc.) disallow `value` to be negative.  This seems to allow for it, and we also have a test for the case `value = -1`.  Is there a reason for this (possibly related to my [other comment](https://github.com/duplicati/duplicati/pull/4469/files#r597851972))?

Duplicati/Library/Utility/ReadLimitLengthStream.cs
tygill: If you're referring to the Length property, I suppose you are right that if the innerStream wasn't starting at 0, then this length wouldn't have been correct. But this class was only used with new streams, so the bug likely didn't ever come up.

Duplicati/Library/Utility/ReadLimitLengthStream.cs
tygill: Indeed we can.

Duplicati/Library/Utility/ReadLimitLengthStream.cs
tygill: The reason I had done this was mostly an attempt at giving some consistent behavior when the underlying stream is positioned before the limited stream's starting point - I had originally been thinking that if the underlying stream was positioned before the start point, that Read() calls should fail (in part because if the stream couldn't be seeked, then we would be in a case where the stream reports a position of 0, but can't be read). However, I realized that even if we can't seek forward, we could at least read until reaching the start position, which removes that oddity. So I've revised this to clamp the reported position to 0 if the base stream is positioned before the start. Attempting to read will cause the stream to be seeked to start position and will then read from there.

Duplicati/Library/Utility/ReadLimitLengthStream.cs
tygill: See the other comment - I've changed this and made the behavior when the base stream is at a Position before the starting point consistent with that.

tygill: (Thanks for all the review comments, by the way!)

Duplicati/UnitTest/UtilityTests.cs
```diff
@@ -125,6 +125,164 @@ public static void ReadLimitLengthStreamWithGrowingFile()
...
+                stream.Position = position;
+                Assert.AreEqual(position, stream.Position);
+                Assert.AreEqual(readByte, stream.ReadByte());
+
+                Assert.AreEqual(position, stream.Seek(position, SeekOrigin.Begin));
+                Assert.AreEqual(position, stream.Position);
+                Assert.AreEqual(readByte >= 0 ? 1 : 0, stream.Read(readBuffer, 0, 1), "Read count");
+                if (readByte >= 0)
+                {
+                    Assert.AreEqual(readBuffer[0], readByte);
```
warwickmm, line 155: Nice job with the tests.  Would it be a slight improvement in coverage to provide an offset of `1` to `stream.Read` on line 152, and check `readBuffer[1]` here to verify that the offset is used correctly?

tygill: That's not a bad idea. I've also set it to clear the array first and make sure the indexes before and after are clear (and make sure nothing is read when Read returns 0.)

Duplicati/Library/Utility/ReadLimitLengthStream.cs
```diff
@@ -4,24 +4,66 @@
...
-    // StreamReadLimitLengthWrapper() based on code from Matt Smith (Oct 26 '15 at 20:38)
-    // https://stackoverflow.com/questions/33354822/how-to-set-length-in-stream-without-truncated
-
+{
+    // StreamReadLimitLengthWrapper() based on code from Matt Smith (Oct 26 '15 at 20:38)
+    // https://stackoverflow.com/questions/33354822/how-to-set-length-in-stream-without-truncated
+
+    /// <summary>
+    /// This class wraps a Stream but only allows reading a limited number of bytes from the underlying stream.
+    /// This can be used to create split a source stream into multiple smaller buffers that can be used as independent streams.
```
warwickmm, line 13: > ... that can be used as independent streams.

I guess we should still be careful with how we use this class.  Given a base stream, I think a second instance of `ReadLimitLengthStream` should not be instantiated until existing instances with the same base stream have been disposed.

For example, I believe the following test will fail:
```
MemoryStream baseStream = new MemoryStream(Enumerable.Range(0, 10).Select(i => (byte) i).ToArray());
Stream stream1 = new ReadLimitLengthStream(baseStream, 0, 5);
Stream stream2 = new ReadLimitLengthStream(baseStream, 5, 5);
Assert.AreEqual(0, stream1.Position);
```

tygill: Yes, though I think that applies almost any time you wrap a stream, since position is an underlying piece of state that all wrappers will be fighting over.

Duplicati/Library/Utility/ReadLimitLengthStream.cs
```diff
@@ -4,24 +4,66 @@
...
+            {
+                throw new ArgumentOutOfRangeException(nameof(start));
+            }
+            if (length < 0 || innerStream.Length < start + length)
             {
-                throw new ArgumentOutOfRangeException(nameof(size));
+                throw new ArgumentOutOfRangeException(nameof(length));
             }
 
             m_innerStream = innerStream ?? throw new ArgumentNullException(nameof(innerStream));
```
warwickmm, line 37: I think since we reference `innerStream.Length` above, `innerStream` can no longer be `null` here.

Duplicati/Library/Utility/ReadLimitLengthStream.cs
```diff
@@ -50,8 +92,24 @@ public override int Read(byte[] buffer, int offset, int count)
         }
 
         public override long Seek(long offset, SeekOrigin origin)
-        {
-            return m_innerStream.Seek(offset, origin);
+        {
+            long innerPosition;
```
warwickmm, line 96: If `innerPosition` is no longer used, can we remove it?

Duplicati/Library/Utility/ReadLimitLengthStream.cs
tygill: To add to that, even if you have multiple ReadLengthLimitStreams active, so long as you aren't using them concurrently in different threads, they should give reasonable behavior (at least, they shouldn't read things outside their own regions) since they look at and clamp the underlying stream's position. Whoever is using it might need seek to the place they want to use it each time (since the position will keep jumping around), but besides that it should pretty much work.

warwickmm: Yes, that's true.  The position jumping around sounds a bit risky, and I'm not sure if seeking it will always work (e.g., if one thread sets the position while another is in the middle of a `CopyTo`).  However, like you said this risk just comes with wrapping streams in general.

Duplicati/Library/Utility/ReadLimitLengthStream.cs
tygill: Right - good catch. Fixed.

Duplicati/Library/Utility/ReadLimitLengthStream.cs
tygill: I'll update the description comment a bit as well to warn about this.

Duplicati/Library/Utility/ReadLimitLengthStream.cs
tygill: Done.

Duplicati/Library/Backend/OneDrive/MicrosoftGraphBackend.cs
```diff
@@ -796,6 +943,100 @@ private T ParseResponse<T>(HttpWebResponse response)
...
+                }
+                else if (retryAfter.Date.HasValue)
+                {
+                    delayUntil = retryAfter.Date.Value;
+                }
+
+                if (delayUntil.HasValue)
+                {
+                    // Set the retry timestamp to the UTC version of the timestamp.
+                    Interlocked.Exchange(ref this.retryAfter, delayUntil.Value.UtcTicks);
```
warwickmm, line 1003: Is it possible for `SetRetryAfter` to be called twice (from two different threads) before the next call to `WaitForRetryAfter`?  In that case, would one session be possibly waiting for the incorrect amount of time?

Duplicati/Library/Utility/ReadLimitLengthStream.cs
```diff
@@ -4,24 +4,73 @@
...
+            : this(innerStream, 0, length)
+        {
+        }
+
+        public ReadLimitLengthStream(Stream innerStream, long start, long length)
         {
-            if (size < 0)
+            if (innerStream == null)
+            {
+                throw new ArgumentNullException(nameof(innerStream))
```
warwickmm, line 33: Missing `;`?

Duplicati/Library/Backend/OneDrive/MicrosoftGraphBackend.cs
tygill: That is true. I've update this to make sure the newly set time is always later than the previous time - this way, in the worst case some thread may end up waiting slightly too long.

warwickmm:
Note that the Travis builds are being [canceled](https://travis-ci.com/github/duplicati/duplicati/jobs/492568705) due to an insufficient credits balance.

I think we should move to the [GitHub-based workflows](https://github.com/duplicati/duplicati/pull/4431).

Duplicati/Library/Backend/OneDrive/MicrosoftGraphBackend.cs
```diff
@@ -796,6 +943,119 @@ private T ParseResponse<T>(HttpWebResponse response)
...
+            }
+            else
+            {
+                // If the date given was in the past then don't wait at all
+                delay = TimeSpan.Zero;
+            }
+
+            if (delay > TimeSpan.Zero)
+            {
+                await Task.Delay(delay).ConfigureAwait(false);
```
warwickmm, line 1055: Would it help to have a logging statement here to help debug jobs that appear to be "stalled"?

warwickmm:
Thanks @tygill.  The code looks good to me.  However, does this fix the throttling for you?  I've been trying to run some basic tests (setting upload to max of 100 KiB/s), and when the PUT operations start, I see a sustained upload of over 10 MiB/s.  It's not until the progress bar reaches the "Waiting for upload" state where I see the upload slowed to around 100 KiB/s.

This is on Linux.

Duplicati/Library/Backend/OneDrive/MicrosoftGraphBackend.cs
tygill: Adding logging to these backends isn't a bad idea - that's something I didn't really look into when originally adding them. It might be worth adding some messages to more than just here though. I'll take a look (probably next week).

tygill:
It seems to throttle on Windows when I add a ThrottledStream to the backend tester. One difference is that on Windows these backends use HttpClient instead of the OAuthHelper most other backends use. Mono doesn't use that by default, but you can force it with `--use-http-client=true`.

I'm suspicious that the non-HttpClient version may not work with throttling (and I wonder if other backends are affected) because pieces are uploaded by writing the stream piece into the request stream and then calling GetResponse after - if it is writing the piece to upload to some buffer instead of starting the request, then this form of throttling won't really work.

The alternative throttling method I mentioned on the forum would be to set a throttling speed and use `--fragment-size` to set a smaller fragment size for large files. The individual uploads may not be throttled, but by uploading smaller pieces at full speed it might come closer to the desired throttling.

warwickmm:
Sorry @tygill, I was making a silly error while testing.  After setting `asynchronous-concurrent-upload-limit` to `1`, I can see the upload respecting the throttle value (not perfectly, but pretty close).  I was seeing a multiple of this previously.  

FYI, to simplify the workflow we're probably going to wait until after the next beta release before merging anything new.

Nice job!

tygill:
That's interesting that each upload is throttled separately. I wonder if ThrottledStream should have some global throttle control since that isn't very intuitive (if I set a max of 100kbps, I'd expect a single upload running to hit that speed but also 4 concurrent uploads to be throttled to that total). But that at least does enable the option.

I'm not in a hurry to get this merged, so that will maybe give me some time to look at adding some logging.

tygill:
I've added a few more changes I came across while playing with this change:

- The cached buffer used by the copy stream helper is now held in a WeakReference, so the GC can collect it if memory is low.
- BackendTester now supports throttling arguments (which let me confirm that throttling works for both the HttpClient and non HttpClient versions of the OneDrive backend on Windows).
- OAuthHttpClient now treats the case where `--http-operation-timeout` is not set as meaning to use an infinite timeout (which should be more consistent with the WebRequest code). When severely throttling, I hit problems with the default 100 second timeout it was using.

I might still try to add some logging though.

Duplicati/CommandLine/BackendTester/Program.cs
```diff
@@ -225,6 +225,28 @@ static bool Run(List<string> args, Dictionary<string, string> options, bool firs
...
+                    if (!(backend is IStreamingBackend) || disableStreaming)
+                    {
+                        Console.WriteLine("Warning: Throttling is only supported in this tool on streaming backends");
+                    }
+
+                    throttleUpload = long.Parse(throttleUploadString);
+                }
+
+                long throttleDownload = 0;
+                if (options.TryGetValue("throttle-upload", out string throttleDownloadString))
```
warwickmm, line 240: Should this be trying to access the value for `"throttle-download"` instead of `"throttle-upload"`?

warwickmm:
> The cached buffer used by the copy stream helper is now held in a WeakReference, so the GC can collect it if memory is low.

Is the benefit of this buffer cache noticeable?  The code isn't complicated, but caching can come with some risks, and sometimes it's not clear if the potential gains are realizable or worth the effort.

Duplicati/CommandLine/BackendTester/Program.cs
tygill: Yes, it should.

Duplicati/Library/Backend/OneDrive/MicrosoftGraphBackend.cs
tygill: Some logging has been added.

tygill:
I haven't measured the benefit of this buffer caching - I figured since this cache is localized to this one small set of methods (and since copying streams is a common occurrence with IStreamingBackend) that trying to reuse this buffer in the common case would be worthwhile. The main gain here would be reduced GC pressure, since we wouldn't be allocating new 64k arrays every time we call this. Though I can remove it if we think it's not worth doing / comes with higher risk.

warwickmm:
I'm always a little wary of introducing caching, especially if the benefits are unclear.  Would you mind if we extract the caching changes to a separate pull request so that we can consider them in isolation of the OneDrive changes?

tygill:
No problem. I've pulled those changes out (looks like there might still be some whitespace changes in the Utility file though). I'll start a separate pull request, and maybe try to run the backend tester with with a profiler with and without the cache to see what impact it makes.

warwickmm:
Thanks!

basrieter:
Are these changes present in  2.0.6.3_beta_2021-06-17? I am trying to restore from a OneDrive (Personal) back-end and keep running into `StatusCode: 429, ReasonPhrase: 'Too Many Requests'` errors. The Retry-After header should prevent this, right?

warwickmm:
This fix is currently only in version 2.0.6.100.

basrieter:
How stable is that build? Since it's a Canary build?

warwickmm:
While the canary builds often contain bug fixes, they can also introduce new problems. It's up to you to decide what risks are acceptable.  It's always a good idea to have multiple backup strategies.

basrieter:
I will wait for a beta then. I have more strategies, so it will be fine.

PR #4471, opened 2021-03-21
Use SeBackupPrivilege on Windows

sergethedev17:
On Windows platforms, try to enable `SeBackupPrivilege` before running backup operations.

In Windows platforms, the `SeBackupPrivilege` privilege allows a process to bypass file and directory object permissions for the purposes of backing up the system.

The privilege is available to processes that run under security context of accounts specified in the **Back up files and directories** group policy (GPO => Computer Configuration => Windows Settings => Security Settings => Local Policies => User Rights Assignment). By default, the **Administrators** and **Backup Operators** security groups are specified in this group policy.

To take advantage of this functionality, configure the Duplicati service to use its own dedicated user account, and add the user account to the **Backup Operators** security group. The process will be able to access all files on the system and will not have any other administrative permissions.

2021-03-21 22_07_18-Computer Management

2021-03-21 21_47_55-Duplicati Server exe_18300 Properties

2021-03-21 21_47_50-Duplicati Server exe_18300 Properties

### DIFF
--- a/Duplicati/Library/Utility/WinTools.cs
+++ b/Duplicati/Library/Utility/WinTools.cs
@@ -18,6 +18,8 @@
 // 
 #endregion
 using Duplicati.Library.Common;
+using Microsoft.Win32.SafeHandles;
+using System;
 
 namespace Duplicati.Library.Utility
 {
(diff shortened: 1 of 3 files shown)


warwickmm:
Thanks @sergethedev17.  Our resources are a bit thin at the moment so it may take some time before someone gets a chance to review and test this. Also, we are in the midst of a feature freeze while we work on the next beta release, so to simplify the workflow we probably won't be merging many changes before then.

I also don't have Windows, so I'll need some other people to help take a look at this.

Duplicati/Library/Utility/WinTools.cs
```diff
@@ -53,5 +55,110 @@ public static string GetWindowsGpgExePath()
...
+        /// 
+        /// To take advantage of this functionality, configure the Duplicati service to use its own dedicated user account,
+        /// and add this user account to the 'Backup Operators' security group. The process will be able to access ALL files on the system,
+        /// without any other admin permissions.
+        /// </remarks>
+        public static void WithBackupPrivileges(Action action)
+        {
+            if (Platform.IsClientWindows)
+            {
+                var addedPrivilege = AddPrivilege(Win32.SE_BACKUP_NAME);
```
warwickmm, line 86: Can `AddPrivilege` throw an exception?  If so, I'm not sure if we want to halt the operation, or log a warning and continue without privileges.  Currently, if we encounter a file that we cannot read, I think we skip it and add a log entry.

The latter option might have implications where the retention policy may delete a "privileged" backup in favor for a "non-privileged" one.

drwtsn32x:
This is a nice improvement. I can help test when I get a chance.

mnaiman:
I took quick look and it seams to me, that you are using SeBackupPrivilege for Backup and Restore.
As stated here https://docs.microsoft.com/en-us/windows/win32/secauthz/privilege-constants you have to use SeRestorePrivilege for Restore tasks imo.

To use it just for tasks is good idea, because it is pretty dangerous permission in terms of security.

Question is if it is necessary to set when VSS is used, because VSS already have SeBackupPrivilege SeRestorePrivilege enabled.

Duplicati/Library/Utility/WinTools.cs
sergethedev17: `AddPrivilege` should never throw.
Logging when privilege enablement occurs (success or failure) is a good idea. I think `warning` severity is too high - it is not a requirement to have this privilege. Ultimately, whether the privilege is available or not does not dictate if reading files will succeed/fail. The process may have full R/W permissions to the files even without this privilege.

sergethedev17:
@mnaiman is correct about requiring `SeRestorePrivilege` for restores. For restore, we need to check if we read and write; we may need both `SeBackupPrivilege` and `SeRestorePrivilege`.
I think we need the privileges regardless of VSS, but this needs to be confirmed with further tests.

sergethedev17:
This PR does not work as expected. Turns out that enabling `SeBackupPrivilege` and `SeRestorePrivilege` privileges is not enough. It is necessary to execute file IO commands with special flags that tell Windows OS to use the privileges to bypass permissions, which was not implemented.

warwickmm:
> We need to re-implement most methods in Duplicati.Library.Common.IO.SystemIOWindows struct to use correct Windows API calls, eg using Alphaleonis.Win32.Filesystem.File.OpenBackupRead.

I noticed [this comment](https://forum.duplicati.com/t/duplicati-as-a-service-more-complicate-than-it-need-be/12278/27) in the forums:

> I did try that early on, as it clearly seems to be the intended way to access files. However, when I tested it, I did not get the raw file data, but rather a packed copy of the file with metadata. This can/could be restored with the corresponding restore method, but I did not like the nonstandard format and worry that it will make it hard/impossible to restore on a new/different machine, so I decided not to use it.

Is this a concern?

sergethedev17:
The latest commit contains a new file IO implementation for Windows that uses the Alphaleonis library. AlphaFS uses `SeBackupPrivilege` and `SeRestorePrivilege` semantics.

I found good code for enabling these privileges on a per-thread basis @ https://stackoverflow.com/a/58533328/1110897. According to SO, the post is licensed [CC BY-SA 4.0](https://stackoverflow.com/help/licensing). I left a comment asking the poster to license his implementation LGPL for us; awaiting reply.

In the Duplicati/Server project, I moved functionality to configure the DataFolder into a [separate class](https://github.com/duplicati/duplicati/pull/4471/commits/0abb9f773c46987f181a4c45a16f2d60ba5d9d92#diff-3f88f7898850c80f9d4e1caccae5a0396c27f6512bb2f360d28dac641ad23fac). When Duplicati runs as privileged process, we may configure a **new** folder with some restrictive permissions. In future, we can also use this class to relocate the DataFolder to %ProgramData% (to avoid issues such as [Restoring Duplicati after a Windows system reinstallation](https://forum.duplicati.com/t/restoring-duplicati-after-a-windows-system-reinstallation/11595)).

This branch has potential to really break stuff on Windows, so we'll need to test extensively. I will write unit tests for regular file I/O, and probably a PowerShell script to test process privilege activation under different security contexts.

Any feedback on code commited to date is welcome.

mgaffigan:
It seems that I am late to the party, but I am the author from https://stackoverflow.com/questions/5510311/how-can-i-manipulate-token-privileges-in-net/58533328 .  Please feel free to use the code as you see fit - though it is derived from http://msdn.microsoft.com/en-us/magazine/cc163823.aspx

kenkendk:
Closing this as I have now merged support for using BackupRead for reading.
The implementation is using Vanara, which is a thin C# mapper for access to the native Windows API calls.

The way it works is that BackupRead opens the file in a backup-mode, where multiple streams for the file are returned.
Besides the main content stream, there are alternate data streams and a metadata stream.
Ideally, we should record all these streams and store them in the backup, then on restore, we can restore all streams and get a very high-quality replication of the original file.

For now, the implementation only handles using BackupRead, and only processes the main contents. The net effect is that you can read locked files without needing to create a VSS snapshot. You still need the SeBackupPrivilege permission for it to work, so you can now choose between a VSS snapshot or the more light-weight BackupRead.

schierlm:
Great, thanks for fixing. 

I hope it will not only work for locked files, but also for files/directories where the user running Duplicati has no Read permissions (e.g. when an admin is backing up user profiles of non-admins). It may also be useful to use both, i.e. use BackupRead on a VSS snapshot (to create consistent backups of multiple files not readable by the backup user).

When #6386 hits a public build, I will do some testing :)

kenkendk:
There is a public build now: https://forum.duplicati.com/t/release-2-1-0-122-canary-2025-07-07/20843/6

PR #4478, opened 2021-04-04
Net5 split/fixups

tsuckow:
The 4.8 builds don't work for me, complaining at runtime about a manifest mismatch for CoCoL that I can't for the life of me track down.

But with the changes below, I can run the .net 5 builds.


tsuckow:
@kenkendk 

Perhaps I should cherry pick the hash changes onto master?

warwickmm:
I don't think we've decided on a proper workflow yet for moving the .NET 5 stuff forward, so I think we should keep `master` clean of these changes.  Unless I'm misunderstanding, and you're suggesting rebasing this branch instead of adding commits to `master`?

tsuckow:
@warwickmm I meant specifically cherrypicking the last two commits (hashalgorithm changes) into a PR targeting master. Those changes should work fine on .net framework but would be one less thing that has to be kept up to date on any netstandard branch

warwickmm:
Thanks for clarifying.  I'll let @kenkendk and others offer their opinions on how best to ease the transition.

kenkendk:
@tsuckow Yes, those two should be safe, we can get them in.

I have not tested the .net or 4.8 versions yet, just made it compile.

I am currently changing all code to have `async` methods on the `IBackend` interface.
This is not difficult, but a lot of manual work and I still have some work left before it compiles again.
The upside is that this now supports efficient listing of large folders and allows cancellation throughout the backend code.

tsuckow:
You using Iasyncenumerable for the large folder stuff?

It should be usable on framework with some hacks.

PR #4483, opened 2021-04-11
Net5 split/more bugfixes

tsuckow:
Can now run a backup to another folder on Windows with the net5 trayicon build.

net4.8 is still defunct for me for reasons I cannot comprehend (CoCoL assembly manifest mismatch).


Duplicati/Server/WebServer/RESTMethods/BackupDefaults.cs
```diff
@@ -38,13 +38,14 @@ select Newtonsoft.Json.Linq.JObject.FromObject(n)
...
-                using(var s = new System.IO.StreamReader(System.Reflection.Assembly.GetExecutingAssembly().GetManifestResourceStream(System.Reflection.Assembly.GetExecutingAssembly().GetName().Name + ".newbackup.json")))
+                using(var s = new System.IO.StreamReader(System.Reflection.Assembly.GetExecutingAssembly().GetManifestResourceStream(typeof(Program), "newbackup.json")))
                     n = (Newtonsoft.Json.Linq.JObject)Newtonsoft.Json.JsonConvert.DeserializeObject(s.ReadToEnd());
 
                 MergeJsonObjects(o, n);
             }
-            catch
+            catch (Exception e)
             {
+                Console.Error.WriteLine(e.Message);
```
warwickmm, line 48: Should we log these messages instead of writing to the console?

tsuckow:
Probably, I'll have to find an example of getting a logger instance

Duplicati/Library/Modules/Builtin/SendMail.cs
```diff
@@ -231,42 +231,10 @@ protected override void SendMessage(string subject, string body)
             List<string> servers = null;
             if (string.IsNullOrEmpty(m_server))
             {
-                var dnslite = new DnsLib.DnsLite();
-                var dnslist = new List<string>();
+                var dnsclient = new LookupClient();
+                var records = dnsclient.Query(toMailDomain, QueryType.MX).Answers.MxRecords();
```
warwickmm, line 235: For some reason, I'm not having any problems with the old `DnsLib.DnsLite` on the `experimental-net5-split` branch on Linux.  I'm not against using `DnsClient` from NuGet especially if it cleans up the code a bit, but would you be able to clarify the issues you were having with `DnsLite`?

tsuckow: Based on my commit message (its been a while) the compiler was emitting a warning that certain functions dnslite used were not implemented on Linux/Mac.

warwickmm: Thanks @tsuckow.  I see it the compiler warning now:
```
duplicati/thirdparty/DnsLite/DnsLite.cs(76,30): warning PC001: UdpClient.Receive(ref IPEndPoint) isn't supported on Linux and macOS
```

Duplicati/Library/Modules/Builtin/SendMail.cs
```diff
@@ -231,42 +231,10 @@ protected override void SendMessage(string subject, string body)
...
-                var records = new List<MXRecord>();
-                foreach (var s in dnslist)
-                {
-                    var res = dnslite.getMXRecords(toMailDomain, s);
-                    if (res != null)
-                        records.AddRange(res.OfType<MXRecord>());
-                }
-
-                servers = records.OrderBy(record => record.preference).Select(x => "smtp://" + x.exchange).Distinct().ToList();
+                servers = records.OrderBy(record => record.Preference).Select(x => "smtp://" + x.Exchange).Distinct().ToList();
```
warwickmm, line 237: I don't have much experience using Duplicati's email options, but on the `master` branch I can get the email reports working simply by using the `send-mail-to` option with a gmail address.  On this branch, I encounter 5 warnings (unfortunately it seems our logging system is broken on this branch), and no email is received.  Are you able to reproduce this?

warwickmm: Looks like it works after setting `send-mail-url` to `smtp://alt4.gmail-smtp-in.l.google.com`.  Any idea why that is now required?

warwickmm: Found some logs (newest on top):

```
Jun 7, 2021 4:15 PM: Failed to send message: MailKit.Security.SslHandshakeException: An error occurred while attempting to establish an SSL or TLS connection. 
The SSL certificate presented by the server is not trusted by the system for one or more of the following reasons: 
1. The server is using a self-signed certificate which cannot be verified. 
2. The local system is missing a Root or Intermediate certificate needed to verify the server's certificate. 
3. The certificate presented by the server is expired or invalid. 
See https://github.com/jstedfast/MailKit/blob/master/FAQ.md#InvalidSslCertificate for possible solutions. --> System.Security.Authentication.AuthenticationException: The remote certificate was rejected by the provided RemoteCertificateValidationCallback.

Jun 7, 2021 4:15 PM: Whole SMTP communication: Connected to smtp://alt4.gmail-smtp-in.l.google.com.:25/?starttls=when-available S: 220 mx.google.com ESMTP x7si11131319ybj.79 - gsmtp C: EHLO [192.168.143.95] S: 250-mx.google.com at your service, [136.24.158.126] S: 250-SIZE 157286400 S: 250-8BITMIME S: 250-STARTTLS S: 250-ENHANCEDSTATUSCODES S: 250-PIPELINING S: 250-CHUNKING S: 250 SMTPUTF8 C: STARTTLS S: 220 2.0.0 Ready to start TLS
```

I get the same error when setting `send-mail-url` to `smtp://alt4.gmail-smtp-in.l.google.com:25`.  However, when running from the `master` branch and without setting `send-mail-url`, the email is successful (via `smtp://gmail-smtp-in.l.google.com:25/?starttls=when-available`).

tsuckow: Maybe relevant?
https://github.com/jstedfast/MailKit/issues/1207

warwickmm: Nice find.  It works now if I include
```
client.ServerCertificateValidationCallback = (s, c, h, e) => true;
```
This is [not recommended](https://github.com/jstedfast/MailKit/blob/master/FAQ.md#SslHandshakeException), but the other options don't seem very satisfactory either.

warwickmm: In the interest of moving the .NET 5 migration forward, I'll propose that for now we will simply accept any certificate using the following patch:
  ```
diff --git a/Duplicati/Library/Modules/Builtin/SendMail.cs b/Duplicati/Library/Modules/Builtin/SendMail.cs
index 29de81cb2..3970c762c 100644
--- a/Duplicati/Library/Modules/Builtin/SendMail.cs
+++ b/Duplicati/Library/Modules/Builtin/SendMail.cs
@@ -270,6 +270,7 @@ namespace Duplicati.Library.Modules.Builtin
                                 if (uri.Scheme.ToLowerInvariant() == "tls")
                                     uri = new System.Uri("smtp://" + uri.Host + ":" + (uri.Port <= 0 ? 587 : uri.Port) + "/?starttls=always");
 
+                                client.ServerCertificateValidationCallback = (s, c, h, e) => true;
                                 client.Connect(uri);
 
                                 if (!string.IsNullOrEmpty(m_username) && !string.IsNullOrEmpty(m_password))
  ```

I created issue #4552 to consider safer approaches for validating the certificates.

The rest of this pull request looks good to me.  Once we add the above, I think we can merge.

Duplicati/Library/Modules/Builtin/SendMail.cs
```diff
@@ -231,42 +231,10 @@ protected override void SendMessage(string subject, string body)
...
-                // If we have no DNS servers, try Google and OpenDNS
-                if (dnslist.Count == 0)
-                {
-                    // https://developers.google.com/speed/public-dns/
-                    dnslist.Add("8.8.8.8");
-                    dnslist.Add("8.8.4.4");
-
-                    //http://www.opendns.com/opendns-ip-addresses/
-                    dnslist.Add("208.67.222.222");
-                    dnslist.Add("208.67.220.220");
```
warwickmm, line 258: I was initially concerned about the removal of these DNS servers, but it looks like the `LookupClient` will fall back to the Google DNS servers if no local DNS servers are found.

https://dnsclient.michaco.net/docs/DnsClient.NameServer.html#DnsClient_NameServer_ResolveNameServers_System_Boolean_System_Boolean_

Duplicati/Library/Modules/Builtin/SendMail.cs
tsuckow: Do you want to push that change or do you need me to do it?

warwickmm: I tested this again, and it now seems to work without the hack.  Looks like it is now using `smtp://gmail-smtp-in.l.google.com` instead of `smtp://alt4.gmail-smtp-in.l.google.com`.  Not sure why that's the case, but I think we can proceed without the hack for now.

Duplicati/Server/WebServer/RESTMethods/BackupDefaults.cs
```diff
@@ -38,13 +41,14 @@ select Newtonsoft.Json.Linq.JObject.FromObject(n)
             {
                 // Add built-in defaults
                 Newtonsoft.Json.Linq.JObject n;
-                using(var s = new System.IO.StreamReader(System.Reflection.Assembly.GetExecutingAssembly().GetManifestResourceStream(System.Reflection.Assembly.GetExecutingAssembly().GetName().Name + ".newbackup.json")))
+                using(var s = new System.IO.StreamReader(System.Reflection.Assembly.GetExecutingAssembly().GetManifestResourceStream(typeof(Program), "newbackup.json")))
```
warwickmm, line 44: Nice, this partially addresses #4554.  Thanks.

PR #4492, opened 2021-04-30
[Storj DCS] Update and rename Tardigrade-Backend

TopperDEL:
Storj just recently renamed their Tardigrade Decentralised Cloud Storage to "Storj DCS". This PR updates the backend accordingly. In Detail, the following is changed:

- Rename all "Tardigrade"-strings to "Storj DCS"
- Update the list of available satellites (endpoints)
- Add an additional passphrase verification field
- Add some input-field-validation
- Update the underlying uplink.NET to the latest version

A huge question from my side is to be discussed: I left all Tardigrade-C#-Object-names the same and also the protocol key. I assume especially the latter would break existing configurations, right? Can someone suggest how to proceed? Shall the backend itself stay "tardigrade"? Is there a way to support both protocols (the old "tardigrade" and the new one tbd)? Should I at least rename all objects and projects from tardigrade to storj?
Any suggestions are highly appreciated. :)


warwickmm:
Thanks @TopperDEL.  Are the old endpoint URLs still valid?  We are about to release our next beta version any day now.

TopperDEL:
The old endpoints are still valid, yes. They get routed to the new one. For new backup-settings, though, it is good to have the new one in place.
You may put this in the next beta or we may wait - this is up to you. From my point there is no need to wait, though. Everything should still work.

warwickmm:
Thanks.  Since we've finished most of our testing for this upcoming release, we will wait before integrating these changes.

I was able to create a free Storj DCS account.  I then created and ran a backup using the old Tardigrade code, then upgraded to this branch, and was able to continue running backups with no issues.  :+1: 

At the very least, it would be nice if all user-facing objects were renamed.  We have an overload of the `CommandLineArgument` constructor that marks an option as deprecated (see pull request #4443 for example).  This would allow us to support both `tardigrade-auth-method` and `storj-auth-method` for some time until the former is removed.

https://github.com/duplicati/duplicati/blob/67aaab15e3a8638806a50cf23b5a085c826a338a/Duplicati/Library/Interface/CommandLineArgument.cs#L258-L263

If it's also possible to rename all internal objects without affecting existing user configurations, that would be a nice bonus for consistency and avoid any confusion for future developers.

warwickmm:
On second thought, I'd also be fine with keeping the internal names as tardigrade, if that simplifies things.  We can save the internal refactoring for a future task.

TopperDEL:
I've now renamed all objects and parameters but led the protocol-key unaffected. The old parameters are marked as "deprecated" now. My tests with old configurations worked as expected, but it might be nice to have a second test from you @warwickmm.

If everything is good from your point, too, than this change could be merged whenever you're ready. Thank you!

Duplicati/Library/Backend/Storj/StorjBackend.cs
```diff
@@ -84,52 +93,52 @@ public Tardigrade(string url, Dictionary<string, string> options)
         {
             InitStorjLibrary();
 
-            var auth_method = options[TARDIGRADE_AUTH_METHOD];
+            var auth_method = options[STORJ_AUTH_METHOD] ?? options[TARDIGRADE_AUTH_METHOD];
```
warwickmm, line 96: I took an existing Tardigrade backup configuration, updated to this branch, and then attempted to run a backup.  This resulted in the following exception:

image

Should we use the `TryGetValue` pattern here instead?

Duplicati/Library/Backend/Storj/StorjBackend.cs
```diff
@@ -139,13 +148,23 @@ public IList<ICommandLineArgument> SupportedCommands
...
+                    new CommandLineArgument(TARDIGRADE_SATELLITE, CommandLineArgument.ArgumentType.String, Strings.Storj.StorjSatelliteDescriptionShort, Strings.Storj.StorjSatelliteDescriptionLong, "us1.storj.io:7777"),
+                    new CommandLineArgument(TARDIGRADE_API_KEY, CommandLineArgument.ArgumentType.String, Strings.Storj.StorjAPIKeyDescriptionShort, Strings.Storj.StorjAPIKeyDescriptionLong),
+                    new CommandLineArgument(TARDIGRADE_SECRET, CommandLineArgument.ArgumentType.Password, Strings.Storj.StorjSecretDescriptionShort, Strings.Storj.StorjSecretDescriptionLong),
+                    new CommandLineArgument(TARDIGRADE_SHARED_ACCESS, CommandLineArgument.ArgumentType.String, Strings.Storj.StorjSharedAccessDescriptionShort, Strings.Storj.StorjSharedAccessDescriptionLong),
+                    new CommandLineArgument(TARDIGRADE_BUCKET, CommandLineArgument.ArgumentType.String, Strings.Storj.StorjBucketDescriptionShort, Strings.Storj.StorjBucketDescriptionLong),
+                    new CommandLineArgument(TARDIGRADE_FOLDER, CommandLineArgument.ArgumentType.String, Strings.Storj.StorjFolderDescriptionShort, Strings.Storj.StorjFolderDescriptionLong),
```
warwickmm, line 167: Should these be the `STORJ_*` versions?  I'm seeing duplicate advanced options:

image

Duplicati/Server/webroot/ngax/scripts/services/EditUriBuiltins.js
```diff
@@ -487,22 +487,42 @@ backupApp.service('EditUriBuiltins', function (AppService, AppUtils, SystemInfo,
...
+		
+        if (options['--storj-auth-method'])
+            scope.storj_auth_method = options['--storj-auth-method'];
+        if (options['--storj-satellite'])
+            scope.storj_satellite = options['--storj-satellite'];
+        if (options['--storj-api-key'])
+            scope.storj_api_key = options['--storj-api-key'];
+        if (options['--storj-secret'])
+            scope.storj_secret = options['--storj-secret'];
+		if (options['--storj-secret-verify'])
```
warwickmm, line 515: Can we fix the indentation here (spaces vs. tabs)?

Duplicati/Server/webroot/ngax/scripts/services/EditUriBuiltins.js
```diff
@@ -749,9 +769,19 @@ backupApp.service('EditUriBuiltins', function (AppService, AppUtils, SystemInfo,
             'tardigrade-satellite': scope.tardigrade_satellite,
             'tardigrade-api-key': scope.tardigrade_api_key,
             'tardigrade-secret': scope.tardigrade_secret,
+			'tardigrade-secret-verify': scope.tardigrade_secret_verify,
             'tardigrade-shared-access': scope.tardigrade_shared_access,
 			'tardigrade-bucket': scope.tardigrade_bucket,
-			'tardigrade-folder': scope.tardigrade_folder
+			'tardigrade-folder': scope.tardigrade_folder,
```
warwickmm, line 775: Can we fix the indentation here (spaces vs. tabs)?

Duplicati/Server/webroot/ngax/scripts/services/EditUriBuiltins.js
```diff
@@ -749,9 +769,19 @@ backupApp.service('EditUriBuiltins', function (AppService, AppUtils, SystemInfo,
...
+			
+			'storj-auth-method': scope.storj_auth_method,
+            'storj-satellite': scope.storj_satellite,
+            'storj-api-key': scope.storj_api_key,
+            'storj-secret': scope.storj_secret,
+			'storj-secret-verify': scope.storj_secret_verify,
+            'storj-shared-access': scope.storj_shared_access,
+			'storj-bucket': scope.storj_bucket,
+			'storj-folder': scope.storj_folder
```
warwickmm, line 784: Can we fix the indentation here (spaces vs. tabs) and above?

Duplicati/Library/Backend/Storj/Strings.cs
```diff
@@ -0,0 +1,33 @@
...
+        public static string StorjSecretVerifyDescriptionLong { get { return LC.L(@"The encryption passphrase verification to make sure you provided the expected value."); } }
+        public static string StorjSharedAccessDescriptionShort { get { return LC.L(@"The access grant"); } }
+        public static string StorjSharedAccessDescriptionLong { get { return LC.L(@"An access grant contains all information in one encrypted string. You may use it instead of a satellite, API key and secret."); } }
+        public static string StorjBucketDescriptionShort { get { return LC.L(@"The bucket"); } }
+        public static string StorjBucketDescriptionLong { get { return LC.L(@"The bucket where the backup will reside in."); } }
+        public static string StorjFolderDescriptionShort { get { return LC.L(@"The folder"); } }
+        public static string StorjFolderDescriptionLong { get { return LC.L(@"The folder within the bucket where the backup will reside in."); } }
+        public static string StorjEncryptionPassphrasesDoNotMatchError { get { return LC.L(@"The encryption passphrases do not match"); } }
```
warwickmm, line 31: The following don't appear to be referenced anywhere:
```
StorjSecretVerifyDescriptionShort
StorjSecretVerifyDescriptionLong
StorjEncryptionPassphrasesDoNotMatchError
```

Did we forget to use these somewhere, or can they be removed?

Duplicati/Library/Backend/Storj/StorjBackend.cs
TopperDEL: I switched to "TryGetValue". May you verify this again? Looks good from my side - but I also said that on the previous try... :)

Duplicati/Server/webroot/ngax/scripts/services/EditUriBuiltins.js
TopperDEL: Fixed

Duplicati/Library/Backend/Storj/StorjBackend.cs
TopperDEL: Nice finding - fixed :)

Duplicati/Server/webroot/ngax/scripts/services/EditUriBuiltins.js
TopperDEL: Fixed

Duplicati/Library/Backend/Storj/Strings.cs
TopperDEL: I removed them - was a misunderstanding from my side. :)

TopperDEL:
@warwickmm Thank you very much for your in-depth review! I fixed all issues, so you may have a look again. Please come back to me if there are any issues left.
Thank you!

Duplicati/Library/Backend/Storj/StorjBackend.cs
```diff
@@ -84,52 +93,66 @@ public Tardigrade(string url, Dictionary<string, string> options)
...
             {
                 _bucket = "duplicati";
             }
 
-            if (options.ContainsKey(TARDIGRADE_FOLDER))
+            if (options.ContainsKey(TARDIGRADE_FOLDER) || options.ContainsKey(STORJ_FOLDER))
             {
-                _folder = options[TARDIGRADE_FOLDER];
+                options.TryGetValue(STORJ_FOLDER, out _folder);
+                if (string.IsNullOrEmpty(_bucket))
```
warwickmm, line 148: Should `_bucket` be `_folder`?

TopperDEL: Indeed, thank you!

warwickmm:
Sorry @TopperDEL, I'm now starting to question my decision to rename everything.  When "upgrading" a Tardigrade backup to this branch, I encounter a bunch of warnings about deprecated options and don't see a trivial way to rename the options in the destination URL.  Moreover, since the protocol is still `Tardigrade`, the command-line help still requires one to refer to it as `Tardigrade` instead of `Storj`, which can be confusing.
```
$ mono Duplicati/GUI/Duplicati.GUI.TrayIcon/bin/Debug/Duplicati.CommandLine.exe help Storj
Topic not found: Storj
```

@ts678, @drwtsn32x, @kenkendk, do you have any preferences here?  While Tardigrade cloud storage has been renamed to Storj DCS, would it be confusing if Duplicati continued to refer to it as Tardigrade?  Or, is there a simple way to update existing configurations to use the new option names, etc.?

warwickmm:
@TopperDEL, I also noticed that the backend files are not visible when I log in to my Storj account and go to "Objects" and click on the bucket name.  Should I expect to see the Duplicati files there?  The backups are working fine, so this is perhaps more just a usability question.

TopperDEL:
This might be due to the reason that your Browser does not have the encryption key. So you see the bucket, but you don't see any content.
It's not a bug, it's a feature. :)

If you are into it you might check the bucket content using uplink CLI, import your key there and list the bucket content. But if duplicati is working, the files are there. :)

ts678:
[Introducing Storj DCS—Decentralized Cloud Storage for Developers](https://www.storj.io/blog/introducing-storj-dcs-decentralized-cloud-storage-for-developers)

[Storj Launches Storj DCS, The Developer’s Gateway to the Decentralized Internet](https://www.businesswire.com/news/home/20210420005963/en/Storj-Launches-Storj-DCS-The-Developer%E2%80%99s-Gateway-to-the-Decentralized-Internet)

I guess the April 20 articles announce this. Storj is the company. STORJ is the token for exchange of value, DCS is the product.
This leaves room for additional products from Storj using STORJ payments. As a side issue, should name be StorjDCS instead?

I'd note that Tencent Cloud Object Storage (also S3 compatible) got named `cos` in Duplicati, so there it used a product name 
possibly because Tencent is a very large multinational holding company, and there are probably lots of products underneath...

I know Duplicati has command aliases, e.g. `list` and `find` are the same, and help will take either one and talk about `find`.
Similarly, `systeminfo` and `system-info` both work, which makes me think that the aliases are controlled by CommandMap:

https://github.com/duplicati/duplicati/blob/7b4f475762928ae72575cb96279364f0784de4e4/Duplicati/CommandLine/Program.cs#L60-L60

It looks like command line arguments can also have aliases, though I don't know/see examples. Might just bad bad searching.

I don't know how configuration editing might work. It's all in the Duplicati-server.sqlite database, but there are also CLI users.
The most user-friendly option seems to be to accept either the old or new names, but I'm not sure protocol is set up for that.

TopperDEL:
Hm, I have no good suggestion here. I do not know Duplicati-internals very good. If there would be one single point where a user-configuration is handled, we could "simply" convert every "tardigrade"-string to "storj" so that it is internally handled like the new protocoll/config/parameters. I would then completely switch over to "storj" or "storj-dcs" and get rid of tardigrade completely.
But I do not have enough knowledge to decide on this.

warwickmm:
As @ts678 recognized, users who use Duplicati via the command-line would also have to reconfigure all their scripts, etc.

Given that the protocol `tardigrade` appears to be fixed, I think the two best options we have are:
1. Keep things named Tardigrade, but update the documentation to mention that the product has been renamed to Storj DCS.  I don't know if this would be frowned upon by Storj, but these are side effects of rebranding.  This could still be a bit confusing to users too.
2. Introduce a new Storj DCS backend and deprecate the Tardigrade backend. Ideally, we would do so by refactoring the code in a way to avoid simply copy/pasting and duplicating everything.

Thoughts?  Have there been previous "rebrandings" of other backends?  If so, how did we treat those?

If @kenkendk has any opinions, that would be much appreciated.

ts678:
[Renaming Tardigrade to Storj](https://forum.rclone.org/t/renaming-tardigrade-to-storj/24308) is the equivalent rclone dev discussion. There might be other projects around having similar discussions.

I haven't gone searching, but I noticed the rclone topic while looking to see how their Jottacloud was, because Duplicati's is [failing](https://forum.duplicati.com/t/errors-after-backup-issues-on-jottacloud/12422) in a strange way for at least one person. So far I'm not seeing a rash of complaints among the few third-party efforts to access Jottacloud.

Support of contributed backends (including this one) is a bit challenging for day-to-day Duplicati helpers who might not have them...

TopperDEL:
Sorry, I have been busy in the last weeks. I would tend to create a new backend "Storj DCS" and leave Tardigrade as-is. Is there maybe a way to send a "warning" to the user that "Tardigrade" is deprecated? Switching over to the then new "Storj DCS" should be quite simple.
Still we would have to think about how exactly we re-use/refactor/centralise the coding for both. So we don't have to maintain both separately. I would highly recommend to make the old one obsolete, maybe show a warning and remove it at some point in  the future. The should not be that much users, yet, though I do not really know.

warwickmm:
I think the number of Tardigrade users is pretty small, so the impact will be limited.  Given that, creating a new "Storj DCS" backend and deprecating the Tardigrade backend is probably the cleanest solution.

I'm not sure what the best way is to consolidate the code.  One option might be to add a new "Storj DCS" backend, and modify the Taridgrade backend to contain an instance of the Storj backend and forward the member implementations as appropriate.  The JavaScript and HTML might be more difficult, but those changes are less common so perhaps duplication there is less of a concern.

Revision 8dae320a7a9622e56979df552630a00611edad6e contains an example of how we warned users about Amazon Cloud Drive being shut down.

TopperDEL:
So finally it should be done. I copied the Tardigrade-Backend over to the new Storj DCS and tried to reduce the duplicate code by internally use the new Storj-Backend from the old Tardigrade-Backend. I would be happy if someone would test this in order to see if I did any mistake - but from my point of view it looks good.

image

warwickmm:
Thanks @TopperDEL.  There appear to be some build errors.  Are you able to reproduce them?

TopperDEL:
Sorry @warwickmm , my bad. I did a quick change and let it go without another test. Let's see what the CI has to say, now. Builds locally here now.

Duplicati/CommandLine/BackendTester/Duplicati.CommandLine.BackendTester.csproj
```diff
@@ -86,8 +86,12 @@
       <Project>{32a74526-3e5f-413a-8cb4-1efdad4c8b91}</Project>
       <Name>Duplicati.Library.Backend.Sia</Name>
     </ProjectReference>
-    <ProjectReference Include="..\..\Library\Backend\Tardigrade\Duplicati.Library.Backend.Tardigrade.csproj">
+    <ProjectReference Include="..\..\Library\Backend\Storj\Duplicati.Library.Backend.Storj.csproj">
       <Project>{ae035e01-c917-4f13-a35e-78f21c1a2f17}</Project>
+      <Name>Duplicati.Library.Backend.Storj</Name>
+    </ProjectReference>
+    <ProjectReference Include="..\..\Library\Backend\Tardigrade\Duplicati.Library.Backend.Tardigrade.csproj">
+      <Project>{ae035e01-c917-4f13-f35e-78f21c1a2f17}</Project>
```
warwickmm, line 94: I'm surprised that the GUIDs for the Storj and Tardigrade projects differ by only one character.  Was this randomly generated?

Also, it appears that `Duplicati/Library/Backend/Tardigrade/Properties/AssemblyInfo.cs` contains Storj's GUID instead of Tardigrade's GUID?

Duplicati/Library/Backend/Tardigrade/Strings.cs
```diff
@@ -9,22 +9,7 @@ namespace Duplicati.Library.Backend.Strings
...
-        public static string TardigradeSharedAccessDescriptionShort { get { return LC.L(@"The access grant"); } }
-        public static string TardigradeSharedAccessDescriptionLong { get { return LC.L(@"An access grant contains all information in one encrypted string. You may use it instead of a satellite, API key and secret."); } }
-        public static string TardigradeBucketDescriptionShort { get { return LC.L(@"The bucket"); } }
-        public static string TardigradeBucketDescriptionLong { get { return LC.L(@"The bucket where the backup will reside in."); } }
-        public static string TardigradeFolderDescriptionShort { get { return LC.L(@"The folder"); } }
-        public static string TardigradeFolderDescriptionLong { get { return LC.L(@"The folder within the bucket where the backup will reside in."); } }
+        public static string DisplayName { get { return LC.L(@"Tardigrade Decentralised Cloud Storage"); } }
```
warwickmm, line 12: Can we append `(Deprecated)` to this to discourage users from creating a new Tardigrade backup configuration?

Duplicati/Library/Backend/Storj/StorjBackend.cs
```diff
@@ -0,0 +1,370 @@
...
+
+namespace Duplicati.Library.Backend.Storj
+{
+    public class Storj : IStreamingBackend
+    {
+        private const string STORJ_AUTH_METHOD = "storj-auth-method";
+        private const string STORJ_SATELLITE = "storj-satellite";
+        private const string STORJ_API_KEY = "storj-api-key";
+        private const string STORJ_SECRET = "storj-secret";
+        private const string STORJ_SECRET_VERIFY = "storj-secret-verify";
```
warwickmm, line 23: How is this (as well as `TARDIGRADE_SECRET_VERIFY`) used in the C# code?  Is it necessary?

Duplicati/Server/webroot/ngax/scripts/services/EditUriBuiltins.js
```diff
@@ -486,23 +513,48 @@ backupApp.service('EditUriBuiltins', function (AppService, AppUtils, SystemInfo,
             delete options[nukeopts[x]];
     }
 	
+	EditUriBackendConfig.parsers['storj'] = function (scope, module, server, port, path, options) {
+		if (options['--storj-auth-method'])
```
warwickmm, line 517: Can we fix the indentation here?

Duplicati/Server/webroot/ngax/scripts/services/EditUriBuiltins.js
```diff
@@ -486,23 +513,48 @@ backupApp.service('EditUriBuiltins', function (AppService, AppUtils, SystemInfo,
...
+            scope.storj_folder = options['--storj-folder'];
+
+        var nukeopts = ['--storj-auth-method','--storj-satellite', '--storj-api-key', '--storj-secret', '--storj-secret-verify', '--storj-shared-access', '--storj-bucket', '--storj-folder'];
+        for (var x in nukeopts)
+            delete options[nukeopts[x]];
+    };
+	
 	EditUriBackendConfig.parsers['tardigrade'] = function (scope, module, server, port, path, options) {
-        if (options['--tardigrade-auth-method'])
+		if (options['--tardigrade-auth-method'])
```
warwickmm, line 540: Can we fix the indentation here?

Duplicati/Server/webroot/ngax/scripts/services/EditUriBuiltins.js
```diff
@@ -486,23 +513,48 @@ backupApp.service('EditUriBuiltins', function (AppService, AppUtils, SystemInfo,
...
             scope.tardigrade_auth_method = options['--tardigrade-auth-method'];
         if (options['--tardigrade-satellite'])
             scope.tardigrade_satellite = options['--tardigrade-satellite'];
         if (options['--tardigrade-api-key'])
             scope.tardigrade_api_key = options['--tardigrade-api-key'];
         if (options['--tardigrade-secret'])
             scope.tardigrade_secret = options['--tardigrade-secret'];
+		if (options['--tardigrade-secret-verify'])
```
warwickmm, line 548: Can we fix the indentation here?

Duplicati/Server/webroot/ngax/scripts/services/SystemInfo.js
```diff
@@ -71,7 +71,8 @@ backupApp.service('SystemInfo', function($rootScope, $timeout, $cookies, AppServ
                 'mssp': null,
                 'dropbox': null,
                 'sia': null,
-                'tardigrade': null,
+                'storj': null,
+				'tardigrade': null,
```
warwickmm, line 75: Can we fix the indentation here?

warwickmm:
I ran several tests and everything is working nicely so far.  Thanks @TopperDEL.

Duplicati/Library/Main/Controller.cs
```diff
@@ -816,8 +816,12 @@ private void ValidateOptions()
             if (string.Equals(m_options.CompressionModule, "7z", StringComparison.OrdinalIgnoreCase))
                 Logging.Log.WriteWarningMessage(LOGTAG, "7zModuleHasIssues", null, "The 7z compression module has known issues and should only be used for experimental purposes");
 
-			//TODO: Based on the action, see if all options are relevant
-		}
+            //Inform the user about the deprecated Tardigrade-Backend. They should switch to Storj DCS instead.
+            if (string.Equals(new Library.Utility.Uri(m_backend).Scheme, "tardigrade", StringComparison.OrdinalIgnoreCase))
+                Logging.Log.WriteWarningMessage(LOGTAG, "TardigradeRename", null, "The Tardigrade-backend got renamed to Storj DCS - please migrate your backups to the new configuration.");
```
warwickmm, line 821: We can probably guide the user a bit by telling them to `... please migrate your backups to the new configuration by changing the destination storage type to Storj DCS.`

Duplicati/CommandLine/BackendTester/Duplicati.CommandLine.BackendTester.csproj
TopperDEL: I copied the whole folder over instead of creating a new one and just changed one character, yes. I hope this is not a problem?
I did not see the GUID in the AssemblyInfo, though. I've changed it accordingly.

Duplicati/Library/Backend/Storj/StorjBackend.cs
TopperDEL: It is not used in code, it is used in the UI. But I got a warning about that parameter not being allowed on that backend so I added it.
I think I have to remove it somewhere so that it is not passed to the backend. But I'm not quite sure if it will be available in the ui-verificator then. Can you guide me a bit here?
It is used in EditUriBuiltins.js for the parser, the builders and the validators. Technically it is only necessary in the validator. Could I simply remove it in the other two occurencies? The parameter should then not be transfered to the backend and I could remove it there then.

Duplicati/Server/webroot/ngax/scripts/services/EditUriBuiltins.js
TopperDEL: Fixed it. I don't understand why Notepad++ does this. I just copied the whole block and the first two lines have wrong indentation then.

Duplicati/Server/webroot/ngax/scripts/services/EditUriBuiltins.js
TopperDEL: Fixed it

Duplicati/Server/webroot/ngax/scripts/services/EditUriBuiltins.js
TopperDEL: Fixed

Duplicati/Server/webroot/ngax/scripts/services/SystemInfo.js
TopperDEL: Fixed

Duplicati/Library/Main/Controller.cs
TopperDEL: Good idea, I've extended the info.

Duplicati/Library/Backend/Tardigrade/Strings.cs
TopperDEL: Not sure what you mean here exactly. Where should that Deprectaed be? On the whole class? And wouldn't it be `[Obsolete]` then?

Duplicati/CommandLine/BackendTester/Duplicati.CommandLine.BackendTester.csproj
ts678: > just changed one character, yes. I hope this is not a problem?

I'm not at all familiar with GUID usage here, but if this was the change to ae035e01-c917-4f13-f35e-78f21c1a2f17 from ae035e01-c917-4f13-a35e-78f21c1a2f17, that encodes an invalid variant value, and I'm not sure what that could break.

[UUID Decoder](https://www.uuidtools.com/decode)
[A Universally Unique IDentifier (UUID) URN Namespace](https://www.ietf.org/rfc/rfc4122.txt) `4.1.1.  Variant` gives interpretation of variable-bit-length field.
[Universally unique identifier](https://en.wikipedia.org/wiki/Universally_unique_identifier#Variants) Variants chose to number variants, but they call the `10` MSB variant 1 while others call it 2...

Looking at the current files, I see the new GUID has gotten around to other places, but I don't know process to correct it.
That's more a developer tool skill, so I'll leave it to developers to sort out. I just happen to know a bit about GUID format.

warwickmm: Thanks @ts678.  I'd prefer that we do not manage the GUIDs manually.  @TopperDEL, can you fix the project references in your IDE?

Duplicati/Library/Backend/Tardigrade/Strings.cs
warwickmm: Sorry, what I meant was to add ` (Deprecated)` to the end of the string returned by `DisplayName`
```
Tardigrade Decentralised Cloud Storage (Deprecated)
```

This way, users can easily see that Tardigrade is deprecated when creating a new configuration, as opposed to only finding out when they encounter a warning after a backup.

Duplicati/Library/Backend/Storj/StorjBackend.cs
warwickmm: It looks like the password verification is a new feature.  Was there a reason why it was added?  I don't think any of the other backends have a verification field, and the "Test Connection" button should give sufficient feedback about the credentials.

Duplicati/CommandLine/BackendTester/Duplicati.CommandLine.BackendTester.csproj
TopperDEL: I've regenerated the GUID, so it is clearly unique.

Duplicati/Library/Backend/Tardigrade/Strings.cs
TopperDEL: Done

Duplicati/Library/Backend/Storj/StorjBackend.cs
TopperDEL: The password verification here is new, yes, and it is really necessary. The password here is not used to access Storj but to encrypt the files before uploading. If you were to make a typo here, it would still work - but you would not be able to successfully decrypt your data afterwards. So I've added a check-field to verify that the password is exactly what the user wanted.
I just don't really need that secret-verify-field in the backend, as the check is already done in the frontend.

warwickmm: Thanks for the explanation, I understand the purpose now.  I ran some tests by removing the 4 lines in the C# code, and I encountered the warnings you described.  If I then remove the following lines (which are involved in building the destination URI), I no longer get the warnings.  Can you try this and confirm if you see the same?

https://github.com/duplicati/duplicati/blob/a2a7e4b7ba1963192d636c53dc01cf6d969db416/Duplicati/Server/webroot/ngax/scripts/services/EditUriBuiltins.js#L804

https://github.com/duplicati/duplicati/blob/a2a7e4b7ba1963192d636c53dc01cf6d969db416/Duplicati/Server/webroot/ngax/scripts/services/EditUriBuiltins.js#L826

TopperDEL: Yes, removing these lines looks good - the check is still working in the frontend but we don't get the value in the backend. So I could remove the secret-verify-parameter in the backend completely.

TopperDEL:
All fixed from my side. :) @warwickmm

warwickmm:
Thanks @TopperDEL.

Issue #4493, opened 2021-05-03
Make compression hints case-insensitive


warwickmm:
- [x] I have searched open and closed issues for duplicates.
- [x] I have searched the [forum](https://forum.duplicati.com) for related topics.

----------------------------------------

## Environment info

- **Duplicati version**: 2.0.6.1
- **Operating system**: All
- **Backend**: All

## Description

The [default_compressed_extensions.txt](https://github.com/duplicati/duplicati/blob/beaf03562fdcf4425e962085bdf7175d6a465f49/Duplicati/Library/Main/default_compressed_extensions.txt) file contains extensions for files that should not be compressed.  However, these are treated in a case-sensitive manner.  While filenames in general should be treated this way for case-sensitive filesystems, it seems unlikely for the extension to be case-sensitive when used to indicate the file type.

We should consider making the compression hints use a case-insensitive match when determining whether to compress a file or not.  This can be done by providing `StringComparer.OrdinalIgnoreCase` as the `IEqualityComparer` to the `Dictionary` of hints here:

https://github.com/duplicati/duplicati/blob/beaf03562fdcf4425e962085bdf7175d6a465f49/Duplicati/Library/Main/Options.cs#L1842

duplicatibot:
This issue has been mentioned on **Duplicati**. There might be relevant details there:

https://forum.duplicati.com/t/default-compressed-extensions-txt-on-linux-system-is-case-sensitive/12084/10

PR #4502, opened 2021-05-08
Compare file extensions case insensitively.

samuel-w:
Fixes https://github.com/duplicati/duplicati/issues/4493

### DIFF
--- a/Duplicati/Library/Main/Options.cs
+++ b/Duplicati/Library/Main/Options.cs
@@ -1839,7 +1839,7 @@ namespace Duplicati.Library.Main
             {
                 if (m_compressionHints == null)
                 {
-                    var hints = new Dictionary<string, CompressionHint>(Library.Utility.Utility.ClientFilenameStringComparer);
+                    var hints = new Dictionary<string, CompressionHint>(StringComparer.OrdinalIgnoreCase);
 
                     string file;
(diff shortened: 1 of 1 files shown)


kenkendk:
Can you add a comment in the source code, such that it is obvious why this is not using the filesystem setting?
Otherwise, I expect that others will try to "fix" it by reverting later.

Duplicati/Library/Main/Options.cs
```diff
@@ -1839,7 +1839,7 @@ public int ConcurrencyCompressors
             {
                 if (m_compressionHints == null)
                 {
-                    var hints = new Dictionary<string, CompressionHint>(Library.Utility.Utility.ClientFilenameStringComparer);
+                    var hints = new Dictionary<string, CompressionHint>(StringComparer.OrdinalIgnoreCase); // Since file extensions are case insensitive
```
warwickmm, line 1842: Thanks @samuel-w.  I think we can try and improve this a little bit by providing some more detail.  Perhaps something like this?
```
// Even if the filesystem is case-sensitive, we will ignore the case when examining the extension for
// compression purposes.  The casing of the extension rarely matters as a file type hint (jpg vs. JPG).
```

samuel-w: ```Ignore file system case sensitivity, since file extensions case rarely indicates type```

PR #4506, opened 2021-05-11
Avoid ECDSA algorithms when using SFTP with Mono

warwickmm:
SSH.NET [relies on](https://github.com/sshnet/SSH.NET/blob/a5bd08d655bb6a3c762306472cec354556dca3a3/src/Renci.SshNet/Security/Cryptography/EcdsaKey.cs#L357) the `System.Security.Cryptography.ECDsaCng` class for ECDSA algorithms, which is [not implemented](https://github.com/mono/mono/blob/a01309d1041ff8ae15b1f8b717420ec139e1cdb8/mcs/class/referencesource/System.Core/System/Security/Cryptography/ECDsaCng.cs#L29-L32) in Mono (as of 6.12.0.144).  This prevents clients from connecting if one of the ECDSA algorithms is chosen as the host key algorithm.  In the event that this causes a connection failure, we will prevent the client from advertising support for ECDSA algorithms and make another connection attempt.

### DIFF
--- a/Duplicati/Library/Backend/SSHv2/SSHv2Backend.cs
+++ b/Duplicati/Library/Backend/SSHv2/SSHv2Backend.cs
@@ -357,6 +357,10 @@ namespace Duplicati.Library.Backend
                         client.ConnectionInfo.HostKeyAlgorithms.Remove(key);
                     }
                 }
+                catch
+                {
+                    // Ignore other exceptions and assume that we have ECDSA support.
+                }
             }
 
             client.Connect();
--- a/Duplicati/Library/Backend/SSHv2/SSHv2Backend.cs
+++ b/Duplicati/Library/Backend/SSHv2/SSHv2Backend.cs
@@ -59,7 +59,31 @@ namespace Duplicati.Library.Backend
 
         private SftpClient m_con;
 
-        private readonly Lazy<bool> supportsECDSA = new Lazy<bool>(SSHv2.SupportsECDSA, true);
+        private static readonly bool supportsECDSA;
+
+        static SSHv2()
+        {
+            // SSH.NET relies on the System.Security.Cryptography.ECDsaCng class for
+            // ECDSA algorithms, which is not implemented in Mono (as of 6.12.0.144).
+            // This prevents clients from connecting if one of the ECDSA algorithms is
+            // chosen as the host key algorithm.  In this case, we will prevent the
+            // client from advertising support for ECDSA algorithms.
+            //
+            // See https://github.com/mono/mono/blob/mono-6.12.0.144/mcs/class/referencesource/System.Core/System/Security/Cryptography/ECDsaCng.cs.
+            try
+            {
+                ECDsaCng unused = new ECDsaCng();
+                SSHv2.supportsECDSA = true;
+            }
+            catch (NotImplementedException)
+            {
+                SSHv2.supportsECDSA = false;
+            }
+            catch
+            {
+                SSHv2.supportsECDSA = true;
+            }
+        }
 
         public SSHv2()
         {
@@ -336,35 +360,9 @@ namespace Duplicati.Library.Backend
             m_con = con;
         }
 
-        /// <summary>
-        /// SSH.NET relies on the System.Security.Cryptography.ECDsaCng class for
-        /// ECDSA algorithms, which is not implemented in Mono (as of 6.12.0.144).
-        /// This prevents clients from connecting if one of the ECDSA algorithms is
-        /// chosen as the host key algorithm.  In this case, we will prevent the
-        /// client from advertising support for ECDSA algorithms.
-        /// </summary>
-        /// <seealso href="https://github.com/mono/mono/blob/mono-6.12.0.144/mcs/class/referencesource/System.Core/System/Security/Cryptography/ECDsaCng.cs">Mono ECDsaCng implementation.</seealso>
-        /// <returns>Whether ECDSA algorithms are supported or not.</returns>
-        private static bool SupportsECDSA()
-        {
-            try
-            {
-                ECDsaCng unused = new ECDsaCng();
-                return true;
-            }
-            catch (NotImplementedException)
-            {
-                return false;
-            }
-            catch
-            {
-                return true;
-            }
-        }
-
         private void TryConnect(SftpClient client)
         {
-            if (!this.supportsECDSA.Value)
+            if (!SSHv2.supportsECDSA)
             {
                 List<string> ecdsaKeys = client.ConnectionInfo.HostKeyAlgorithms.Keys.Where(x => x.StartsWith("ecdsa")).ToList();
                 foreach (string key in ecdsaKeys)
--- a/Duplicati/Library/Backend/SSHv2/SSHv2Backend.cs
+++ b/Duplicati/Library/Backend/SSHv2/SSHv2Backend.cs
@@ -351,7 +351,7 @@ namespace Duplicati.Library.Backend
                 // See https://github.com/mono/mono/blob/mono-6.12.0.144/mcs/class/referencesource/System.Core/System/Security/Cryptography/ECDsaCng.cs
                 if (Utility.Utility.IsMono)
                 {
-                    IEnumerable<string> ecdsaKeys = client.ConnectionInfo.HostKeyAlgorithms.Keys.Where(x => x.StartsWith("ecdsa"));
+                    List<string> ecdsaKeys = client.ConnectionInfo.HostKeyAlgorithms.Keys.Where(x => x.StartsWith("ecdsa")).ToList();
                     foreach (string key in ecdsaKeys)
                     {
                         client.ConnectionInfo.HostKeyAlgorithms.Remove(key);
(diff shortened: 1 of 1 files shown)


Duplicati/Library/Backend/SSHv2/SSHv2Backend.cs
```diff
@@ -328,11 +328,44 @@ private void CreateConnection()
...
+                // This prevents clients from connecting if one of the ECDSA algorithms is
+                // chosen as the host key algorithm.  In the event that this causes a
+                // connection failure, we will prevent the client from advertising support
+                // for ECDSA algorithms and make another connection attempt.
+                //
+                // See https://github.com/mono/mono/blob/mono-6.12.0.144/mcs/class/referencesource/System.Core/System/Security/Cryptography/ECDsaCng.cs
+                if (Utility.Utility.IsMono)
+                {
+                    IEnumerable<string> ecdsaKeys = client.ConnectionInfo.HostKeyAlgorithms.Keys.Where(x => x.StartsWith("ecdsa"));
+                    foreach (string key in ecdsaKeys)
```
drwtsn32x, line 355: I had to change `ecdsaKeys` to `ecdsaKeys.ToList()` otherwise this exception was getting thrown:

`System.InvalidOperationException: Collection was modified; enumeration operation may not execute.`

Not sure if adding the `.ToList()` is the best way to solve this but it works.

drwtsn32x (changes requested):
This solution works in my testing.  The code successfully removes all the key algorithms that start with `ecdsa` if the initial connection fails.  Subsequent reattempt works and RSA key was used.

Duplicati/Library/Backend/SSHv2/SSHv2Backend.cs
warwickmm: Thanks for testing and noticing this.  That makes sense, we need to enumerate the items before modifying the collection.  I fixed this in revision b80845a17652fe333145d90bc2509a732d02a517.  Would you mind testing again, just to be safe?

drwtsn32x: Just tested...works great!

kenkendk:
Looks good to me, thanks for figuring out the problem.

Right now we rely on the connection to fail, but this happens _after_ we actually connect.
I worry that there might be cases where the connection fails for another reason, and we downgrade the algorithm without needing to  (should throw another exception though).

A test with Mono and .Net5 works for me with:

```csharp
try
{
    ECDsa.Create(default(ECCurve));
}
catch(NotImplementedException)
{
    // Mono
    Console.WriteLine("Missing implementation of ECDsa");
}
catch
{
    // Net5
    Console.WriteLine("Failed to create algorithm");
}
```

I can add this test to the code if this is preferred?

warwickmm:
Thanks @kenkendk.  I like your suggestion of checking for the `ECDsa` implementation explicitly, as opposed to my indirect approach.

warwickmm:
Let me know if you are able to make the changes, or if you'd like me to.

warwickmm:
Revision 3d8817608e81bbb06e6fc58ac68323d7769e6043 attempts to make the test for ECDSA support more explicit.

@drwtsn32x, would you mind running another test please?

drwtsn32x:
Yes, I'll see if I can test tomorrow.

drwtsn32x:
@warwickmm - latest revision works great in my testing.

warwickmm:
Thanks!

kenkendk:
In my tests, the code:

```csharp
ECDsa.Create(default(ECCurve));
```

Throws an exception if `ECDsa` is supported, because `default(ECCurve)` is not a valid curve description.
In the code as it stands, this is not checked, so the backend will fail _if_ `ECDsa` is supported, but only running on Mono.

I think there should simply be a `catch {}` afterwards to silently ignore the exception. Also, since it is a system-static property it would be nice if this information was cached. This can then be done without checking for Mono, because it will work as expected on Windows.

The property looks something like:
```csharp
private bool? m_supportsECD;

private bool SupportsECD
{
    get
    {
        if (m_supportsECD.HasValue)
            return m_supportsECD.Value;

        try
        {
            ECDsa.Create(default(ECCurve));
            return m_supportsECD = true;
        }
        catch(NotImplementedException)
        {
            return m_supportsECD = false;
        }
        catch
        {
            return m_supportsECD = true;
        }       
    }
}
```

Then the `TryConnect` method would be:
```csharp
private void TryConnect(SftpClient client)
{ 
    if (!SupportsECD)
    {
        foreach (var key in client.ConnectionInfo.HostKeyAlgorithms.Keys.Where(x => x.StartsWith("ecdsa")).ToList())
            client.ConnectionInfo.HostKeyAlgorithms.Remove(key);
    }

    client.Connect();
}
```

warwickmm:
Thanks @kenkendk.

>In my tests, the code:
>
>ECDsa.Create(default(ECCurve));
>
>Throws an exception if ECDsa is supported, because default(ECCurve) is not a valid curve description.
>In the code as it stands, this is not checked, so the backend will fail if ECDsa is supported, but only running on Mono.

I modified the code to try `new ECDsaCng()` as I believe that more closely reflects what SSH.NET does.  Let me know if you think there are issues with this.  Can someone test this on a system (Windows) that has ECDSA support?

> I think there should simply be a catch {} afterwards to silently ignore the exception.

Done.

> Also, since it is a system-static property it would be nice if this information was cached.

I initially (22564bb2a332e6455afd0ea66780044ffcbf035b) opted to use `Lazy<bool>` initialization.  However, I decided that it might be simpler to use a static constructor (c246976fbbd937164e2ef60a2c61e762b6fe075b).  Let me know what your preferences are.

drwtsn32x:
> Can someone test this on a system (Windows) that has ECDSA support?

I tested the prior commits with Windows 10 talking to a Linux backend SFTP server that had an ECDSA cert. Is that what you mean?  (Also tested same back end with a Linux/mono front end of course.) I can test the latest commits as well.

warwickmm:
Sorry, I should clarify.  The new implementation utilizes a new check for ECDSA support, and doesn't have special code paths for Mono, so we should test that Windows is still allowed to use an ECDSA host key.

kenkendk:
I tried `new ECDsaCng()` with .Net 5 and it throws a different exception, but the effect is clearly the same:

```
System.PlatformNotSupportedException: Windows Cryptography Next Generation (CNG) is not supported on this platform.
```

So if the code `new ECDsaCng()` works on Windows (it appears so) and this is what SSH.Net does, I think your test is much better. But then I would suggest that any exception would set `SSHv2.supportsECDSA = false;` as it will break SSH.Net.

It then becomes:
```csharp
try
{
   ECDsaCng unused = new ECDsaCng();
   SSHv2.supportsECDSA = true;
}
catch
{
   SSHv2.supportsECDSA = false;
}
```

I think static constructors are exactly for this single-initialization purpose, but they tend to make debugging harder if something goes wrong (failed to load type, missing methods, exceptions, etc). But this _should_ be safe to do.

warwickmm:
I simplified the exception handling in revision a5fbee033536d54a937d7dbc7ff935efcbe10b7c.  I believe in .NET 5 the responsibility of implementing the cryptographic algorithms is pushed to the [operating system](https://docs.microsoft.com/en-us/dotnet/standard/security/cross-platform-cryptography), so it's possible that we will encounter more of these types of issues in the future.

drwtsn32x:
Tested latest commit... works as expected for me.  On Linux/mono I get an RSA key with SFTP back end (that has both ECDSA and RSA keys).  On Windows I get the ECDSA key.

warwickmm:
Thanks @drwtsn32x.

kenkendk:
I think the PR is good as-is, and we can merge it now to fix the issues in the current beta.

Taking at step back, it is really a problem in SSH.Net. Calling any of the crypto CNG methods directly will limit the library to work on Windows. I can see from the [source code](https://github.com/sshnet/SSH.NET/blob/a5bd08d655bb6a3c762306472cec354556dca3a3/src/Renci.SshNet/Security/Cryptography/EcdsaKey.cs#L281) that the `.netstandard2.0` version of the library deals with this and does not call the CNG version directly. 

I _think_ this is done because the default implementation in `.net4.5` would use a managed (slow) version, where `.net5` will default to wrap the native version (meaning CNG for Windows, OpenSSL for Linux).

An alternative solution could then be to force the `.netstandard2.0` SSH.Net library instead of the `.net45` version. This would allow the Mono implementation to use the ECD algorithms. It _could_ however have more implications that will not show up before wider deployment. The current PR will work such that things that work now (Windows) will continue to work, and Mono will start to work (without ECD).

warwickmm:
I agree, we should limit the scope of the changes to avoid more unanticipated issues.

Note that the destination branch of this pull request is `master`, which contains other changes we don't want to release in a beta yet.  We should probably create a new release branch from v2.0.6.1 to integrate this as well as the changes from pull request #4513.

ts678:
> We should probably create a new release branch from v2.0.6.1

I thought a release branch was standard procedure. I see a May 3 [https://github.com/duplicati/duplicati/tree/release/2.0.6.1_beta](https://github.com/duplicati/duplicati/tree/release/2.0.6.1_beta).

warwickmm:
Sorry, what I meant was that the branch should be based on `v2.0.6.1-2.0.6.1_beta_2021-05-03` and not `master`.  Or, whatever is easiest for @kenkendk to integrate just the desired changes (I tend to prefer merge commits rather than cherry-picking).

kenkendk:
I will make a new experimental release then.  
The new experimental `v2.0.6.2` will be based on `v2.0.6.1-2.0.6.1_beta_2021-05-03` with #4506 and #4513 cherry picked in.

PR #4509, opened 2021-05-15
On Azure backend, allow use of SAS token

sergethedev17:
1. Fixes a bug in AzureBlobBackend constructor.
2. Adds a feature to allow users to authenticate with Azure Storage using a Shared access signature (SAS) token. SAS tokens are more secure because they:
 * have a start and expiry date/time
 * can be restricted to specified IP addresses
 * can limit access scope to a single Storage blob container

SAS tokens can still be cancelled/revoked by changing/rotating the storage account Access Key.
This feature adds a new parameter named `azure-access-sas-token`, which can be used instead of `azure-access-key`. When both parameters are specified, the `azure-access-sas-token` parameter takes precedence.

More info about SAS tokens @ [Grant limited access to Azure Storage resources using shared access signatures (SAS)](https://docs.microsoft.com/en-us/azure/storage/common/storage-sas-overview).

Screenshot from Azure showing how to generate a SAS token:
2021-05-15 21_29_36-Window

### DIFF
--- a/Duplicati/Library/Backend/AzureBlob/AzureBlobBackend.cs
+++ b/Duplicati/Library/Backend/AzureBlob/AzureBlobBackend.cs
@@ -53,10 +53,17 @@ namespace Duplicati.Library.Backend.AzureBlob
                 storageAccountName = options["auth-username"];
             if (options.ContainsKey("auth-password"))
                 accessKey = options["auth-password"];
-            if (options.ContainsKey("azure_account_name") || options.ContainsKey("azure-account-name"))
-                storageAccountName = options["azure_account_name"] ?? options["azure-account-name"];
-            if (options.ContainsKey("azure_access_key") || options.ContainsKey("azure-access-key"))
-                accessKey = options["azure_access_key"] ?? options["azure-access-key"];
+
+            if (options.ContainsKey("azure_account_name"))
+                storageAccountName = options["azure_account_name"];
+            else if (options.ContainsKey("azure-account-name"))
+                storageAccountName = options["azure-account-name"];
+
+            if (options.ContainsKey("azure_access_key"))
+                accessKey = options["azure_access_key"];
+            else if (options.ContainsKey("azure-access-key"))
+                accessKey = options["azure-access-key"];
+
             if (!string.IsNullOrEmpty(uri.Username))
                 storageAccountName = uri.Username;
             if (!string.IsNullOrEmpty(uri.Password))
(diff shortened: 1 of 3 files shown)


sergethedev17:
Don't know why macos tests failed. I think there may be an issue with the unit test.

warwickmm:
> 1. Fixes a bug in AzureBlobBackend constructor.

Thanks for noticing this :+1: .  I created a separate pull request (see #4513) to fix the handling of the deprecated options to make it easier to quickly create an updated beta release if necessary.

warwickmm:
> Don't know why macos tests failed. I think there may be an issue with the unit test.

Yes, there's an issue with the tests (or perhaps actual code) on macos that we haven't figured out yet.

sergethedev17:
> 
> 
> > 1. Fixes a bug in AzureBlobBackend constructor.
> 
> Thanks for noticing this 👍 . I created a separate pull request (see #4513) to fix the handling of the deprecated options to make it easier to quickly create an updated beta release if necessary.

Thanks @warwickmm . I didn't know about those aws deprecated options. Also, your logic (`if`-`if`, instead of `if`-`else if`) is probably better.
What do you want to do with this PR?

sergethedev17:
Question about testing connectivity.
In the `Duplicati.Library.Backend.AzureBlob.AzureBlobBackend` class, the `Test()` method simply enumerates all objects. There are 2 issues with this:

1. The backend may contain 100s (or 1000s?) of files. Enumerating all files is not necessary and can take a lot of time. Some cloud-based remotes charge users per API request.
2. The implementation does not test write functionality. With SAS tokens in particular, it is possible that enumeration is allowed, but writing or deleting blobs is not. (There was a similar issue with FTP; https://github.com/duplicati/duplicati/pull/2500).

Can I have a suggestion to improve the `Test()` method? Perhaps, enumerate first 10 files, create a test file (random name), write 1KB, read the file, delete the file.

warwickmm:
> What do you want to do with this PR?

After pull request #4513 is merged, we can then merge the upstream branch into this one and resolve any conflicts.

> Can I have a suggestion to improve the Test() method? Perhaps, enumerate first 10 files, create a test file (random name), write 1KB, read the file, delete the file.

I think that's a good idea.  However, we should make such changes in a separate pull request since they are unrelated to this one.  The [AlternativeFTPBackend](https://github.com/duplicati/duplicati/blob/0a43410535d89e3e8ca93e78f799b9cec8812310/Duplicati/Library/Backend/AlternativeFTP/AlternativeFTPBackend.cs#L460) and [Tardigrade](https://github.com/duplicati/duplicati/blob/0a43410535d89e3e8ca93e78f799b9cec8812310/Duplicati/Library/Backend/Tardigrade/TardigradeBackend.cs#L297) backends have tests that do more than a simple list if you want to see existing examples.

kenkendk:
I merged the commandline option fixes. Once the tests passes for this, I think it is good to merge.
(And yes, the `Test()` issue should go in another PR)

warwickmm:
This looks good to me.  Thanks @sergethedev17 for this, and for noticing the bug in the `AzureBlobBackend` constructor.

PR #4511, opened 2021-05-15
Fix ReSharper's "Use string interpolation expression" / Code Smell

marodev:
Hello,

This pull request is _very simple_ and aims to reduce the technical debt. It specifically targets the"Use string interpolation expression" [rule](https://www.jetbrains.com/help/resharper/UseStringInterpolation.html) reported by ReSharper.

### DIFF
--- a/Duplicati/Library/Backend/Rclone/Rclone.cs
+++ b/Duplicati/Library/Backend/Rclone/Rclone.cs
@@ -249,7 +249,7 @@ namespace Duplicati.Library.Backend
         {
             try
             {
-                RcloneCommandExecuter(rclone_executable, $"copyto {local_repo}:{filename} {remote_repo}:{Path.Combine(this.remote_path, remotename)}", CancellationToken.None).Await();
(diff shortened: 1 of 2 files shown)


Duplicati/Library/Backend/Rclone/Rclone.cs
```diff
@@ -249,7 +249,7 @@ public void Get(string remotename, string filename)
         {
             try
             {
-                RcloneCommandExecuter(rclone_executable, String.Format("copyto {2}:{3} {0}:{1}", local_repo, filename, remote_repo, Path.Combine(this.remote_path, remotename)), CancellationToken.None).Await();
+                RcloneCommandExecuter(rclone_executable, $"copyto {local_repo}:{filename} {remote_repo}:{Path.Combine(this.remote_path, remotename)}", CancellationToken.None).Await();
```
warwickmm, line 252: Are these arguments in the wrong order?

marodev: Nice catch! Fixed.

PR #4513, opened 2021-05-15
Fix bug in handling deprecated Azure and S3 options

warwickmm:
In pull request #4443, we renamed some of the Azure and S3 options to use dashes `-` instead of underscores `_` for consistency.  However, our treatment of the options in the backend constructors still required that the underscore versions be present even if the dashed version was provided.

Now, we allow for either option to be present, but the value for the dashed version will take precedence if both are provided.

Thanks to @sergethedev17 for noticing this.


warwickmm:
I based this branch on v2.0.5.114 to make it easier to merge into different destination branches.

PR #4548, opened 2021-05-31
Check contents of written test file

sergethedev17:
Before performing test write, reset stream Position to 0, otherwise test performs zero-byte write.
When performing test read, verify that file was written correctly.

### DIFF
--- a/Duplicati/Library/Backend/AlternativeFTP/AlternativeFTPBackend.cs
+++ b/Duplicati/Library/Backend/AlternativeFTP/AlternativeFTPBackend.cs
@@ -446,15 +446,6 @@ namespace Duplicati.Library.Backend.AlternativeFTP
             get { return new string[] { new Uri(_url).Host }; }
         }
 
-        private static Stream StringToStream(string str)
-        {
-            var stream = new MemoryStream();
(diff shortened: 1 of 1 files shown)


Duplicati/Library/Backend/AlternativeFTP/AlternativeFTPBackend.cs
```diff
@@ -451,6 +451,7 @@ private static Stream StringToStream(string str)
             var stream = new MemoryStream();
             var writer = new StreamWriter(stream) { AutoFlush = true };
             writer.Write(str);
+            stream.Position = 0;
             return stream;
```
warwickmm, line 455: It seems strange to me that the caller is responsible for disposing the returned `Stream`, and that the `StreamWriter` is not disposed.  What are your thoughts on moving these statements to the one usage below and then removing this method?

sergethedev17: Good catch. I removed `StreamWriter` altogether.

warwickmm: Thanks @sergethedev17.  Looks good to me.

PR #4549, opened 2021-05-31
Feature/aureblob backend test

sergethedev17:


Duplicati/Library/Backend/AzureBlob/AzureBlobBackend.cs
```diff
@@ -195,11 +201,85 @@ public string Description
...
+            {
+                for (int i = 0; i < TEST_FILE_ENUM_COUNT && entriesEnumerator.MoveNext(); i++)
+                {
+                    var fileEntry = entriesEnumerator.Current;
+                }
+            }
+            finally
+            {
+                entriesEnumerator.Dispose();
+            }
```
warwickmm, line 221: Is there a reason to prefer `try-finally` over a simpler `using` statement?
```
using (var entriesEnumerator = _azureBlob.ListContainerEntries().GetEnumerator())
{
    for (int i = 0; i < TEST_FILE_ENUM_COUNT && entriesEnumerator.MoveNext(); i++)
    {
        var fileEntry = entriesEnumerator.Current;
    }
}
```

Duplicati/Library/Backend/AzureBlob/AzureBlobBackend.cs
```diff
@@ -195,11 +201,85 @@ public string Description
...
+
+
+
```
warwickmm, line 224: Can we remove redundant blank lines here (and below as well)?  We prefer not to have more than one consecutive blank line.

Duplicati/Library/Backend/AzureBlob/AzureBlobBackend.cs
```diff
@@ -210,6 +290,16 @@ public void CreateFolder()
...
+        }
+
+        private static Stream StringToStream(string str)
+        {
+            var stream = new MemoryStream();
+            var writer = new StreamWriter(stream) { AutoFlush = true };
+            writer.Write(str);
+            stream.Position = 0;
+            return stream;
+        }
```
warwickmm, line 302: Similar to in pull request #4548, can we remove this method and move these statements to the single use case above?

warwickmm:
Thanks @sergethedev17.  Can you add a description to this pull request so that if we later revisit it, we can easily get a sense of its purpose without having to inspect the code?

Duplicati/Library/Backend/AzureBlob/AzureBlobBackend.cs
sergethedev17: Definitely. Just wanted to get feedback on #4548 . Thanks for the other feedback above.

Duplicati/Library/Backend/AzureBlob/AzureBlobBackend.cs
sergethedev17: No good reason here. I used the same approach as in `AzureBlobWrapper.ListContainerEntries()` by inertia.

Duplicati/Library/Backend/AzureBlob/AzureBlobWrapper.cs
```diff
@@ -112,46 +112,58 @@ public void DeleteObject(string keyName)
...
+                    throw new FolderMissingException(Strings.AzureBlobBackend.MissingContainerError(_containerName, ex.Message), ex);
+                }
+                catch
+                {
+                    itemsEnumerator.Dispose();
+                    throw;
+                }
+
+                yield return ParseBlobItem(itemsEnumerator.Current);
+            }
```
sergethedev17, line 141: This is the only way I could think of to return an `IEnumerable<>`. C# does not allow `yield return` from inside `try-catch` blocks.

warwickmm: I don't have the ability to test this, but does a `using` statement allow us to remove the calls to `Dispose` as well as the final `catch` block ,while still disposing of everything correctly (in particular, with how we enumerate the items in the `AzureBlobBackend.Test` method)?
```
using (IEnumerator<IListBlobItem> itemsEnumerator = listBlobItems.GetEnumerator())
{
    while (true)
    {   
        try 
        {   
            if (!itemsEnumerator.MoveNext())
                break;
        }   
        catch (StorageException ex) when (ex.RequestInformation.HttpStatusCode == 404)
        {   
            throw new FolderMissingException(Strings.AzureBlobBackend.MissingContainerError(_containerName, ex.Message), ex);
        }   

        yield return ParseBlobItem(itemsEnumerator.Current);
    }   
}
```

Duplicati/Library/Backend/AzureBlob/AzureBlobBackend.cs
```diff
@@ -195,11 +201,76 @@ public string Description
         public string[] DNSName
         {
             get { return _azureBlob.DnsNames; }
-        }
+        }
+
```
warwickmm, line 205: Extra blank line.

Duplicati/Library/Backend/AzureBlob/AzureBlobBackend.cs
```diff
@@ -195,11 +201,76 @@ public string Description
...
+                {
+                    Delete(TEST_FILE_NAME);
+                }
+                catch (Exception e)
+                {
+                    throw new Exception(string.Format(Strings.AzureBlobBackend.ErrorDeleteFile, e.Message), e);
+                }
+            }
+
+
```
warwickmm, line 234: Extra blank line.

kenkendk:
Closing this as the regular Test operation now does this

Issue #4526, opened 2021-05-27
Fix SQLite issues on Linux and MacOS


warwickmm:
There are some SQLite issues due to resolve.  The separate code path due to

https://github.com/duplicati/duplicati/blob/bd172d39c87704d1eb8815850add9ef49cdbcdbb/Duplicati/Library/SQLiteHelper/SQLiteLoader.cs#L168

results in errors like
```
Native library pre-loader failed to load native SQLite library "/home/ken/documents/code/duplicati/Executables/net5/Duplicati.GUI.TrayIcon/bin/Debug/net5.0/SQLite/win64/SQLite.Interop.dll" (getLastError = 203): System.DllNotFoundException: Unable to load shared library '__Internal' or one of its dependencies. In order to help diagnose loading problems, consider setting the LD_DEBUG environment variable: lib__Internal: cannot open shared object file: No such file or directory
   at System.Data.SQLite.UnsafeNativeMethodsPosix.dlopen(String fileName, Int32 mode)
   at System.Data.SQLite.NativeLibraryHelper.LoadLibraryPosix(String fileName)
   at System.Data.SQLite.NativeLibraryHelper.LoadLibrary(String fileName)
   at System.Data.SQLite.UnsafeNativeMethods.PreLoadSQLiteDll(String baseDirectory, String processorArchitecture, Boolean allowBaseDirectoryOnly, String& nativeModuleFileName, IntPtr& nativeModuleHandle)
```

We may be able to simply use the `System.Data.SQLite.Core` package from NuGet.  This already seems to done in https://github.com/tsuckow/duplicati/commit/a759995e7bb3472be524b76b3e0c9d5751a0d68b.

warwickmm:
This is related to issue #4525.

warwickmm:
With some minor changes (related to #4527 and #4526), I was able to get the tests to run:

https://github.com/warwickmm/duplicati/actions/runs/883834505

The macOS tests are failing for the known (weird) timestamp issue that is "fixed" in `master` (by allowing for a larger tolerance).  The UI tests are failing because I think no encryption modules are detected (and hence the "passphrase" element cannot be found).  When I run the tray icon project, I encounter similar issues:

![image](https://user-images.githubusercontent.com/17345532/119912105-9fb84380-bf0f-11eb-835b-8c39641467ef.png)

Sometimes refreshing the page helps, other times it doesn't.

ts678:
[...]

Issue #4524, opened 2021-05-27
Remove usage of __MonoCS__ symbol


warwickmm:
Usages of the `__MonoCS__` symbol were removed in pull request #4212.  There are some remaining instances of this symbol in the `netstandard` branch that need to be addressed.

PR #4550, opened 2021-06-02
Remove Utility.IsMono, related Mono-specific Code 

### LINKED ISSUE #4525: Replace usages of Utility.IsMono
We should remove usages of `Utility.IsMono`, as it will always return `false` in .NET 5.  The usages need to be studied to see if separate code paths are still necessary (most likely not, but there may be some usages that were proxies for os-specific code).

drehtisch:
I removed `Utility.IsMono`, `Utility.MonoVersion`, `Utility.MonoDisplayVersion` and everything that referenced those properties. 

Also removed a bool flag `USE_HTTP_CLIENT_DEFAULT` because of if it always being true after the removal of `Utility.IsMono`

Resolves #4525 

## Remarks

It might be a idea instead of removing `%MONO%` from help.txt, to replace it with `dotnet`. Which should work for all platforms.


warwickmm:
Thanks @drehtisch.  Does this compile for you?  There appears to be a usage of `USE_HTTP_CLIENT_DEFAULT` remaining here:

https://github.com/drehtisch/duplicati/blob/be658da277185d7e90bbfe24de3eb669d897cd7c/Duplicati/Library/Backend/OneDrive/MicrosoftGraphBackend.cs#L184

Also, there is JavaScript code that references mono.  Here's one example (there may be more):

https://github.com/drehtisch/duplicati/blob/be658da277185d7e90bbfe24de3eb669d897cd7c/Duplicati/Server/webroot/ngax/scripts/directives/backupEditUri.js#L147

Would you be able to remove these references as well?

drehtisch:
Thanks for your feedback @warwickmm. Sorry, I probably missed that. I will remove these references of course. And I will also look again for more mono references.

drehtisch:
While searching the solution for more mono specific code, some questions came up for me. I would like to hear your opinion / feedback on these. 

1. I suppose "__MonoCS__" should be removed in another pull request, because of it being object of another issue (#4524) or should i include that here ?
2. What do we do with the dependency `MonoMac`, see `GUI.TrayIcon.Program:TryGetMonoMac()`, remove it ?
3. Do we still need `Library.Utility.Strings.SslCertificateValidator.MonoHelpSSL`?. It might still be relevant under linux. If that is the case the message needs to be reworded and the property renamed.

warwickmm:
> 1. I suppose __MonoCS__ should be removed in another pull request, because of it being object of another issue (#4524) or should i include that here ?

I think it's fine to remove `__MonoCS__` in this pull request.

> 2. What do we do with the dependency MonoMac, see GUI.TrayIcon.Program:TryGetMonoMac(), remove it ?

Based on [this comment](https://github.com/duplicati/duplicati/pull/3124#issuecomment-436915441), it sounds like we can remove MonoMac.

> 3. Do we still need Library.Utility.Strings.SslCertificateValidator.MonoHelpSSL?. It might still be relevant under linux. If that is the case the message needs to be reworded and the property renamed.

I think it would be ok to remove.

Perhaps this pull request should just try and remove as much mono as we can, and then we can get  some other opinions on what parts need to stay, if any.

/cc @mnaiman, @tsuckow, @kenkendk

Duplicati/GUI/Duplicati.GUI.TrayIcon/Program.cs
```diff
@@ -319,14 +314,6 @@ private static TrayIconBase RunTrayIcon(string toolkit)
 
         [System.Runtime.CompilerServices.MethodImpl(System.Runtime.CompilerServices.MethodImplOptions.NoInlining)]
         private static TrayIconBase GetWinformsInstance() { return new Win32Runner(); }
-#if __MonoCS__ || __WindowsGTK__ || ENABLE_GTK
```
warwickmm, line 322: I think we only want to remove the `__MonoCS__` part of the `#if` condition.  The rest of this code should remain.

drehtisch: Alright. Will restore that then.

Duplicati/Library/SQLiteHelper/SQLiteLoader.cs
```diff
@@ -165,67 +165,34 @@ public static Type SQLiteConnectionType
...
 
+                // If we have a new path, try to force load the mixed-mode assembly for the current architecture
+                // This can be avoided if the preload in SQLite works, but it is easy to do it here as well
+                if (loadMixedModeAssembly)
+                {
+                    try { PInvoke.LoadLibraryEx(Path.Combine(assemblyPath, "SQLite.Interop.dll"), IntPtr.Zero, 0); }
+                    catch (Exception ex) { Logging.Log.WriteExplicitMessage(LOGTAG, "LoadMixedModeSQLiteError", ex, "Failed to load the mixed mode SQLite database: {0}", Path.Combine(assemblyPath, "SQLite.Interop.dll")); }
+                }
+
+
```
warwickmm, line 195: This will be cleaned up further when we address issue #4526.

warwickmm:
This looks good to me.  Thanks @drehtisch.

PR #4568, opened 2021-06-24
Update localization_fr.po

ccastello:
typography errors


warwickmm:
Thanks @ccastello.  Our translations are actually maintained in Transifex. If you make the changes there, we will generated the resulting translation files.

https://www.transifex.com/duplicati/duplicati/

PR #4569, opened 2021-06-24
Update localization-fr.po

ccastello:
Typography error


warwickmm:
Thanks @ccastello.  Our translations are actually maintained in Transifex. If you make the changes there, we will generated the resulting translation files.

https://www.transifex.com/duplicati/duplicati/

PR #4571, opened 2021-06-26
Add support for mega.nz two factor authentication

### LINKED ISSUE #3696: Mega.nz with 2FA not working

## Environment info

- **Duplicati version**: 2.0.4.2_experimental_2018-11-12
- **Operating system**: Win10

## Description

I want to backup a local folder to my mega.nz-account. Everything worked fine, until I activated 2FA for my mega.nz-account. 
After deactivating 2FA again, the backup works again. 

There also seems to be no way to configure Duplicati using a generated API-Key from [Mega.nz/sdk](https://mega.nz/sdk).

## Steps to reproduce

1. Activate 2FA in Mega.nz
2. Create Backup to Mega.nz data provider
3. Run Backup

- **Actual result**:
API error -26
- **Expected result**:
Working backup

## Log

> Fatal error
> CG.Web.MegaApiClient.ApiException: API response: -26
>    bei Duplicati.Library.Main.BackendManager.List()
>    bei Duplicati.Library.Main.Operation.FilelistProcessor.RemoteListAnalysis(BackendManager backend, Options options, LocalDatabase database, IBackendWriter log)
>    bei Duplicati.Library.Main.Operation.FilelistProcessor.VerifyRemoteList(BackendManager backend, Options options, LocalDatabase database, IBackendWriter log)
>    bei Duplicati.Library.Main.Operation.BackupHandler.PreBackupVerify(BackendManager backend)
>    bei Duplicati.Library.Main.Operation.BackupHandler.Run(String[] sources, IFilter filter)

vfrz:
Fixes #3696

How it works:
- I've added an option called `auth-two-factor-key` in which you set the TOTP generation key given by mega.nz
- I've added a reference to the NuGet package `Otp.NET` to generate a valid TOTP code using the `auth-two-factor-key` when we create the `MegaApiClient`

I have tested with my personal mega.nz account and it's working fine!

I believe the only downside of the PR is that we are adding a new external library and I don't know what your policy is on this. Maybe we should add it on a "higher level" csproj so other backends could use it too?

### DIFF
--- a/Duplicati/Library/Backend/Mega/Strings.cs
+++ b/Duplicati/Library/Backend/Mega/Strings.cs
@@ -6,8 +6,8 @@ namespace Duplicati.Library.Backend.Strings {
         public static string AuthPasswordDescriptionShort { get { return LC.L(@"Supplies the password used to connect to the server"); } }
         public static string AuthUsernameDescriptionLong { get { return LC.L(@"The username used to connect to the server. This may also be supplied as the environment variable ""AUTH_USERNAME""."); } }
         public static string AuthUsernameDescriptionShort { get { return LC.L(@"Supplies the username used to connect to the server"); } }
-        public static string AuthTwoFactorKeyDescriptionShort { get { return LC.L(@"Supplies the two-factor key used to connect to the server"); } }
-        public static string AuthTwoFactorKeyDescriptionLong { get { return LC.L(@"The two-factor authentication key used to connect to the server."); } }
+        public static string AuthTwoFactorKeyDescriptionShort { get { return LC.L(@"The shared secret used to generate two-factor TOTP codes."); } }
+        public static string AuthTwoFactorKeyDescriptionLong { get { return LC.L(@"For accounts with two-factor authentication enabled, this is the shared secret used to generate the two-factor TOTP codes."); } }
         public static string NoPasswordError { get { return LC.L(@"No password given"); } }
         public static string NoUsernameError { get { return LC.L(@"No username given"); } }
         public static string Description { get { return LC.L(@"This backend can read and write data to Mega.co.nz. Allowed formats are: ""mega://folder/subfolder"""); } }
--- a/Duplicati/Library/Backend/Mega/Strings.cs
+++ b/Duplicati/Library/Backend/Mega/Strings.cs
@@ -6,6 +6,8 @@ namespace Duplicati.Library.Backend.Strings {
         public static string AuthPasswordDescriptionShort { get { return LC.L(@"Supplies the password used to connect to the server"); } }
         public static string AuthUsernameDescriptionLong { get { return LC.L(@"The username used to connect to the server. This may also be supplied as the environment variable ""AUTH_USERNAME""."); } }
         public static string AuthUsernameDescriptionShort { get { return LC.L(@"Supplies the username used to connect to the server"); } }
+        public static string AuthTwoFactorKeyDescriptionShort { get { return LC.L(@"Supplies the two-factor key used to connect to the server"); } }
+        public static string AuthTwoFactorKeyDescriptionLong { get { return LC.L(@"The two-factor authentication key used to connect to the server."); } }
         public static string NoPasswordError { get { return LC.L(@"No password given"); } }
         public static string NoUsernameError { get { return LC.L(@"No username given"); } }
         public static string Description { get { return LC.L(@"This backend can read and write data to Mega.co.nz. Allowed formats are: ""mega://folder/subfolder"""); } }
--- a/Duplicati/Library/Backend/Mega/packages.config
+++ b/Duplicati/Library/Backend/Mega/packages.config
@@ -1,5 +1,6 @@
 ﻿<?xml version="1.0" encoding="utf-8"?>
 <packages>
-  <package id="MegaApiClient" version="1.7.1" targetFramework="net471" />
+  <package id="MegaApiClient" version="1.9.0" targetFramework="net471" />
   <package id="Newtonsoft.Json" version="12.0.2" targetFramework="net471" />
+  <package id="Otp.NET" version="1.2.2" targetFramework="net471" />
 </packages>
--- a/Duplicati/License/Duplicati.License.csproj
+++ b/Duplicati/License/Duplicati.License.csproj
@@ -72,12 +72,15 @@
     </Content>
     <Content Include="..\..\thirdparty\Otp.NET\Homepage.txt">
       <Link>licenses\Otp.NET\Homepage.txt</Link>
+      <CopyToOutputDirectory>PreserveNewest</CopyToOutputDirectory>
     </Content>
     <Content Include="..\..\thirdparty\Otp.NET\license.txt">
       <Link>licenses\Otp.NET\license.txt</Link>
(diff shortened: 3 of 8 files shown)


ts678:
Thanks! This will make some people happy. It reminds me of the #3696 oathtool idea, except portable, which is a very good thing.

mega.nz seems to have the idea of a session, but I'm not sure how long it lasts (any idea?). Storing a long-lived secret to generate TOTP code when needed is convenient, seems to be less secure than using a separate app, but might be the only path. Any ideas?

Whatever method is used should (I think) be able to work in unattended mode, e.g. when running Duplicati from a command line.

warwickmm:
Thanks @vfrz.  I ran some tests and in general it is working well.

I did notice a few things after I enabled 2FA on my Mega account:
1. If I don't provide the `auth-two-factor-key` option, I encounter a `TwoFactorAuthenticationError` error.  However, if I provide an incorrect secret key, the MegaApiClient issues a `RequestFailedPermanetly` error.  I don't know if there's anything we can do to improve this error message, but it would've been nice if the error hinted at an authentication issue.  Do you know if `RequestFailedPermanetly` is also encountered when 2FA is disabled, and the wrong username/password is provided?
2. I encountered a lot of `ResourceExpired` errors that required operations to be retried.  It sounds like this is an [existing issue](https://forum.duplicati.com/t/mega-nz-error-when-check-connection/5316) and Duplicati's retry mechanism was able to recover from these.  But it does make the Mega backend feel less reliable.

Duplicati/Library/Backend/Mega/Strings.cs
```diff
@@ -6,8 +6,10 @@ internal static class MegaBackend {
         public static string AuthPasswordDescriptionShort { get { return LC.L(@"Supplies the password used to connect to the server"); } }
         public static string AuthUsernameDescriptionLong { get { return LC.L(@"The username used to connect to the server. This may also be supplied as the environment variable ""AUTH_USERNAME""."); } }
         public static string AuthUsernameDescriptionShort { get { return LC.L(@"Supplies the username used to connect to the server"); } }
+        public static string AuthTwoFactorKeyDescriptionShort { get { return LC.L(@"Supplies the two-factor key used to connect to the server"); } }
+        public static string AuthTwoFactorKeyDescriptionLong { get { return LC.L(@"The two-factor authentication key used to connect to the server."); } }
```
warwickmm, line 10: I think "used to connect to the server" could be a little confusing to users that are new to 2FA and TOTP codes.  Perhaps we can improve this a little by saying something like

Short:
```
The shared secret used to generate two-factor TOTP codes.
```

Long:
```
For accounts with two-factor authentication enabled, this is the shared secret used to generate the two-factor TOTP codes.
```

Duplicati/Library/Backend/Mega/packages.config
```diff
@@ -1,5 +1,6 @@
 ﻿<?xml version="1.0" encoding="utf-8"?>
 <packages>
-  <package id="MegaApiClient" version="1.7.1" targetFramework="net471" />
+  <package id="MegaApiClient" version="1.9.0" targetFramework="net471" />
   <package id="Newtonsoft.Json" version="12.0.2" targetFramework="net471" />
-</packages>
+  <package id="Otp.NET" version="1.2.2" targetFramework="net471" />
```
warwickmm, line 5: Can you add the corresponding license information for Otp.NET to the `Duplicati.License` project?

Nullvaettr:
Is it now available?

Duplicati/Library/Backend/Mega/packages.config
vfrz: Done!

Duplicati/Library/Backend/Mega/Strings.cs
vfrz: I agree with you and I made the changes.

vfrz:
> Thanks @vfrz. I ran some tests and in general it is working well.
> 
> I did notice a few things after I enabled 2FA on my Mega account:
> 
> 1. If I don't provide the `auth-two-factor-key` option, I encounter a `TwoFactorAuthenticationError` error.  However, if I provide an incorrect secret key, the MegaApiClient issues a `RequestFailedPermanetly` error.  I don't know if there's anything we can do to improve this error message, but it would've been nice if the error hinted at an authentication issue.  Do you know if `RequestFailedPermanetly` is also encountered when 2FA is disabled, and the wrong username/password is provided?
>    EDIT: When 2FA is disabled, an incorrect username/password results in `ResourceNotExists`, which is not great either.  I think `RequestFailedPermanetly` is a slight improvement, so we can leave this as-is for now.
> 2. I encountered a lot of `ResourceExpired` errors that required operations to be retried.  It sounds like this is an [existing issue](https://forum.duplicati.com/t/mega-nz-error-when-check-connection/5316) and Duplicati's retry mechanism was able to recover from these.  But it does make the Mega backend feel less reliable.

1. I believe I can't do much about this since it's handled by the MegaApiClient library.
2. Since it's an already existing issue I am tempted to answer the same thing as for the first point and the problem is surely the MegaApiClient library. I don't have much time to investigate this but I could try to whenever I got free time (and maybe open an issue in the client library if needed?).

vfrz:
> Thanks! This will make some people happy. It reminds me of the #3696 oathtool idea, except portable, which is a very good thing.
> 
> mega.nz seems to have the idea of a session, but I'm not sure how long it lasts (any idea?). Storing a long-lived secret to generate TOTP code when needed is convenient, seems to be less secure than using a separate app, but might be the only path. Any ideas?
> 
> Whatever method is used should (I think) be able to work in unattended mode, e.g. when running Duplicati from a command line.

I sure understand your point and I partially agree but I wouldn't say it's less secure. In the worst case (duplicati) it just doesn't add more security (because the token is saved on your computer) but your Mega account is still more protected until someone has access to your computer's files and know where to look for the key.
I don't know about oathtool but it would require to store the secret too.

Duplicati/License/Duplicati.License.csproj
```diff
@@ -70,6 +70,15 @@
...
       <CopyToOutputDirectory>PreserveNewest</CopyToOutputDirectory>
     </Content>
+    <Content Include="..\..\thirdparty\Otp.NET\Homepage.txt">
+      <Link>licenses\Otp.NET\Homepage.txt</Link>
+    </Content>
+    <Content Include="..\..\thirdparty\Otp.NET\license.txt">
+      <Link>licenses\Otp.NET\license.txt</Link>
+    </Content>
+    <Content Include="..\..\thirdparty\Otp.NET\licensedata.json">
+      <Link>licenses\Otp.NET\licensedata.json</Link>
```
warwickmm, line 80: Each of these needs the `<CopyToOutputDirectory>PreserveNewest</CopyToOutputDirectory>` element.

vfrz: My bad, it's fixed!

warwickmm:
Thanks @vfrz, this looks good to me.

antivirtel:
Hi, When is the next beta scheduled to be released? I don't see a new Beta tag from last June, it was almost a year ago, this would be a really nice feature... Thanks

jonahdami:
Any updates regarding this release?

Issue #4223, opened 2020-06-11
System.PlatformNotSupportedException


DaPlague:
When running Duplicati with an argument in docker-compose

`CLI_ARGS= --webservice-sslcertificatefile=/config/certs/server.pfx`

the webserver won't start an give the following error:

```
A serious error occurred in Duplicati: System.PlatformNotSupportedException: Operation is not supported on this plat
form.
  at Duplicati.Server.Database.ServerSettings.get_ServerSSLCertificate () [0x00115] in <c5f097a49c0a4f1fb0f93cf3f5f218b1>:0
  at Duplicati.Server.WebServer.Server..ctor (System.Collections.Generic.IDictionary`2[TKey,TValue] options) [0x002a4] in <c5f097a49c0a4f1fb0f93cf3f5f218b1>:0
  at Duplicati.Server.Program.StartWebServer (System.Collections.Generic.Dictionary`2[TKey,TValue] commandlineOptions) [0x00000] in <c5f097a49c0a4f1fb0f93cf3f5f218b1>:0
```

in a previous version a did not got an error and i could use my ssl cert.

Environment:

Ubuntu server 18.04
Docker

FishDog5000:
This looks very similar to this issue: linuxserver/docker-duplicati#30 (where a workaround was found).

duplicatibot:
This issue has been mentioned on **Duplicati**. There might be relevant details there:

https://forum.duplicati.com/t/ssl-works-only-the-first-time/9843/3

duplicatibot:
This issue has been mentioned on **Duplicati**. There might be relevant details there:

https://forum.duplicati.com/t/ssl-works-only-the-first-time/9843/7

PR #4580, opened 2021-07-21
Fix server certificate linux

mnaiman:
fix for https://github.com/duplicati/duplicati/issues/4223 and discussion and root cause is here https://forum.duplicati.com/t/ssl-works-only-the-first-time/9843/19


mnaiman:
God know why there is so many commits :/

warwickmm:
> God know why there is so many commits :/

Looks like your `master` branch contains these extra merge commits.  Can you create a new branch based on the upstream `master` branch and cherry pick the relevant commit?

Issue #4593, opened 2021-08-16
Warnings when restoring duplicate files with no-local-db and no-local-blocks


warwickmm:
- [x] I have searched open and closed issues for duplicates.
- [x] I have searched the [forum](https://forum.duplicati.com) for related topics.

----------------------------------------

## Environment info

- **Duplicati version**: 2.0.6.100
- **Operating system**: Linux
- **Backend**: Local

## Description
Attempts to restore with `no-local-db` and `no-local-blocks` set to `"true"` result in warnings if two source files contain the same contents.

## Steps to reproduce

1. Create an empty source directory.
2. Add a file `file1` to the source directory.
3. Copy `file1` so that there are two source files with the same contents, `file1` and `file2`.
4. Run a backup.
5. Restore all files (`Controller.Restore(new[] { "*" })`).

- **Actual result**:
No errors and no warnings.

- **Expected result**:
```
2021-08-16 13:31:32 -07 - [Warning-Duplicati.Library.Main.Operation.RestoreHandler-PatchingFileLocalFailed]: Failed to patch file: "<path to test data>/restored/file1" with data from local file "<path to test data>/backup-data/file2", message: Unexpected result when marking block.
2021-08-16 13:31:32 -07 - [Warning-Duplicati.Library.Main.Operation.RestoreHandler-PatchingFileLocalFailed]: Failed to patch file: "<path to test data>/restored/file2" with data from local file "<path to test data>/backup-data/file2", message: Unexpected result when marking block.
```

PR #4594, opened 2021-08-16
Add tests for restoring without local data

warwickmm:
This adds some tests for restoring files without local data (local database, local blocks, patching with local blocks).  This revealed an issue when `no-local-db` and `no-local-blocks` are "true" and two source files contain the same contents (see issue #4593).  Once that issue is fixed, we can revert revision 75a6efddf09a7ce1f26f195e96c01690457e19f9 to test the additional code paths.


kenkendk:
Closing this in favor of #5752

PR #4601, opened 2021-09-03
Fix implementation of Options.PatchWithLocalBlocks

warwickmm:
Previously, `Options.PatchWithLocalBlocks` would return `true` as long as the key was present in the `Dictionary`, regardless of what the value parsed to.


warwickmm:
The tests appear to be failing due to an issue downloading the test data from https://testfiles.duplicati.com/data.zip.

PR #4603, opened 2021-09-09
Tencent COS: support setting storage class

moesoha:


Duplicati/Library/Backend/TencentCOS/COSBackend.cs
```diff
@@ -81,6 +82,11 @@ public class CosOptions
             /// A path or subfolder in a bucket
             /// </summary>
             public string Path { get; set; }
+            /// <summary>
+            /// Storage class of the object
+            /// https://cloud.tencent.com/document/product/436/33417
```
warwickmm, line 87: Thank you @moesoha.  Can we provide a link to an English-language page?  That would make this a little more accessible to our current team of developers.  The same link appears in the `Strings.cs` class as well.

moesoha: since Tencent Cloud provides two languages dividually, I have put both en/zh links on it. `Strings.cs` also updated.

warwickmm:
Thanks @moesoha!  :+1:

PR #4610, opened 2021-09-22
Add delays between backup pages in test

samuel-w:
Should I add delays anywhere else?
Hopefully fixes the UI tests randomly failing.

### DIFF
--- a/guiTests/guiTest.py
+++ b/guiTests/guiTest.py
@@ -107,36 +107,42 @@ driver.find_element_by_link_text("No, my machine has only a single account").cli
 driver.find_element_by_link_text("Add backup").click()
 
 # Choose the "add new" option
+time.sleep(0.5)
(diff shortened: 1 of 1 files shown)


warwickmm:
Thanks @samuel-w.  Doesn't hurt to try and see if this works.

warwickmm:
Looks like that [didn't work](https://github.com/duplicati/duplicati/runs/3677573046?check_suite_focus=true).  I'm not sure what the issue is exactly.  Are you able to run the test locally?  Perhaps we can try using [explicit waits](https://selenium-python.readthedocs.io/waits.html) with a larger wait time to see if that helps.

samuel-w:
Seems to fail with explicit waits as well https://github.com/samuel-w/duplicati/runs/3706495365?check_suite_focus=true

PR #4618, opened 2021-10-01
Fixed backup export to CLI Fail for space on Windows.

Sanitiy:
Exporting a backup configuration to CLI in Windows failed if the configuration had an option/filter which contained a space.
Cause was that options/filters weren't surrounded completely by brackets, but like --option="...".

Since I have no Linux installation, somebody please check 1 or 2 exports to command line in Linux


warwickmm:
@Sanitiy, can you provide an example of an option that reveals the issue?  I can test this on Linux.

Sanitiy:
Turns out I've been fixing a phantom error. The error appears only if you start the commandline.cs  from Rider, in which case any password with a space fails. The same configuration pasted into cmd works flawlessly...

PR #4640, opened 2021-11-12
Renames relating to the extended device path prefix, \\?\

dferreyra:
The renames are based on the constant names used in the .NET Runtime repository; see https://github.com/dotnet/runtime/blob/7211aa01b34bb55ca67bdddd6e80ce23ee201bd2/src/libraries/Common/src/System/IO/PathInternal.Windows.cs#L28-L46.

### DIFF
--- a/Duplicati/Library/Common/IO/SystemIOWindows.cs
+++ b/Duplicati/Library/Common/IO/SystemIOWindows.cs
@@ -29,20 +29,24 @@ namespace Duplicati.Library.Common.IO
 {
     public struct SystemIOWindows : ISystemIO
     {
-        private const string UNCPREFIX = PathInternalWindows.ExtendedDevicePathPrefix;
-        private const string UNCPREFIX_SERVER = PathInternalWindows.UncExtendedPathPrefix;
-        private const string PATHPREFIX_SERVER = PathInternalWindows.UncPathPrefix;
-        private const string PATHPREFIX_SERVER_ALT = @"//";
+        // Based on the constant names used in
+        // https://github.com/dotnet/runtime/blob/v5.0.12/src/libraries/Common/src/System/IO/PathInternal.Windows.cs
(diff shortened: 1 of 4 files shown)


Duplicati/Library/Common/IO/SystemIOWindows.cs
```diff
@@ -29,57 +29,69 @@ namespace Duplicati.Library.Common.IO
...
-        private const string UNCPREFIX = PathInternalWindows.ExtendedDevicePathPrefix;
-        private const string UNCPREFIX_SERVER = PathInternalWindows.UncExtendedPathPrefix;
-        private const string PATHPREFIX_SERVER = PathInternalWindows.UncPathPrefix;
-        private const string PATHPREFIX_SERVER_ALT = @"//";
+        // Based on the constant names used in
+        // https://github.com/dotnet/runtime/blob/v5.0.12/src/libraries/Common/src/System/IO/PathInternal.Windows.cs
+        private const string ExtendedDevicePathPrefix = @"\\?\";
+        private const string UncPathPrefix = @"\\";
+        private const string AltUncPathPrefix = @"//";
+        private const string UncExtendedPathPrefix = @"\\?\UNC\";
```
dferreyra, line 37: The previous constants were all caps.  In this commit I thought we'd just use the constant names from the .NET Runtime sources verbatim.

Also, I changed from referring to PathInternalWindows for the constants to just copying over the constants.  My thought was that since these constants are not exposed in .NET, once Duplicati is on, say, .NET 6, the .NET Runtime sources can be removed and all that would need to be changed is to change the call to `DotNetRuntimePathWindows.IsPathFullyQualified` to `Path.IsPathFullyQualified` instead.

warwickmm: Good idea.  Thanks @dferreyra.

PR #4651, opened 2021-12-15
Don't explicitly call reload() in BackupList

ltGuillaume:
serverstatechanged.lastDataUpdateId is already triggered, so reload() was called twice.

This also caused the backup list to always "flicker" (show - hide - show) upon opening the web gui.


warwickmm:
Thanks @ltGuillaume.

Issue #2399, opened 2017-03-22
Use exponential backoff for retries


shoeper:
I have:
- [ x ] searched open and closed issues for duplicates

----------------------------------------

### Version info

**Duplicati Version:** 2.0.1.48_canary_2017-02-23
**Operating System:** Linux, Debian Jessie 64-Bit with Docker and this image: https://hub.docker.com/r/intersoftlab/duplicati/
**Backend:** OneDrive (personal)

### Bug description
Unauthorized 401 and backups aborted

### Steps to reproduce
Don't know where the issue was. Likely a temporary issue with either my internet connection, OneDrive itself or the OAuth handler instance (as the log from another backup indicates). Was the first issue after continuously uploading for something like 3 weeks.

**Actual result:** It says 401 unauthorized and fails to retry (likely) because retries are done without timeout. Restarting it manually did continue the upload without any further actions taken.
**Expected result:** After failure Duplicati tries again after a few seconds with an incrementing time. E.g. Duplicati waits 2 seconds. If it fails again 4, then 8, 16, ... 1024 (aborts if it also fails after 1024s)

### Screenshots

none

### debug log

```
The remote server returned an error: (401) Unauthorized.
```

```
21. Mrz. 2017 19:02: Message
Fatal error
System.Net.WebException: The remote server returned an error: (401) Unauthorized.
  at Duplicati.Library.Main.BackendManager.List () [0x00041] in <removed>:0 
  at Duplicati.Library.Main.Operation.FilelistProcessor.RemoteListAnalysis (Duplicati.Library.Main.BackendManager backend, Duplicati.Library.Main.Options options, Duplicati.Library.Main.Database.LocalDatabase database, Duplicati.Library.Main.IBackendWriter log, System.String protectedfile) [0x0000d] in <removed>:0 
  at Duplicati.Library.Main.Operation.FilelistProcessor.VerifyRemoteList (Duplicati.Library.Main.BackendManager backend, Duplicati.Library.Main.Options options, Duplicati.Library.Main.Database.LocalDatabase database, Duplicati.Library.Main.IBackendWriter log, System.String protectedfile) [0x00000] in <removed>:0 
  at Duplicati.Library.Main.Operation.BackupHandler.PreBackupVerify (Duplicati.Library.Main.BackendManager backend, System.String protectedfile) [0x0006c] in <removed>:0 
21. Mrz. 2017 19:02: Retry
[...]

PR #4661, opened 2022-01-15
Implement exponential backoff for backend errors

vmsh0:
Hi,

I'm opening this pull request with an initial proof-of-concept implementation of exponential backoff for backend interactions and uploads. In my evaluation, the feature might be useful as shown by #2399 and https://forum.duplicati.com/t/scheduled-backup-failure-google-drive-403-forbidden/13160.

I've experienced similar issues myself (I have an ~8TB backup which has been going since the beginning of December, more details in #2399), which I mitigated by increasing the number of retries and the timeout, but the solution is sub-optimal, because /all/ retries are of course penalized by the increased delay.

The implementation I'm presenting here is very simple and straightforward. Please forgive my unfamiliarity with the project style an conventions. I'm hoping someone from the project could guide me towards producing further iterations of this PR and reaching a solution suitable for inclusion.

### DIFF
--- a/Duplicati/Library/Main/Operation/Backup/BackendUploader.cs
+++ b/Duplicati/Library/Main/Operation/Backup/BackendUploader.cs
@@ -318,7 +318,10 @@ namespace Duplicati.Library.Main.Operation.Backup
             for (retryCount = 0; retryCount <= m_options.NumberOfRetries; retryCount++)
             {
                 if (m_options.RetryDelay.Ticks != 0 && retryCount != 0)
-                    await Task.Delay(m_options.RetryDelay).ConfigureAwait(false);
+                {
+                    var delay = Library.Utility.Utility.GetRetryDelay(m_options.RetryDelay, retryCount, m_options.RetryWithExponentialBackoff);
+                    await Task.Delay(delay).ConfigureAwait(false);
+                }
 
                 if (cancelToken.IsCancellationRequested)
                     return false;
--- a/Duplicati/Library/Main/Options.cs
+++ b/Duplicati/Library/Main/Options.cs
@@ -280,6 +280,7 @@ namespace Duplicati.Library.Main
 
                     new CommandLineArgument("number-of-retries", CommandLineArgument.ArgumentType.Integer, Strings.Options.NumberofretriesShort, Strings.Options.NumberofretriesLong, "5"),
                     new CommandLineArgument("retry-delay", CommandLineArgument.ArgumentType.Timespan, Strings.Options.RetrydelayShort, Strings.Options.RetrydelayLong, "10s"),
+                    new CommandLineArgument("retry-with-exponential-backoff", CommandLineArgument.ArgumentType.Boolean, Strings.Options.RetrywithexponentialbackoffShort, Strings.Options.RetrywithexponentialbackoffLong, "false"),
 
                     new CommandLineArgument("synchronous-upload", CommandLineArgument.ArgumentType.Boolean, Strings.Options.SynchronousuploadShort, Strings.Options.SynchronousuploadLong, "false"),
                     new CommandLineArgument("asynchronous-upload-limit", CommandLineArgument.ArgumentType.Integer, Strings.Options.AsynchronousuploadlimitShort, Strings.Options.AsynchronousuploadlimitLong, "4"),
@@ -814,7 +815,7 @@ namespace Duplicati.Library.Main
         public bool DisablePipedStreaming { get { return GetBool("disable-piped-streaming"); } }
 
         /// <summary>
-        /// Gets the timelimit for removal
+        /// Gets the delay period to retry uploads
         /// </summary>
         public TimeSpan RetryDelay
         {
@@ -825,6 +826,14 @@ namespace Duplicati.Library.Main
                 else
                     return Library.Utility.Timeparser.ParseTimeSpan(m_options["retry-delay"]);
             }
+        }
+
+        /// <summary>
+        /// Gets whether exponential backoff is enabled
+        /// </summary>
+        public Boolean RetryWithExponentialBackoff
+        {
+            get { return Library.Utility.Utility.ParseBoolOption(m_options, "retry-with-exponential-backoff"); }
         }
 
         /// <summary>
--- a/Duplicati/Library/Utility/Utility.cs
+++ b/Duplicati/Library/Utility/Utility.cs
@@ -1497,5 +1497,34 @@ namespace Duplicati.Library.Utility
         {
             return task.GetAwaiter().GetResult();
         }
+
+        /// <summary>
+        /// Utility that computes the delay before the next retry of an operation, optionally using exponential backoff.
+        /// Note: when using exponential backoff, the exponent is clamped at 10.
+        /// </summary>
+        /// <param name="retryDelay">Value of one delay unit</param>
+        /// <param name="retryAttempt">The attempt number (e.g. 1 for the first retry, 2 for the second retry, etc.)</param>
+        /// <param name="useExponentialBackoff">Whether to use exponential backoff</param>
+        /// <returns>The computed delay</returns>
+        public static TimeSpan GetRetryDelay(TimeSpan retryDelay, int retryAttempt, bool useExponentialBackoff)
+        {
+            if (retryAttempt < 1)
+            {
+                throw new ArgumentException("The attempt number must not be less than 1.", nameof(retryAttempt));
+            }
+
+            TimeSpan delay;
+            if (useExponentialBackoff)
+            {
+                var delayTicks = retryDelay.Ticks << Math.Min(retryAttempt - 1, 10);
+                delay = TimeSpan.FromTicks(delayTicks);
+            }
+            else
+            {
+                delay = retryDelay;
+            }
+
+            return delay;
+        }
     }
 }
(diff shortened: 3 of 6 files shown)


Duplicati/Library/Main/Operation/Backup/BackendUploader.cs
```diff
@@ -318,7 +318,15 @@ private async Task<bool> DoWithRetry(Func<Task> method, FileEntryItem item, Work
             for (retryCount = 0; retryCount <= m_options.NumberOfRetries; retryCount++)
             {
                 if (m_options.RetryDelay.Ticks != 0 && retryCount != 0)
-                    await Task.Delay(m_options.RetryDelay).ConfigureAwait(false);
+                {
+                    TimeSpan delay;
+                    if (m_options.RetryDelayExponential)
+                        delay = TimeSpan.FromTicks(m_options.RetryDelay.Ticks * (2L << Math.Min(63, retryCount - 1)));
```
warwickmm, line 324: Was there a reason to prefer the form
```
x * (2L << (y - 1))
```
instead of just
```
x << y
```

Duplicati/Library/Main/Operation/Backup/BackendUploader.cs
```diff
@@ -318,7 +318,15 @@ private async Task<bool> DoWithRetry(Func<Task> method, FileEntryItem item, Work
             for (retryCount = 0; retryCount <= m_options.NumberOfRetries; retryCount++)
             {
                 if (m_options.RetryDelay.Ticks != 0 && retryCount != 0)
-                    await Task.Delay(m_options.RetryDelay).ConfigureAwait(false);
+                {
+                    TimeSpan delay;
+                    if (m_options.RetryDelayExponential)
+                        delay = TimeSpan.FromTicks(m_options.RetryDelay.Ticks * (2L << Math.Min(63, retryCount - 1)));
+                    else
+                        delay = m_options.RetryDelay;
```
warwickmm, line 326: To avoid duplicating the logic in `BackendManager`, could we extract this to a method in the `Duplicati.Library.Utility.Utility` class?  Something like
```
public static TimeSpan GetRetryDelay(TimeSpan retryDelay, int retryAttempt, bool useExponentialBackoff)
{
    if (retryAttempt < 0)
    {
        // throw exception
    }
    ...
}
```
This would also make it easier to add a really simple unit test to the `Duplicati.UnitTest.UtilityTest` class to verify the behavior.

warwickmm:
Thanks @vmsh0.  Our style conventions have evolved and are not consistent throughout the solution.  However, you did the right thing by being consistent with the style of the surrounding code.

Duplicati/Library/Main/Options.cs
```diff
@@ -827,6 +828,14 @@ public TimeSpan RetryDelay
             }
         }
 
+        /// <summary>
+        /// Gets whether exponential backoff is enabled
+        /// </summary>
+        public Boolean RetryDelayExponential
```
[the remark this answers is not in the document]
vmsh0: Yes, I believe `RetryWithExponentialBackoff` makes it both clearer that it's a boolean, and what it does. I'm changing it to that. Same for the argument.

Duplicati/Library/Main/Operation/Backup/BackendUploader.cs
vmsh0: I'm taking `(y - 1)` as opposed to `y` because `y` (i.e. the retries counter) is first incremented after the first failure. So, taking `y` would result in the first delay being double of what the user has set with the `retry-delay` argument, which I don't think is the behaviour the user would expect.

As for using the form `x * (2L << (y-1))` instead of `x << (y - 1)`, I have now changed it to `x << (y -1)`, which is indeed more concise and means the same thing.

Duplicati/Library/Main/Operation/Backup/BackendUploader.cs
vmsh0: Done. I used your template code to check the `retryAttempt` parameter. Not being familiar with the project conventions, I'm not sure whether the other parameters should be checked as well.

I also added the test.

vmsh0:
I have pushed a new version of the patch, addressing the previous comments, as well as introducing a clamp to the exponential backoff exponent. I think 10 is a sane default, but should we add a parameter to allow changing it?

Duplicati/UnitTest/UtilityTests.cs
```diff
@@ -495,6 +495,34 @@ public static void ThrottledStreamWrite()
...
+            int[] testValues = { 1, 2, 63, 64, int.MaxValue };
+            double[] expect =  { 1, 1,  1,  1,            1 };
+
+            for (int i = 0; i < testValues.Length; i++)
+                Assert.AreEqual(TimeSpan.FromSeconds(expect[i]), Utility.GetRetryDelay(baseDelay, testValues[i], false));
+        }
+
+        [Test]
+        [Category("Utility")]
+        public static void RetryDelayExponentialBackoffTest()
```
warwickmm, line 515: Minor issue, but I think we can omit `Test` from the end of the method name.

warwickmm:
> I think 10 is a sane default, but should we add a parameter to allow changing it?

I think `10` is a reasonable default.  To avoid having too many options, we can leave this as-is until someone makes the case for this to be customizable.

Duplicati/Library/Utility/Utility.cs
```diff
@@ -1497,5 +1497,34 @@ public static T Await<T>(this Task<T> task)
...
+        /// </summary>
+        /// <param name="retryDelay">Value of one delay unit</param>
+        /// <param name="retryAttempt">The attempt number (e.g. 1 for the first retry, 2 for the second retry, etc.)</param>
+        /// <param name="useExponentialBackoff">Whether to use exponential backoff</param>
+        /// <returns>The computed delay</returns>
+        public static TimeSpan GetRetryDelay(TimeSpan retryDelay, int retryAttempt, bool useExponentialBackoff)
+        {
+            if (retryAttempt < 1)
+            {
+                throw new Exception("todo");
```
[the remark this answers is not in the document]
vmsh0: Done!

Duplicati/UnitTest/UtilityTests.cs
vmsh0: Done.

vmsh0:
I implemented the suggested changes, and rebased my patch on your latest master.

warwickmm:
Looks good to me.  Thanks @vmsh0!

Paul-5:
Hi @vmsh0!, warwickmm just let me know about this fix after I implemented something similar but it was specifically for Google Services & Drive. A couple of things I've observed about this fix. It appears that the first thing it does in DoWithRetry is delay and then run the operation, unless I'm misreading the code it should be the other way round with the delay at the end of the loop, possibly within the exception handler.

The other thing I've noticed is that the fix will retry for any exception, whether or not it's a retriable exception. I realise that each backend has its own way of alerting errors, particularly retriable errors. I was thinking that a base class for backends could implement a virtual method maybe called "IsRetriableException(Exception ex)" that returns true in the base class. When overridden in a backend handler it could interrogate the exception and determine whether the exception was truly retriable for that backend. If the exception is retriable then the code would enter the retry delay loops otherwise it would simply rethrow the exception.

Would you be happy for me make a few modifications to this fix accordingly?

vmsh0:
Hi Paul-5,

Thank you for your feedback.

> It appears that the first thing it does in DoWithRetry is delay and then run the operation, unless I'm misreading the code it should be the other way round with the delay at the end of the loop, possibly within the exception handler.

That is not correct. [The conditional block at line 320](https://github.com/vmsh0/duplicati/blob/528478fda480da09deb7db3cf74d5f027224c843/Duplicati/Library/Main/Operation/Backup/BackendUploader.cs#L318-L324) skips the delay for the first iteration.

> 
> The other thing I've noticed is that the fix will retry for any exception, whether or not it's a retriable exception. I realise that each backend has its own way of alerting errors, particularly retriable errors. I was thinking that a base class for backends could implement a virtual method maybe called "IsRetriableException(Exception ex)" that returns true in the base class. When overridden in a backend handler it could interrogate the exception and determine whether the exception was truly retriable for that backend. If the exception is retriable then the code would enter the retry delay loops otherwise it would simply rethrow the exception.

I would suggest you to discuss it with any of the more active Duplicati developers. I have only made this one fix, and gave a small contribution to a discussion related to a completely different issue, and as such I'm afraid I have nothing interesting to say! Fwiw, it does sound like a sensible idea.

> 
> Would you be happy for me make a few modifications to this fix accordingly?

Please note that my PR only changes the way the duration of the delays is computed. It does not change or address any of the delay logic, which from what I can see [has been in place for ~3 years](https://github.com/vmsh0/duplicati/commit/4f4529a88a31a34a623287e7e6c2a97351552862) now. As such, I think it might be a better idea to open a separate pull request for your proposed improvements regarding retriable exceptions (possibly after discussing it on the forum!) rather than contributing to this one.

Paul-5:
> That is not correct. [The conditional block at line 320](https://github.com/vmsh0/duplicati/blob/528478fda480da09deb7db3cf74d5f027224c843/Duplicati/Library/Main/Operation/Backup/BackendUploader.cs#L318-L324) skips the delay for the first iteration.

Ah, yes, thank you, I see it now.

> As such, I think it might be a better idea to open a separate pull request for your proposed improvements regarding retriable exceptions (possibly after discussing it on the forum!) rather than contributing to this one.

Will do, thank you. I just want to kick around the idea before implementing as it affects backend handlers, the majority of which I have not looked at... I've only played with Google Drive.

Thank you for your help and for your fix.

PR #4663, opened 2022-01-20
Exec full path in .desktop file

kdejaeger:
I added this to /home/xxx/.config/autostart and it didn't work unless I had the full path in the Exec property.


warwickmm:
@drwtsn32x, @ts678 do any of you have a Debian installation to test this?

drwtsn32x:
@warwickmm I can help with testing.  @kdejaeger what version of Debian did you experience the problem, and what GUI (guessing KDE from your username)

kdejaeger:
I used the rpm on opensuse tumbleweed. kde indeed

drwtsn32x:
@kdejaeger - I thought this fix was for a debian based distribution?

kdejaeger:
No please test this in debian as well. Check is the absolute path is always the same.

ts678:
> I thought this fix was for a debian based distribution?

A source search found fedora and others copying file from debian tree. 7-Zip can see it in `.rpm` but I don't have it on a system.
/usr/share/applications looks like its usual install place. I can't think offhand of a Linux where full path to `duplicati` is different.

[The Exec key](https://specifications.freedesktop.org/desktop-entry-spec/desktop-entry-spec-latest.html#exec-variables) (if that is the relevant spec) explains this as

> If no full path is provided the executable is looked up in the $PATH environment variable used by the desktop environment.

but what's a little odd is -- for a given desktop (on which I'm no expert) does it now work in menu system but fail in autostart?

kdejaeger:
It's not a big issue for me if this doesn't get fixed. But yea, if you have monthly backups, you just want the service to be running in the background right.

warwickmm:
I'm surprised this hasn't been mentioned before.  That `Exec` line was changed from `/usr/bin/duplicati` to `duplicati` in [2013](https://github.com/duplicati/duplicati/commit/0f54d8bf61d8af6fa2faae53ce2b3f2dd2be426e).

drwtsn32x:
@warwickmm - wonder if there was a good reason it was changed. Hopefully changing it back won't lead to some regression.  (By the way, I have not yet had time to test anything..)

warwickmm:
My guess is that it could be easier for a user to add something to their path than it is to add something to `/usr/bin`.  2013 was quite a while ago, so a lot has changed since then.

Given that there haven't been any other reported issues with this, I'm a bit reluctant to make this change without better understanding how it might affect existing users.

warwickmm:
I'm going to close this since there haven't been many complaints about this for 9 years.  We can revisit later if necessary.

kdejaeger:
This product is super complex. Why can you test this simple issue? If you want, I could test it in a debian VM.

warwickmm:
Thanks @kdejaeger.  I only run Arch and `/usr/bin/duplicati` does not exist so will need help from others to test other distributions.  The fact that it works from the start menu and not from autostart is strange.  Could they be running as different users with different paths?

warwickmm:
If it helps, the Debian and Fedora systemd service files we provide reference `/usr/bin/duplicati-server`

https://github.com/duplicati/duplicati/blob/master/Installer/debian/debian/duplicati.service
https://github.com/duplicati/duplicati/blob/master/Installer/fedora/duplicati.service

warwickmm:
I'll reopen this as a draft.

kdejaeger:
Ok so on arch the path is different. Then it's better to keep it like it is. I should be using the service files better then. Thanks

kdejaeger:
What works as well in kde, instead of adding the preconfigured .desktop file that comes with the installation. Is to add the path to duplicati-server. KDE then seems to create a .desktop file in the .autostart folder with the command you have provided via the GUI. So everyone on KDE can do this, not necessary to play around with sysctl.

koen@localhost ~ $ ll .config/autostart
-r--r--r-- 1 koen users 850 Aug 10  2020 org.opensuse.opensuse_welcome.desktop
total 12
-rw------- 1 koen users 747 Nov 20 12:40 brave-browser.desktop
-rw------- 1 koen users 113 Feb  8 12:15 duplicati-server.desktop
koen@localhost ~ $ cat .config/autostart/duplicati-server.desktop 
[Desktop Entry]
Exec=/usr/bin/duplicati-server
Icon=
Name=duplicati-server
Path=
Terminal=False
Type=Application
koen@localhost ~ $

Issue #4106, opened 2020-02-17
Filters textboxes are too small horizontally


Germs2004:
- [x ] I have searched open and closed issues for duplicates.

----------------------------------------

## Environment info
- **Duplicati version**: 2.0.5.101_canary_2020-01-23
- **Operating system**: Windows 10

## Description
When you edit a backup job, on the "Source data" page, it shows a list of "Filters" as textboxes.  On my screen, those textboxes are so short horizontally that I can't see the whole path.  If the textboxes would be edited to fill the available horizontal space (between the filter type dropdown and the X button) then it would help to see the filter text much better.

## Screenshots
![screenshot](https://user-images.githubusercontent.com/5519340/74678197-4d269e00-5188-11ea-9f12-11ce9a4f7196.PNG)

warwickmm:
If anyone's interested in helping out, this will probably involve modifying `style.less` here

https://github.com/duplicati/duplicati/blob/8bb7b27c1f3c3ac0947c279692b2aa08a2be22be/Duplicati/Server/webroot/ngax/less/style.less#L1231-L1235

and then running
```
lessc Duplicati/Server/webroot/ngax/less/style.less --clean-css > Duplicati/Server/webroot/ngax/styles/style.css
```
The [Duplicati/Server/webroot/ngax/templates/addoredit.html](https://github.com/duplicati/duplicati/blob/v2.0.5.104-2.0.5.104_canary_2020-03-25/Duplicati/Server/webroot/ngax/templates/addoredit.html) file might need to be modified as well.

PR #4665, opened 2022-01-23
Fix Filters texboxes width - Issue #4106

idotj:
**#4106 Filters textboxes are too small horizontally**
This changes will allow the input field to use all the empty space dynamically.
Tested also in mobile screen


warwickmm:
Thanks @idotj!  Looks good to me on Brave 1.34.81 (Chromium: 97.0.4692.99).  @drwtsn32x, are you able to test this with a few other browsers?

idotj:
Thanks @warwickmm, yes I did, under Firefox 96 and Chromium 89 (both at different resolutions between 360px and 1920px width)

warwickmm:
Sounds good!  Thanks @idotj.

Issue #3833, opened 2019-07-21
[Feature request] Resizable log view


funnym0nk3y:
- [x] I have searched open and closed issues for duplicates.

----------------------------------------

## Environment info
- **Duplicati version**: 2.0.4.23_beta_2019-07-14
- **Operating system**: Windows 10 
- **Backend**: local

## Description
This is a feature request. When viewing the logs generated by Duplicati the column in which the text is displayed is very narrow. This leads to a wordwrap which makes Messages, Warnings and Errors difficult to read. Therefor I am suggesting a adaptable width of that column either manually or automatically.

BlueBlock:
What if there as a link to download the log and then you can view it in your preferred editor?

funnym0nk3y:
@BlueBlock  I'm a little torn between those two options. Making the log-files downloadable from the user interface adds many possibilities to customize to following workflow. But when having a quick look at the logs downloading and opening a file is a little awkward. Viewing the log inside the web interface maybe adds a feel of cleanliness and seclusion. Maybe both solutions? Just thinking

BlueBlock:
Perhaps having the log color coded might help a bit a bit since it is JSON.

ts678:
> But when having a quick look at the logs downloading and opening a file is a little awkward.

Less awkward (but still not as nice as resize) is copy and paste, which now has a button to select just the log text so you can copy.

![image](https://user-images.githubusercontent.com/40771998/102032705-89fef980-3d87-11eb-914c-3cebcc54e045.png)

This (I think) came in on the [2.0.4.6](https://github.com/duplicati/duplicati/releases/tag/v2.0.4.6-2.0.4.6_canary_2018-12-11) redesign. [2.0.4.23](https://github.com/duplicati/duplicati/releases/tag/v2.0.4.23-2.0.4.23_beta_2019-07-14) is actually just 2.0.4.5 and a warning about Amazon Cloud Drive retirement.

> Reworked the log UI, thanks @LacunaSoftware

I'm not positive if that added the selection button. Regardless, please look at 2.0.5.1 Beta and see if it helps with log view reading.
Warnings and Errors are now broken out in separate UI sections without having to look through the `Complete log` to find them.
There's still unfortunately an issue that fatal backup errors wind up in the server log at About --> Show log, but it's nice and wide.

PR #4669, opened 2022-01-29
Resizable log view #3833

idotj:
Replaced the `<div>` containing the complete log by a `<textarea readonly>`
This will allow the user to resize the width and height if clicks and drag at the bottom right corner. Also vertical and horizontal scroll appears when needed (native behavior for `<textarea>`)

I also pushed the copy log link to the right side and added the label/text.


warwickmm:
Thanks @idotj!  When I click "Copy log", the content is copied to the clipboard, but the text in the textarea is also selected.  This might give users the impression that this merely selects the text, and they still have to manually perform the copy.  Is it possible for the text to be copied to the clipboard without visually indicating the selection?

image

warwickmm:
Sorry, I just noticed that's the existing behavior as well.  We can leave that for a separate pull request in the future.

Localizations/webroot/localization_webroot-bn.po
```diff
@@ -449,6 +449,10 @@ msgstr ""
 msgid "Complete log"
 msgstr ""
 
+#: templates/backup-result/box.html:40
+msgid "Copy log"
+msgstr ""
+
```
warwickmm, line 455: I believe the `.po` and `.pot` files are automatically created when we update the translations prior to a release, so we do not need to modify them here.

idotj: Ok, I removed all the changes for .po and .pot files, should be fine.

warwickmm:
Thanks @idotj!.

PR #4676, opened 2022-02-04
Added null check to fix exception reading vhd settings xml.

awerghcpc:
When I attempted to backup a single Hyper-V VM on my Win10 machine it failed with the following error:
**System.ArgumentNullException: Value cannot be null.
Parameter name: s**

This was caused by outParams["SettingsData"] being null for one of the vhds.
**doc.LoadXml((string)outParams[_wmiv2Namespace ? "SettingData" : "Info"]);**

For this particular vhd multiple paths were returned an invalid path twice:
U:\vhds\NetBSD PXE Test.vhdx
and the actual path once: 
U:\vhds\NetBSD_PXE_Test.vhdx 

When the path was invalid the null occurred.
Adding the null check for outParams prevents this ArgumentNullException from occurring when reading the XML.

### DIFF
--- a/Duplicati/Library/Snapshots/HyperVUtility.cs
+++ b/Duplicati/Library/Snapshots/HyperVUtility.cs
@@ -302,7 +302,16 @@ namespace Duplicati.Library.Snapshots
                                        select ((string[])systemBaseObj["Connection"])[0]).ToList();
 
                         foreach (var vhd in tempvhd)
-                            result.Add(vhd);
+                        {
+                            if (File.Exists(vhd))
+                            {
+                                result.Add(vhd);
+                            }
+                            else
+                            {
+                                Logging.Log.WriteWarningMessage(LOGTAG, "HyperVInvalidVhd", null, "Invalid VHD file detected, file does not exist: {0}", vhd);
+                            }
+                        }
                     }
             }
 
--- a/Duplicati/Library/Snapshots/HyperVUtility.cs
+++ b/Duplicati/Library/Snapshots/HyperVUtility.cs
@@ -319,11 +319,16 @@ namespace Duplicati.Library.Snapshots
                         if (outParams != null)
                         {
                             var doc = new System.Xml.XmlDocument();
-                            doc.LoadXml((string)outParams[_wmiv2Namespace ? "SettingData" : "Info"]);
-                            var node = doc.SelectSingleNode("//PROPERTY[@NAME = 'ParentPath']/VALUE/child::text()");
+                            var propertyValue = (string)outParams[_wmiv2Namespace ? "SettingData" : "Info"];
 
-                            if (node != null && File.Exists(node.Value))
-                                ParentPaths.Add(node.Value);
+                            if (propertyValue != null)
+                            {
+                                doc.LoadXml(propertyValue);
+                                var node = doc.SelectSingleNode("//PROPERTY[@NAME = 'ParentPath']/VALUE/child::text()");
+
+                                if (node != null && File.Exists(node.Value))
+                                    ParentPaths.Add(node.Value);
+                            }
                         }
                     }
                 }


warwickmm:
Thanks @awerghcpc.

> For this particular vhd multiple paths were returned an invalid path twice:
U:\vhds\NetBSD PXE Test.vhdx
and the actual path once:
U:\vhds\NetBSD_PXE_Test.vhdx

Do you know what might have caused the incorrect path to be returned?

mnaiman:
Can you please dump, how XML with invalid path look like?

awerghcpc:
> Do you know what might have caused the incorrect path to be returned?

I'm not sure, it is possible I created the vhdx with spaces first and then later changed it to underscores but I don't remember.
The VM itself is a Generation 1 VM with a Configuration Version of 9.0.
Looking through the WMI queries the objects both exist so presumably they get registered by Hyper-V in some way.

> Can you please dump, how XML with invalid path look like?

The SettingsData value is null.
image

Using WMI Explorer I can compare both the valid and invalid path (looks like all invalid paths look the same):
U:\vhds\NetBSD PXE Test.vhdx
image

instance of __PARAMETERS
{
	Job = "\\\\ACPW01\\root\\virtualization\\v2:Msvm_StorageJob.InstanceID=\"49224CFA-935F-40CD-8DBC-85662A03AEB9\"";
	ReturnValue = 4096;
	SettingData = NULL;
};

For the valid path:
U:\vhds\NetBSD_PXE_Test.vhdx
image

instance of __PARAMETERS
{
	Job = NULL;
	ReturnValue = 0;
	SettingData = "<INSTANCE CLASSNAME=\"Msvm_VirtualHardDiskSettingData\"><PROPERTY NAME=\"BlockSize\" TYPE=\"uint32\"><VALUE>33554432</VALUE></PROPERTY><PROPERTY NAME=\"Caption\" TYPE=\"string\"><VALUE>Virtual Hard Disk Setting Data</VALUE></PROPERTY><PROPERTY NAME=\"DataAlignment\" PROPAGATED=\"true\" TYPE=\"uint64\"></PROPERTY><PROPERTY NAME=\"Description\" TYPE=\"string\"><VALUE>Setting Data for a Virtual Hard Disk.</VALUE></PROPERTY><PROPERTY NAME=\"ElementName\" TYPE=\"string\"><VALUE>NetBSD_PXE_Test.vhdx</VALUE></PROPERTY><PROPERTY NAME=\"Format\" TYPE=\"uint16\"><VALUE>3</VALUE></PROPERTY><PROPERTY NAME=\"InstanceID\" TYPE=\"string\"><VALUE>D139015C-9C20-4938-AED1-D7491DB4B7D2</VALUE></PROPERTY><PROPERTY NAME=\"IsPmemCompatible\" TYPE=\"boolean\"><VALUE>FALSE</VALUE></PROPERTY><PROPERTY NAME=\"LogicalSectorSize\" TYPE=\"uint32\"><VALUE>512</VALUE></PROPERTY><PROPERTY NAME=\"MaxInternalSize\" TYPE=\"uint64\"><VALUE>10737418240</VALUE></PROPERTY><PROPERTY NAME=\"ParentIdentifier\" PROPAGATED=\"true\" TYPE=\"string\"></PROPERTY><PROPERTY NAME=\"ParentPath\" TYPE=\"string\"><VALUE></VALUE></PROPERTY><PROPERTY NAME=\"ParentTimestamp\" PROPAGATED=\"true\" TYPE=\"datetime\"></PROPERTY><PROPERTY NAME=\"Path\" TYPE=\"string\"><VALUE>U:\\vhds\\NetBSD_PXE_Test.vhdx</VALUE></PROPERTY><PROPERTY NAME=\"PhysicalSectorSize\" TYPE=\"uint32\"><VALUE>4096</VALUE></PROPERTY><PROPERTY NAME=\"PmemAddressAbstractionType\" TYPE=\"uint16\"><VALUE>0</VALUE></PROPERTY><PROPERTY NAME=\"Type\" TYPE=\"uint16\"><VALUE>3</VALUE></PROPERTY><PROPERTY NAME=\"VirtualDiskId\" TYPE=\"string\"><VALUE>313FD1D9-625F-4968-8FFF-780D28DB13BE</VALUE></PROPERTY></INSTANCE>";
};

warwickmm:
I'm not familiar with this part of the code (and I don't have Windows).  @mnaiman, would you be able to review this when you have time?

mnaiman:
I think I found out where problem is, but I think VM config XML is corrupted.
ReturnValue = 4096 means file not found
ReturnValue = 0 means ok

I don't know how is it possible that Hyper-V is starting this machine. Please check event log if there is any 0x80070003 error.

If you export machine and examine machine XML is there again old path?

mnaiman:
New version of hyper-v is not exporting XML :/
But .vmcx file even as binary is still "readable"....

awerghcpc:
The VM in question is starting and running correctly (not the VM I was trying to backup incidentally).
I do seem to have some problems with Hyper-V on my machine, it is probably time to patch and restart and see if that makes a difference.
I have somewhat been avoiding touching the VM in question in case I fix it preventing me from reproducing this issue.

awerghcpc:
A warning makes a lot of sense. 
I have effectively changed it from throwing an exception that tells you nothing to hiding the problem from the user entirely.

The warning would show either the name of the disk or the disk and the original vm it was assigned to.
It is probably easier to put the warning in earlier (such as after finding the vhd doesn't exist) rather than as apart of the null check.

kenkendk:
The fix looks good to me, it simply defers loading the xml.
Did you want to update with a log message?

awerghcpc:
Added a check that the vhd file exists so that a warning can be returned to the user.
image

mnaiman:
That's great, no silent skipping files 👍

PR #4687, opened 2022-02-27
Add indexes to improve backup query performance

jedthe3rd:
Here is the PR for the indexes I identified to fix some query issues. I mentioned it all here: https://forum.duplicati.com/t/identified-another-slow-query-during-backup/14005

These probably should be included in the next beta release because they seem resolve some of the query issues caused by certain datasets combined with the default blocksize, which almost caused me to find a different backup solution.

Let me know if you have any questions.


jedthe3rd:
I am not sure why the UI tests are failing. None of my code touch the part of the test it is failing on.  @warwickmm  any ideas?

warwickmm:
Thanks @jedthe3rd!  The UI tests have some issues and fail randomly.  I'll re-run them.

Do you have any measurements on how these indexes impact the performance of the slow queries you identified, as well as the size of the database?

jedthe3rd:
Runtime Before index - 3:42:24.304 - Location LocalDatabase.cs Line 756
Runtime After index - ~28.5 seconds
```
SELECT "CalcLen", "Length", "A"."BlocksetID", "File"."Path"
FROM ( 
    SELECT "A"."ID" AS "BlocksetID", IFNULL("B"."CalcLen", 0) AS "CalcLen", "A"."Length"
    FROM "Blockset" A 
    LEFT OUTER JOIN ( 
        SELECT "BlocksetEntry"."BlocksetID", SUM("Block"."Size") AS "CalcLen"
        FROM "BlocksetEntry" 
        LEFT OUTER JOIN "Block" 
            ON "Block"."ID" = "BlocksetEntry"."BlockID"
        GROUP BY "BlocksetEntry"."BlocksetID" 
    ) B 
    ON "A"."ID" = "B"."BlocksetID" 
) A, "File"
WHERE "A"."BlocksetID" = "File"."BlocksetID" 
    AND "A"."CalcLen" != "A"."Length"
```

Runtime Before Index - 0:12:47:47.152 - Location LocalBackupDatabase.cs Line 586 
Runtime After Index - ~2.1 seconds
This index will improve runtime of all queries from Line 586 - 590 in LocalBackupDatabase.cs
```
SELECT COUNT(*)
FROM "File" 
INNER JOIN "FilesetEntry" ON "File"."ID" = "FilesetEntry"."FileID"
WHERE "FilesetEntry"."FilesetID" = 2 
    AND "File"."BlocksetID" = -200 
    AND NOT "File"."Path" IN (
        SELECT "Path"
        FROM "File" 
        INNER JOIN "FilesetEntry" 
            ON "File"."ID" = "FilesetEntry"."FileID"
        WHERE "FilesetEntry"."FilesetID" = 1
    )
```

Runtime Before Index - 0:18:45:54.738 - Line 602 & 603 LocalBackupDatabase.cs
Runtime After Index - ~3.5 seconds
```
SELECT COUNT(*)
FROM "File"
    INNER JOIN "FilesetEntry"
    ON "File"."ID" = "FilesetEntry"."FileID"
WHERE "FilesetEntry"."FilesetID" = 2
    AND "File"."BlocksetID" != -100
    AND "File"."BlocksetID" != -200
    AND NOT "File"."Path" IN (
        SELECT "Path"
    FROM "TmpFileList-8E37B6CB073EF54FA0A00784FD07EDD7"
    );
```

I don't see a noticeable change in DB size before or after the indexes are added, and from my testing the updates/inserts/deletes don't seem to be impacted performance wise at all.

The main goal of these indexes were to cause the query optimizer to not use table scans wherever possible. The First query has some table scans it does still which is why it has a 30 second runtime but it is about as efficient as it can get without restructuring how the data is obtained and possibly stored.

jedthe3rd:
Let me know if you need anything else for this to be merged!

Duplicati/Library/Main/Database/Database schema/12. Performance Indexes.sql
```diff
@@ -0,0 +1,12 @@
+CREATE INDEX "nnc_Metadataset" ON Metadataset ("ID","BlocksetID");
+CREATE INDEX "nn_FilesetentryFile" on FilesetEntry ("FilesetID","FileID");
```
warwickmm, line 2: I don't think we have a convention for the index names, but is there a meaning behind the `nnc_` and `nn_` prefixes?

image

jedthe3rd: Ya, nn = Non Unique non clustered, nc would be non unique clustered, the nnc stands for non-unique non-clustered and covered (covered meaning it has all the columns in a table in the index).

warwickmm: Sounds good, thanks!

warwickmm:
Looks good to me.  Thanks @jedthe3rd!

PR #4689, opened 2022-03-14
Enhanced error reporting for all Google Services & fix to Google Drive 403 (forbidden) error relating to malware and spam

PaulAHopkins:
Enhanced error reporting for all Google Services by unpacking and displaying the HTTP response that Google Drive and Google API returns. These changes have been made across all Google web methods.

In addition to this I have looked further into the 403 (forbidden) error being returned from Google Drive. In my case the error being returned was "This file has been identified as malware or spam and cannot be downloaded". It appears that Google Drive checks files before allowing them to be downloaded. In order to bypass this error Google Drive has a query string parameter called "acknowledgeAbuse" that must be set if a file is detected as having spam/malware (the flag can't be set generally, it can only be set for files detected as having spam/malware). I have modified the code to add this flag if this error is detected but only if a new Google Drive Duplicati option called "acknowledge-abuse" is also enabled.

Finally I have replaced (automatically) all "var" statements with their native types. The "var" statements were making the code difficult to read particularly from GitHub because it created uncertainty around variable types, removing the use of "var" makes the code let ambiguous.

### DIFF
--- a/Duplicati/Library/Backend/GoogleServices/GoogleCloudStorage.cs
+++ b/Duplicati/Library/Backend/GoogleServices/GoogleCloudStorage.cs
@@ -22,6 +22,7 @@ using Newtonsoft.Json;
 using System;
 using System.Collections.Generic;
 using System.Net;
+using System.IO;
 using System.Text;
 using System.Threading;
 using System.Threading.Tasks;
@@ -52,7 +53,7 @@ namespace Duplicati.Library.Backend.GoogleCloudStorage
 
         public GoogleCloudStorage(string url, Dictionary<string, string> options)
         {
-            var uri = new Utility.Uri(url);
+            Utility.Uri uri = new Utility.Uri(url);
 
             m_bucket = uri.Host;
             m_prefix = Util.AppendDirSeparator("/" + uri.Path, "/");
@@ -106,22 +107,31 @@ namespace Duplicati.Library.Backend.GoogleCloudStorage
                 if (wex.Response is HttpWebResponse response && response.StatusCode == HttpStatusCode.NotFound)
                     throw new FolderMissingException();
                 else
-                    throw;
+                {
+                    using (Stream exstream = wex.Response.GetResponseStream())
+                    {
+                        using (StreamReader reader = new StreamReader(exstream))
+                        {
+                            string error = reader.ReadToEnd();
+                            throw new WebException(error, wex, wex.Status, wex.Response);
+                        }
+                    }
+                }
             }
         }
 
         #region IBackend implementation
         public IEnumerable<IFileEntry> List()
         {
-            var url = WebApi.GoogleCloudStorage.ListUrl(m_bucket, Utility.Uri.UrlEncode(m_prefix));
+            string url = WebApi.GoogleCloudStorage.ListUrl(m_bucket, Utility.Uri.UrlEncode(m_prefix));
             while (true)
             {
-                var resp = HandleListExceptions(() => m_oauth.ReadJSONResponse<ListBucketResponse>(url));
+                ListBucketResponse resp = HandleListExceptions(() => m_oauth.ReadJSONResponse<ListBucketResponse>(url));
 
                 if (resp.items != null)
-                    foreach (var f in resp.items)
+                    foreach (BucketResourceItem f in resp.items)
                     {
-                        var name = f.name;
+                        string name = f.name;
                         if (name.StartsWith(m_prefix, StringComparison.OrdinalIgnoreCase))
                             name = name.Substring(m_prefix.Length);
                         if (f.size == null)
@@ -132,17 +142,17 @@ namespace Duplicati.Library.Backend.GoogleCloudStorage
                             yield return new FileEntry(name, f.size.Value, f.updated.Value, f.updated.Value);
                     }
 
-                var token = resp.nextPageToken;
+                string token = resp.nextPageToken;
                 if (string.IsNullOrWhiteSpace(token))
                     break;
                 url = WebApi.GoogleCloudStorage.ListUrl(m_bucket, Utility.Uri.UrlEncode(m_prefix), token);
             }
         }
 
-        public async Task PutAsync(string remotename, string filename, CancellationToken cancelToken)
+        public Task PutAsync(string remotename, string filename, CancellationToken cancelToken)
         {
             using (System.IO.FileStream fs = System.IO.File.OpenRead(filename))
-                await PutAsync(remotename, fs, cancelToken);
+                return PutAsync(remotename, fs, cancelToken);
         }
 
         public void Get(string remotename, string filename)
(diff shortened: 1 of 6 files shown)


Duplicati/Library/Backend/GoogleServices/GoogleDrive.cs
```diff
@@ -253,10 +307,10 @@ public IEnumerable<IFileEntry> List()
             }
         }
 
-        public async Task PutAsync(string remotename, string filename, CancellationToken cancelToken)
+        public Task PutAsync(string remotename, string filename, CancellationToken cancelToken)
         {
             using (System.IO.FileStream fs = System.IO.File.OpenRead(filename))
```
kenkendk, line 312: I don't think this works. The using statement will dispose the file stream when exiting the scope (returning the task). At this point the task may not have completed, and an error will occur when the code tries to process the disposed filestream. Using `await` inside the using will generate a state machine that will defer the dispose call until after the `PutAsync` operation has completed.

Duplicati/Library/Backend/GoogleServices/GoogleCommon.cs
```diff
@@ -195,26 +219,36 @@ private static async Task<T> ChunkedUploadAsync<T>(OAuthHelper oauth, string upl
...
 
                             return res;
                         }
                         else
                         {
-                            throw new WebException(string.Format("Unexpected status code: {0}", code), null, WebExceptionStatus.ServerProtocolViolation, resp);
+                            /*
+                             * Grab the error detail from the response stream
+                             */
+                            using (Stream exstream = resp.GetResponseStream())
```
kenkendk, line 233: I don't think you can guarantee there to be a response stream present? If the error is "Network disconnected" there will be no response stream and the user will get an exception here.

Duplicati/Library/Backend/GoogleServices/GoogleCommon.cs
```diff
@@ -104,18 +107,39 @@ private static long QueryUploadRange<T>(OAuthHelper oauth, string uploaduri, lon
...
+                }
+            }
+            /*
+             * Catch any web exceptions and grab the error detail from the repose stream
+             */
+            catch (WebException wex)
             {
-                if (resp.StatusCode != HttpStatusCode.OK || string.IsNullOrWhiteSpace(resp.Headers["Location"]))
-                    throw new WebException("Failed to start upload session", null, WebExceptionStatus.UnknownError, resp);
+                using (Stream exstream = wex.Response.GetResponseStream())
```
kenkendk, line 130: I don't think you can guarantee there to be a response stream present? If the error is "Network disconnected" there will be no response stream and the user will get an exception here.

Duplicati/Library/Backend/GoogleServices/GoogleCloudStorage.cs
```diff
@@ -249,37 +259,46 @@ public void Get(string remotename, System.IO.Stream stream)
...
                     Library.Utility.Utility.CopyStream(rs, stream);
             }
             catch (WebException wex)
             {
                 if (wex.Response is HttpWebResponse response && response.StatusCode == HttpStatusCode.NotFound)
                     throw new FileMissingException();
                 else
-                    throw;
+                {
+                    using (Stream exstream = wex.Response.GetResponseStream())
```
kenkendk, line 276: I don't think you can guarantee there to be a response stream present? If the error is "Network disconnected" there will be no response stream and the user will get an exception here.

Duplicati/Library/Backend/GoogleServices/GoogleCloudStorage.cs
```diff
@@ -106,22 +107,31 @@ private T HandleListExceptions<T>(Func<T> func)
                 if (wex.Response is HttpWebResponse response && response.StatusCode == HttpStatusCode.NotFound)
                     throw new FolderMissingException();
                 else
-                    throw;
+                {
+                    using (Stream exstream = wex.Response.GetResponseStream())
```
kenkendk, line 111: I don't think you can guarantee there to be a response stream present? If the error is "Network disconnected" there will be no response stream and the user will get an exception here.

Since this logic is used multiple places, it might be a good idea to create a single `ParseErrorResponse(WebException wex)` helper.

Duplicati/Library/Backend/OAuthHelper/JSONWebHelper.cs
```diff
@@ -150,7 +150,28 @@ public virtual async Task<HttpWebResponse> PostMultipartAsync(string url, Action
...
+            try
+            {
+                return (HttpWebResponse)(await req.GetResponseAsync().ConfigureAwait(false));
+            }
+            /*
+             * Catch any web exceptions and grab the error detail from the respose stream
+             */
+            catch (WebException wex)
+            {
+                using (Stream exstream = wex.Response.GetResponseStream())
```
kenkendk, line 162: I don't think you can guarantee there to be a response stream present? If the error is "Network disconnected" there will be no response stream and the user will get an exception here.

Duplicati/Library/Backend/OAuthHelper/JSONWebHelper.cs
```diff
@@ -367,6 +388,20 @@ public HttpWebResponse GetResponseWithoutException(AsyncHttpRequest req, object
                 if (wex.Response is HttpWebResponse response)
                     return response;
 
+                /*
+                 * If not an HttpWebResponse then grab the error detail from the response stream
+                 */
+                using (Stream exstream = wex.Response.GetResponseStream())
+                {
+                    using (StreamReader reader = new StreamReader(exstream))
```
kenkendk, line 396: I don't think you can guarantee there to be a response stream present? If the error is "Network disconnected" there will be no response stream and the user will get an exception here.

Duplicati/Library/Backend/OAuthHelper/JSONWebHelper.cs
```diff
@@ -420,7 +455,17 @@ public async Task<HttpWebResponse> GetResponseWithoutExceptionAsync(AsyncHttpReq
                 if (wex.Response is HttpWebResponse response)
                     return response;
 
-                throw;
+                /*
+                 * If not an HttpWebResponse then grab the error detail from the response stream
+                 */
+                using (Stream exstream = wex.Response.GetResponseStream())
+                {
+                    using (StreamReader reader = new StreamReader(exstream))
```
kenkendk, line 463: I don't think you can guarantee there to be a response stream present? If the error is "Network disconnected" there will be no response stream and the user will get an exception here.

Duplicati/Library/Backend/OAuthHelper/JSONWebHelper.cs
```diff
@@ -468,6 +513,20 @@ public HttpWebResponse GetResponse(AsyncHttpRequest req, object requestdata = nu
...
                 return (HttpWebResponse)req.GetResponse();
             }
+            /*
+             * If a web exception then grab the error detail from the response stream
+             */
+            catch (WebException wex)
+            {
+                using (Stream exstream = wex.Response.GetResponseStream())
+                {
+                    using (StreamReader reader = new StreamReader(exstream))
```
kenkendk, line 523: I don't think you can guarantee there to be a response stream present? If the error is "Network disconnected" there will be no response stream and the user will get an exception here.

Duplicati/Library/Backend/OAuthHelper/JSONWebHelper.cs
```diff
@@ -515,6 +574,20 @@ public async Task<HttpWebResponse> GetResponseAsync(AsyncHttpRequest req, Cancel
...
```
kenkendk, line 584: I don't think you can guarantee there to be a response stream present? If the error is "Network disconnected" there will be no response stream and the user will get an exception here.

kenkendk:
Excellent detective work! 

I would like there to be a common error parsing function instead of the replicated error handling, but otherwise I like the PR.

I am not a fan of dropping the `var` keywords as it replicates types and makes it harder to maintain.

kenkendk:
Closing this for lack of activity

PR #4690, opened 2022-03-17
Remove python 2 dependencies

warwickmm:
This is an attempt to remove our dependencies on python 2, which is no longer included as of macOS 12.3.

### DIFF
--- a/Duplicati/GUI/Duplicati.GUI.TrayIcon/OSXTrayHost/osx-trayicon-rumps.py
+++ b/Duplicati/GUI/Duplicati.GUI.TrayIcon/OSXTrayHost/osx-trayicon-rumps.py
@@ -41,7 +41,7 @@ def transform_app_type(toForeground):
 def item_clicked(sender):
     for k in lookup:
         if lookup[k].title == sender.title:
-            print "click:%s" % k
+            print("click:%s" % k)
             sys.stdout.flush()
 
 
@@ -49,7 +49,7 @@ def parsemenus(cfg):
     menuitems = []
     for n in cfg["Menus"]:
         mn = rumps.MenuItem(n["Text"])
-        if n.has_key("Enabled") and not n["Enabled"]:
+        if "Enabled" in n and not n["Enabled"]:
             mn.set_callback(None)
         else:
             mn.set_callback(item_clicked)
@@ -69,7 +69,7 @@ def get_input():
             if line == '':
                 continue
 
-            print "info:Read %s" % line
+            print("info:Read %s" % line)
             sys.stdout.flush()
             cfg = None
             try:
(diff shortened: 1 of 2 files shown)


warwickmm:
**Someone (with a Mac) needs to take over and fix this branch.**  On my system (Linux with python 3.10), I am unable to install the rumps dependencies (specifically, `pyobjc-framework-Cocoa`).

Our current `rumps.py` was derived from revision [ae11371bddc](https://github.com/jaredks/rumps/commit/ae11371bddcafbabc5f62d1eadddf83bfbd653dc) by taking the `rumps.py` file, embedding its dependencies, and then making some modifications to allow images in the notifications.

Below is a patch that illustrates the difference between Duplicati's `rumps.py` and the original version.  Someone will need to make the analogous modifications using a newer python 3 compatible version of rumps.

```
diff --git a/rumps/rumps.py b/rumps/rumps.py
index f93fbaf..a1ff64c 100644
--- a/rumps/rumps.py
+++ b/rumps/rumps.py
@@ -1,11 +1,20 @@
-#!/usr/bin/env python
+#!/usr/bin/env python2
 # -*- coding: utf-8 -*-
 #
 # rumps: Ridiculously Uncomplicated Mac os x Python Statusbar apps.
 # Copyright: (c) 2015, Jared Suttles. All rights reserved.
 # License: BSD, see LICENSE for details.
+#
+# This file is a join of the rumps.py and utils.py source from:
+# https://github.com/jaredks/rumps/commit/ae11371bddcafbabc5f62d1eadddf83bfbd653dc
+#
+# The method _nsimage_from_file has been modified to also allow 
+# loading a manually constructed NSImage
+# The notification class has also been modified to allow an image
 # - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - -
 
+from collections import OrderedDict
+
 _NOTIFICATIONS = True
 try:
     from Foundation import NSUserNotification, NSUserNotificationCenter
@@ -20,12 +29,29 @@ from PyObjCTools import AppHelper
 import os
 import weakref
 from collections import Mapping, Iterable
-from .utils import ListDict
 
 _TIMERS = weakref.WeakKeyDictionary()
 separator = object()
 
 
+# ListDict: OrderedDict subclass with insertion methods for modifying the order of the linked list in O(1) time
+# https://gist.github.com/jaredks/6276032
+class ListDict(OrderedDict):
+    def __insertion(self, link_prev, key_value):
+        key, value = key_value
+        if link_prev[2] != key:
+            if key in self:
+                del self[key]
+            link_next = link_prev[1]
+            self.OrderedDict[key] = link_prev[1] = link_next[0] = [link_prev, link_next, key]
+        dict.__setitem__(self, key, value)
+
+    def insert_after(self, existing_key, key_value):
+        self.__insertion(self.OrderedDict[existing_key], key_value)
+
+    def insert_before(self, existing_key, key_value):
+        self.__insertion(self.OrderedDict[existing_key][0], key_value)
+
 def debug_mode(choice):
     """Enable/disable printing helpful information for debugging the program. Default is off."""
     global _log
@@ -69,7 +95,7 @@ def alert(title=None, message='', ok=None, cancel=None):
     return alert.runModal()
 
 
-def notification(title, subtitle, message, data=None, sound=True):
+def notification(title, subtitle, message, data=None, sound=True, image=None):
     """Send a notification to Notification Center (Mac OS X 10.8+). If running on a version of Mac OS X that does not
     support notifications, a ``RuntimeError`` will be raised. Apple says,
 
@@ -96,6 +122,8 @@ def notification(title, subtitle, message, data=None, sound=True):
     notification.setUserInfo_({} if data is None else data)
     if sound:
         notification.setSoundName_("NSUserNotificationDefaultSoundName")
+    if image != None:
+        notification.setContentImage_(image)
     notification.setDeliveryDate_(NSDate.dateWithTimeInterval_sinceDate_(0, NSDate.date()))
     NSUserNotificationCenter.defaultUserNotificationCenter().scheduleNotification_(notification)
 
@@ -122,6 +150,10 @@ def quit_application(sender=None):
 
 def _nsimage_from_file(filename, dimensions=None):
     """Take a path to an image file and return an NSImage object."""
+
+    if filename != None and type(filename) == NSImage:
+        return filename
+
     try:
         _log('attempting to open image at {0}'.format(filename))
         with open(filename):
```

tgorgdotcom:
Hello.  I did some research on this and it seems tricky because Python 3 is not included by default on most Macs.  It does look like there's code for a native status icon.  Is there a specific reason why that isn't being used over the Python implementation?

tgorgdotcom:
I think I answered my own question actually!  The native code is using the MonoMac library, which is (in my limited research) not 64-bit enabled and outmoded by the Xamarin.Mac library.  I wonder how hard it would be to update this code...

tgorgdotcom:
Related to #4802

gpatel-fr:
The rumps code has been updated to Python 3 in canary 106

PR #4702, opened 2022-04-01
Delay creation of database directory until we know it is needed

albertony:
This avoids creating directory %LocalAppData%\Duplicati when it would just be left
empty, and it makes running portable mode (e.g. debug builds) stealthier.


warwickmm:
Thanks @albertony!b

Issue #4700, opened 2022-03-30
Not a valid Win32 FileTime


blastehh:
- [x] I have searched open and closed issues for duplicates.
- [x] I have searched the [forum](https://forum.duplicati.com) for related topics.

----------------------------------------

## Environment info

- **Duplicati version**: 2.0.6.3_beta_2021-06-17
- **Operating system**: Win10
- **Backend**: SMB via TrueNAS-12.0-U5.1
## Description

A backup creates new files that does not have any visible access time causing the backend validation part to fail.
![image](https://user-images.githubusercontent.com/8324677/160834035-b1a4381d-02f7-451a-b03a-3c35a0a290c0.png)
![image](https://user-images.githubusercontent.com/8324677/160834107-bc352abf-c43d-497f-ab74-b179efc2de4e.png)

The backup job is considered a fail and no logs are saved via the GUI.
The email report contains:
```
Failed: Not a valid Win32 FileTime.
Parameter name: fileTime
Details: System.ArgumentOutOfRangeException: Not a valid Win32 FileTime.
Parameter name: fileTime
   at Duplicati.Library.Main.BackendManager.List()
   at Duplicati.Library.Main.Operation.FilelistProcessor.RemoteListAnalysis(BackendManager backend, Options options, LocalDatabase database, IBackendWriter log, IEnumerable`1 protectedFiles)
   at Duplicati.Library.Main.Operation.FilelistProcessor.VerifyRemoteList(BackendManager backend, Options options, LocalDatabase database, IBackendWriter log, IEnumerable`1 protectedFiles)
   at Duplicati.Library.Main.Operation.BackupHandler.PostBackupVerification(String currentFilelistVolume)
   at Duplicati.Library.Main.Operation.BackupHandler.<RunAsync>d__20.MoveNext()
--- End of stack trace from previous location where exception was thrown ---
   at System.Runtime.ExceptionServices.ExceptionDispatchInfo.Throw()
   at CoCoL.ChannelExtensions.WaitForTaskOrThrow(Task task)
   at Duplicati.Library.Main.Controller.<>c__DisplayClass14_0.<Backup>b__0(BackupResults result)
   at Duplicati.Library.Main.Controller.RunAction[T](T result, String[]& paths, IFilter& filter, Action`1 method)

Log data:
2022-03-30 10:03:00 +01 - [Error-Duplicati.Library.Main.Operation.BackupHandler-FatalError]: Fatal error
[...]

PR #4703, opened 2022-04-02
Ignore exception when file access time is not valid

albertony:
When getting the value from FileSystemInfo.LastAccessTime, it will internally convert the FILETIME value from Windows API to a DateTime. If the value represents a date after 12/31/9999 it will throw ArgumentOutOfRangeException, because this is not supported by DateTime. Some file systems seem to set strange access timestamps on files, which may lead to this exception being thrown. Since the last accessed timestamp is not important such exceptions are now just silently ignored.

Issue #4700 discusses such a case, and external sources like [this](https://discourse.mcneel.com/t/parameter-is-incorrect-rhino-6/74814/48) mentions similar experiences.

I have not been able to reproduce the exact case from the issue. But I was able to "provoke" the same result: By modifying the access timestamp of the dlist.zip.aes file of an existing backup, using a C++ snippet which calls Windows API function SetFileTime, and setting a date for a year after 9999 (in my case 09/01/‎20840), when I later trigger the verify operation in Duplicati on the backup the result is identical ("System.ArgumentOutOfRangeException: Not a valid Win32 FileTime." etc.).

The exception is thrown from here: https://github.com/dotnet/runtime/blob/ffc5a34196aa3d0a1f5c8f2a90535d0fd8d73a7a/src/libraries/System.Private.CoreLib/src/System/DateTime.cs#L756

Now the question is if this PR really fixes the problem described in the issue. Since I have not reproduced a complete real world situation, I don't know. Also; I don't know if it is *one* problem - there are several users describing similar problems commenting on the issue, and also in a forum thread, but does all of them have the same cause? I guess the only way to find out is to wait for the affected users to test their cases on a beta build... And I think this change is relevant to keep as a precaution, even it should proove to not solve all reported issues.

PS: I don't know how your policy is regarding comments. I chose the verbose alternative; included more or less the same description as in the commit message as code comment, but if you are on the opposite stand just give me a hint and I can remove the comment, or just reduce it to a one-liner (or, of course, feel free to change it yourself as you like).

### DIFF
--- a/Duplicati/Library/Common/IO/SystemIOWindows.cs
+++ b/Duplicati/Library/Common/IO/SystemIOWindows.cs
@@ -508,7 +508,19 @@ namespace Duplicati.Library.Common.IO
         public IFileEntry FileEntry(string path)
         {
             var fileInfo = new FileInfo(AddExtendedDevicePathPrefix(path));
-            return new FileEntry(fileInfo.Name, fileInfo.Length, fileInfo.LastAccessTime, fileInfo.LastWriteTime);
+            var lastAccess = new DateTime();
+            try
+            {
+                // Internally this will convert the FILETIME value from Windows API to a
+                // DateTime. If the value represents a date after 12/31/9999 it will throw
+                // ArgumentOutOfRangeException, because this is not supported by DateTime.
+                // Some file systems seem to set strange access timestamps on files, which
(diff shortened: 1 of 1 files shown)


warwickmm:
Thanks @albertony for your help debugging this!

PR #4704, opened 2022-04-07
Bypass auto-update when running in a debug environment

AxeOfMen:
Debugging the source won't work unless `AUTOUPDATER_Duplicati_SKIP_UPDATE=1` is set as an environment variable. This is not an obvious step for those new to the codebase. This change would check whether the code is running in a debugger and bypass the auto-update feature allowing users to debug the code.

### DIFF
--- a/Duplicati/Library/AutoUpdater/UpdaterManager.cs
+++ b/Duplicati/Library/AutoUpdater/UpdaterManager.cs
@@ -24,6 +24,7 @@ using System.Threading.Tasks;
 using Duplicati.Library.Interface;
 using Duplicati.Library.Common.IO;
 using Duplicati.Library.Common;
+using System.Diagnostics;
 
 namespace Duplicati.Library.AutoUpdater
 {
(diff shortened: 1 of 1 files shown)


warwickmm:
Thanks @AxeOfMen!

ts678:
Thanks for the change. I just started trying out the Visual Studio debugger, got frustrated, but didn't bypass its failing yet.

> This is not an obvious step for those new to the codebase.

Yes, for almost all people using a debugger, this seems more likely to land the right way. See bottom for a caveat though.

Even for those not-so-new, Visual Studio 2022 seemingly and surprisingly (to newbie me) can't set environment variables.

[Project settings for C# debug configurations (.NET Core, .NET 5+, and ASP.NET Core)](https://docs.microsoft.com/en-us/visualstudio/debugger/project-settings-for-csharp-debug-configurations-dotnetcore?view=vs-2022) Here you have it (but doesn't apply).
[Project settings for C# debug configurations](https://docs.microsoft.com/en-us/visualstudio/debugger/project-settings-for-csharp-debug-configurations?view=vs-2022) Here you don't.

Possible answer and confusing talk about SDK-projects:

[Debugging program with environment variables](https://developercommunity.visualstudio.com/t/debugging-program-with-environment-variables/353814)
[Debugging program with environment variables #5667](https://github.com/dotnet/project-system/issues/5667)

Although I don't know if I'll ever try it, what's the easiest tested recipe to run through auto-update code under debugger?
If there's no easy way, I wonder if the bypass should only be done if, for example, the environment variable does not exist.

warwickmm:
> ... I wonder if the bypass should only be done if, for example, the environment variable does not exist.

Yes, this is a better approach.  @AxeOfMen, care to make this change?

Issue #4608, opened 2021-09-20
Empty Popup "dialog" opening web UI


kahomono:
- [X] I have searched open and closed issues for duplicates.
- [X] I have searched the [forum](https://forum.duplicati.com) for related topics.

----------------------------------------

## Environment info

- **Duplicati version**: 2.0.6.3_beta_2021-06-17
- **Operating system**: Linux
- **Backend**: N/A - UI issue

## Description

When opening the web UI on a given instance for the first time in a while, get an empty popup dialog.  Need to refresh the browser tab to dismiss it.

## Steps to reproduce

1. Freshly open browser tab to Duplicati web GUI
2. Happens on both localhost and remote instances

- **Actual result**:
![Screenshot at 2021-09-20 03-54-37](https://user-images.githubusercontent.com/20188009/133973000-abe5bc4f-92ad-4ca5-b4f8-4834c68ada24.png)

- **Expected result**:
![Screenshot at 2021-09-20 04-10-37](https://user-images.githubusercontent.com/20188009/133973018-3d9a399d-81c2-4161-bd0c-281d6b2f698a.png)

duplicatibot:
This issue has been mentioned on **Duplicati**. There might be relevant details there:

https://forum.duplicati.com/t/empty-popup-on-opening-web-gui/12670/4

warwickmm:
I can't reproduce this with Brave 1.29.80 (Chromium: 93.0.4577.63) on Linux (Mono 6.12.0).  Do you have any advanced settings configured?

kahomono:
Using Firefox 78.14.0esr on Debian 11.  I have the SimpleTabGroups extension and I encounter this whenever I re-open my Duplicati group, which has three servers' UIs in it.  I get it on every one.

Not sure what advanced settings you might be interested in?

warwickmm:
> Not sure what advanced settings you might be interested in?

Anything that's not the default.

warwickmm:
Based on @ts678's [observation](https://forum.duplicati.com/t/empty-popup-on-opening-web-gui/12670/4), I noticed that if I perform a "right click" -> "Duplicate" in Brave, I do occasionally encounter the following:

![image](https://user-images.githubusercontent.com/17345532/134054584-7b75433a-92c6-46ff-9dc3-0a45b7e0982e.png)

ts678:
[...]

PR #4705, opened 2022-04-08
Issue #4608: Empty Popup "dialog" opening web UI

AxeOfMen:
Add additional condition to prevent empty dialog from displaying in Firefox when duplicating a Duplicati tab.

### DIFF
--- a/Duplicati/Server/webroot/ngax/index.html
+++ b/Duplicati/Server/webroot/ngax/index.html
@@ -283,9 +283,9 @@
     <notification-area></notification-area>
 
     <div ng-controller="DialogController">
-        <div ng-show="state.CurrentItem != null" class="blocker" ng-click=""></div>
+        <div ng-show="state.CurrentItem != null && state.CurrentItem.title != null" class="blocker" ng-click=""></div>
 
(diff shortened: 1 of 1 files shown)


warwickmm:
:+1:

ts678:
Thanks for the change and the merge. I was just wishing a few days ago that I had this. I had two Duplicati running, and Firefox is where I run the second one (because I think putting two Duplicati in one browser causes them to trip up on each others cookies).

RalfRoemling:
Thanks also from my side but I don't know how to implement it in my Duplicati-installation. Anybody can give me a hint?

ts678:
> I don't know how to implement it in my Duplicati-installation

If your desire is to not wait for a release, the file is in your Duplicati installation folder, and you edit the two changed lines.
Installation folder varies by OS. If your main installation is an old version, it's probably safest to reinstall it then do editing.
About --> System info shows BaseVersionName (latest OS-style installation) and ServerVersionName after an autoupdate.

RalfRoemling:
@ts678 thx a lot. Have been blind. 😄 
Yeah sure, it were only those two lines. 🤦‍♂️

Issue #4645, opened 2021-12-02
Backup gets stuck executing a query


AxeOfMen:
- [x] I have searched open and closed issues for duplicates.
- [x] I have searched the [forum](https://forum.duplicati.com) for related topics.

----------------------------------------

## Environment info

- **Duplicati version**: 2.0.6.1_beta_2021-05-03
- **Operating system**: Windows Server 2008 R2 Standard 64-bit (6.1, Build 7601) Service Pack 1 (7601.win7sp1_ldr.170427-1518)
- **Backend**: Minio on local machine
- **Processor**: Intel(R) Pentium(R) CPU G3258 @ 3.20GHz (2 CPUs), ~3.2GHz
- **Memory**: 8192MB RAM
- **.NET version**: 4.7.03062

## Description

[Forum post](https://forum.duplicati.com/t/query-during-backup-takes-too-long-does-not-finish/13416)
I have a large backup, over 300 gigs. The initial backup succeeded. Subsequent backups never complete. They get stuck on this query:

```
[...]

PR #4706, opened 2022-04-09
Add indexes to improve performance

AxeOfMen:
Add performance index

### DIFF
--- a/Duplicati/Library/Main/Database/Database schema/13. Performance Indexes.sql
+++ b/Duplicati/Library/Main/Database/Database schema/13. Performance Indexes.sql
@@ -1,0 +1,4 @@
+﻿CREATE INDEX "FileLookupMetadataID" ON "FileLookup" ("MetadataID");
+CREATE INDEX "FilesetEntryFilesetID" ON "FilesetEntry" ("FilesetID");
+
+UPDATE "Version" SET "Version" = 13;
--- a/Duplicati/Library/Main/Database/Database schema/Schema.sql
+++ b/Duplicati/Library/Main/Database/Database schema/Schema.sql
@@ -289,4 +289,4 @@ CREATE TABLE "ChangeJournalData" (
     "ConfigHash" TEXT NOT NULL  
 );
 
-INSERT INTO "Version" ("Version") VALUES (12);
+INSERT INTO "Version" ("Version") VALUES (13);
--- a/Duplicati/Library/Main/Database/Database schema/Schema.sql
+++ b/Duplicati/Library/Main/Database/Database schema/Schema.sql
@@ -82,6 +82,7 @@ CREATE TABLE "FilesetEntry" (
 /* Improved reverse lookup for joining Fileset and File table */
 CREATE INDEX "FilesetentryFileIdIndex" on "FilesetEntry" ("FileID");
 CREATE INDEX "nn_FilesetentryFile" on FilesetEntry ("FilesetID","FileID");
+CREATE INDEX "FilesetEntryFilesetID" ON "FilesetEntry" ("FilesetID")
 
 
 
@@ -112,6 +113,7 @@ CREATE TABLE "FileLookup" (
 /* Fast path based lookup, single properties are auto-indexed */
 CREATE UNIQUE INDEX "FileLookupPath" ON "FileLookup" ("PrefixID", "Path", "BlocksetID", "MetadataID");
 CREATE INDEX "nn_FileLookup_BlockMeta" ON FileLookup ("BlocksetID", "MetadataID");
+CREATE INDEX "FileLookupMetadataID" ON "FileLookup" ("MetadataID");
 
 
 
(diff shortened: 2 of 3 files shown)


jedthe3rd:
Hey, can you validate that adding :
CREATE INDEX "FilesetEntryFilesetID" ON "FilesetEntry" ("FilesetID");
Improves the performance over the indexes already in Performance Indexes.sql.

This index: CREATE INDEX "nn_FilesetentryFile" on FilesetEntry ("FilesetID","FileID"); should meet the same responsibility as your FilesetEntryFilesetID index.

Also any new database changes need to be in their own subsequent .sql file. So you should add your changes to a 13. Additional index.sql or something as any existing databases created with the current canary would need to have another version so they know to apply the new indexes.

AxeOfMen:
Thanks for the feedback. I have moved the indexes to a separate file and incremented the db version to 13.

Yes, as mentioned in the comments for [Issue 4645](https://github.com/duplicati/duplicati/issues/4645), these new queries are necessary. I figured your nn_FilesetentryFile index would cover the same ground as my FilesetEntryFilesetID index, but for whatever reason, mine is necessary as well. I'm curious whether you see this using the query I posted in 4645 against your 10G db.

AxeOfMen:
@jedthe3rd  Running `ANALYZE;` after your indexes eliminates the need for my indexes. Should I change my pull request to just do an analyze? It seems a little silly, but otherwise the next analyze won't occur until the end of the next run when [the database connection is closed](https://github.com/duplicati/duplicati/blob/master/Duplicati/Library/Main/Database/LocalDatabase.cs#L1379). That could take hours. In my case, I waited for weeks and the query never finished. Certainly worse in your case with a 10G db.

jedthe3rd:
> @jedthe3rd  Running `ANALYZE;` after your indexes eliminates the need for my indexes. Should I change my pull request to just do an analyze? It seems a little silly, but otherwise the next analyze won't occur until the end of the next run when [the database connection is closed](https://github.com/duplicati/duplicati/blob/master/Duplicati/Library/Main/Database/LocalDatabase.cs#L1379). That could take hours. In my case, I waited for weeks and the query never finished. Certainly worse in your case with a 10G db.

This is what I was thinking was the case. Ya, I definitely think we should run analyze before each backup. I actually made this change in my code but haven't gotten around to creating the pull request because I am working on a few changes I wanted to batch.

Just change this request to add analyze to run before the backup runs and that should be good. This should have a minimal time impact overall. Can you also see if you can update the backup status to show it's "optimizing the database" as a step. I think that would be a good addition to this change. If it seems like a bunch of work then we can leave that for a different time.

AxeOfMen:
I'm not in love with the idea of analyzing before each run. It's not best practice and could set a bad precedent for others looking to do something similar. While analyzing before each run may not be excessive overhead, ideally the analyze would occur immediately after adding the index. But since your PR has already been released to canary, I think that ship has sailed. As goofy as it looks, I think adding version 13 to do an analyze is the most viable option. Hmm.

jedthe3rd:
> I'm not in love with the idea of analyzing before each run. It's not best practice and could set a bad precedent for others looking to do something similar. While analyzing before each run may not be excessive overhead, ideally the analyze would occur immediately after adding the index. But since your PR has already been released to canary, I think that ship has sailed. As goofy as it looks, I think adding version 13 to do an analyze is the most viable option. Hmm.

Ya, I didn't think that thought all the way through. Actually the best bet is probably to add an analyze after a database is upgraded here: https://github.com/duplicati/duplicati/blob/master/Duplicati/Library/SQLiteHelper/DatabaseUpgrader.cs
Then once we release the next version of changes an analyze is a part of the process of the upgrade. I have a number of SQL changes I need to finish testing so we can have the next db version here fairly soon.

kenkendk:
@AxeOfMen & @jedthe3rd Do you want to include the changes here, or did you have a better fix in mind?

jedthe3rd:
> @AxeOfMen & @jedthe3rd Do you want to include the changes here, or did you have a better fix in mind?

These indexes are basically duplicates. The real solution to the issue he is facing is to add an analyze when the database is upgraded.

Duplicati/Library/Main/Database/Database schema/Schema.sql
```diff
@@ -82,6 +82,7 @@ CREATE TABLE "FilesetEntry" (
 /* Improved reverse lookup for joining Fileset and File table */
 CREATE INDEX "FilesetentryFileIdIndex" on "FilesetEntry" ("FileID");
 CREATE INDEX "nn_FilesetentryFile" on FilesetEntry ("FilesetID","FileID");
+CREATE INDEX "FilesetEntryFilesetID" ON "FilesetEntry" ("FilesetID")
```
mr-russ, line 85: The line above has an index on the same leading column.  The cost of maintaining this index will be higher shan using the original one as it will skipscan.

Is there a specific reason for this index or crn it be removed?

Duplicati/Library/Main/Database/Database schema/13. Performance Indexes.sql
```diff
@@ -0,0 +1,4 @@
+﻿CREATE INDEX "FileLookupMetadataID" ON "FileLookup" ("MetadataID");
```
mr-russ, line 1: I think there is value in adding sql comments indicating the operations this is known to improve so others working on this later can see the reason for the addition.

mr-russ (commented):
Is this still relevant in the same way, the comments on the pull request seem to indicate the addition of analyze on upgrade would be useful.  That is true, should that be part of this chrnge?

Also note that vacuum appears to be called during operation startup as well, I cant recall without looking it uf if that does any analysis work.

Issue #4666, opened 2022-01-24
DuplicatiVerify.py generates incorrect missing file errors


dm413:
- [x] I have searched open and closed issues for duplicates.
- [x] I have searched the [forum](https://forum.duplicati.com) for related topics.

----------------------------------------

## Environment info

- **Duplicati version**: 2.0.6.3_beta_2021-06-17
- **Operating system**: Windows 10
- **Backend**: Local USB drive

## Description

DuplicatiVerify.py does not correctly handle entries in the `duplicati-verification.json` which have a State of 5. It reports these files as "missing", while a state of 5 corresponds to "deleted", and the file is no longer present.

This was first reported in forum post [Client Verification File Contains Expired Files](https://forum.duplicati.com/t/client-verification-file-contains-expired-files/13459/6). There is a proposed code fix in that discussion that resolved the problem for the OP and for me as well. 

This is related to issue #2989. That issues reported and fixed the problem for the powershell script DuplicatiVerify.ps1, but did not address the issue for the python script DuplicatiVerify.py.

## Steps to reproduce

1. Perform a backup with the --upload-verification-file  option set true.
2. Repeat until you get a `duplicati-verification.json` file that contains a file with `State` = 5. I don't know how to force that, but it happens occasionally. The OP indicates that it happens when files are deleted due to the file retention policy. 
3. Notice that the file with `State` 5 is not actually present in the back end.
3. Run DuplicatiVerify.py in a python interpreter (at the back end).

- **Actual result**:
The script reports a missing file and list the filename of the file with `State` 5.
- **Expected result**:
No error message.

duplicatibot:
This issue has been mentioned on **Duplicati**. There might be relevant details there:

https://forum.duplicati.com/t/client-verification-file-contains-expired-files/13459/8

PR #4707, opened 2022-04-09
fix for handling state 5 in DuplicatiVerify.py

mr-flibble:
fix for issue https://github.com/duplicati/duplicati/issues/4666
Same logic as at powershell verify file (at line 78 in ps1)
Credits for code ["ts678" in Duplicati forum](https://forum.duplicati.com/t/client-verification-file-contains-expired-files/13459/2)
I tested it and test output fir json with error 5 is now same as in powershell verify


warwickmm:
Thanks @mr-flibble!

PR #4712, opened 2022-04-21
Added a new s3 provider and some other minor changes

MuhammadAbrar1984:
**1. Added a new s3 provider**
_Found IDrive e2: the cheapest cloud storage in the market with no fees for egress. We use duplicati for backups so tested extensively with duplicati and found it to be way faster than other s3 providers like Backblaze B2, Wasabi etc. It deserves a place in this amazing tool_

**2. Made "Use SSL" check checked by default on S3 form**
_The user would rarely be interested in HTTP. Even a few providers like e2 do not support this._

**3. Renamed label "AWS access ID"  to "Access Key ID" on s3 form**
 _"AWS Access ID" is confusing for non-aws s3 user. "Access Key Id" is a more appropriate term and all providers even AWS itself use this._ 

**4. Renamed "AWS Access Key" to "Secret Access key"** 
_"AWS Access Key" is confusing for non-aws s3 user. "Secret Access key" is a more appropriate term and all providers even AWS itself use this._ 

**5. Fixed AppVeyor build** 
_Removed an unnecessary bindingRedirect statement to fix UnitTest on AppVeyor_ 

It's my first ever contribution to any open source and this would be really an honor for me if I could become a contributor to this great tool. 

Thanks,
Abrar


MuhammadAbrar1984:
@kenkendk
FYI

PR #4713, opened 2022-04-22
show progress when restoring files

flawiddsouza:
Show this:
image
Instead of this:
image


warwickmm:
Thanks @flawiddsouza.

PR #4699, opened 2022-03-29
Jottacloud backend oauth authentication

albertony:
Changes Jottacloud from basic username/password authentication to OAuth authentication, via authid and the oauth handler service.

**This is a breaking change**: Users must modify their backend configuration. But it is really a breaking change from Jottacloud's part: They recently discontinued username/password authentication, but agreed to re-enable it for a limited time until June 1st 2022 after request from Duplicati users.

**Depends on changes in the oauth handler**, which must be merged first: https://github.com/duplicati/oauth-handler/pull/7.

Fixes #4697.

Not done:
- Update manual (https://duplicati.readthedocs.io/en/latest/05-storage-providers/#jottacloud)

PR #4715, opened 2022-04-24
Jottacloud upload implementation changed to use new api

albertony:
Note: Builds on top of #4699, which adds oauth authentication. Will rebase if/when that gets merged.

This PR changes upload implementation from an old API into the same that current official client (and rclone) uses. The new api has built-in support for deduplication and resuming.

Includes an additional commit that avoids changing the type when using an existing custom device in Jottacloud.

### DIFF
--- a/Duplicati/Library/Backend/Jottacloud/Jottacloud.cs
+++ b/Duplicati/Library/Backend/Jottacloud/Jottacloud.cs
@@ -20,6 +20,7 @@
 using Duplicati.Library.Common.IO;
 using Duplicati.Library.Interface;
 using Duplicati.Library.Localization.Short;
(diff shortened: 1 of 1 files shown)


kenkendk:
@albertony PR #4699 is now merged.

albertony:
Great news, thank you!
Rebased this PR, but I may need to work on it a bit more, at least do some manual testing, i.e. still work in progress.

kenkendk:
Not sure if this should be closed? I think we can update Jottacloud to the new API, but the backend was rewritten, so this PR is hard to continue on.

kenkendk:
Closing for lack of activity

PR #4730, opened 2022-06-01
Added statistics to DuplicatiVerify.py

mr-flibble:
Hello, this commit is adding new statistics feature to DuplicatiVerify.py. Statistics will be displayed after verification is complete. The **goal was for the python version to have the same functionality as Powershell version** ([DuplicatiVerify.ps1](https://github.com/duplicati/duplicati/blob/master/Tools/Verification/DuplicatiVerify.ps1)) without any possible problematic changes to original Python code. 

About code:

- no new import modules dependency
- no changes to original verify code (except new optional parametr)
- Statistics will start after verify task is done - so a potential problem can't affect verifying
- Unlike the Powershell version of the statistics, the "total size" is now displayed and the units (MB,GB,TB) are automatic

I tested it on multiple computers (python 3.9.6, 3.3.4 etc) and it works fine. The displayed statistics also match - the numbers are the same as in the Powershell version

Images:

**Original [DuplicatiVerify.ps1](https://github.com/duplicati/duplicati/blob/master/Tools/Verification/DuplicatiVerify.ps1)** statistics output:
image

**New DuplicatiVerify.py** statistics:
image

### DIFF
--- a/Tools/Verification/DuplicatiVerify.py
+++ b/Tools/Verification/DuplicatiVerify.py
@@ -84,22 +84,142 @@ if __name__ == "__main__":
     parser.add_argument("path", type=str, nargs='?',
                     help="""path to the verification file or folder containing the verification file.\
                         Defaulf is curent path""")
+    parser.add_argument("--stats_only", action='store_true', help="display statistics from json only, no verification")
     args = parser.parse_args()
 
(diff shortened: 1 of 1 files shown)


kenkendk:
Looks nice!

PR #4749, opened 2022-06-16
Improved warning message about Hyper-V VSS Writer

Germs2004:
Minor grammar improvements, and added the actual class id for Hyper-V VSS Writer that is needed to disable the warning from happening every time.


Germs2004:
I didn't actually test this to ensure it displays correctly.  It increases the length of the displayed text, so please make sure it all fits and looks as expected before accepting.

PR #4754, opened 2022-06-21
Jottacloud use username from token

albertony:
Refactoring and small performance improvement. See commit messages for details.

This was done after debugging on the issue we have with the new oauth authentication in the Jottacloud backend (stale token in oauth service). These changes reduces the number of requests to the oauth service, hence the frequency of usage of the access token. If the stale token is caused by some race condition, perhaps this may have a positive effect, although I don't really have high expectations... I think the change is an improvement regardless.


kenkendk:
Closing this as it can no longer be applied. I think we may have fixed some of the underlying issues in the OAuth handler, where it detects a new refresh token being issued and using that going forward.

PR #4755, opened 2022-06-21
[Storj DCS] Update to latest uplink.NET

TopperDEL:
This PR updates the Storj DCS (and the depreciated Tardigrade) backend(s) to the lates uplink.NET v2.9.2710. There is no functional change, but it now contains the latest updates regarding performance and security.


TopperDEL:
Can anyone help me with this PR? Anything I can do regarding the failing Tests?

DeflateAwning:
Please merge this PR, as it may help fix other issues with Storj.

TopperDEL:
I've just updated this PR to the latest uplink.NET-version.

gpatel-fr:
Hello @TopperDEL 

I have a bit of trouble with merging this PR, the Windows build generates about 300 warnings, along the lines of:
```
2023-03-10T11:01:16.7783844Z   C:\Program Files\Microsoft Visual Studio\2022\Enterprise\MSBuild\Current\Bin\Microsoft.Common.CurrentVersion.targets(2352,5): warning MSB3277: Found conflicts between different versions of "System.Numerics.Vectors" that could not be resolved. [D:\a\testdupl\testdupl\Duplicati\Library\Backend\Storj\Duplicati.Library.Backend.Storj.csproj]
2023-03-10T11:01:16.7785957Z C:\Program Files\Microsoft Visual Studio\2022\Enterprise\MSBuild\Current\Bin\Microsoft.Common.CurrentVersion.targets(2352,5): warning MSB3277: There was a conflict between "System.Numerics.Vectors, Version=4.1.3.0, Culture=neutral, PublicKeyToken=b03f5f7f11d50a3a" and "System.Numerics.Vectors, Version=4.1.4.0, Culture=neutral, PublicKeyToken=b03f5f7f11d50a3a". [D:\a\testdupl\testdupl\Duplicati\Library\Backend\Storj\Duplicati.Library.Backend.Storj.csproj]
2023-03-10T11:01:16.7787383Z C:\Program Files\Microsoft Visual Studio\2022\Enterprise\MSBuild\Current\Bin\Microsoft.Common.CurrentVersion.targets(2352,5): warning MSB3277:     "System.Numerics.Vectors, Version=4.1.3.0, Culture=neutral, PublicKeyToken=b03f5f7f11d50a3a" was chosen because it was primary and "System.Numerics.Vectors, Version=4.1.4.0, Culture=neutral, PublicKeyToken=b03f5f7f11d50a3a" was not. [D:\a\testdupl\testdupl\Duplicati\Library\Backend\Storj\Duplicati.Library.Backend.Storj.csproj]
2
```
I don't include them all but you can find the build here:

https://github.com/gpatel-fr/testdupl/actions/runs/4383803389

I don't have much time for Duplicati this afternoon so if you have a ready made fix it would be welcome.
If not I have all this week end to investigate it :-)

gpatel-fr:
I have done some progress since my previous comment, I now see that it's a DLL hell and it involves far less warnings that I thought at first - it was the rather difficult to read format of the log output by Github build that made me misread the problem scope - but nonetheless I don't like that the build involves more warnings without clear understanding of the cause. Still digging.

gpatel-fr:
I created a free account with storj.io, followed the link to documentation:

https://docs.storj.io/dcs/how-tos/backup-with-duplicati

that unfortunately get nowhere, read a web.archive.org version, tried to fill the api key, the encryption passphrase (why is it necessary to enter an encryption passphrase before getting an api key, then entering again the passphrase after entering the api key in the third party client, is totally clear as mud to me...), entered the bucket I created, no directory, clicked on test and got invalid macaroon version. Oh well. Is it supposed to work with current Duplicati ?

gpatel-fr:
While I could not make head or tail from the authentication methods 'Authentication method()' and 'Api key', I think I got to grant an access on Storj.io site and get an 'access credentials' that I then pasted in the 'Access grant' of the Duplicati Storj config dialog, entered the bucket name I had created on the Storj site, then either with or without a directory name, and either with Canary 104 or my version compiled with this PR got the same answer after a full minute or so: 'Failed to connect. One or several errors did happen'. 

Note as a double failure, the error message 'One or several errors did happen' was displayed in the native language of the Windows installed, not in English - while I had asked for the Duplicati language to be set to English, suggesting that the backend user interface is localized independently fro the Duplicati settings. 

Whatever, in any case I'll try again next day but no more than that.

gpatel-fr:
Finally success. With the BackendTool I noticed that it was a plain connection denial on port 7777. I opened the firewall port and things got better immediately. I noticed also that while I had registered on the EU node, the default connection was on US1 - but I don't think it was the reason since I did not edit the backup in any way and it succeeded. Since it was with the new version, the outlook of including Storj is quite better. Just have to remove the warning by doing a BindingRedirect or something like that and it should be good to ship.

DeflateAwning:
Can confirm that backups via Duplicati (2.0.6.104_canary_2022-06-15) do work when it's all setup correctly!

gpatel-fr:
After the false joy of doing a very very small backup that succeeded, this morning I tried a more normal size backup and it failed. 
Note that I used the command line as I struggle to configure the UI. But the command line I used works (more or less) with Duplicati Canary 104, and fails with the new build. 
I tried your previous version of this PR and got the same problem. There is no error reported, the command ends seemingly without displaying anything (however it's necessary I hit Enter to show a new prompt, it is not shown by itself suggesting a bad end without any backtrace...), and trying to restart a backup displays that a file is missing. In the Storj interface I see indeed 2 index files and one block file.

now, under Linux (Ubuntu 20.04 LTS), using the June version of your PR, I see no problem. A backup (still with command line) seems to succeed, so maybe the build warning under Windows is indeed showing a real issue.

DeflateAwning:
You may have discovered this open issue: https://github.com/duplicati/duplicati/issues/4832

TopperDEL:
Hi all! Sorry, on weekend is family time so I could not get back to this topic.

The update on uplink.NET should not lead to any (new) issues with backups and such. Though I would do some tests again on my end. I will also have a look at #4832.
But unfortunately my son is sick currently, so I don't have that much spare time. Will come back to this ASAP though - hope until the end of this week.

gpatel-fr:
@DeflateAwning 
very much doubt it. All my tests are done on the same computer (so same network) with an assortment of VM, containers. All succeed with Linux with backups of 50-100 Mb. All fails under Win10 even with backup size as low as 50 KB.

@TopperDEL 
Thanks for answering, hope it will get better for you.
I'll try to find the explanation in the mean time.

gpatel-fr:
I have now come back on this. I realized (belatedly) that the 'Api key' config is not really functional, so I am basing all tests now on the 'Access grant' config. I have regular failings on Windows and sometimes on Linux on my system. I have also tried to test on another Windows system on a faster network (fiber). I have found that it works correctly on this network, so performance (speed) of Internet access matters. To confirm that, I got success on the slow network by invalidating parallel transfers.
I have noticed that in the driver code there is no logging, and no testing for lack of success in uploading. Also I wonder about the Dispose() code. Not sure if the call to Dispose in ResetBackend when there are network failures could be the reason for #4832.
Whatever, I am now building for a release without Storj update. If it does not work, I'll know that it is an interaction with the other changes - or the building platform itself.

gpatel-fr:
Right, it's time for me to move on.
I  have come to the conclusion that the warnings are not the worst problem.
I have found this ref on the Web:
https://github.com/dotnet/runtime/issues/45921
that seems related. If it is the case, there is not much to be done, short of using a different software not  needing these libraries.
I have tried to update libraries, without much success. Sql drivers can be updated but don't change anything, I have not succeeded in finding a combination of other libraries that could be updated successfully.

About the main problem, the backup uploading all blocks without any apparent error but Duplicati finding all blocks missing (the only files being uploaded successfully to Storj being the small files, index and dlist), the only mitigation I have found is to disable parallel upload and to get a faster link, hardly a satisfying fix.

In short, I think it is better to not update the driver at the moment, since the old one is at least working. I'll be busy now to get back to cut a new release and hope it get better for you in the meantime.

gpatel-fr:
Well, I guess that's all water under the bridge now. 
Anyway, I have finally understood that with Duplicati being based on C# and relying often (directly or undirectly) on the HttpClient implementation, it's basically useless to test any backend based on Http - such as Storj - on a slow link. The default Http timeout in Ms implementation is 100s, and it can happen with some backends that the HttpClient is created under the hood inside the driver with this default timeout. If the network link is slow, and the driver puts large objects, again something the Duplicati backend may not have much control short of setting the dblock size to very low values, timeouts will happen often - and will be badly handled by Duplicati to compound the misery. 

I should have known that and not tried at all to test the Storj driver on a slow link, the only meaningful test was the one I did on a fiber link. I may come back to try to integrate this PR into next Canary.

DeflateAwning:
Duplicati works fine for me on a very slow link. Are you suggesting that this PR breaks that compatibility?

gpatel-fr:
@DeflateAwning 

what do you mean exactly by 'slow' ? By my rough calculations, a packet from Minio (16 Mb size) will upload in about 120 s with Adsl (about 125 kb /s). It's far beyond what can run correctly. If I slow down my network to only 10Mbits/, Minio runs fine for small backups. Note that I am talking about Minio because that's what I am testing currently, both Minio and Storj are Http based backups so can suffer from timeouts, but Storj can have smaller packets, or your upload can be a bit faster than Adsl.

DeflateAwning:
Slow as in 7.5 mbps. It has worked very reliably for me even on that slow of internet, even for rather large backups. 

A known issue is that the first time a new backup is completed, it must be very small or it fails. After that, subsequent backups (with more added files) succeed fine.

gpatel-fr:
IMO 7.5 Mbits/s upload is marginal. Yes it should work most of the time. Do the math; with a 16 Mbytes Http put, that means about 17 s, a bit less than 20% of the default timeout. However the margin is not wide, so small problems (air borne problem - magnetism, humidity - or network congestion or config problems at the FAI level or whatever) could push it in danger territory.  Note that I don't know the Put size with Storj. It's not the total backup size that matters, that's the size of a single Put operation.

TopperDEL:
Sorry, I did not yet find the time to investigate if there is any trouble with the new version provided in this PR. It should not change anything to the features itself, it just updates the library. I still want to do some in-depth checks especially for the problems with disrupted backups as mentioned above.

I'm not sure if you can say that the Storj backend is HTTP-based. It uses a completely native Go-based library under the hood using it's own protocolls, so I'm not sure if HttpClient timeouts and such apply here. Furthermore, backing up X MB to storj using Uplink.NET means some overheade (expansion factor) for erasure coding and encryption stuff - so in the end it might upload more that you would expect from the raw backuped data.

I still have this on my list an hope to come back to it soon! Sorry for the delay!

TopperDEL:
I just updated the Storj DCS-backend to the latest version of upink.NET 2.10.3175. But now we have a different issue: uplink.NET is now based on net6 and will switch to net8, soon. So it is not compatible to netframework anymore.

The backend will still work as-is with Storj DCS itself, but I won't be able to update until Duplicati uses a newer .Net-version.

DeflateAwning:
How hard is it to upgrade to v8? Were there many breaking changes?

gpatel-fr:
Hello

I had given up on Storj and was preparing to write a deprecation warning for next Canary, however going to nuget.org I see that last version (2.12.33) is supposed to support Net 4.7 and Net standard. If true, it's sad because now it's too late to retest for next Canary.

DeflateAwning:
I really hope that Duplicati doesn't deprecate support for Storj

gpatel-fr:
@DeflateAwning
Don't take my post as a sign of hostility, I don't understand where is going uplink.net in the short term now, that's all.
I have done a quick test of storj with next Canary and it has worked, however it's still a pathetic 2.7 driver (and there are known problems (issues) when the amount of data is significant with this version)

TopperDEL:
Storj/uplink.NET is updated regularly when the underlying uplink-c (from storj itself) gets updates. The latest updates where to support Mac Catalyst and took me some tries to get it working.
My latest updates to duplicati regarding uplink.NET generated problems around the .Net-version. I don't remember the details, but I guess as I switched to .Net6 for uplink.NET it was not longer supported from Duplicati as Duplicati still uses .Net framework. But my knowledge around .Net and the different versions is not that deep.

From my point of view Storj DCS must never be deprecated. I would do all to further support it. I'm using it by myself on all of my machines for my personal backups.

Older versions of uplink.NET should still work, though. The Storj-network itself did not change that much and is backwards compatible. So if the switch to .Net6/8 from uplink.NET is still a problem, we could simply use an older Nuget from uplink.NET for Duplicati.

gpatel-fr:
I did a quick test with 2.9.2958, downloading the Linux dll from nuget (renaming it to libstorj_uplink.so) and it seems to upload a tiny backup. 
why did you upgrade to net6 from netstandard2.0 ? I was thinking that any 'modern' net application can use a netstandard library (that's the point of building for netstandard after all). Or did you need a specific function that is not available in netstandard ?
also, why were you uploading manually the native libraries ? Was it not possible to automate copying directly from the nuget loaded binaries to the bin/Debug directory with some msbuild magic ?

TopperDEL:
> why did you upgrade to net6 from netstandard2.0 ? 

To be honest: I thought this would bring the features of the higher .Net-versions to my library, first of all higher speeds. And it always complained that uplink.NET was meant for .Net Framework and "might not work as expected". But you maybe right: NetStandard might have been ok, still.

> also, why were you uploading manually the native libraries ? Was it not possible to automate copying directly from the nuget loaded binaries to the bin/Debug directory with some msbuild magic ?

If I remeber correctly the files did not get added to the installer this way. They needed to be present for the installer to pick them up.

gpatel-fr:
>>NetStandard might have been ok

well, for next Canary (not the one planned for before year end, the next one after that) I'll try to test the last netstandard version, if you can investigate getting back to Netstandard and see if your Net (core,5...) users can live with it it would be great also. Upgrading Duplicati to 'modern' Net has been planned for years when it still had a working maintenance team, and it never got anywhere, so it's unlikely to happen real soon now.
I have more pressing matters for beginning of next year. Maybe someone else will fork Duplicati and do it, but I'm not holding my breath for that.

>>the files did not get added to the installer this way.

if you look at what is doing the build-release.sh file, circa line 188, some files (alphavss) are added manually, if needed the install script could be enhanced in the same way - IMO it would be far less pain than having to think to upload the right files at the same place everytime.

ts678:
> NetStandard might have been ok, still.

As a non-C# developer, I see [https://www.nuget.org/packages/uplink.NET#supportedframeworks-body-tab](https://www.nuget.org/packages/uplink.NET#supportedframeworks-body-tab) claims a lot, but see:

image

but I don't want to make any estimate on how awkward it will be to actually make everything work. To be continued, apparently.

gpatel-fr:
@ts678 

it comes probably from the uplink.NET.nuspec file. A possible guess could be that the netstandard 2.0 claim comes before the net6.0 one, so that's the one taken in account by nuget.org . The csproj file claims only net6 however, so the project configuration looks a bit strange.

gpatel-fr:
>> I've reverted back to NetStandard2.0

very nice, thank you !
Unfortunately it's too late for next Canary. I was intending for next Beta to come fast after Canary, but finally I may search for a next Canary in January to get a few more fixes and among them an upgrade to Storj  :-)

gpatel-fr:
The update is [here](https://github.com/duplicati/duplicati/pull/5104), it should (hopefully) obsolete this PR.

TopperDEL:
Looks good to me!

DeflateAwning:
Hmm should we close this PR? Seems like this change was actually done already in other PRs?

TopperDEL:
Yeah, let's close it!

PR #4785, opened 2022-07-07
Improved restore performance

snamds:
This PR was motivated after I had to cancel a restore in a NAS after 1.5 days waiting for it to begin. I had to cancel it and set the "no-local-blocks" option

Devices used for test:
NAS: Synology DS418. HDD. Restore was 1.4TB. version 2.0.5.1. I can't update to the latest release because the last mono version for dsm 6 for this architecture is 5.8.0.108
PC: intel i7-8700 16GB RAM and SSD. Restore was 1.7TB. Github checkout (28/12/2021)
The PC restore is 1.7TB data
The NAS restore is 1.4TB data. A subset of the 1.7TB PC data

1. FindMissingBlocks: 
	NAS: 9h
	PC: 7min
	PC after optimization: 1m20s
Problem: 
	Very slow table insertion
Solution: 
	Create indexes after insertion.

2. ScanForExistingSourceBlocksFast: (No local source was present)
	NAS: It didn't finish after 1.5 days. It had to be skipped via "no-local-blocks".
	PC: 1h52m
	PC after optimization: 2s
Problem:
	The fact that "ExecuteReader" gives you a cursor immediately doesn't mean that the query traversal is going to be fast. This query has a traversal overhead of 1h52m. Regardless of whether or not the source files exist.
Solution:
	2a. Modify sql query to improve performance:
		This query joins 5 tables:
			"Fileset-XXXXX" (Temporary)
			"Blocks-XXXXX" (Temporary) 
			"LatestBlocksetIds-XXXXX" (Temporary) 
			"Block", 
			"BlocksetEntry"
		"LatestBlocksetIds-XXXXX" is used to get the latest blockId of every file. This info can be previously stored in the Fileset-XXXXX table. This way this join doesn't need to be done
		"Block" table is used to get the BlockId. This info can be previously stored in the Blocks-XXXXX almost for free. This way this join doesn't need to be done
		With this optimization, this function takes 70s (PC) to complete
	2b. We can previously look for the files that exist and store this info in Fileset-XXXXX so this blocks don't need to be queried.
		After this optimization, as no source files were present, the ScanForExistingSourceBlocksFast took 3s (PC)

3. GetMissingVolumes: 
	NAS: 14h
	PC: 3min
	PC after optimization: 7s
Problem:
	The query of this function is very slow.
Solution:
	There is a inner query that has to match the temporary block table ("Blocks-XXXXX") with the block table ("Block") "just" to retrieve B.VolumeId.
	The VolumeId can be added to the "Blocks-XXXXX" when is created. This way this join is not needed


kenkendk:
Moved this to wishlist because it is not critical and the `--no-local-blocks` option is now default on (renamed to `--use-local-blocks`)

carljohnsen:
I did some testing with this PR and I was able to reproduce some speedup on a restore of 5163 files totaling ~7 TB: 

| name   | legacy restore | files already restored | skip metadata |       time (speedup) |
|--------|----------------|------------------------|---------------|----------------------|
| master |                |                        |               | 07:11:08.024 (1.000) |
| 4785   |                |                        |               | 07:02:33.103 (1.020) |
|        |                |                        |               |                      |
| master |              x |                        |               | 09:35:40.653 (1.000) |
| 4785   |              x |                        |               | 09:17:06.292 (1.033) |
|        |                |                        |               |                      |
| master |              x |                      x |             x | 03:54:56.619 (1.000) |
| 4785   |              x |                      x |             x | 03:58:00.226 (0.987) |

While the speedup for my particular case wasn't as great as outlined in this PR, I do see value in this. I'll merge it when the tests pass.

PR #4796, opened 2022-07-22
Correct library loading problem with Backend.OneDrive

jherig:
While starting to learn the codebase, I ran across the test failures in Duplicati.UnitTest.BackendToolTests.Get, and traced them back to Backend.OneDrive loading problems that I've seen on a personal project related to NuGet and System.Net.HTTP.  The fix that I finally got to work was erasing the .csproj and recreating it.  Note: I do not use OneDrive as a backend, so would appreciate more testing, but the code itself has not been affected, just the build process.


jherig:
Hmm... the appveyor build shows the errors I fixed, yet they don't occur on my system.  Perhaps I need to understand the appveyor build process better first.

jherig:
Does not fix issue, not sure exactly why, need more investigation. Closing for now.

Issue #4668, opened 2022-01-27
"Duplicati needs to be updated" warning on upgrade to MacOS Monterrey 12.1


bob787:
- [X] I have searched open and closed issues for duplicates.
- [X] I have searched the [forum](https://forum.duplicati.com) for related topics.

----------------------------------------

## Environment info
- **Duplicati version**: 2.0.6.3_beta_2021-06-17
- **Operating system**: MacOS Monterey 12.1
- **Backend**: OneDrive

## Description
After installing MacOS Monterey 12.1, a warning pops up: "Duplicati needs to be updated" This app will not work with Future versions of MacOS... 

See the screenshot. The "Learn more" link points to: https://www.python.org/doc/sunset-python-2/ 

## Steps to reproduce
1. Install Duplicati on a MacOS prior to 12.1
2. Upgrade MacOS to 12.1

This prompt is triggered by a process calling: /usr/bin/python. It could be the pkg, the pkg’s preinstall, or postinstall script. These can be difficult to locate, as evidenced here: https://macmule.com/2021/10/25/macos-monterey-prompt-needs-to-be-updated/

This issue was lightly touched in https://github.com/duplicati/duplicati/issues/4316, but as noted in the Macmule link above, this issue can not be solved by simply changing to /usr/bin/python3

The intrepid ts678 in the forums suspects the issue has to do with the TrayIcon Rumps, https://github.com/jaredks/rumps 

- **Actual result**:
See error above
- **Expected result**:
Error-free upgrade

## Screenshots
![66f7e3020baebe8485f497a887e30dc82f2541d8](https://user-images.githubusercontent.com/98560908/151455008-3c9e1233-26de-4bcd-b3fe-1fc61dae757e.png)

duplicatibot:
This issue has been mentioned on **Duplicati**. There might be relevant details there:

https://forum.duplicati.com/t/macos-monterey-this-app-will-not-work-with-future-versions-of-macos/13307/12

bob787:
Pkgcheck s a good package inspection script, https://github.com/scriptingosx/pkgcheck, that can help isolate the parts of Duplicati that still call /usr/bin/python. 

Starting with macOS 12.3, (https://developer.apple.com/documentation/macos-release-notes/macos-12_3-release-notes), the Python 2.7 binary /usr/bin/python will not work anymore. Calls to python will fail.

I attempted running the script against the Duplicati .dmg file with limited success, but I certain its just because I have no idea what I'm doing.

[...]

PR #4584, opened 2021-07-27
Fix server certificate on Linux

mnaiman:

Issue #4468, opened 2021-03-13
Restore impossible permissions in Rootless Docker don't result in warning or error


schklom:
- [x] I have searched open and closed issues for duplicates.
- [x] I have searched the [forum](https://forum.duplicati.com) for related topics.

----------------------------------------

## Environment info

- **Duplicati version**: 2.0.5.1_beta_2020-01-18
- **Operating system**: Debian 10
- **Backend**: Google Drive

## Description

The UID and GID are restored incorrectly when using Rootless Docker.

## Steps to reproduce

1. Install Rootless Docker on a new account B (1001:1001), with main one being A (1000:1000)
2. Run Duplicati on it
3. Subuid and subgid are
A:100000:65536
B:165536:65536
[more details](https://rootlesscontaine.rs/how-it-works/userns/#subuids--subgids-newuidmap--newgidmap) if you need
4. Have files with UID/GID outside the UID/GID range allowed by the user namespace (in /etc/subuid and /etc/subgid)
e.g. file having UID:GID = 1001:1000
5. Files restored with wrong UID/GID

- **Actual result**:
GID = 231069
- **Expected result**:
Error or at least warning

I know that in this scenario, Duplicati cannot give UID 1000 because it's outside its namespace.
Therefore, it should yield a error about incorrect permissions instead of working with wrong info anyway.
Or at least a warning and set the wrong UID or GID to root (in this case 1001).

This isn't vital, but Rootless Docker should become more and more used, and might become the norm in a few years. So you might want to treat this as low priority, but imo you shouldn't just ignore it or people will complain later ^^

PS: I love your tool, it's amazing 😃

PR #4800, opened 2022-08-01
Fix compile for vs mac

tgorgdotcom:
This is a "Part 1" to fix the issues with newer versions of MacOS and Python.  This part enables us to get duplicati compiled on VS for Mac (and perhaps Xamarin Studio, etc.).  It primarily adds bouncycastle back in as a referenced packaged (as it's still being used in MailKit/MimeKit packages).  It also upgrades a few packages (a side effect of the process I was using to reinstall packages).

"Part 2" will come from @dgileadi, and include fixes to get the Duplicati.GUI.TrayIcon CocoaRunner class working, hopefully enabling us to deprecate the RumpsRunner class, and free us of the Python dependency.

### DIFF
--- a/Duplicati/Library/Backend/Idrivee2/Duplicati.Library.Backend.Idrivee2.csproj
+++ b/Duplicati/Library/Backend/Idrivee2/Duplicati.Library.Backend.Idrivee2.csproj
@@ -40,10 +40,10 @@
   </PropertyGroup>
   <ItemGroup>
     <Reference Include="AWSSDK.Core, Version=3.3.0.0, Culture=neutral, PublicKeyToken=885c28607f98e604, processorArchitecture=MSIL">
-      <HintPath>..\..\..\..\packages\AWSSDK.Core.3.3.103.37\lib\net45\AWSSDK.Core.dll</HintPath>
+      <HintPath>..\..\..\..\packages\AWSSDK.Core.3.3.103.71\lib\net45\AWSSDK.Core.dll</HintPath>
     </Reference>
     <Reference Include="AWSSDK.S3, Version=3.3.0.0, Culture=neutral, PublicKeyToken=885c28607f98e604, processorArchitecture=MSIL">
-      <HintPath>..\..\..\..\packages\AWSSDK.S3.3.3.104.25\lib\net45\AWSSDK.S3.dll</HintPath>
+      <HintPath>..\..\..\..\packages\AWSSDK.S3.3.3.104.43\lib\net45\AWSSDK.S3.dll</HintPath>
     </Reference>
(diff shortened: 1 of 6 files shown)


warwickmm:
Are the build issues specific to macOS 12?  The [tests](https://github.com/duplicati/duplicati/runs/7481911990?check_suite_focus=true) are running on macOS 11 without any problems and I believe @kenkendk builds and creates the releases on a mac (not sure what version).

tgorgdotcom:
Hello @warwickmm.  Not sure if it's MacOS 12 specific, but I get issues when starting from pulling fresh from the repository.    There's [a reference in issue #4468](https://github.com/duplicati/duplicati/issues/4668#issuecomment-1189385567) that details what's going on from another user.  I think the lines removed in pull request #4584 might have introduced the bug (removed reference to BouncyCastle).

warwickmm:
I don't have a mac, but a [test build](https://github.com/warwickmm/duplicati/runs/7623897217?check_suite_focus=true) with macOS 12.5 appears to compile without error.  It would be nice if we can get a better understanding of how the build environments differ.

warwickmm:
Sorry, I just saw the stack trace [here](https://github.com/duplicati/duplicati/issues/4668#issuecomment-1189385567).  Is the issue a runtime error and not a compile time error?

tgorgdotcom:
It seems to happen as the debug is starting, after the build process.

tgorgdotcom:
Hmm.  I pulled a fresh copy of the codebase and I ran the tests manually via command line.  I get 6 errors, (only 3 provide stack traces), but it does look like duplicati runs completely, and the stack traces are not related to the errors I've been getting in Visual Studio for Mac.  

I also tried an additional fresh copy of the codebase and opened that in VS, in the hopes doing so would determine if VS is doing something different from `nuget restore Duplicati.sln`.  It does do a restore of the packages automatically, and running a build works okay (up to the point that it errors when it's looking for Python).

**However**, it's when I start making changes to Duplicati.GUI.TrayIcon and CocoaRunner when I start getting the error that @dgileadi also gets in [issue #4668](https://github.com/duplicati/duplicati/issues/4668#issuecomment-1189385567).  I essentially modify this to use Xamarin.Mac instead of Mono.Mac (including modifying references files to the correct dlls).  Perhaps I should ask @dgileadi to include the reference in his pull request that will include the switch to Xamarin.Mac for CocoaRunner?

dgileadi:
Yes, it sounds like the reference changes are tied up with using Xamarin.Mac. I didn't expect that.

I'll go ahead and make a PR with my changes, including the reference changes, and see how that goes. Thanks @tgorgdotcom for all your help with this!

PR #4849, opened 2022-11-19
fix #4716 by falling back to System hasher

### LINKED ISSUE #4716: OneDrive backup error message: EVP_MD_size assembly:<unknown assembly> type:<unknown type> member:(null)
- [x ] I have searched open and closed issues for duplicates.
- [ x] I have searched the [forum](https://forum.duplicati.com) for related topics.

## Environment info
- **Duplicati version**:Duplicati - 2.0.6.3_beta_2021-06-17
- **Operating system**: Ubuntu 22.04 LTS
- **Backend**: OneDrive

## Description
Every backup attempt results in an error message on the bottom of the screen stating "Error while running OneDrive - 
EVP_MD_size assembly:<unknown assembly> type:<unknown type> member:(null)". This behaviour has started after I upgraded to Ubuntu 22.04 LTS yesterday.

## Screenshots
image

## Debug log
No log data is generated after the error message.

vilaureu:
# Description
When only having _OpenSSL_ 3 installed, `FasterHashing.FasterHash.Create()` currently throws an `EntryPointNotFoundException`.
This issue is discussed in #4716.
This PR mitigates the issue by falling back to `HashAlgorithm.Create` in the error case.

# Open Questions
- Should a warning be emitted if the fallback route is taken?
- Looking forward, [_FasterHashing_](https://github.com/kenkendk/FasterHashing) should be extended to support _OpenSSL_ 3.

### DIFF
--- a/Duplicati/Library/Utility/HashAlgorithmHelper.cs
+++ b/Duplicati/Library/Utility/HashAlgorithmHelper.cs
@@ -20,11 +20,13 @@ using System.Security.Cryptography;
 
 namespace Duplicati.Library.Utility
 {
-	/// <summary>
-	/// This class helps picking the fastest hash algorithm implementation,
+    /// <summary>
+    /// This class helps picking the fastest hash algorithm implementation,
     /// which is what <seealso cref="System.Security.Cryptography.HashAlgorithm.Create()"/> should do, but does not.
-	/// </summary>
-	public static class HashAlgorithmHelper
+    /// It falls back to the aforementioned System library when creation of
+    /// the faster hash function fails.
+    /// </summary>
+    public static class HashAlgorithmHelper
     {
         /// <summary>
         /// Create the hash algorithm with the specified name.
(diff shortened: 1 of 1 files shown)


ts678:
Thanks!

Very good to see an attempt (is it proven?) to fall back to normal hashing without having to change FasterHashing to do that.
IMO ideally FasterHashing gets changed, but that developer has little time now, so the less time taken, the more likely to ship.

vilaureu:
I confirmed this patch working at least on an _Arch Linux_ _Docker_ container.
I tested it by making a backup and a successful restore with only _OpenSSL_ 3 installed. (It did not work without the patch.)
I was not able to build _Duplicati_ on a _Ubuntu_ nor a _Debian_ container because the packaged _NuGet_ version was too old.

gpatel-fr:
@umgfoin 

The Duplicati project is currently in a wonky situation with the original author limiting contribution to publishing new releases. I'm trying to push a new release as a mere contributor, but this library situation is rather awkward. If you can't get your fix in, a possibility could be pushing the whole lib into Duplicati proper, a sort of 'vendoring' even if the author is the same. It's not as this lib has a wide use outside of Duplicati given the low number of forks and stars.
The code would be implicitly relicensed I think but as far as I know it's possible given that proper aknowledgments are kept in the modified project. Could you consider taking this task ? It's not urgent because at the moment I have more pressing concerns. Thanks even if all you do is to inform me that it will not be possible.

gpatel-fr:
@umgfoin 

thanks for the reply and sorry to be late.
At the moment I am trying to get a release out to keep Duplicati active before the Net Core version is available.
Faster hashing is pulled from Nuget as a binary. External lib is as you can see not really maintained.
My feeling about git based dependencies is not so good since it will involve the same hassle of building Duplicati + library but with an additional step of downloading (git is not available everywhere) but I may be overstating the problems, I'll give more thought to your idea since I had very little time available these last days.

PR #4890, opened 2023-02-09
EXPERIMENTAL rebuild database optimization

gpatel-fr:
rewrite of FindMissingBlocklistHashes with simpler queries add pragma to boost sqlite memory cache by 1000
behaviour based on the environment variable EXPERIMENTAL_RECREATEDB=1

### DIFF
--- a/Duplicati/Library/Main/Database/LocalRecreateDatabase.cs
+++ b/Duplicati/Library/Main/Database/LocalRecreateDatabase.cs
@@ -43,10 +43,10 @@ namespace Duplicati.Library.Main.Database
         private readonly System.Data.IDbCommand m_findHashBlockCommand;
         private readonly System.Data.IDbCommand m_insertBlockCommand;
         private readonly System.Data.IDbCommand m_insertDuplicateBlockCommand;
-        
+
         private string m_tempblocklist;
         private string m_tempsmalllist;
-        
+
         /// <summary>
         /// A lookup table that prevents multiple downloads of the same volume
         /// </summary>
@@ -101,9 +101,9 @@ namespace Duplicati.Library.Main.Database
         public LocalRecreateDatabase(LocalDatabase parentdb, Options options)
             : base(parentdb)
         {
-            m_tempblocklist = "TempBlocklist-" + Library.Utility.Utility.ByteArrayAsHexString(Guid.NewGuid().ToByteArray());
-            m_tempsmalllist = "TempSmalllist-" + Library.Utility.Utility.ByteArrayAsHexString(Guid.NewGuid().ToByteArray());
-                        
+            m_tempblocklist = "TempBlocklist_" + Library.Utility.Utility.ByteArrayAsHexString(Guid.NewGuid().ToByteArray());
+            m_tempsmalllist = "TempSmalllist_" + Library.Utility.Utility.ByteArrayAsHexString(Guid.NewGuid().ToByteArray());
+
             using(var cmd = m_connection.CreateCommand())
             {
                 cmd.ExecuteNonQuery(string.Format(@"CREATE TEMPORARY TABLE ""{0}"" (""BlockListHash"" TEXT NOT NULL, ""BlockHash"" TEXT NOT NULL, ""Index"" INTEGER NOT NULL)", m_tempblocklist));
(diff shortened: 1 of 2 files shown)


gpatel-fr:
Rationale for the changes
with the following test script:
```
export classicdb=/usr/lib/duplicati/data/CLASSIC_DB.sqlite
export expdb=/usr/lib/duplicati/data/EXP_DB.sqlite

echo experimental recreate

export EXPERIMENTAL_RECREATEDB=1
rm $expdb -f
time mono /shared/duplicati/Duplicati/CommandLine/bin/Debug/Duplicati.CommandLine.exe repair "s3://myuser-monbucket/?s3-server-name=q1900%3A9000&s3-location-constraint=&s3-storage-class=&s3-client=minio&auth-username=myuser&auth-password=myuser56" --dbpath=$expdb --no-encryption=true --compression-module=zip --console-log-level=warning >~/logexp.txt

echo classic recreate

export EXPERIMENTAL_RECREATEDB=0
rm $classicdb -f
time mono /shared/duplicati/Duplicati/CommandLine/bin/Debug/Duplicati.CommandLine.exe repair "s3://myuser-monbucket/?s3-server-name=q1900%3A9000&s3-location-constraint=&s3-storage-class=&s3-client=minio&auth-username=myuser&auth-password=myuser56" --dbpath=$classicdb --no-encryption=true --console-log-level=warning --compression-module=zip >~/logclassic.txt

echo differences between the 2 databases...
sqldiff --table block $classicdb $expdb 
sqldiff --table blocklist $classicdb $expdb 
sqldiff --table filelookup $classicdb $expdb 
sqldiff --table blockset $classicdb $expdb 
sqldiff --table filesetentry $classicdb $expdb 
sqldiff --table blocksetentry $classicdb $expdb 
sqldiff --table blocklisthash $classicdb $expdb 
sqldiff --table remotevolume $classicdb $expdb 
```
I get the following output with a database having 1.2 million blocks

```
database in a normal state

experimental recreate

real    13m54.819s
user    13m0.541s
sys     0m56.933s

classic recreate

real    18m43.928s
user    14m40.016s
sys     2m48.337s

same backend with 3 index files removed

experimental recreate

real    15m6.798s
user    13m58.805s
sys     1m13.157s

classic recreate

real    26m38.209s
user    19m2.095s
sys     5m18.429s
``

ts678:
FYI [Backups suddenly taking much longer to complete](https://forum.duplicati.com/t/backups-suddenly-taking-much-longer-to-complete/12994) tried to tune a DB with [PRAGMA page_size](https://www.sqlite.org/pragma.html#pragma_page_size) and [PRAGMA cache_size](https://www.sqlite.org/pragma.html#pragma_cache_size).

gpatel-fr:
@ts678 
Thanks for the comment. 

I don't quite see (maybe I missed it, it's a long thread) where someone is saying that they actually tried to change these Pragma. Talking about it, yes, but doing it ? I am not sure that it's possible to set them without changing the code. 

Following your post I toyed with adding to Duplicati a generic possibility of setting custom Pragma, unfortunately it's not totally trivial to do it right.

Anyway, I did not try changing page size, I saw such a large effect of cache_size that I stopped at that. And simplifying the queries also plays a role. I seem to remember that for a given query, I got 23 s from the original query, 16 from the simplified version, and 8 with memory_cache limit raised, so it all matters (needed memory and query complexity can be linked also).

gpatel-fr:
Updated to make the env var name more explicit and remove the expand cache feature to be replaced by the more general PR #4907

mr-russ:
I've been looking sqlite stuff as I've used sqlite in embedded systems and looked at performance.
cache_size should be much larger than the default and set on db open.  2000*4k = 8MB.  Which for a large backup application is no memory allowance at all.  Doubling the page size doubles the cache as it's measured in pages, so performance increased and the disk cliff moved.  You could use db disk size to guess the memory cache size.  But allowing 32MB would improve performance with little other impact.

I'm not sure of the threading model used for connections yet.  Multiple threads can really block each other in journal mode (the default).  WAL mode is much better on disk usage and churn.  It also allows unlimited readers while one writer does work. 
There is less slow operations because of this and fsync is only caused on checkpoint and commit.  Journal fsyncs every operation to ensure no corruption.  That is expensive. This concept also flows to how temporary tables are created and managed, and those are made for queries that make intermediate tables.  Encouraging memory with the tmp location pragma by setting it to memory will improve speed further for the complex selects I see in the rebuild code (and other places).

I don't understand all of the SQLs well enough, but using left joins unless required stops table reordering and can reduce performance a lot.  It also reduces the possibility of using indexes as it has to return all rows.  Changing LEFT JOIN to join, and placing them before all left joins will allow better optimization and runtime where that's possible.

I've not had time to setup and development anh performance environment for a large backup.  So I can't prove what I'm saying.  Even if I did, the rules for getting code reviewed and merged are not clear.  Without some expectation of getting code accepted and merged in a reasonable time, it's hard to know if the effort is worth it. There are 37 outstanding pull requests, going back 3 years, many seeming to be irrelevant now, lacking feedback or really simple with no feedback at all.  That does not give me a new developer encouragement that any effort put in will be worth it.  I saw forum posts on the trying to release stable, but pull requests and issues aren't categorised to know if they are for a proposed stable, or for a new dev branch.  The .net5 branch is 3 years old. It's all a bit confusing and overwhelming for a new developer, even one who's worked on a number of open source projects.

gpatel-fr:
@mr-russ 

I don't have really the time to answer to all your points in detail, about the db_cache the PR #4907 has been merged and allows to set the db_cache (and any other pragma) when opening the database. As of setting a bigger default, well, a good default is hard to find since the product should be able to run well on small (memory constrained) computers and servers. Keeping the database default seem a good compromise to me. Thanks for the other ideas, I don't have the time to test them myself currently.

About the net 5 situation, there is a fork that aims at continuing the work: 
https://github.com/tsuckow/duplicati/tree/feature/net5-kestrel
If it succeeds (feel free to help, I don't have time myself at the moment), it could be merged back into Duplicati if the author is willing, since I am still able to push changes. I don't have the privilege to grant merge rights myself.

mr-russ (commented):
What methods do you use for testing these changes?

PR #4892, opened 2023-02-15
fix --all-versions=false being actually ignored

gpatel-fr:
Small but irritating problem: asking for a list with '--version=0 --all-versions=false' while having no match on the request produces a 'not found, trying in all versions'. If I ask explicitly for NOT all-version, I expect the software to do what I asked for.

### DIFF
--- a/Duplicati/CommandLine/Commands.cs
+++ b/Duplicati/CommandLine/Commands.cs
@@ -394,7 +394,7 @@ namespace Duplicati.CommandLine
                     (res.Files == null || !res.Files.Any()) &&
                     !compareFilter.Empty;
 
-                if (isRequestForFiles && !Library.Utility.Utility.ParseBoolOption(options, "all-versions"))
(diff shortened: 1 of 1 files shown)


Duplicati/CommandLine/Commands.cs
```diff
@@ -394,7 +394,7 @@ public static int List(TextWriter outwriter, Action<Duplicati.Library.Main.Contr
                     (res.Files == null || !res.Files.Any()) &&
                     !compareFilter.Empty;
 
-                if (isRequestForFiles && !Library.Utility.Utility.ParseBoolOption(options, "all-versions"))
+                if (isRequestForFiles && !(options.ContainsKey("all-versions")))
```
mr-russ, line 397: Is it wise the "help" the user by adjusting the all-versions, if there are no matching files?  It seems surprising.  Why not just exit in that case?

PR #4924, opened 2023-04-20
Don't distinct where it's not needed.

mr-russ:
DISTINCT and UNION will remove duplicates, requiring a forced sort, unique and temporary table step.  It can't be flattened or optimized in other ways.  So they are removed to allow SQLite to do optimization of searches.

### DIFF
--- a/Duplicati/Library/Main/Database/LocalBugReportDatabase.cs
+++ b/Duplicati/Library/Main/Database/LocalBugReportDatabase.cs
@@ -48,10 +48,13 @@ namespace Duplicati.Library.Main.Database
                 {
                 
                     upcmd.Transaction = tr;
-                    upcmd.ExecuteNonQuery(string.Format(@"CREATE TEMPORARY TABLE ""{0}"" (""ID"" INTEGER PRIMARY KEY, ""RealPath"" TEXT NOT NULL, ""Obfuscated"" TEXT NULL)", tablename));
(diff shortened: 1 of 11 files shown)


ts678:
I'm wondering how these specific ones were chosen and by what process they're guaranteed to not change behavior badly.
Long ago I tested two (possibly just the DISTINCT part rather than final result) on an example database and result changed.

Jojo-1000:
I started to run through all of these queries on a copy of my main backup database (with 100kB blocksize, so lots of blocks) to see what improvement is there. It seems like in some cases it can be 30% faster while in others it is just the same.

> what process they're guaranteed to not change behavior badly

What I see in a lot of places is a `NOT IN (SELECT DISTINCT ...)`. I am pretty sure at least those are save to change (IN should not care about duplicates). The others need some careful consideration, which is why I am trying to see if it is worth it in terms of performance.

Jojo-1000:
I tried to run each query individually to see how much faster it is, but I gave up in the middle so I don't have times for the others. However, in my test case that runs all database queries there was no significant performance improvement from this PR.

kenkendk:
Closing this PR as we rewrote pretty much all queries with the update to the SQLite library, and it is no longer realistic that this PR can be updated to a state where we can merge it.

If there are still places where we should not use `DISTINCT` please open a new PR.

PR #4933, opened 2023-05-06
try to work around problems in downloading test file by retrying

gpatel-fr:


Duplicati/UnitTest/CommandLineOperationsTests.cs
```diff
@@ -82,23 +82,43 @@ private void DownloadS3FileIfNewer(string destinationFilePath, string url)
             try
             {
                 // check if the file should be downloaded, exception if not
-                using ((HttpWebResponse)webRequest.GetResponse())
+                using (var wr=(HttpWebResponse)webRequest.GetResponse())
 
                 using (WebClient client = new WebClient())
-                {
-                    client.DownloadFile(url, destinationFilePath);
+                { // try to workaround a weird intermittent bug where downloaded size
```
mr-russ, line 88: These files are quite large, and likely error prone. 

Is the reason we download them all the time so that automated checks can work?

I've considered making their existence and sha256 a requirement for the test to run successfully.

Overall I don't like the surprise that running tests can download nearly 1GB of data without you knowing that's going to happen.

Duplicati/UnitTest/CommandLineOperationsTests.cs
```diff
@@ -82,23 +82,43 @@ private void DownloadS3FileIfNewer(string destinationFilePath, string url)
...
+                    Console.WriteLine("ex.Response is null !");
                     throw;
                 }
 
                 if (((HttpWebResponse) ex.Response).StatusCode != HttpStatusCode.NotModified)
                 {
+                    Console.WriteLine("file to download {0} already exists and is up to date", destinationFilePath);
                     throw;
                 }
+                Console.WriteLine("download exception: {0}", ex.Response);
```
mr-russ, line 121: Any chance we could consider NLog for logging?  Much nicer for test debugging and sending data to different locations.  Even if it's just the console.

Duplicati/UnitTest/CommandLineOperationsTests.cs
```diff
@@ -82,23 +82,43 @@ private void DownloadS3FileIfNewer(string destinationFilePath, string url)
...
             catch (WebException ex)
             {
                 if (ex.Response == null){
+                    Console.WriteLine("ex.Response is null !");
                     throw;
                 }
 
                 if (((HttpWebResponse) ex.Response).StatusCode != HttpStatusCode.NotModified)
                 {
+                    Console.WriteLine("file to download {0} already exists and is up to date", destinationFilePath);
```
mr-russ, line 118: Any reason not to use $@"file to download {destinationFilePath}...".   It's easier to read and understand, especially when there are multiple variables.

Duplicati/UnitTest/CommandLineOperationsTests.cs
```diff
@@ -82,23 +82,43 @@ private void DownloadS3FileIfNewer(string destinationFilePath, string url)
             try
             {
                 // check if the file should be downloaded, exception if not
-                using ((HttpWebResponse)webRequest.GetResponse())
+                using (var wr = (HttpWebResponse)webRequest.GetResponse())
 
                 using (WebClient client = new WebClient())
```
mr-russ, line 87: If WebClient is causing issue on mac, is it worth considering using HttpClient?

Issue #4831, opened 2022-10-15
FTPS (both TLS 1.2, TLS 1.3) broken with new FileZilla FTP server versions on Windows clients


taz-il:
- [x] I have searched open and closed issues for duplicates.
- [x] I have searched the [forum](https://forum.duplicati.com) for related topics.

----------------------------------------

## Environment info

- **Duplicati version**: 2.0.6.3_beta_2021-06-17
- **Operating system**: Win 11 Pro, 22H2
- **Backend**: FTP and Alternative FTP

## Description

I've recently upgraded my FTP server from FileZilla 0.9.6 to 1.5.1
The newer version requires TLS session resumption, protects against session truncation, and only supports TLS 1.2, TLS 1.3 for secure FTP.

Neither FTP (native FtpWebRequest) nor Alternative FTP (FluentFTP) worked with TLS 1.2 or TLS 1.3:
- TLS 1.2 - The file transfers fine, but the server returns an error due to client not properly terminating the session.
- TLS 1.3 - The file transfer doesn't start, server complains about TLS session not being resumed.

Note that after searching online a bit, I initially suspected NAT / Router / firewall configuration issues, however I ruled these out since the failures reproduce against localhost (i.e. 127.0.0.1; Duplicati and FZ Server running on the same machine).

I believe this isn't directly a Duplicati issue (rather Fluent FTP / underlying .Net libs), but posting here in case others run into this and/or have more ideas.

Created [simple console apps that reproduce this](https://github.com/taz-il/Repro-FTP-TLS-issue), and [posted the issue on FileZilla Server forum](https://forum.filezilla-project.org/viewtopic.php?f=6&t=55147), waiting for a reply.

## Steps to reproduce

1. Set up a backup with FTP / Alternative FTP backend, Duplicati running on Win11
2. Configure it to use TLS 1.2 / TLS 1.3 (the only ones that newer FileZilla servers support, at least on Windows).
3. Run it against newer versions of FileZilla Server (I used the latest, 1.5.1, but if I understand correctly this should happen on any version after FZ migrated from OpenSSL to GnuTLS)
4. Consider running the server on the same machine, and use 127.0.0.1 (to rule out NAT / Router / firewall issues), though this will also fail between remote computers.
5. Voila it fails (at least for me).

## Screenshots

Sample failure:
![image](https://user-images.githubusercontent.com/77799839/196006830-27472d0e-ba23-4c74-aeac-f55b4191ce1b.png)

## Debug log

```
[...]

PR #4844, opened 2022-11-08
Fix runtime exceptions loading System.Net.Http

taz-il:
The continuous integration tests have been failing for a long time, due to loading errors of the OneDrive backends.
I've also encountered the same runtime errors when building and running a fresh copy of Duplicati locally.
The failures seem to be rooted at version conflicts of System.Net.Http.

https://github.com/dotnet/runtime/issues/24382 seems to touch on the core issue, but it left me utterly confused.

I'm not an expert at this at all, tweaked removing/adding binding redirects, as well as any other solution I could find online, and nothing else solved the issue.

After a lot of tweaking, landed on the commit here - which fixed the Runtime errors locally and the continuous integration tests. I'm not sure this is the right approach, but it was the only one that fixed the issue (and therefore the tests) for me.

Fixes #4636

PR #4935, opened 2023-05-12
Update FluentFTP to latest version (46.0.2)

taz-il:
Update FluentFTP to the latest version, as a step towards resolving TLS session issues (#4831)

While the new version doesn't resolve the issues (they seem to be [known issues](https://github.com/robinrodricks/FluentFTP/wiki/FTPS-Connection-using-GnuTLS) and stem from the underlying .net implementation) - FluentFTP recently added support for GnuTLS (not yet using GnuTLS here - leaving this to a future pull request, I haven't yet tried to make the switch; upgrading to the lastest FluentFTP seems like the right thing to do regardless of the TLS issues).

Note the migration to FluentFTP v40 onwards wasn't backward compatible, followed these [migration instructions](https://github.com/robinrodricks/FluentFTP/wiki/v40-Migration-Guide).

Also added in this PR is a commit migrating to NUnit 3 nuget adapter (VSIX test adapters are [deprecated for several years](https://marketplace.visualstudio.com/items?itemName=NUnitDevelopers.NUnit3TestAdapter))

### DIFF
--- a/Duplicati/Library/Backend/AlternativeFTP/AlternativeFTPBackend.cs
+++ b/Duplicati/Library/Backend/AlternativeFTP/AlternativeFTPBackend.cs
@@ -21,6 +21,7 @@
 using Duplicati.Library.Common.IO;
 using Duplicati.Library.Interface;
 using FluentFTP;
+using FluentFTP.Client.BaseClient;
 using System;
 using System.Collections.Generic;
 using System.IO;
@@ -61,9 +62,7 @@ namespace Duplicati.Library.Backend.AlternativeFTP
 
         private readonly string _url;
         private readonly bool _listVerify = true;
-        private readonly FtpEncryptionMode _encryptionMode;
-        private readonly FtpDataConnectionType _dataConnectionType;
-        private readonly SslProtocols _sslProtocols;
+        private readonly FtpConfig _ftpConfig;
         private readonly TimeSpan _uploadWaitTime;
 
         private readonly byte[] _copybuffer = new byte[CoreUtility.DEFAULT_BUFFER_SIZE];
@@ -86,7 +85,7 @@ namespace Duplicati.Library.Backend.AlternativeFTP
             get { return "aftp"; }
         }
 
-        private FtpClient Client
+        private AsyncFtpClient Client
         { get; set; }
 
         public IList<ICommandLineArgument> SupportedCommands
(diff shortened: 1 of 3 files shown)


Duplicati/UnitTest/packages.config
```diff
@@ -1,4 +1,5 @@
 ﻿<?xml version="1.0" encoding="utf-8"?>
 <packages>
   <package id="NUnit" version="3.12.0" targetFramework="net471" />
-</packages>
+  <package id="NUnit3TestAdapter" version="4.2.1" targetFramework="net471" />
```
mr-russ, line 4: This is a unit test change, that I think we should have.  If we are doing it, can we use the latest?

mr-russ (commented):
Are there unit tests covering this?  If so, it would be good to say so to increase confidence.

I have upgraded to the 40's for a work project (100000+ connections per day), and had good results where we had one failure at least once a week where the system.  I personally agree and believe the benefits are better than risks of upgrading.  But I'm in the have tests and take slightly more risk than others.

Duplicati/UnitTest/packages.config
taz-il: Good catch - 4.2.1 was the latest when I authored this commit (several months ago).
Added a new commit that bumps both NUnit3TestAdapter and NUnit to their latest (stable) versions - 4.4.2, 3.13.3 respectively.

taz-il:
> Are there unit tests covering this? If so, it would be good to say so to increase confidence.

The PR doesn't introduce any new test failures. To be honest, I'm a newbie to this repo, so not totally sure about unit test coverage for the core change here (note no functionality is added or removed, just upgrading FluentFTP package version).

I have tested this PR manually (through the UI, set up a test backup with Alternative FTP, etc), and it worked fine.

Separately - testing related - I've noted that the continuous build has been failing for a very long time. I've created #4844 which fixes Runtime errors while loading OneDrive backends, which caused BackendToolTests to fail. Please take a look there too if you can (unrelated to this change here).

mr-russ:
@gpatel-fr I'm struggling to find the information you found with no 32bit support for FluentFtp, can you post a reference that I can look at to see what's going on.

In regard to the mixing, I've submitted another PR, that does just unit test changes. #4939.  Maybe that's mergable to enable better unit testing capabilities in visual studio.

gpatel-fr:
@mr-russ 

please read [this](https://github.com/robinrodricks/FluentFTP/wiki/FTPS-Connection-using-GnuTLS)

where this nugget can be found:
`Currently only supports x64 architectures.`

taz-il:
Regarding mixing unrelated changes:

You're absolutely right - what happened here is the current code doesn't support running the unit tests in VS (due to VSIX test adapter deprecation), and I needed this to be able to easily test locally. Given I have perceived the Unit test adapter change to be minimal risk and generally useful, I have left it there. Thanks @mr-russ for splitting it off to a different PR.

Regarding the core change (upgrading FluentFTP version):

I think there's confusion here that I can help clear up. Duplicati already depends on FluentFTP (just on a really old version). FluentFTP supports (and seems to continue supporting) both 32 and 64 bit architectures.

FluentFTP.GnuTLS is a new **optional** add-on, that I'm not using in this PR. It currently supports only 64 bit architectures. We **may** want in a future PR to add that as an option (that will only work on x64 for now), but we don't have to. This is why I wrote that the current PR doesn't fix the TLS 1.3 issues

taz-il:
(I accidentally closed the PR. Reopened now)

gpatel-fr:
@taz-il 

thanks for the provided details. I agree that upgrading to latest upstream is valuable.
I suggest that you remove the commits for tests from this PR, and submit them in a separate one (if you want)

None of these changes will go in upcoming beta, it's too late, I don't want to add any substantive stuff to Duplicati at this point.

For next Canary, I'll commit this PR in a separate branch since I can't build binaries without that. Once it's done, I will test them - I'll also post the information on the forum, if anyone else want to test the new Ftp library. 
If nothing bad comes up I'll merge the branch.
I'll look up the tests and if I don't see anything concerning I'll commit them directly for next Canary.

taz-il:
Sounds good! thanks.

I've removed the NUnit changes from here - so this PR now only contains upgrading FluentFTP to latest.

@gpatel-fr any changes I needed from my end to request merging to a different (canary?) branch?

gpatel-fr:
@taz-il 
no, when you create a PR it's always based on the main branch, only a contributor can redirect to another branch.

gpatel-fr:
The binaries are created here:

https://github.com/duplicati/duplicati/actions/runs/5008478766

I think that I will test them tomorrow. Under Windows at least.

gpatel-fr:
See this [post](https://forum.duplicati.com/t/ftp-with-alternate-ftp-testers-wanted/16280).

PR #4936, opened 2023-05-14
Update tests workflow - fix warnings that will soon break the tests

taz-il:
The tests workflow is using actions which use patterns that are currently deprecated. The commits here update the versions of the various actions used by the tests workflow, thus fixing all 17 warnings (about set-output, save-state, Node.js 12-->16 migration).

The soft deprecation (warnings) will become a hard deprecation (errors) and break the tests according to the documentation:
- June 1st 2023 - [set-output, save-state](https://github.blog/changelog/2022-10-11-github-actions-deprecating-save-state-and-set-output-commands/)
- Summer 2023 - [node16](https://github.blog/changelog/2022-09-22-github-actions-all-actions-will-begin-running-on-node16-instead-of-node12/)

[e.g. test run before](https://github.com/taz-il/duplicati/actions/runs/4967727082):
image

[e.g. test run after](https://github.com/taz-il/duplicati/actions/runs/4969568960):
image

### DIFF
--- a/.github/workflows/tests.yml
+++ b/.github/workflows/tests.yml
@@ -17,13 +17,13 @@ jobs:
     steps:
     - name: Add msbuild to PATH (Windows)
       if: matrix.os == 'windows-latest'
-      uses: microsoft/setup-msbuild@v1.0.2
+      uses: microsoft/setup-msbuild@v1
 
     - name: Install NuGet
-      uses: nuget/setup-nuget@v1.0.5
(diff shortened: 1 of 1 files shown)


mr-russ:
Given this does not touch any production code.  I have limited github actions knowledge so can't say much more about correctness.

My vote is to merge if it's passing tests.

gpatel-fr:
Thanks for your contribution!

PR #4938, opened 2023-05-16
Normalize line endings to lf.

mr-russ:
lf is used on linux and Mac.  Most of the code is lf already, with items of crlf mixed in.  This will change future checkouts to be lf only.


luixxiul:
This should be taken care of by https://github.com/duplicati/duplicati/pull/5272.

PR #4939, opened 2023-05-16
Update NUnit and use NUnit Adapter.

mr-russ:
This allows tests to be run in VisualStudio,
and also using dotnet when projects support that.


mr-russ:
@gpatel-fr are you able to review this, there changes seem safe to me and are slowing my ability to contribute in other ways.

PR #4945, opened 2023-05-19
FluentFTP version bump - Fix handling of non-existing base path

taz-il:


taz-il:
Not picking up the changes I meant. Closing.

PR #4946, opened 2023-05-20
FluentFTP version update - Fix handling of missing base folder

taz-il:
1. Improve error descriptiveness (by avoiding unnecesary wrapping in AggregateException, using InnerException where relevant)
2. Improved detection of missing base folder, now triggers prompting the user for new folder creation
3. Fix folder creation (CreateClient invoked right before folder creation previously failed since it always tried to set working directory to the missing folder).

### DIFF
--- a/Duplicati/Library/Backend/AlternativeFTP/AlternativeFTPBackend.cs
+++ b/Duplicati/Library/Backend/AlternativeFTP/AlternativeFTPBackend.cs
@@ -226,88 +226,88 @@ namespace Duplicati.Library.Backend.AlternativeFTP
             var list = new List<IFileEntry>();
             string remotePath = filename;
 
-                var ftpClient = CreateClient();
+            var ftpClient = CreateClient();
 
-                // Get the remote path
-                var url = new Uri(this._url);
-                remotePath = "/" + this.GetUnescapedAbsolutePath(url);
+            // Get the remote path
+            var url = new Uri(this._url);
+            remotePath = "/" + this.GetUnescapedAbsolutePath(url);
 
-                if (!string.IsNullOrEmpty(filename))
+            if (!string.IsNullOrEmpty(filename))
+            {
+                if (!stripFile)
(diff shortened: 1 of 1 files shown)


taz-il:
Some more details on the change:

a. Look inside AggregateException, and throw the correct missing folder exception
b. Consolidated this try/catch block to a single location - the SetWorkingDirecotry in CreateClient (was originally in both List and PutAsync)
c. Fixed the condition that detects the missing folder (was highly dependent on specific error string from the server, changed to depend on error code)
d. Fixed CreateFolder (by having CreateClient not try to SetWorkingDirectory when called from CreateFolder)

Diff seems ugly, due to two large try blocks that were removed in (b). The changed indentation confuses Github text diff? The actual diff is less than what it seems in List and PutAsync.

taz-il:
Marking as draft, until I have time to look into Issue 2 that was mentioned on [the post](https://forum.duplicati.com/t/ftp-with-alternate-ftp-testers-wanted/16280/5) - might then add another commit here.

taz-il:
@gpatel-fr, @ts678 - ready for review / merge

I've separated to 4 commits for easier reviewing.
Note Github text diff is slightly confused by indentation (e.g. "Fix indentation" commit only de-indents two large blocks, no actual code changes, unlike what it may appear at a first glance).

It now fixes / improves everything that was pointed out in [Issue 1](https://forum.duplicati.com/t/ftp-with-alternate-ftp-testers-wanted/16280/2) and [Issue 2](https://forum.duplicati.com/t/ftp-with-alternate-ftp-testers-wanted/16280/5), namely:
- Test connection with a missing folder now results in offering the user to create a folder, rather than an error message
- Test connection with missing write permissions results in a more descriptive error**: "Failed to connect: Error writing file: Code: 550 Message: Permission denied"
- As a bonus 2, folder creation now works in AFTP (previously broken due to a bug)

Issue #4636, opened 2021-11-07
Unit tests can fail trying to load Duplicati.Library.Backend.OneDrive.dll when running backups


dferreyra:
- [x] I have searched open and closed issues for duplicates.
- [x] I have searched the [forum](https://forum.duplicati.com) for related topics.

----------------------------------------

## Environment info

- **Duplicati version**: "master" beyond 2.0.6.100_canary_2021-08-11
- **Operating system**: Windows 10
- **Backend**: N/A

## Description

I've often had a problem running unit tests from the command line where a unit test performs a backup, and the backup results in three warnings.  The unit test in question fails because it wants to see zero warnings.

Here's an example command I'm using to run unit tests, along with the error output:
```
...>nunit3-console.exe --test Duplicati.UnitTest.RecoveryToolTests Duplicati\UnitTest\bin\Debug\Duplicati.UnitTest.dll
NUnit Console Runner 3.12.0 (.NET 2.0)
Copyright (c) 2021 Charlie Poole, Rob Prouse
Saturday, November 6, 2021 11:44:38 PM

Runtime Environment
   OS Version: Microsoft Windows NT 6.2.9200.0
   Runtime: .NET Framework CLR v4.0.30319.42000

Test Files
    Duplicati\UnitTest\bin\Debug\Duplicati.UnitTest.dll

Test Filters
    Test: Duplicati.UnitTest.RecoveryToolTests
[...]
Errors, Failures and Warnings

1) Failed : Duplicati.UnitTest.RecoveryToolTests.Recover("false")
  Expected: 0
  But was:  3
   at Duplicati.UnitTest.RecoveryToolTests.Recover(String buildIndexWithFiles) in ...\Duplicati\UnitTest\RecoveryToolTests.cs:line 70

Run Settings
    DisposeRunners: True
    WorkDirectory: ...
    ImageRuntimeVersion: 4.0.30319
    ImageTargetFrameworkName: .NETFramework,Version=v4.7.1
    ImageRequiresX86: False
    ImageRequiresDefaultAppDomainAssemblyResolver: False
    TargetRuntimeFramework: net-4.0
    NumberOfTestWorkers: 4

Test Run Summary
  Overall result: Failed
  Test Count: 2, Passed: 1, Failed: 1, Warnings: 0, Inconclusive: 0, Skipped: 0
    Failed Tests - Failures: 1, Errors: 0, Invalid: 0
  Start time: 2021-11-07 06:44:38Z
[...]

PR #4954, opened 2023-06-02
net 472

gpatel-fr:


taz-il:
While not necessarily the original motivation of this PR, I can verify it fixes #4636 - runtime exceptions that occurred when loading System.Net.Http in OneDrive backends (MicrosoftGroup, OneDriveV2, SharePointV2)

With this PR these are resolved everywhere I encountered them, namely:
- Locally built binary
- Local tests (BackendToolTests.Get was consistently failing): 
image
- Appveyor continuous integration tests (same, BackendToolTests.Get was consistently failing), see below: 
image

PR #4962, opened 2023-06-04
Fix filter UI

### LINKED ISSUE #2682: Search shows blank tree.
Duplicati - 2.0.2.1_beta_2017-08-01

If I use Search and enter say *.js the the tree is blank as shown below.

image

### LINKED ISSUE #2976: Exclude filter ignoring all except last filter
- [X] I have searched open and closed issues for duplicates.

## Environment info

- **Duplicati version**: 2.0.2.17_canary_2018-01-23
- **Operating system**: Windows 7
- **Backend**: GUI Command Line

## Description

Filter/exclude is not working properly (using the GUI to populate).  It does seem to work if it's the last line of the advanced options or at least the last filter.  All filters prior to the last filter shown in Advanced options are ignored.

## Steps to reproduce

1. Create a backup set, including a source folder that has a few subdirectories
2. Under Filters, choose "Exclude directories whose name contains" and enter part of a name from a subdirectory in step 1 (in my example, "--exclude=*Previews.lrdata*\".  Finish and save the backup set.
3. Select the backup set, then Commandline...  Choose Test-Filters (removing any Target URL) and run the command.  Note that filter works.  (**Expected result**)
4. Either edit the backup set to add another Filter another subdirectory (Exclude Folder).  Note that this results in a full path, C:\dir\dir2\
5. Run Test-Filters in Commandline... and ensure that the --exclude from step 4 is the after the one from step 2.  Note that the excluded directories whose name contains is now shown as included when running the test.  (**Actual result**)
6.  Run Test-Filters in Commandline... again, but this time edit the advanced options to put the "whose name contains" exclusion last.  The exclusion/filter now works properly again.

Upon further investigation, it seems that all filters are ignored except the last one.  It is not specifically a "directories whose name contains" issue.

### LINKED ISSUE #3473: Filters GUI not detecting some regular expressions

## Environment info
- **Duplicati version**: 2.0.3.13_canary_2018-10-31
- **Operating system**: Windows 10, Firefox 63

## Description
The "Filters" web GUI when editing a backup job does not correctly identify some filters which are valid regular expressions. The backup engine correctly parses the regular expression so this is just a GUI issue.

## Steps to reproduce
1. Click "Edit as text" in the Filters section under Source Data.

2. Enter the following:
`-[.:\\Users\\[^\\]+\\OneDrive\\]`

3. Click "Edit as list".

- **Actual result**:
The GUI shows the entry as "Exclude File" instead of "Exclude regular expression". It also keeps the square brackets, which are usually removed for "Exclude regular expression".

- **Expected result**:
The GUI should show "Exclude regular expression" on the left and `.:\\Users\\[^\\]+\\OneDrive\\` on the right. It appears to be checking for the existence of * in the string to decide whether it is a regular expression or not. Shouldn't it just check if it starts with -[ and ends with ].

### LINKED ISSUE #3961: Execlude not works with "/" as last character
- **Duplicati version**: 2.0.4.32_canary_2019-10-24
- **Operating system**: Ubuntu 18.04
- **Backend**: Any

## Description

There is an inconsistency on the backup source editor page. When I add a backup path directory, I need to put a "/" at the end of the path - otherwise, it's just a file (the GUI warns for this).
But when I want to exclude a directory, I need to remove the "/' - for example, "/var/cache/" won't work,  need to use "/var/cache". Why? And why the GUI didn't warns me??

- **Actual result**:
Exclude /var/cache/ not works
- **Expected result**:
Gui warning or just remove the last character if it is a "/"

## Reproduce
https://youtu.be/oUWyPOUIzQk

## Screenshots
When adding a path, GUI warns me:

The only way it works, untraceable where I need to put "/" as the last character:

Jojo-1000:
Closes #2976, closes #3961, closes #3473, closes #2682.

Fixes bugs regarding filters in the web UI:
- The display of how filters with wildcards apply was bugged in the file selection list
- The input fields for filters could lead to inconsistent results in flags (e.g. double slash at end of directory filters, different data when switching types) (#3961)
- On the command line page, only the last exclude or include was applied (#2976)
- Regular expressions without wildcards were recognized as file excludes (#3473)
- Used regular expressions instead of globbing in restore picker (#2682)

All of these changes only impact the web UI, the backend is not affected.
The filters can still be inconsistent, but that would best be fixed by combining type select and input field in one component in a newer angular version. However, I did not find any with particularly unexpected behavior.
Example: enter `*.txt` in exclude filter, type changes to exclude extension but text still shows `*.txt` instead of `txt`, but internal flags are correct unless the text is changed again)

### DIFF
--- a/Duplicati/Server/webroot/ngax/scripts/directives/parseFilterType.js
+++ b/Duplicati/Server/webroot/ngax/scripts/directives/parseFilterType.js
@@ -3,11 +3,13 @@ backupApp.directive('parseFilterType', function(AppUtils, SystemInfo) {
         restrict: 'A',
         require: ['ngModel'],
         link: function(scope, element, attr, ctrl) {
-            var body = null;
-
             ctrl[0].$parsers.push(function(txt) {
                 var dirsep = SystemInfo.state.DirectorySeparator || '/';
-                return AppUtils.buildFilter(txt, body, dirsep);
+                // Store type in scope for parseFilterContent directive
+                // This works because ng-repeat creates a child scope for each iteration
+                // It is still glitchy in some edge cases, but should be more consistent
+                scope.filterType = txt;
+                return AppUtils.buildFilter(txt, scope.filterBody, dirsep);
             });
 
             ctrl[0].$formatters.push(function(src) {
@@ -18,8 +20,7 @@ backupApp.directive('parseFilterType', function(AppUtils, SystemInfo) {
                     return null;
                 }
 
-                body = parts[1];
-
+                scope.filterType = parts[0];
                 return parts[0];
             });
         }
@@ -31,23 +32,23 @@ backupApp.directive('parseFilterContent', function(AppUtils, SystemInfo) {
         restrict: 'A',
         require: ['ngModel'],
         link: function(scope, element, attr, ctrl) {
-
-            var type = null;
-
             ctrl[0].$parsers.push(function(txt) {
                 var dirsep = SystemInfo.state.DirectorySeparator || '/';
-                return AppUtils.buildFilter(type, txt, dirsep);
+                // Store type in scope for parseFilterType directive
+                scope.filterBody = txt;
+                return AppUtils.buildFilter(scope.filterType, txt, dirsep);
             });
 
             ctrl[0].$formatters.push(function(src) {
                 var dirsep = SystemInfo.state.DirectorySeparator || '/';
+                // Remove double trailing slashes
                 var parts = AppUtils.splitFilterIntoTypeAndBody(src, dirsep);
                 if (parts == null) {
                     type = null;
                     return null;
                 }
 
-                type = parts[0];
+                scope.filterBody = parts[1];
                 return parts[1];
             });
         }
--- a/Duplicati/Server/webroot/ngax/scripts/services/AppUtils.js
+++ b/Duplicati/Server/webroot/ngax/scripts/services/AppUtils.js
@@ -593,13 +593,25 @@ backupApp.service('AppUtils', function($rootScope, $timeout, $cookies, DialogSer
             return [type, body];
         }
 
-        for(var i in this.filterClasses) {
+        var m = [];
+        for (var i in this.filterClasses) {
             var n = matches(src, this.filterClasses[i]);
             if (n != null)
-                return n;
+                m.push(n);
         }
-
-        return null;
+        if (m.length == 0) {
+            return null;
+        }
+        // Select match with shortest body, meaning longest prefix and suffix
+        var shortestIdx = 0;
+        var shortestLen = src.length;
+        for (i = 0; i < m.length; ++i) {
+            if (m[i][1].length < shortestLen) {
+                shortestIdx = i;
+                shortestLen = m[i][1].length;
+            }
+        }
+        return m[shortestIdx];
     };
 
     this.buildFilter = function(type, body, dirsep) {
--- a/Duplicati/Server/webroot/ngax/scripts/services/AppUtils.js
+++ b/Duplicati/Server/webroot/ngax/scripts/services/AppUtils.js
@@ -422,6 +422,13 @@ backupApp.service('AppUtils', function($rootScope, $timeout, $cookies, DialogSer
       return str.replace( new RegExp( "(" + this.preg_quote(pattern) + ")" , 'g' ), replacement );
     };
 
+    this.globToRegexp = function (str) {
+        // Escape special chars, except ? and *
+        str = (str + '').replace(/([\\\.\+\[\^\]\$\(\)\{\}\=\!\<\>\|\:])/g, "\\$1");
+        // Replace ? and * with .? and .*
+        return str.replace(/(\?|\*)/g, ".$1");
+    };
+
     this.format = function() {
         if (arguments == null || arguments.length < 1)
             return null;
@@ -609,7 +616,7 @@ backupApp.service('AppUtils', function($rootScope, $timeout, $cookies, DialogSer
             if (rx)
                 filter = filter.substr(1, filter.length - 2);
             else
-                filter = this.replace_all(this.replace_all(this.preg_quote(filter), '*', '.*'), '?', '.');
+                filter = this.globToRegexp(filter);
 
             try {
                 res.push([flag == '+', new RegExp(filter, caseSensitive ? 'g' : 'gi')]);
@@ -621,9 +628,11 @@ backupApp.service('AppUtils', function($rootScope, $timeout, $cookies, DialogSer
     };
 
     this.evalFilter = function(path, filters, include) {
-        for(var i = 0; i < filters.length; i++) {
+        for (var i = 0; i < filters.length; i++) {
             var m = path.match(filters[i][1]);
-            if (m && m.length == 1 && m[0].length == path.length)
+            // Regex such as .* might match empty string at the end which is unwanted
+            // Check that the first match covers the full string
+            if (m && m.length >= 1 && m[0].length == path.length)
                 return filters[i][0];
         }
 
--- a/Duplicati/Server/webroot/ngax/scripts/services/AppUtils.js
+++ b/Duplicati/Server/webroot/ngax/scripts/services/AppUtils.js
@@ -259,12 +259,14 @@ backupApp.service('AppUtils', function($rootScope, $timeout, $cookies, DialogSer
             return new Date(dt);
     };
 
-    this.parseOptionStrings = function(val, dict, validateCallback) {
+    this.parseOptionStrings = function (val, dict, validateCallback) {
+        // Parse options and return a dict with dict['--key']=value options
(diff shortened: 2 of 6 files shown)


gpatel-fr:
Thanks for your contribution. Please understand that the team for reviewing PRs is very limited and priorities for next release and testing them are involving quite some work so I can't give a fixed time to read and study this PR unfortunately. 
It sure looks interesting but there are quite a lot of interesting PRs that I can't find time to look at.

Jojo-1000:
> Thanks for your contribution. Please understand that the team for reviewing PRs is very limited and priorities for next release and testing them are involving quite some work so I can't give a fixed time to read and study this PR unfortunately. It sure looks interesting but there are quite a lot of interesting PRs that I can't find time to look at.

That is understandable.

If anyone wants to test these changes without building from source, you can copy the [webroot](https://github.com/Jojo-1000/duplicati/tree/fix-filter-ui/Duplicati/Server/webroot) folder into the install folder (make a backup). This only works with installs that were not updated. You may need to restart the application and refresh the browser cache.

runmode:
@Jojo-1000 Thanks for your fix, I would gladly like to test it, but can't copy your files (no download zip available). How to?

runmode:
@Jojo-1000 Thank you for your contribution, however in my test with a leading asteriks I could not find an improvement. Hopefully my test was done correctly: 
quit Duplicati from Tray icon
backed up C:\Program Files\Duplicati 2\webroot 
downloaded you duplicati fork
unzipped webroot (415 files)
copied to C:\Program Files\Duplicati 2\
startet Duplicati2
loaded http://localhost:8200/ngax/index.html without browser cache
configuration - restore files
search for files: *2022*.jpg + Search button 
result: odd tree without text
search for files: 2022*.jpg + Search button 
result: perfectly filtered tree with text
search for files: *2022*.jpg + Search button (repeated)
result: odd tree without text

Jojo-1000:
> @Jojo-1000 Thank you for your contribution, however in my test with a leading asteriks I could not find an improvement.

Getting the patched files to work can be a bit difficult. Your approach is correct, but it might be that you are using an update which overrides the files.

@runmode Can you check that you are getting the updated sources? If you open the debugger in the browser on the GUI (inspect element) and go to Sources, ngax/scripts/services/AppUtils.js you should be able to find a `globToRegexp` function in the file (line 443). If that is not there, you are still getting the old scripts.

If you still get the old scripts, you should check in %localappdata%\Duplicati whether there are updates in the update folder. Then you would need to delete the update folder (those files cannot be changed) and install the latest Duplicati version from the installer, then repeat your steps above to patch the webroot.

gpatel-fr:
Hello

I have compiled this PR (with some others) into a private build and it seems to fix the 'odd tree' symptom of #2682. 
So @runmode I think that you have not done the changes right.

runmode:
Yes, I followed Jojo's instructions and checked the globToRegexp function was missing in line 443, found the update folder from the recent auto update, deleted it and installed v2.0.7. from the msi file. This way the search test worked well with the leading asteriks, also on my PC. :-)

But sadly, I noticed another strange behavior, while testing, without wildcards this time: 
FILES including ...
"photoshop" are found *),
"photo" not found,
"shop" neither.
"20221030" is found,
"2022" is not.
It looks as if the search term must have a minimum length of 6 or 7 characters to work.
Some files did'nt work that don't fit into this pattern.

FOLDERS were generally not found.

*) With "found/working" I mean that a filtered and expanded tree is presented with the search terms highlited.
"Not found/working" means that search terms were highlited, but not filtered out from the total tree which did not expand, so that it is not really possible to select the files.
Folders generally were highlited, but not filtered 
Every search was done after a reload without chache.

Jojo-1000:
@runmode I can reproduce your problems, but they are not caused by my changes. This already happens on the released version. There seem to be some other bugs in there. Also, when searching for part of a folder name, an empty '/' item appears in that folder.

runmode:
No doubt, good work, thank you again. I just thougt, perhaps you can also fix the other issues, which have not been documented before, as fas as I know. I hope your PR makes it into the next version!

Jojo-1000:
> perhaps you can also fix the other issues, which have not been documented before

@runmode It would be good if you can create a new issue for that with a minimal test backup set (which folders and files are needed and the search terms that do/don't work).

Jojo-1000:
The problems about searching folders reported by @runmode and also the appearing empty '/' items should be fixed by #4998.

Duplicati/Server/webroot/ngax/scripts/services/AppUtils.js
```diff
@@ -621,9 +628,11 @@ backupApp.service('AppUtils', function($rootScope, $timeout, $cookies, DialogSer
     };
 
     this.evalFilter = function(path, filters, include) {
-        for(var i = 0; i < filters.length; i++) {
+        for (var i = 0; i < filters.length; i++) {
             var m = path.match(filters[i][1]);
-            if (m && m.length == 1 && m[0].length == path.length)
+            // Regex such as .* might match empty string at the end which is unwanted
+            // Check that the first match covers the full string
```
gpatel-fr, line 667: I don't see when there could be several matches, could you please explain ?

Duplicati/Server/webroot/ngax/scripts/directives/parseFilterType.js
```diff
@@ -31,23 +32,23 @@ backupApp.directive('parseFilterContent', function(AppUtils, SystemInfo) {
...
+                scope.filterBody = txt;
+                return AppUtils.buildFilter(scope.filterType, txt, dirsep);
             });
 
             ctrl[0].$formatters.push(function(src) {
                 var dirsep = SystemInfo.state.DirectorySeparator || '/';
+                // Remove double trailing slashes
                 var parts = AppUtils.splitFilterIntoTypeAndBody(src, dirsep);
                 if (parts == null) {
                     type = null;
```
gpatel-fr, line 47: type is not defined anywhere, what could this be doing now ?

Duplicati/Server/webroot/ngax/scripts/services/AppUtils.js
```diff
@@ -302,6 +314,10 @@ backupApp.service('AppUtils', function($rootScope, $timeout, $cookies, DialogSer
             if (d['--' + k] !== undefined) {
                 DialogService.dialog(gettextCatalog.getString('Error'), gettextCatalog.getString('Duplicate option {{opt}}', { opt: k }));
                 return false;
+            } else if (k == 'include' || k == 'exclude') {
+                // Cannot specify filters in extra options
+                DialogService.dialog(gettextCatalog.getString('Error'), gettextCatalog.getString('Cannot specify filter include or excludes in extra options'));
+                return false;
```
gpatel-fr, line 320: I don't see how this can be called, could you please explain this code function ? as far as I can tell, it's not called when using the test-filter called and I don't see how it could ever be called since the filters are not part of extra options, but of the backup object.

Duplicati/Server/webroot/ngax/scripts/directives/parseFilterType.js
```diff
@@ -31,23 +32,23 @@ backupApp.directive('parseFilterContent', function(AppUtils, SystemInfo) {
         restrict: 'A',
         require: ['ngModel'],
         link: function(scope, element, attr, ctrl) {
-
-            var type = null;
-
             ctrl[0].$parsers.push(function(txt) {
                 var dirsep = SystemInfo.state.DirectorySeparator || '/';
-                return AppUtils.buildFilter(type, txt, dirsep);
+                // Store type in scope for parseFilterType directive
```
gpatel-fr, line 37: Should this comment not be 'Store content in scope' instead of 'store type in scope' ?

Duplicati/Server/webroot/ngax/scripts/services/AppUtils.js
Jojo-1000: You can test this on a regex website with JS flavor:
```js
m = "test".match(".*")
m[0] = "test"
m[1] = ""
```

`*` is greedy so it matches as much as possible. After that an empty string is left, which also matches, even though it is counter intuitive.

Duplicati/Server/webroot/ngax/scripts/services/AppUtils.js
Jojo-1000: If I remember correctly this can be triggered when you go to advanced options in the backup settings, edit as text and add `--include=x`, which is why I added this check.
Otherwise someone might add multiple include or exclude lines there, which would be ignored.

Duplicati/Server/webroot/ngax/scripts/services/AppUtils.js
gpatel-fr: From the comment I had assumed that .* actually meant "something.*" as I could not imagine it was a straight ".*" because such a criterium does not make any sense either for exclusion or inclusion. I had done a test with something actually useful like "a.*" and matches only returns one occurence. 
Could you make the comment more explicit by adding quotes around the RE ? It's nice to add comments but having not so clear comments does not help much in reading the code.

Issue #4579, opened 2021-07-20
duplicati-2.0.6.3-2.0.6.3_beta_20210617 just destoryed one month worth of backup


digulla:
- [x] I have searched open and closed issues for duplicates.
- [x] I have searched the [forum](https://forum.duplicati.com) for related topics.

----------------------------------------

## Environment info

- **Duplicati version**: 2.0.6.3
- **Operating system**: openSUSE 15.3
- **Backend**: Replace me

## Description

I had a failure of my main hardisk.
I restored most of the content from an old full-disk backup which was about one month old.
Then I installed duplicati to restore my daily backups.

There were all the config files and index locally. On the remote server, I had backups until two days ago!

When I clicked "repair database", duplicati marked all remote backups that it couldn't see in the LOCAL index for DELETION instead of recognizing that the REMOTE files were MORE RECENT and that the LOCAL index was OUTDATED.

The restore "failed" with 500+ errors like this one:

2021-07-20 21:58:37 +02 - [Error-Duplicati.Library.Main.Operation.RepairHandler-FailedNewIndexFile]: Failed to accept new index file: duplicati-i66607ecd144b438b8386a86f6794db3e.dindex.zip.aes, message: Volume duplicati-b390ddb4921f14b5288436c1656f42c7a.dblock.zip.aes has local state Deleting

When I checked on the remote server, the file mentioned was gone.

I checked my remote backups and there is no file left from the last month.

**THIS IS UNACCEPTABLE BEHAVIOR FOR A BACKUP SOFTWARE**

I'm very upset and disappointed.

Please advice how I should proceed.

Another problem: I couldn't stop duplicati from trying to create "new" backups. As soon as I started it, it began to create new daily backups which means that I get restore points which have a huge delta because of all the files which seem to be "lost".

## Steps to reproduce

1. Set up a backup
2. Wait a couple of days
3. Copy the local index to a safe place
4. Wait a couple more days
5. Delete the local index and restore it from the the copy in step 3.
6. Start duplicati2
7. Watch it destory your data

warwickmm:
Clearly, this is undesired behavior.  For help on how to proceed, please create a post in the forums.  There are more users there with experience with recovering from these types of situations.  From there, we can diagnose what went wrong and hopefully uncover some flaws that can be fixed.

https://forum.duplicati.com/

warwickmm:
[...]

PR #4968, opened 2023-06-16
Add safety check to repair that prevents deleting unknown new backups

Jojo-1000:
If any dlist files have a newer timestamp than the last database backup, the repair is aborted before remote files are deleted. This prevents data loss like in #4579, when an outdated version of the local database is used with a newer remote backup.

This affects both the CLI and GUI repair operations. An override is not provided. The user has to delete the additional dlist files manually if the data loss is intended. After that, leftover dblock files are cleaned up as usual.

### Steps to reproduce error:
- Create a backup
- Make copy of database
- Change files in backup and create a second version
- Replace database with old copy
- Run backup results in an error message that prompts to repair

### Current behavior:
- Repair finds additional files on the remote and deletes them
- The second version of the backup is lost

### New behavior:
- Repair checks that newer backup version exists and results in an error:
  `The remote files seem to be newer than the local database. Repair would cause data loss. Run recreate instead if the remote files are not corrupted.`
- The user has to run a database recreate
- No data is lost

I am not sure whether the new error message needs to be translated or not (most others in the repair command do not seem to be).


Jojo-1000:
Fixed by #5946

PR #4967, opened 2023-06-14
Fix missing file error caused by interrupted compact

Jojo-1000:
Before the dindex files are deleted on the remote, update the state in the database to Deleting. Previously, only dblock files were handled this way. Patch by warwickmm

Closes #4129

PR #4838, opened 2022-11-01
Top level log stats

taz-il:
Added top-level stats (duration, size uploaded, backup size) to the job-specific log page.

While one can expand each row and dig-up these stats, this change will make it easy to glance over multiple runs, and see that things are ok (e.g. durations are stable / proportional to upload size, backup size not exploding, etc).

image

Closes #4830 

(also - added NUnit 3 nuget test adapter, VSIX test adapters are deprecated)

PR #4978, opened 2023-06-23
Improve result reporting

### LINKED ISSUE #4344: Run-script-before with exit 1 saves metadata of aborted backup to statistics

## Environment info

- **Backend**: Local SATA Harddrive, USB Harddrive, USB Stick

## Description

I set up a `--run-script-before` batch script to return `exit 1` when my drive is not attached so that I will not get warnings. The backup information in the home tab of the web server does however change to a 'zero state' after silently aborting the backup due to exit 1 code.

On the forum I made a thread where the problem is discussed: [https://forum.duplicati.com/t/warning-folder-x-does-not-exist/10894/2](https://forum.duplicati.com/t/warning-folder-x-does-not-exist/10894/2) .

Another issue on Github is: [https://github.com/duplicati/duplicati/issues/4222](https://github.com/duplicati/duplicati/issues/4222) which is about a fatal error, also the `--run-script-after` command is also invoked. My issue is about the zero job statistics. When solving one might want to tackle both issues at the same time since there is a good chance the same parts of the code are involved. 

## Steps to reproduce

1. Create new backup, as basic as possible
2. Run the new backup in order to fill the current backup state information
2. Add run-script-before as advanced step, point it to `C:\Duplicati\check_if_X_exists.bat`: 
```
exit 1
```
3. Run backup again.

- **Actual result**:
The backup information which contained the size, date of last backup and versions changes to all zeros:
```
Last successful backup:
    Today at 12:45 PM (took 00:00:00)
    Run now 
Next scheduled run:
    Tomorrow at 12:45 PM
Source:
    0 bytes
Backup:
    0 bytes / 0 Versions
```
- **Expected result**:
I expect that the information of the last 'successful' / 'not aborted' backup will be displayed.

## Screenshots

A screenshot of an empty backup state 'Backup_Noord' and a backup with information 'Test_backup'
image

## Debug log

This is the log from About -> Show log -> Live. I think my log looks normal.
```
Oct 14, 2020 5:55 PM: Aborting operation by request, requested result: Normal
Oct 14, 2020 5:54 PM: Cannot open WMI provider \\localhost\root\virtualization\v2. Hyper-V is probably not installed.
Oct 14, 2020 5:54 PM: Cannot find any MS SQL Server instance. MS SQL Server is probably not installed.
Oct 14, 2020 5:54 PM: Using WMI provider \\localhost\root\virtualization\v2 
```

Many thanks for reading this issue. I am happy to assist in any way.

### LINKED ISSUE #4829: Failed runs missing from the job-specific log page

## Environment info

- **Duplicati version**: 2.0.6.3_beta_2021-06-17
- **Operating system**: Win 11 Pro, 22H2
- **Backend**: FTP

## Description

When a backup job fails, the failures only appear on the general server page (the one for all backups), and not on the log page of the individual backup job that failed. As an example, I have this job running daily. It failed consecutively on Oct 1, 2, 3, etc, but these failures don't show up (is it only for me? is this the intended behavior for some reason?)

## Steps to reproduce
This happens with every backup failure, so in order to reproduce, just cause a backup to fail (e.g. wrong user/pass for FTP; or any other way).

- **Actual result**:
image
- **Expected result**:
image

Expected: all backup attempts show up in the log (not only the successful ones), with a minimum of "Backup failed" title, ideally with some short description of the error (e.g. "host unreachable"), but this latter part is of course lower prio.

Jojo-1000:
Closes #4829, closes #4344

Add a new 'Interrupted' field to `IBasicResult`. This is set to true if the operation was interrupted (e.g. by `run-script-before`).
Interrupted and failed backups are recorded in the job log (#4829).
When the backup is interrupted or fails (`Interrupted==true` or `ParsedResult=Fatal`), metadata for the server database is not updated (#4344).

If exceptions are caught from an operation this previously resulted in an error message in the server log. Now the error message is also recorded in the job log. This has to create a new Operation in the database for now, so the logs are not linked to the performed changes in the database yet.

There are two new translation needed (to display "Fatal error" or "Interrupted" in the job log). I put these messages above the time, please suggest other options:
screenshot of log entry with fatal error

Also, maybe a darker shade of red and a different icon could be used for the icons of fatal errors, so they can be spotted easier. I don't have the tools installed to change the stylesheet.

### DIFF
--- a/Duplicati/Server/Runner.cs
+++ b/Duplicati/Server/Runner.cs
@@ -754,18 +754,18 @@ namespace Duplicati.Server
                 backup.Metadata["LastRestoreFinished"] = Library.Utility.Utility.SerializeDateTime(result.EndTime.ToUniversalTime());
             }
 
-            if (result is IParsedBackendStatistics r2)
+            if (result is IParsedBackendStatistics r2 && !result.Interrupted)
             {
                 UpdateMetadata(backup, r2);
             }
 
-            if (result is IBackendStatsticsReporter r3)
+            if (result is IBackendStatsticsReporter r3 && !result.Interrupted)
             {
                 if (r3.BackendStatistics is IParsedBackendStatistics statistics)
                     UpdateMetadata(backup, statistics);
             }
 
-            if (result is ICompactResults r4)
+            if (result is ICompactResults r4 && !result.Interrupted)
             {
                 UpdateMetadataLastCompact(backup, r4);
 
@@ -773,25 +773,28 @@ namespace Duplicati.Server
                     UpdateMetadataLastVacuum(backup, r4.VacuumResults);
             }
 
-            if (result is IVacuumResults r5)
+            if (result is IVacuumResults r5 && !result.Interrupted)
             {
                 UpdateMetadataLastVacuum(backup, r5);
             }
 
             if (result is IBackupResults r)
             {
-                backup.Metadata["SourceFilesSize"] = r.SizeOfExaminedFiles.ToString();
-                backup.Metadata["SourceFilesCount"] = r.ExaminedFiles.ToString();
-                backup.Metadata["SourceSizeString"] = Duplicati.Library.Utility.Utility.FormatSizeString(r.SizeOfExaminedFiles);
-                backup.Metadata["LastBackupStarted"] = Library.Utility.Utility.SerializeDateTime(r.BeginTime.ToUniversalTime());
-                backup.Metadata["LastBackupFinished"] = Library.Utility.Utility.SerializeDateTime(r.EndTime.ToUniversalTime());
-                backup.Metadata["LastBackupDuration"] = r.Duration.ToString();
+                if (!result.Interrupted)
+                {
+                    backup.Metadata["SourceFilesSize"] = r.SizeOfExaminedFiles.ToString();
+                    backup.Metadata["SourceFilesCount"] = r.ExaminedFiles.ToString();
+                    backup.Metadata["SourceSizeString"] = Duplicati.Library.Utility.Utility.FormatSizeString(r.SizeOfExaminedFiles);
+                    backup.Metadata["LastBackupStarted"] = Library.Utility.Utility.SerializeDateTime(r.BeginTime.ToUniversalTime());
+                    backup.Metadata["LastBackupFinished"] = Library.Utility.Utility.SerializeDateTime(r.EndTime.ToUniversalTime());
+                    backup.Metadata["LastBackupDuration"] = r.Duration.ToString();
 
-                if (r.CompactResults != null)
-                    UpdateMetadataLastCompact(backup, r.CompactResults);
+                    if (r.CompactResults != null)
+                        UpdateMetadataLastCompact(backup, r.CompactResults);
 
-                if (r.VacuumResults != null)
-                    UpdateMetadataLastVacuum(backup, r.VacuumResults);
+                    if (r.VacuumResults != null)
+                        UpdateMetadataLastVacuum(backup, r.VacuumResults);
+                }
 
                 if (r.FilesWithError > 0 || r.Warnings.Any() || r.Errors.Any())
                 {
@@ -837,19 +840,19 @@ namespace Duplicati.Server
                     );
                 }
             }
-            else if (result.ParsedResult != Library.Interface.ParsedResultType.Success)
+            else if (result.ParsedResult != ParsedResultType.Success)
             {
-                var type = result.ParsedResult == Library.Interface.ParsedResultType.Warning
+                var type = result.ParsedResult != ParsedResultType.Warning
                             ? NotificationType.Warning
                             : NotificationType.Error;
 
-                var title = result.ParsedResult == Library.Interface.ParsedResultType.Warning
+                var title = result.ParsedResult != ParsedResultType.Warning
                                 ? (backup.IsTemporary ?
                                 "Warning" : string.Format("Warning while running {0}", backup.Name))
                             : (backup.IsTemporary ?
                                 "Error" : string.Format("Error while running {0}", backup.Name));
 
-                var message = result.ParsedResult == Library.Interface.ParsedResultType.Warning
+                var message = result.ParsedResult != ParsedResultType.Warning
                                     ? string.Format("Got {0} warning(s)", result.Warnings.Count())
                                     : string.Format("Got {0} error(s)", result.Errors.Count());
 
--- a/Duplicati/Library/Interface/ResultInterfaces.cs
+++ b/Duplicati/Library/Interface/ResultInterfaces.cs
@@ -39,6 +39,7 @@ namespace Duplicati.Library.Interface
         IEnumerable<string> Warnings { get; }
         IEnumerable<string> Messages { get; }
         ParsedResultType ParsedResult { get; }
+        bool Interrupted { get; }
     }
 
     public interface IBackendStatstics
--- a/Duplicati/Library/Main/Controller.cs
+++ b/Duplicati/Library/Main/Controller.cs
@@ -434,6 +434,10 @@ namespace Duplicati.Library.Main
                     if (result.EndTime.Ticks == 0)
                         result.EndTime = DateTime.UtcNow;
                     result.SetDatabase(null);
+                    if(result is BasicResults r)
+                    {
+                        r.Interrupted = false;
+                    }
 
                     OnOperationComplete(result);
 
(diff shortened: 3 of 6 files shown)


taz-il:
Thanks for taking the time to fix #4829 😃

> I put these messages above the time, please suggest other options:
I would like to suggest different formatting, but first would like to discuss the various operation result status types:

If I understand correctly, you are referring to 4 types (success, warning, error, fatal error) - and thus also suggest a darker red hue for fatal errors.

What is the difference between an error and a fatal error?

Due to #4829, I only operations with either 0 warnings and errors or operations with warnings only, listed on the job specific log page. So suggesting any job with at least one error is equivalent to a fatal error (operation was not completed)

I would like to suggest we use 3 result types (at least when displaying to the user):
- green/success: operation completed with no warnings
- yellow/warnings: operation completed, but encountered at least one warning
- red/failure: operation not completed, due to encountering at least one error

This is aligned e.g. with compiler warning/error classification (where warning means the operation can continue, whereas error means the operation will fail).

I'm not familiar enough with Duplicati's codebase, suggesting sticking to 3 types of operation completion status from a user perspective.

Am I missing anything?

Jojo-1000:
> What is the difference between an error and a fatal error?

From a programming perspective (as implemented now):
A fatal error throws an exception that escapes the function that is supposed to run the operation. For a normal error, the operation returns normally, but there are error messages in the result log.

From a user perspective, maybe there are errors with only limited impact, so some of the backup still worked?

The fatal result code already existed, it was just never used. It is also useful to know that this run probably has no valid metadata (backup size etc.). We can of course display it the same as normal errors in the UI.

> Due to #4829, I only operations with either 0 warnings and errors or operations with warnings only, listed on the job specific log page. So suggesting any job with at least one error is equivalent to a fatal error (operation was not completed)

I don't think this is true. If you can produce an error that only writes an error message and does not throw an exception, or the exception is caught internally, you should get a result with errors in the job log. I think I had one of those for testing, but I would have to find it again.

taz-il:
Got it. In this case, I propose the following:

image

When expanded, each result contains the list of fatal errors (if any), errors, and warnings.

The "header row" (i.e. what's seen when collapsed) reflects the most severe status:
- Fatal error: Date, time - Operation: Blah failed
- Error: Date, time - Operation: Blah (x errors)    [no matter how many warnings, only errors show at the collapsed state]
- Warning: Date, time - Operation: Blah (y warnings)
- Success: Date, time - Operation: Blah

Proposed styling for the new "Fatal" state (ResultIcon):
`"fa fa-stop-circle fatal-color"`

Adding the following to base.less:
```
.fatal-color {
    color: #990000;
}
```

**Note:** I only have very partial understanding of your code change, but it seems to me the content of the fatal error itself (i.e. the message of the exception that caused the fatal error) might not currently be propagated and saved?
(the content that would appear once "Fatal errors" is expanded in my mock screenshot)

Jojo-1000:
@taz-il I like the design. There was also #4838 that integrates some stats. I guess those would only show for success then?
Do you think the square stop icon would be confused with a manually stopped job?

> Note: I only have very partial understanding of your code change, but it seems to me the content of the fatal error itself (i.e. the message of the exception that caused the fatal error) might not currently be propagated and saved?
(the content that would appear once "Fatal errors" is expanded in my mock screenshot)

Yes, right now it is written to the other error messages. I am not sure how the fatal error message could be distinguished from a normal error. There are separate fields for messages, warnings and errors in the result object, but not for fatal errors. Also the logger does not have a fatal category.

I could maybe add a single string for fatal error message (I don't think there could be two), but the whole result code is pretty tangled, with nested results which all need to be propagated.

Another compromise would be to replace the error section with fatal errors and assume that all of the error messages are relevant.

Jojo-1000:
@taz-il do you have an idea where to put the message for interrupted backups? Maybe it would be enough in the title line, or it could replace the right side (Source files), because those statistics would be collected after and are always zero.

I would also like to discuss how to call the "Interrupted" result. This currently appears when the script indicates that the backup should not run (with an exit code). Since that is *before* the actual job starts, maybe it should better be called "Skipped"? I also thought about "Aborted", but that has a connotation of an error which is not always the case.

Jojo-1000:
I decided to hide the source file statistics for interrupted backups and leave the fatal errors only for the header for now:
screenshot updated graphics

gpatel-fr:
Hello

Thanks, very interesting change.

A few observations: it's logical indeed to hide stats when the job has been interrupted; however when the job has failed from a fatal error, the stats are wrong too and should be hidden IMO. See below

Also I have tried to find another icon because the one you picked is difficult to make apart of the one for error, warning..., while this case is different. What do you think of this one ? (also advice wanted from @ts678, @taz-il ; if you want to search for a better one, the source is here: https://fontawesome.com/v4/icons/ - we can't use v5 and the brand icons are verboten. I have also considered the flash, the broken link; no icon has made me think 'ahah this is the one !')

image

As you can see I have also mixed #4976 to test. 

BTW I have used the procedure by @ts678 in https://github.com/duplicati/duplicati/issues/4129#issuecomment-599149786 to test this change, because I still want to commit #4967 and add an error message hinting to check the backend, something along the lines of 'if failed or interrupted log error message "consider to test your backend"', and I think this PR is the best place to add it (if you don't want to I can add a separate PR of course once this one has landed)

ts678:
> What do you think of this one ?

I guess that's "fire" on their page. I suppose that idea fits some cases. Are there any where this would come up upon request?

A more neutral approach to operations that just don't finish is to use minus-circle (to keep the circle theme) or minus-square. Minus suggests something is missing (specifically, completion is missing, right?). times-rectangle might work for an emphatic version of the current error which I think uses times-circle. I'm not sure I like tweaking the shade of red a little to make a fatal.

Jojo-1000:
An icon that would be noticably different and still fit with the circle would be "ban" 🚫, but the meaning is slightly off. "fire" works for the meaning, but I couldn't identify the icon on first glance and the shape doesn't fit the others.

> it's logical indeed to hide stats when the job has been interrupted; however when the job has failed from a fatal error, the stats are wrong too and should be hidden IMO.

I agree they are most likely wrong, but depending on when the failure occurred the stats might be helpful for debugging. So I'm not sure if we should hide them completely.

> something along the lines of 'if failed or interrupted log error message "consider to test your backend"',

That is a good idea, but I would only do it for failed, since interrupted is intended behavior and shouldn't break anything. I think you should be able to commit on this branch directly, if not just tell me the changes.

ts678:
> An icon that would be noticably different and still fit with the circle

I don't think it absolutely must fit with the circle, which is why one of my thoughts was a square which is noticeably different.
Regarding meaning, I proposed a minus (which seems to fit better than the reverse-slash inside a ban circle), but x might do, provided (ideally) something else changes to make it distinctive. Two other ways to indicate something was incomplete may include not shading in the circle (sort of like "ban" above), but are those harder to see? Maybe they stand out as "different"? "times-circle-o" would qualify, if you buy that argument. It uses usual error "x" but indicates it didn't finish, by being unfilled.

At the risk of complicating things, I think the tray icon is a Duplicati icon, and I don't think it distinguishes fatal from an error. Someone could look at the code. Maybe it's only red on a fatal error because there was also an error logged and recognized.

[Feature request: Additional tray icon status styles #3861](https://github.com/duplicati/duplicati/issues/3861) and associated PRs show some history about the visual tweaks there.

Following that thought into the total experience, I got an email about a failed backup (another Backblaze 503 with no retries). Looking closer, the status bar said `Error!` while my slow script after backup ran, and the tray icon was green while script ran.
Amusing temporary lack of coordination on the different signaling, but not a huge deal IMO, and eventually I got a lot of red.

I have `send-mail-level` set to `Warning,Error,Fatal`, but I now have a better idea of the difference between last two. Thanks!

gpatel-fr:
@ts678 

Here are your 2 ideas and 2 others based on the exclamation mark on a red background. I think that I like the most the last one, the triangle with an exclamation mark, because I think it's all right to have a different form for a different case (as in opposed to the 3 others, already existing cases: success, warning, error). Also the triangle reminds vaguely of the road sign 'stop' in most countries.

image

------------------------------------------

image

------------------------------------------

image


------------------------------------------


image

Thanks for your input on this subject; I don't trust my designer instincts...

@Jojo-1000 

with the case I use to test (the @ts678 test) the stats are wrong, but when going to the details, we can get at the backend stats for debugging:

image

on: only add a hint about verifying the backend in the case of failed, not interrupted you are right of course.

I'd prefer you do the needed changes on your side, it would save me the css recompilation.

drwtsn32x:
I like that triangle one the best, too.  Nice work ....

ts678:
I think the triangle has the advantage of being not-a-circle (so distinctive by shape), specifically suggestive (varies by country), filled with a color (so we can debate that -- and I still don't fully understand what we want to convey relative to an error color).

References from Wikipedia:

[Warning sign](https://en.wikipedia.org/wiki/Warning_sign)
[Hazard symbol](https://en.wikipedia.org/wiki/Hazard_symbol)
[Comparison of traffic signs in English-speaking territories](https://en.wikipedia.org/wiki/Comparison_of_traffic_signs_in_English-speaking_territories)
[Warning triangle](https://en.wikipedia.org/wiki/Warning_triangle)

Jojo-1000:
I created two separate messages, so that it is not confusing whether an interruption is a fatal error:
 
screenshot

x2rz90porgh3:
Hello

thanks for the effort and sorry for the delay.
I am closing this PR as it has been committed in https://github.com/duplicati/duplicati/commit/71eb03130aeba946c1845c7270488d9486e11da0

gpatel-fr:
Thanks for the effort, thanks all for the ideas. 
Sorry for the delay.
I'm closing this PR as it has been committed in https://github.com/duplicati/duplicati/commit/71eb03130aeba946c1845c7270488d9486e11da0

Duplicati/Server/Runner.cs
```diff
@@ -837,19 +840,19 @@ private static void UpdateMetadata(Duplicati.Server.Serialization.Interface.IBac
                     );
                 }
             }
-            else if (result.ParsedResult != Library.Interface.ParsedResultType.Success)
+            else if (result.ParsedResult != ParsedResultType.Success)
             {
-                var type = result.ParsedResult == Library.Interface.ParsedResultType.Warning
+                var type = result.ParsedResult != ParsedResultType.Warning
```
gpatel-fr, line 845: @Jojo-1000 

Err, I know I'm bit late but I suddenly wonder about this change. Why did you replace == by != exactly (same for the 2 following changes) ? 
It seems wrong. And when running a repair from the Web UI with a warning, it displays a red window saying '0 errors'

Jojo-1000: Yes I didn't catch that. The intention was to handle fatal the same as error, so I wanted to reverse the ternary operator. Apparently I forgot to reverse the order of the condition. But now that I look at this it would have worked in the original version anyways.

PR #4982, opened 2023-06-27
Fix "test" error: extra hashes

### LINKED ISSUE #4693: test all with full-remote-verification shows "Extra" hashes from error in compact

## Environment info

- **Duplicati version**: 2.0.6.100 Canary but problem is very old
- **Operating system**: Windows 10 Professional Version 21H2
- **Backend**: Local folder

## Description

Running the test command over `all` files with the full-remote-verification option may find errors, displayed as Extra: \<hash>
This has had quite a few forum reports, has been thought harmless, has bothered some people a lot, but was not understood.

Tracking it down well took a time-series (move/copy) of databases and an `rclone sync` with --backup-dir option to save files.
After each backup, I sync, test, and error if exit code is nonzero. I can then trace hash to compact, and with luck can go earlier.
A profiling log is not enough because even with profile-all-database-queries, some details one might want are just not visible.

## Steps to reproduce

1. Set up an unencrypted backup of a source folder, with a retention of 1 backup and with full-remote-verification option.
1. Download Duplicati 2.0.6.3, 2.0.6.100, 2.0.6.101 as `.zip` files. Consider having `.zip` and `.sqlite` tools around to look.
1. Create A.txt containing A. Without a line ending, the file's hash is VZrq0IJk1XldOQlxjN0Fq9SVcuhP5VWQ7vMaiKCP3/0=
1. Create B.txt containing B.
1. Create duplicati-2.0.6.3_beta_2021-06-17.zip
1. Backup 1
1. Delete A.txt
1. Create C.txt
1. Create duplicati-2.0.6.100_canary_2021-08-11.zip
1. Backup 2 and consider looking in the DeletedBlock table to find A.txt hash, or look in the dblock or dindex vol file for it
1. Create A.txt
1. Create duplicati-2.0.6.101_canary_2022-03-13.zip
1. Backup 3 and consider looking in the Block table to find A.txt hash, or look in the dblock or dindex vol file for it.
1. Delete duplicati-2.0.6.3_beta_2021-06-17.zip
1. Delete duplicati-2.0.6.100_canary_2021-08-11.zip
1. Backup 4 which results in a compact of the files from backup 1 and backup 2, but leaves the files from backup 3.
1. Run the `test` command on `all` files to see if the backup 3 files and backup 4 files coexist, or both claim to have A.txt

- **Actual result**:
Extra: VZrq0IJk1XldOQlxjN0Fq9SVcuhP5VWQ7vMaiKCP3/0= from dindex file and dblock file from the colliding backups
- **Expected result**:
No errors

The test sequence has each backup have a large file to make wasted space on demand, plus a small file to stop full deletion.
The A.txt file is special because it comes, goes, and returns. When it goes, then backup 2 happens, it goes into DeletedBlock.
When it returns, then backup 3 happens, it is not recovered from DeletedBlock (should it be?) thus gets a new entry in Block.
Deleting the large files initially in backups 1 and 2 makes wasted space to compact backups 1 and 2 together after backup 4.
Compact checks which blocks are still used, moving those to the new volume. A.txt came in backup 1 and again in backup 3.
Code confusion appears to be that the compact check looks in the Block table, sees the second A.txt, and so also keeps first.

https://github.com/duplicati/duplicati/blob/2f3b0e83a0c3e5dbdd8ccba0a08e3623d64ca231/Duplicati/Library/Main/Operation/CompactHandler.cs#L158-L162

https://github.com/duplicati/duplicati/blob/2f3b0e83a0c3e5dbdd8ccba0a08e3623d64ca231/Duplicati/Library/Main/Database/LocalDeleteDatabase.cs#L315-L332

I don't understand the design theory well enough to define a fix, but it seems like there were two possible points to stop this.
First might be to not let the A.txt block get duplicated from the second appearance of A.txt. Another might be to fix compact.

Once the error gets into the destination files, it can survive database recreate. 2.0.2.1 breaks backups but `test` doesn't detail.
Pointing a more recent Duplicati (e.g. 2.0.4.5) at the 2.0.2.1 destination for DB recreate and then test command details "Extra".

I'm still not sure if any further damage comes from this problem though. Now that I look for it, I see that I get these regularly.

Jojo-1000:
Closes #4693

This fixes multiple issues with deleted / duplicated blocks:
- Reuse existing deleted blocks to prevent uploading duplicated copies that can cause problems later
- When running recreate database, fill the DeletedBlock table for correct space calculations
- When running compact, check that there is no duplicate block in another volume which would be modified

## Steps to reproduce
[Error with compact (comment)](https://github.com/duplicati/duplicati/issues/4693#issue-1174148653)
[Error with recreate (comment)](https://github.com/duplicati/duplicati/issues/4693#issuecomment-1608125127)

### Duplicate blocks:
- Create backup with `--no-auto-compact` and `--no-encrypt`
- Create file A.txt with content A, backup 1
- Delete A.txt, backup 2
- Recreate A.txt with same content, backup 3

**Expected result:**
Backup 3 should not upload block for A.txt (hash `VZrq0IJk1XldOQlxjN0Fq9SVcuhP5VWQ7vMaiKCP3/0=`), as it is already contained in backup 1.

**Actual result:**
The dblock file for backup 3 contains a block with hash `VZrq0IJk1XldOQlxjN0Fq9SVcuhP5VWQ7vMaiKCP3/0=`, as does the dblock file for backup 1.

### Performance impact
Check the performance impact of reusing existing deleted blocks. If necessary, add an index over (Hash, Size) to the DeletedBlock table.
A test of the before/after performance on a large backup with deleted blocks would be appreciated.

The worst case performance could be tested as follows:
- Create a test scenario where large amounts of new files are added every backup
- 24% of those are deleted in the next version (less than the compact threshold, so those blocks are kept). It would need to be ensured that the deleted files are spread over the dblock volumes.
- Compare backup times before and after this change

### Validate database recreate change
The change in database recreate works in my tests, but I am not completely sure that the query catches all usages of blocks (I  first missed the BlocklistHash table, for example). It could be possible that some blocks are moved to deleted incorrectly.


ts678:
> I first missed the BlocklistHash table, for example

I meant to comment in this issue, where you wrote:

> check the Block table for blocks that do not appear in any file.

but you got the PR out first. I think that's all the flavors. A blockset might be data or metadata though (two aspects of a file).
[Reducing Time Spent Deleting Blocks](https://forum.duplicati.com/t/reducing-time-spent-deleting-blocks/16403/6?u=ts678) SQL shows that idea, although my proposed speedup rewrite didn't get much testing...

ts678:
> When running recreate database, fill the DeletedBlock table for correct space calculations

Does the current brute-force Deleting Blocks SQL in topic linked above reduce such needs?

https://github.com/duplicati/duplicati/blob/d0f1498bd41b151d8512fd2acb57739f6a05587f/Duplicati/Library/Main/Database/LocalDeleteDatabase.cs#L97

Of course it has to run, but it likely will eventually run. Regardless, are the goals equivalent?
That would give an opportunity to pick the best scheme, and use it in both of the locations.

Jojo-1000:
> Does the current brute-force Deleting Blocks SQL in topic linked above reduce such needs?

If I understand it correctly, the suggested change runs the query when specific files or filesets are deleted, so blocks don't have to be looked up in the full table.

In a database recreate, all of the blocks have to be examined to see if they are used, so I don't think the change can be applied here.

The current queries are very similar, although I use a temporary table to avoid looking up the block IDs twice.

ts678:
> the suggested change runs the query when specific files or filesets are deleted

PoC change would probably not fit here, which is why I suggest brute-force plan:

> all of the blocks have to be examined

Better formatted (courtesy of poorsql.com) version of the cited query looks like it looks at all of the blocks.
The part above UNION is blockset (file data and metadata) oriented. Below the UNION are blocklist blocks.

```
INSERT INTO "DeletedBlock" (
 	"Hash",
 	"Size",
 	"VolumeID"
 	)
SELECT "Hash",
 	"Size",
 	"VolumeID"
FROM "Block"
WHERE "ID" NOT IN (
 	 	SELECT DISTINCT "BlockID" AS "BlockID"
 	 	FROM "BlocksetEntry"
 	 	
 	 	UNION
 	 	
 	 	SELECT DISTINCT "ID"
 	 	FROM "Block",
 	 	 	"BlocklistHash"
 	 	WHERE "Block"."Hash" = "BlocklistHash"."Hash"
 	 	)
```

Questions about DISTINCT and UNION ALL versus UNION got many query variations benchmarked here:
[Backup Runtime after 2.0.7.1 update](https://forum.duplicati.com/t/backup-runtime-after-2-0-7-1-update/16378/69?u=ts678). It would be interesting to time your plan against the favorites there,
assuming of course that all give the correct answer. Maybe whichever way runs best gets put to wide use.
In the other use, there was also some moaning about having to do a slow query twice in a row. Any help?

ts678:
> When running recreate database, fill the DeletedBlock table

> Reuse existing deleted blocks

These actually pair well together for one occasional use case:

[Migrate from Linux to Windows](https://forum.duplicati.com/t/migrate-from-linux-to-windows/10292)
or more generally with a longer writeup across two requests:
[Changing source OS](https://forum.duplicati.com/t/changing-source-os/1010)

and there might be others, e.g. if database gets very broken.

Idea is to reattach source file blocks rather than reuploading.
Ideally a recreate can know them by just reading dindex files.
If new design puts the entire destination in DeletedBlock, it's
good that an initial backup on new OS etc. can get them out.

https://github.com/duplicati/duplicati/blob/d0f1498bd41b151d8512fd2acb57739f6a05587f/Duplicati/Library/Main/Operation/BackupHandler.cs#L413

kenkendk:
> This fixes multiple issues with deleted / duplicated blocks:
> 
> - Reuse existing deleted blocks to prevent uploading duplicated copies that can cause problems later
> - When running recreate database, fill the DeletedBlock table for correct space calculations
> - When running compact, check that there is no duplicate block in another volume which would be modified

I have taken the part "When running recreate database, fill the DeletedBlock table for correct space calculations" and implemented in #5624 as they were easy to test.

If have implemented the fix to remove broken files in #5625 
I will see if I can get the other two parts up as well.

kenkendk:
I added the final pieces in two PRs:
- #5626 
- #5627 
I am closing this PR as it has been split into multiple PRs

PR #4989, opened 2023-07-09
Improve source file picker

### LINKED ISSUE #4751: CRITICAL FEATURE REQUEST - Visual indication that folders are being excluded even when duplicati shows them as included
Since #4487 was closed and I can't comment on it, which deals with the fact that Nextcloud is considered a "system" folder, but unchecking the "Exclude -> System" box doesn't actually fix it, with the expected result of having those files included...  Instead, there is a note in the last comment indicating that you have to add a custom "Exclude Filter Group -> System" to get this to work.  

There is NO visual indicator until you actually go to restore a file that happens to be in the Nextcloud file structure, and then only that the folder tree isn't listed, despite it being listed when you go back to check your configuration and the list of included folders. 

I just lost weeks worth of work because of this due to a simple accidentally overwritten file, and on top of that have found that 2 YEARS worth of incremental backups were not taking place on my mission critical work files.... it's dumb luck that I didn't have a complete failure and need those files!  My business would have folded.

The "not my problem" response to that scenario in the other bug was even more concerning.  It didn't have to be a Nextcloud specific fix... the issue is that Duplicati shows that the file tree is included in one place, but then simply doesn't include it, and doesn't produce even a warning that files you've selected are not actually included.  This would happen with any folder with those attributes.... silently.

When selecting folders in the Source list, it should be exceedingly clear that when I select a folder that is considered a System folder, and I've unchecked the exclude System  filter, that those files are STILL NOT INCLUDED.

PLEASE ADDRESS THIS, you can't imagine how many people out there are "backing up" Nextcloud without realizing that they're not actually backing anything up.

Kevin.

Jojo-1000:
Closes #4751

Draft because #4962 has changes that are a prerequisite for this PR. Once that is merged, this PR can be rebased and merged.

## Improvements to source file picker
- Double click on a path in the `Source Data` section to jump to that position in the tree
- Fix a bug where sometimes the `Source Data` folders could be opened and show contents
- Also apply `Exclude` section in source file picker to show ❌ in the file list
  - Files with System/Temporary/Hidden file attributes are shown as excluded, *unless* they are explicitly included as a source. This follows the filter logic for the backup.
  - Files larger than the maximum size are shown as excluded (this can't be overruled by specifying as a source).
- Apply `Exclude filter group` rules to show ❌ in the file list
- Improve in-/exclude logic in file picker (click to toggle ❌/✔️):
  - Add filters for specific files to the beginning of the list. This will take precedence over more general filters lower in the list
  - If an include/exclude rule specifies one file to be in-/excluded, remove that rule to toggle
  - If a file is currently excluded by a general exclude filter, add an include filter to include it
  - If a file is currently excluded by an exclude attribute, add it as a source to include it
  - If a file is currently excluded by maximum size, show a message that it will not be included due to size

These changes make the source file selection more intuitive by showing more accurately what will actually be included (100% accuracy probably can't be achieved, especially because files change). This should resolve some of the confusion between the `System files` exclude (that works by a file attribute) and the system file filter group (that works based on path). It will also make it clear which files each of the filter groups includes.

There are also changes to simplify including/excluding single files, without needing to mess around with the details of filters.

## Impact
Most changes are in the web UI, but it was also required to modify the REST API to return more details in the file listing (should be backwards compatible) and add `/systeminfo/filtergroups` to get system specific filter groups. There is also a change in file enumeration to fix a warning when source files are included multiple times. Translations are needed for the new file size info dialog.

## Screenshots
**Excluding system file (S file attribute on Win):**
screenshot excluding system file

**Automatically added include filters for `B.txt` and `C.txt` after clicking on ❌ :**
screenshot with include rules

**Info dialog when attempting to include a file that is too large:**
dialog with file size info

**Showing files excluded with filter group:**
screenshot with excluded OS files


kenkendk:
@npodbielski This changes a bit in the REST methods for the FileSystem.

npodbielski:
Ok @kenkendk  thanks for info.

Issue #4988, opened 2023-07-07
list-broken-files doesn't notice missing folder metadata block


ts678:
- [x] I have searched open and closed issues for duplicates.
- [x] I have searched the [forum](https://forum.duplicati.com) for related topics.

----------------------------------------

## Environment info

- **Duplicati version**: 2.0.7.2 Canary
- **Operating system**: Windows 10
- **Backend**: Local folder

## Description

A missing folder metadata block is enough to break DB recreate which then says to `list-broken-files` but that finds no problem.
We're trying to debug some Sami_Lehtinen DB recreate problems [in the forum](https://forum.duplicati.com/t/ftp-delete-not-atomic-verified-transactional/12936/51?u=ts678), so I was doing some manual fault insertion testing...

## Steps to reproduce

1. Make an empty folder called "folder metadata".
1. Do an unencrypted backup of that empty folder.
1. View the dblock contents. Note the block name.
1. Delete the block and note the new file's size.
1. Open the dindex and copy file from vol folder.
It has the dblock name, so name might collide.
I often use a subfolder to avoid that problem.
1. Open file in text editor to delete that block.
Update the dblock file size, as noted earlier.
Copy and paste the edited text back in dindex.
1. Recreate button will complain about file hash.
Watch in About --> Show log --> Live --> Retry
Note actual hash and return to dindex to edit.
1. Recreate button should now recreate broken DB.
"Recreated database has missing blocks and 1 broken filelists. Consider using "list-broken-files" and "purge-broken-files" to purge broken data from the remote store and the database."
1. Run `list-broken-files`

- **Actual result**:
Nothing reported
- **Expected result**:
Problem reported

In comparison, forcing a missing block in file data or metadata reports the bad file. Is the problem here that this is a folder, I wonder?

## Debug log

Because `list-broken-files` and `purge-broken` files [give little output data since logging code was redone](https://forum.duplicati.com/t/error-during-auto-backup/16437/44?u=ts678), here's text from live log:

```
Jul 7, 2023 4:32 PM: The operation PurgeBrokenFiles has completed
Jul 7, 2023 4:32 PM: Purge completed, and consistency checks completed, marking database as complete
Jul 7, 2023 4:32 PM: Database was previously marked as in-progress, checking if it is valid after purging files
Jul 7, 2023 4:32 PM: No remote filesets were deleted
Jul 7, 2023 4:32 PM: Preventing removal of last fileset, use --allow-full-removal to allow removal ...
[...]

PR #4990, opened 2023-07-09
include files without blockset in broken files (issue #4988)

gpatel-fr:


Jojo-1000:
If I understand this change correctly, it was caused because empty files/folders don't have a blockset, but can still be broken due to metadata.

I guess the BROKEN_FILE_IDS subquery correctly finds the files, but they were discarded in the external query? Did you check all usages that they can handle `B.Length=null`, or should it be set to zero instead?

ts678:
Thanks for fast response, all. I'll stay pretty quiet since I think your SQL is better than mine.
I followed the general left join idea but didn't want to try to ask questions about the NULL.

My higher-level wonder was how well purge-broken-files purges a folder that's found bad.
Possibly that requires a build with the proposed change and a fancier backup to test with...

gpatel-fr:
@Jojo-1000 
normal files have 2 blocksets, one for data, one for metadata, directories have one blockset for metadata.
Yes the internal query finds bad files (it's named for that in LocalListBrokenFilesDatabase.cs) and it's working fine as far as I understand it.
I don't think the Null value is really used elsewhere. A quick test shows it displaying -1 for the length. Not very cute but it's better that not displaying anything.

@ts678 
I don't think that this is used to purge broken files. From browsing (quickly) the code in PurgeBrokenFileHandler.cs, the purge is not using the database at all. That's probably the reason it was working in your case while the list is not.

Jojo-1000:
@gpatel-fr
I made a test database with an empty file and it does have a blockset with size 0 in the database, so this is only an issue for folders and symlinks. I think for those we can keep the -1 size, this is consistent with the result from list files.
Otherwise you could use `IFNULL(B.Length,A.BlocksetID)` to output -100 for folders and -200 for symlinks, if it is necessary to distinguish the two.

ts678:
> I don't think that this is used to purge broken files.

Purge needs to figure it out too. Stack might look like:

https://github.com/duplicati/duplicati/blob/d0f1498bd41b151d8512fd2acb57739f6a05587f/Duplicati/Library/Main/Database/LocalListBrokenFilesDatabase.cs#L78-L80

https://github.com/duplicati/duplicati/blob/d0f1498bd41b151d8512fd2acb57739f6a05587f/Duplicati/Library/Main/Operation/ListBrokenFilesHandler.cs#L52-L55

https://github.com/duplicati/duplicati/blob/d0f1498bd41b151d8512fd2acb57739f6a05587f/Duplicati/Library/Main/Operation/PurgeBrokenFilesHandler.cs#L51-L57

> That's probably the reason it was working in your case while the list is not.

I'm not sure how well it was working, but I updated the issue with output after remembering --console-log-level must be raised, otherwise all the useful details get tossed. This is an unpleasant regression (IMO), introduced by the change of the logging code.

gpatel-fr:
@ts678 

I have changed the message log level to a Warning. It seems appropriate because asking for a purge that would remove the last fileset without specifying that one wants to do it is something of a mistake (while not being quite an error).

Jojo-1000:
The issue is #4988, it has steps to reproduce but the way it appeared in actual backups is not stated (probably a compact related bug that might be fixed by now).

Repair complains about a broken file, but list-broken-files does not show any.

Issue #4994, opened 2023-07-18
Alibaba oss error 阿里云对象存储报错


wodepig:
- [x ] I have searched open and closed issues for duplicates.
- [x ] I have searched the [forum](https://forum.duplicati.com) for related topics.

----------------------------------------

## Environment info
- **Duplicati version**:Duplicati - 2.0.7.1_beta_2023-05-25
- **Operating system**: unraid with docker 
- **Backend**: s3 compatible

## Description

When adding Alibaba Cloud Object Storage (OSS) with S3 Compatible, an error is reported:
```
The request signature we calculatedl does not match the signature you provided. Check your key and signing method.
Aws MultiChunkedEncoding is notsupported
```

## Steps to reproduce
1. Add a backup and set it to S3 Compatible
2. You must set this advanced option for OSS instances of Alibaba Cloud(OSS keys and secrets are normal)

4. Start the backup,stuck...

- **Actual result**:
stuck,and return 403
- **Expected result**:
backup and upload

## Screenshots
 Screenshots of errors, Alibaba Cloud OSS documentation, and S3 compatible configuration are attached:
![72b6d0cc04c892357988d39243d11e9](https://github.com/duplicati/duplicati/assets/42103087/a980a8dd-c687-4acc-bc83-bdba984a6ebf)
![de533df01147e964a777b94f49a4010](https://github.com/duplicati/duplicati/assets/42103087/42fa2af6-676a-47ac-919a-a6fab3da4651)
![cc74520cb8dd4623d7e9afcc31fb08c](https://github.com/duplicati/duplicati/assets/42103087/6e7fb6ea-326b-4e89-adb9-77211a904f06)
You must set this advanced option for OSS instances of Alibaba Cloud
![5a0a4f452a5fc286a0e9fc648cc7525](https://github.com/duplicati/duplicati/assets/42103087/1e546b01-536d-42f2-a33d-61abf8d9226a)
upload is stuck
![image](https://github.com/duplicati/duplicati/assets/42103087/2386f4ac-83dc-4391-922f-910da7660bcd)

## Debug log

```
[...]

Issue #523, opened 2014-08-05
"Object reference not set to an instance of an object" error for GoogleDocs backend


kenkendk:
_From [l.kwul...@gmail.com](https://code.google.com/u/117220724268324045587/) on December 27, 2011 08:03:05_

What steps will reproduce the problem? 1. Have a collection shared with you in Google Docs (see GoogleDocs.png)
2. Create a new backup to the Google Docs account
3. Click "Test connection" What is the expected output? What do you see instead? Expected result: "The connection succeeded" message
Actual result: "Object reference not set to an instance of an object" error (see error.png) What version of Duplicati are you using? On what operating system? I'm using Duplicati 1.3 and the latest from the trunk. Windows 7 x64. What backend (destination) are you using? Google Docs Please provide any additional information below. There is a NullReferenceException in \Duplicati\Library\Backend\GoogleDocs\GoogleDocs.cs. d.AtomEntry.EditUri is null for shared collection. I've created a quick workaround for the issue - GoogleDocs.cs.patch

 private static Dictionary<string, Google.Documents.Document> ParseFolderTree(IEnumerable<Google.Documents.Document> data)
        {
            Dictionary<string, Google.Documents.Document> parentlookup = new Dictionary<string, Google.Documents.Document>();
            Dictionary<string, Google.Documents.Document> dict = new Dictionary<string, Google.Documents.Document>();

```
        foreach (Google.Documents.Document d in data)
            parentlookup.Add(d.AtomEntry.EditUri.Content, d);
```

**Attachment:** [error.png GoogleDocs.png GoogleDocs.cs.patch](http://code.google.com/p/duplicati/issues/detail?id=523)

_Original issue: http://code.google.com/p/duplicati/issues/detail?id=523_

PR #4999, opened 2023-07-22
Add option to disable chunked encoding for AWS S3.

Jojo-1000:
Some S3 providers do not support MultiChunkedEncoding (#4994).

This adds an option for the S3 backend:
```
--s3-disable-chunk-encoding = false
  This disables chunk encoding for the aws client, which is not supported by
  all S3 providers.
```

It only affects the AWS client, because Minio either does not use this encoding, or does not allow to configure it.

### DIFF
--- a/Duplicati/Library/Backend/Idrivee2/Idrivee2Backend.cs
+++ b/Duplicati/Library/Backend/Idrivee2/Idrivee2Backend.cs
@@ -76,7 +76,7 @@ namespace Duplicati.Library.Backend
             string host= GetRegionEndpoint("https://api.idrivee2.com/api/service/get_region_end_point/" + accessKeyId);
 
 
-            m_s3Client = new S3AwsClient(accessKeyId, accessKeySecret, null, host, null, true, options);
+            m_s3Client = new S3AwsClient(accessKeyId, accessKeySecret, null, host, null, true, false, options);
 
         }
 
--- a/Duplicati/Library/Backend/S3/S3AwsClient.cs
+++ b/Duplicati/Library/Backend/S3/S3AwsClient.cs
@@ -42,11 +42,12 @@ namespace Duplicati.Library.Backend
         private readonly string m_locationConstraint;
         private readonly string m_storageClass;
         private AmazonS3Client m_client;
+        private readonly bool m_useChunkEncoding;
 
         private readonly string m_dnsHost;
 
         public S3AwsClient(string awsID, string awsKey, string locationConstraint, string servername,
-            string storageClass, bool useSSL, Dictionary<string, string> options)
+            string storageClass, bool useSSL, bool disableChunkEncoding, Dictionary<string, string> options)
         {
             var cfg = new AmazonS3Config
             {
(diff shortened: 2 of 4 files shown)


gpatel-fr:
Thanks for your contribution.

ts678:
> which is not supported by all S3 providers.

I wonder if our new issue on OCI (Oracle Cloud Infrastructure) is another such case, but have just some web research findings.
[Back Up to Cloudflare R2 Storage Fails](https://forum.duplicati.com/t/back-up-to-cloudflare-r2-storage-fails/15511) might be another, and had help from a Cloudflare person. At high level, notice request:

> Is there a possibility of adding an advanced option to allow streaming sig4 to be disabled on a given backup job?

which sounds like what this PR does, except by setting `UseChunkEncoding` = false instead of `DisablePayloadSigning` = true.
[PutObjectRequest](https://docs.aws.amazon.com/sdkfornet/v3/apidocs/items/S3/TPutObjectRequest.html) describes these. My forum talk with Cloudflare got too broad, but we could ask directly or maybe test them. They seemed interested in helping. Changing code just for them was unappealing, but this might be a broad industry problem.

Before we go further on this, has it been tested under throttling? S3 is now an IStreamingBackend. Does this option stop that?

https://github.com/duplicati/duplicati/blob/ec9d49d6849b3b37f965ac80be45e1460d21417c/Duplicati/Library/Interface/IStreamingBackend.cs#L25-L30

I'm dragging in streaming because common usage seems to speak of concepts like streaming sig4 (see Cloudflare above), and
[https://docs.aws.amazon.com/AmazonS3/latest/API/sigv4-streaming.html](https://docs.aws.amazon.com/AmazonS3/latest/API/sigv4-streaming.html) is what I suspect you're turning off with this option.
That page also mentions `STREAMING-AWS4-HMAC-SHA256-PAYLOAD` which showed up with Cloudflare as a nice clear can't-do-it:

> STREAMING-AWS4-HMAC-SHA256-PAYLOAD not implemented

Circling back to Oracle, a search of its STREAMING-AWS4-HMAC-SHA256-PAYLOAD status finds a number of results including
[AWS S3 To OCI Object Storage Bucket Transfer Failing (Doc ID 2858031.1)](https://support.oracle.com/knowledge/Oracle%20Cloud/2858031_1.html) (needs support account, but maybe OCI user has it):

> Error message- STREAMING-AWS4-HMAC-SHA256-PAYLOAD is not supported 

[Object storage](https://docs.gitlab.com/ee/administration/object_storage.html) at GitLab has a setting for `enable_signature_v4_streaming` with a default of `true`, but a warning for Oracle:

> Set to true to enable HTTP chunked transfers with [AWS v4 signatures](https://docs.aws.amazon.com/AmazonS3/latest/API/sigv4-streaming.html). Oracle Cloud S3 needs this to be false.

There were also some Google findings at GitHub on the topic. I didn't read all, but `fog-aws` needed to invent a similar option.
`Make S3 Signature v4 Streaming Optional #523` (not linking directly to avoid GitHub reference spam while we discuss this)

PR #5005, opened 2023-08-12
Enable write-ahead logging and memory mapped IO

samuel-w:
Prior pull request #4612, I broke it with a force push.

I set locking_mode to exclusive for highest compatibility with WAL, and got the mmap number from the SQLite documentation.


kenkendk:
Closing this in favor of having custom options configurable.

PR #5006, opened 2023-08-14
Modernize version of .NET P1

npodbielski:
…ing.csproj to netstandard2.0

First in a series PRs. Target here is to modernize version of .NET in duplicati. 
Process would be to:
1. Move all projects that we can to .netstandard 2.0 which is compatible with .net framework 4.7.1 and .net7.
2. Rewrite as little as possible just to make it work.
3. Move all the rest .net 4.7.1 projects to .net7 
4. Move all .net standard and .net7 projects to .net 8 which will be LTS pretty soon.

I did ran the tests and results were the same as in master - 2 fails, exact the same tests.


veleek (approved):

TopperDEL:
I just update the Storj DCS-backend to the latest uplink.NET-library, that needs net6 at least so does not work with Duplicati using netframework 4.7. So your PR would be good for the Storj-Backend. :)

npodbielski:
Can you merge it? 
Honestly I gave up on this. No maintainers seems to be interested in this. Which is perplexing to me. .net framework is supported only because Microsoft have it tied up so badly to whole OS that they can't get rid of it. But if they do this project is dead in the water. 

Working with this repository code is horrible experience on Linux - which is not the case with newer version of .net - because Azure uses it and Linux machines are way better for cloud. So moving to new version of .net seems like the only choice to keep this project forward. 

IMHO the maintainers only keep maintaining it because it have so many connectors to backends and people keep using it. If people are using it, when they are being ask for money ('you restored your files! great! pay us 10 euros!") probably some people are paying for this... thing, even if this application is constantly throwing errors and warnings and breaks remote backups by loosing files all the time - with S3 or Swift I had to check on daily basis if backup actually succeeded or not. 

And because it is .net framework, running through mono, on linux it is also very slow. Ah and probably SQL lite databases do not help either. I had like 1,3 TB of data to backup and this DB was like 3GB. I mean W T F. 

I personally moved to to [kopia](https://kopia.io/). It is much, much faster. Incremental snapshot of data takes like 3 mins for my 1,3TB (there are not that many files to change). Restoring 33GB of data takes like 10 mins, while for Duplicati it was 30-40 mins or hour, hour and a half depending on backend type. 
The only downside is that kopia runs only one backend at the time. So if you want 3 backends - you need 3 instances. 

Anyway if you guys want to i.e. fork it and move to .net 7 there - I totally can help. It is really a shame that such great project (Duplicati is mentioned on all '10 ten OS backup applications' articles) is going to waste because so lame people are maintaining it.

veleek:
I understand your frustration @npodbielski - I also spent some time and effort attempting to modernize some portion of this bit was stymied by being unable to get changes merging - but it's important to try and understand the position of the original maintainers.  

They don't have a responsibility for anything.  If they don't want to accept changes they don't have to. If they don't want to add maintainers they don't have to. This is super frustrating and I get it, but name calling doesn't really help.  

It's possible that the best way forward would be to fork it and start maintaining it yourself if you're interested. Triplicati anybody? 😁

npodbielski:
> They don't have a responsibility for anything. 

Oh they have! They have the responsibility to all the users using their software this way or another. And part of this responsibility is to actually 'maintain' the software, and part of this is to update the freaking framework :smile: 

> name calling doesn't really help.

"lame"? Hardly name calling, considering they supposed to maintain this project but they are not. Just asking for money. 
But yeah probably I should restrain myself, from it... but this is so hard, when I see code that was written 11 years ago! 

> Triplicati anybody? 😁

Yeah, why not :smile:

kenkendk:
@npodbielski If you have already built Triplicati, maybe we can use it :)

Otherwise, I have written my considerations and a plan for finally getting this fixed: https://forum.duplicati.com/t/battle-plan-for-migrating-to-net8/17482

I appreciate any input.

npodbielski:
> @npodbielski If you have already built Triplicati, maybe we can use it :)
Haha, no unfortunately I was not working on that since I am using Kopia and it is quite resilient - though I must say that I hate that it is not possible to use it with OAuth and multiple backup backends. 
This is what I really like about Duplicati, but it was in turn producing invalid backups :D I guess it is better to have resilient backup than multiple invalid backups :)

kenkendk:
Closing this as the changes to .NET8 are now merged

PR #5011, opened 2023-08-23
make the delay for retry-after static 

gpatel-fr:
so it's kept when the backend is recreated - it happens every time the retry-after is set. 
Also, as having the cloud throttling the connection is a serious event, make it a warning instead of a profiling event so it will not be hidden from the user.

### DIFF
--- a/Duplicati/Library/Backend/OAuthHelper/RetryAfterHelper.cs
+++ b/Duplicati/Library/Backend/OAuthHelper/RetryAfterHelper.cs
@@ -34,7 +34,9 @@ namespace Duplicati.Library
         // Whenever a response includes a Retry-After header, we'll update this timestamp with when we can next
         // send a request. And before sending any requests, we'll make sure to wait until at least this time.
         // Since this may be read and written by multiple threads, it is stored as a long and updated using Interlocked.Exchange.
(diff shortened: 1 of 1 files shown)


gpatel-fr:
I don't parse the meaning for 'each backend that uses it'. There is only one backend that uses it: OneDrive. Maybe it was intended to be shared with other drivers but so far it has not happened. That's why I have modified it as if it was under the OneDrive directory, because for all intends and purposes it's the case.

tygill:
A simple way to make it static per backend would be to have the helper take a `ref` parameter to track the state in. Each backend that wants to use it could then have its own static that it passes in each time it is called.

kenkendk:
The intention of the `RetryAfterHelper` is that it should be a reusable component.
I see that this fix solves a problem, but I agree with @Jojo-1000 that this mixes the use of timeouts between backends.
If we add a new backend that uses the class it will now cause problems because the two backends can set each others wait times.

kenkendk:
I added PR #5619 which has another take on this.
For the severity of the log message, I think it would be nice to have it visible, but on the other hand, it is something that can show up in a normal run, and is not really something the user can fix.

PR #5037, opened 2023-09-28
allow full-remote-verification to target only list and index files

gpatel-fr:
The option is turned into an enum with values True, False, ListIndexes
True and False are for compatibility with the existing backups.
ListIndexes is equivalent to True but don't handle block volumes. 
This is to make more workable a strategy of full remote testing after each backup.


gpatel-fr:
@ts678 

I agree that this is not limpid - however the facts on the ground are complicated to say the least..
I have spent a few hours trying to understand all the intricacies of this stuff.

I have attempted to make it clearer in (unfortunately not committed) https://github.com/kees-z/DuplicatiDocs/pull/104/commits/cda150979b2bbe51948944bf6202d73846b67be1

ts678:
The manual update confuses me even more, assuming it's trying to describe current working, not the one after this change.

> For tests launched independently of backups (test in command line mode), if defined, it is set by default to false. However, if not defined, a full remote verification is performed.

`--full-remote-verification` being like `--full-remote-verification=false` is opposite the usual boolean meaning, but maybe it's so (unless it's not so until the change is made). Proposed extension of the option makes it a non-boolean anyway.

"if not defined", to me, means option was not mentioned at all. I don't recall this doing internals check, but I could be wrong.

gpatel-fr:
This describes the current reality. Yes it's confusing but it's how it is.
'Not defined' means that the option is not included when starting the test. In this case the default is true and the full remote test is done unless I'm very mistaken.
When defining the option, it is set by default to false (invalidating it). That's kinda logical in a way. If you want to invalidate the default behaviour you have to add the option. If you check the option you just added, it's the same as if you had done nothing. 
It would have easier to explain if the behaviour were to NOT do the test by default and when adding the option to check it by default to enable full testing, but it was not done this way.
If you have a better idea on how to explain this, I'm all ears. The PR to the Duplicati Docs is still uncommitted.

ts678:
> In this case the default is true

Strange, but seemingly so, contrary to what source code would have me think, and the (maybe) auto-generated help answer.

https://github.com/duplicati/duplicati/blob/f6440a8e706b3c30ac430205853b21d50796d6cd/Duplicati/Library/Main/Options.cs#L359

Look at the `false` in the fifth argument position there, and I'm assuming that corresponds to `defaultValue` in code below

https://github.com/duplicati/duplicati/blob/f6440a8e706b3c30ac430205853b21d50796d6cd/Duplicati/Library/Interface/CommandLineArgument.cs#L201-L209

Also contrary to the convention of boolean options defaulting to false (thus the annoying need to think in double negatives).

Manual doesn't appear to discuss how boolean options work, but kenkendk does, clearly suggesting the default is false, [here](https://forum.duplicati.com/t/can-some-explain-verify-parameter-full-remote-verification/2644/2?u=ts678).

> all boolean options are false by default

gpatel-fr:
it was changed here:

commit b362614bf5161e5973582b108157108a343ed04d
Author: Kenneth Skovhede <kenneth@hexad.dk>
Date:   Fri Mar 3 20:52:04 2017 +0100

    Updated the verification to use deep verification to test archive readability when the verify is invoked directly.
    When invoking the verification after the backup, the deep verification is off by default.

It's off by default when backuping because it's slow, so it would be an pain to do it with every backup if the user specify to test the full backend with the backup-test-percentage option set to 100. Now if you think that this change was not correct, sorry but it has been in Duplicati since 6 years so it's not possible to change it now even if it makes documenting it rather awkward.

ts678:
Thanks for the finding. I guess there's an exception to the general way that needs to be kept for compatibility.

One other question was whether changing a Boolean to an Enumeration could break on old export file values,
however testing a current release suggests that everything's reasonable unless one's really asking for trouble.

GUI `Edit as text` to `--full-remote-verification` seems to always bounce back to true and export that.
If it was easy to export something other than `true` or `false` (e.g. an empty string), I'd wonder if it imports.
I can still force a Warning for either a Boolean or an Enumeration by forcing bogus value in, but it's diffiicult...

> 2023-10-05 19:42:02 -04 - [Warning-Duplicati.Library.Main.Controller-OptionValidationError]: The value "reallytrue" supplied to --full-remote-verification does not parse into a valid boolean, this will be treated as if it was set to "true"

> 2023-10-05 19:42:02 -04 - [Warning-Duplicati.Library.Main.Controller-OptionValidationError]: The option --console-log-level does not support the value "invalid", supported values are: ExplicitOnly, Profiling, Verbose, Retry, Information, DryRun, Warning, Error

Duplicati/Library/Main/Options.cs
```diff
@@ -1620,11 +1642,22 @@ public bool Dryrun
...
+        /// Gets a value indicating if the remote verification is deep
         /// </summary>
-        public bool FullRemoteVerification
+        public RemoteTestStrategy FullRemoteVerification
         {
-            get { return Library.Utility.Utility.ParseBoolOption(m_options, "full-remote-verification"); }
+            get
+            {
+                string policy;
+                if (!m_options.TryGetValue("full-remote-verification", out policy))
```
Jojo-1000, line 1652: To best preserve the old behavior, `--full-remote-verification` (with an empty string in the options dict) would need to result in true. Otherwise, existing configurations might silently change from true to false. 
However, this is different from the other enum options and might lead to other confusion. Maybe check for an empty string, set the result to true and print a warning that the option was changed and true should be specified explicitly (also I think an invalid enum should also print a warning).

gpatel-fr: @Jojo-1000 
what do you mean by 'an empty string in the options dict' ?
There is no dict in the database, only rows. If the option is true in the Web UI, we get this in the Option table:
1||--full-remote-verification|true
if it false, we get this:
1||--full-remote-verification|false
if the option is removed from the UI there is no row.
For the export json files, it is the same:

/bin/grep -A 1 -i remote sftp-duplicati-config-opttrue.json 
        "Name": "--full-remote-verification",
        "Value": "true",
~/Téléchargements$ /bin/grep -A 1 -i remote sftp-duplicati-config-optfalse.json
        "Name": "--full-remote-verification",
        "Value": "false",
~/Téléchargements$ /bin/grep -A 1 -i remote sftp-duplicati-config-noconfig.json 

So there should not be any problem when using the Web UI. I think that the problem you are referring to could be with people scripting the command line directly. I have tried to address this case with this update:

                string policy;
                if (!m_options.TryGetValue("full-remote-verification", out policy))
                    policy = "False";

                RemoteTestStrategy r;
                if (!Enum.TryParse(policy, true, out r))
                    r = RemoteTestStrategy.True;

                return r;

gpatel-fr:
@Jojo-1000 

>>What do you think about warning for invalid enum values?

Err, I think that a warning *is* generated in Command Line mode.

PR #5039, opened 2023-10-04
[WIP] Migrate AngularJS to Angular

Jojo-1000:
The web UI currently uses AngularJS which is [out of support](https://docs.angularjs.org/misc/version-support-status) since 2022. This means it gets no security fixes anymore. The version we are using is even older (from 2015).

There were a few attempts to rewrite the UI ([question about updating](https://forum.duplicati.com/t/angularjs-web-ui-is-very-old/13042?u=jojo-1000), [redesign in Vue](https://forum.duplicati.com/t/simple-web-ui-design-study/10148)) but they did not seem to go anywhere, be it due to lack of initiative or lack of encouragement.

I chose to use the successor Angular over Vue, because I did not want to change anything about the site structure or design (this can be tackled at a later stage, if resources permit). Everything is as close to a 1:1 copy as possible, while using more modern code patterns where it is feasible.

I know this is not high on the priority list, but I created this draft to keep track of progress, for others to contribute, and to prevent anyone from doing duplicated work if they also want to update the UI.

## Progress
- [x] convert html pages to the new template format
- [x] convert old directives and controllers to components
- [x] add, edit and restore wizards (including file pickers)
- [x] dialogs, status bar and notifications
- [x] editor components for all backend types in original UI
- [x] UI themes can be changed
- [x] translation support (different locales with single app build)
- [ ] update translations to new format (WIP, majority of translations can be automatically converted via a python script)
- [ ] verification and testing
- [ ] documentation
- [ ] check whether we need third party libraries for date formatting and cryptography
- [ ] verify that changes in XSRF handling can't be exploited
- [ ] incorporate UI changes made since

## Nice to have
- [ ] improve form validation (show obvious errors before running Test Connection, for example)
- [ ] keep state during some navigation transitions (e.g. when going back from restore to restoredirect)
- [ ] split less used parts into lazy-loaded modules to improve load times

## Usage
The new UI is running in parallel to the existing UI for now.

The process to set up the webserver pages is more complicated, because it is required to build the app into a precompiled form. The process is described in `webroot/angular/README.md`. It is also possible to serve via the angular CLI (for development) and use a proxy to forward to the Duplicati REST API.

Once built, the UI can be reached on the address `http://HOSTNAME/ng/` instead of `http://HOSTNAME/ngax/`. A browser cookie can be set on `http://HOSTNAME/theme.html` to automatically redirect to the new version.


AlexFRAN:
If anyone merges this request, i'll do a graphical redesign of the UI, since my old one is really really dated, but i agree with Jojo, that it makes no sense keeping AngularJS and his codebase seems good.

gpatel-fr:
@AlexFRAN 

>>If anyone merges this request

it will not be for next Canary for sure, but the PR is very interesting nonetheless.

>> i'll do a graphical redesign of the UI

Thanks for the proposal ! 

>>since my old one is really really dated

wow, 8 years ago.
Once this is into Duplicati, enhancements could be discussed on the forum, but there are more important things to fix first.

kenkendk:
Amazing progress @Jojo-1000 !
I think the size of this PR explains why it has just been left as AngularJS for so long :)

luixxiul:
Would you be so kind as to share the time schedule plan for this PR to be landed, if any?

I would like to work on #5379 and #5393 but I am wondering if I should wait until this will be implemented eventually.

Jojo-1000:
That is a matter of prioritization by @kenkendk. Right now there is a focus on the .NET update, and some webserver changes as well that would need to be included in this. Except for that, the majority of the work should be done for a 1:1 change.

I am also not completely happy with the localization, because the previous format for translations was much nicer (now there is just a lookup by ID in a json file, no context information at all).

There are a lot more proposed changes than time to test and merge them (just look at the open PRs).

kenkendk:
I am closing this as it has been inactive for 2 years.

My priority is to work on ngclient, but if there is still a desire to keep ngax going we can re-open this PR.

PR #5041, opened 2023-10-08
Fix leftover/extra index volume after interrupted upload and more tests

Jojo-1000:
- Fixes part of #5023
- Also adds more test cases to track some known reproducible issues (those that fail are category "Bug" and explicit only).

## Steps to reproduce
- See `IssueTests.Issue5023ReferencedFileMissing` and `RepairHandlerTests.RepairExtraIndexFiles`

# Current behavior
- After interrupted dindex upload (but the upload finished), no index-block link is created
- The next run uploads a new dindex file for the dblock
- When repairing with an extra index file, the invalid ID -1 is inserted into the link table
# New behavior
- Index-block link is always created
- The next run recognizes that the dindex file already exists and no extra is uploaded
- When repairing with an extra index file, the correct ID is inserted into the link table

## Impact
- Order of operations in repair handler was changed to fix the ID bug, but wrapped in a transaction to ensure that no inconsistencies occur.
- The index block link is created *before* the upload is known to be completed (instead of after the upload). It will be removed together with the `RemoteVolume` entry at the start of the next run.


kenkendk:
Closing this PR as it is included in #5610

PR #5059, opened 2023-11-12
set default log level to Dryrun if dryrun option is set

gpatel-fr:


Jojo-1000:
This certainly makes sense.

I also think that maybe the default console level (and also log file level) should be information. I don't know why this was chosen, but information is usually always the default.

It might be for backups this level spams too much, every backend operation for example, but it erases a lot of useful messages for all the other operations. Also I think console users should be able to handle the additional output, or increase the level manually.

gpatel-fr:
>>I don't know why this was chosen

me neither, that's why I'm not keen to change it. Before doing it, someone should go over the code to ponder the usefulness of each Information level logging message. I have no strong urge to procrastinate more.

ts678:
> This certainly makes sense.

and the clue to the user's intended output is painfully obvious. Figuring out what the user wants in other uses takes guessing.

Especially for a log file, I can see an argument for sticking to Warning and more severe, although one might regret choice later.
Context for the output is sometimes helpful. There was a nice display of one command showing vital output versus its context:

[Found 33 files that are missing from the remote storage, please run repair](https://forum.duplicati.com/t/found-33-files-that-are-missing-from-the-remote-storage-please-run-repair/16937/14?u=ts678)

I'm not an expert in logging internals, but it seems to me that some of the same output could go to the user watching live, or console output shell-redirected (or with Linux `tee`) to file, or log file option, or an email or such report, or LogData, or server.

There are different ways of making outputs. I looked at some context messages, and they used MessageEvent, thus WriteLine.
Output that seems usually useful only to a live watcher is the progress messages, for example those flow when I run a backup:

```         
Backup started at 11/12/2023 5:21:25 PM
Checking remote backup ...
  Listing remote folder ...
Scanning local files ...
  0 files need to be examined (0 bytes)
  375 files need to be examined (31.37 MB) (still counting)
  457 files need to be examined (574.11 MB) (still counting)
```

I don't use CommandLine that often, but the current default messaging in it looks pretty good to me without Information level. Other commands definitely fall short of ideal, especially those where we suspect a non-expert live view is asking for information.
Those should be informative about what's being done, and not only issues found, but what things were done that had no issue.

> I don't know why this was chosen, but information is usually always the default.

```
  --console-log-level (Enumeration): Console information level
    Console information level
    * values: ExplicitOnly, Profiling, Verbose, Retry, Information, DryRun,
    Warning, Error
    * default value: Warning
```

but making the console default to Information might be tolerable, maybe a bit noisier than it is now, e.g. on remote operations.

If you really wanted something that wasn't Warning (which can trigger alarms) you could add a `Console` level for console data:

https://github.com/duplicati/duplicati/blob/ec9d49d6849b3b37f965ac80be45e1460d21417c/Duplicati/Library/Logging/Log.cs#L47-L58

PR #5064, opened 2023-11-28
Added option to show sources for jobs on Home page

Pegoth:
Added the possibility to show sources on the Home page to be able to find a specific job easier.

image
image

### DIFF
--- a/Duplicati/Server/WebServer/RESTMethods/Backups.cs
+++ b/Duplicati/Server/WebServer/RESTMethods/Backups.cs
@@ -36,6 +36,12 @@ namespace Duplicati.Server.WebServer.RESTMethods
             var schedules = Program.DataConnection.Schedules;
             var backups = Program.DataConnection.Backups;
 
+            // Read sources for each backup if configured
+            var showSources = Program.DataConnection.GetSettings(Database.Connection.SERVER_SETTINGS_ID).FirstOrDefault(s => s.Name == "show-sources");
+            if (showSources?.Value != null && bool.TrueString.Equals(showSources.Value, StringComparison.OrdinalIgnoreCase))
+                foreach (var backup in backups)
+                    backup.Sources = Program.DataConnection.GetSources(long.Parse(backup.ID));
+            
             var all = from n in backups
                 select new AddOrUpdateBackupData {
                 IsUnencryptedOrPassphraseStored = Program.DataConnection.IsUnencryptedOrPassphraseStored(long.Parse(n.ID)),
--- a/Duplicati/Server/webroot/ngax/scripts/controllers/SystemSettingsController.js
+++ b/Duplicati/Server/webroot/ngax/scripts/controllers/SystemSettingsController.js
@@ -62,6 +62,7 @@ backupApp.controller('SystemSettingsController', function($rootScope, $scope, $r
         $scope.usageReporterLevel = data.data['usage-reporter-level'];
         $scope.disableTrayIconLogin =  AppUtils.parseBoolString(data.data['disable-tray-icon-login']);
         $scope.remoteHostnames = data.data['allowed-hostnames'];
+        $scope.showSources = AppUtils.parseBoolString(data.data['show-sources']);
         $scope.advancedOptions = AppUtils.serializeAdvancedOptionsToArray(data.data);
         $scope.servermodulesettings = {};
 
@@ -82,7 +83,8 @@ backupApp.controller('SystemSettingsController', function($rootScope, $scope, $r
             'startup-delay': $scope.startupDelayDurationValue + '' + $scope.startupDelayDurationMultiplier,
             'update-channel': $scope.updateChannel,
             'usage-reporter-level': $scope.usageReporterLevel,
-            'disable-tray-icon-login': $scope.disableTrayIconLogin
+            'disable-tray-icon-login': $scope.disableTrayIconLogin,
+            'show-sources': $scope.showSources
         };
 
         if ($scope.requireRemotePassword && $scope.rawdata['server-passphrase'] != $scope.remotePassword) {
--- a/Duplicati/Server/webroot/ngax/templates/home.html
+++ b/Duplicati/Server/webroot/ngax/templates/home.html
@@ -47,6 +47,13 @@
                         <p ng-click="doShowLog(item.Backup.ID)" translate>Show log …</p>
                         <p ng-click="doCreateBugReport(item.Backup.ID)" translate>Create bug report …</p>
                     </dd>
+
+                    <dt ng-show="item.Backup.Sources != null" translate>Sources:</dt>
+                    <dd ng-show="item.Backup.Sources != null">
+                        <ul>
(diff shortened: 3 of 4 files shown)


gpatel-fr:
Thanks for your contribution !
About the UI, the fact that the option in the User interface settings is above the language and color settings is hardly justified. My feeling is that these last 2 settings are more important. So it could be below these 2 settings.

I'd even tend to think that this change does not justify making the UI settings heavier and that this enhancement could just be the new behaviour (that is, without any parameter to enable/disable it: Duplicati already has a gazillion options, far too much in my opinion...)
If anyone is really using the UI to automate Duplicati so that their solution would be broken by the new UI, anyone is just wrong, the Python project is an example to do it right.

Thoughts (@ts678, @Jojo-1000, ... or anyone interested) ?

Jojo-1000:
> this enhancement could just be the new behaviour

I wanted to ask: Was there a specific motivation for this change?

From my point of view, the benefit of this is not that big. You already have the description field if you want to add more info about the backup. In my main backup, this source list is about 20 entries long. They also don't give a full picture about the contents, because filters still need to be looked up on the edit page.

I don't mind it as a visual option, but if it is always on I would expect a scroll area so that the statistics are not obscured when it is visible. However, @gpatel-fr is right that the number of options should not needlessly be increased.

ts678:
Basically, ditto, but I'll post anyway. I didn't yet get a chance to write the part about the filters also controlling the backup...

> to be able to find a specific job easier.

Wouldn't showing the optional Description be a more concise way to do that? For example, show it under Name on expand.
The example here only has 2 lines, but my rather small backup has 27. I could see someone having more than screen can fit.
Likely the end just drops off the screen end, but it should be tested. A long list will also mean people may want to unexpand
rather than just leaving job expanded. We've got some requests to be able to rearrange jobs order. This could increase need.

Pegoth:
The motivation was just that I wanted to see the exact sources for each job, nothing fancy.

Description did not show new lines properly, so it was not an option for me to create a list there, therefore I made this PR.

I placed the setting under UI, because it is relevant to the user interface, but I'm fine removing it from there and just placing it under advanced or something.

gpatel-fr:
>>Description did not show new lines properly
yes and that's not pretty (even if it's concise but when it's forced by software that's not a quality IMO)

I have tested this solution https://stackoverflow.com/questions/19684708/preserve-line-breaks-in-angularjs and it works but it has the drawback of adding 2 useless line feeds before the description.

To get back to your proposal, it could be changed easily by moving the ng-repeat="source in item.Backup.Sources">{{source}} thingy under the description field in the home.html template, then add an extended option named show-sources-in-job-description (just an idea, you can name it otherwise if it makes more sense) that could be added in the settings page, that would avoids making the settings page more heavy for an option that is not felt useful by everyone  (my original reservation).
so:
@Pegoth : could this change be easy enough for you ? you can add an extended option to Duplicati in Main/Options.cs and Main/Strings.cs, see an example in https://github.com/duplicati/duplicati/commit/68c36edbf7307f632aff41fec25d53f3edadc5db#diff-f2f3de51cabef256347ac28378ca3913bcfd30688540668a711b4029fab6fe6cR423 , and you can test for it in a template code with the Options['name-of-option'] syntax I think. Ideally the change could be made simpler, even if it means adding the sources to each backup regardless of the option (the load is not significant IMO).

@ts678 and @Jojo-1000: could it be more acceptable for you ?

Pegoth:
I would prefer if the UI of the job would be like what I did. The place where the config is doesn't really matter for me as I'll just set it to enabled once and done.

gpatel-fr:
sooooooo: @ts678  and @Jojo-1000 : would the same patch but with an extended option instead of the UI option be acceptable for you ?

Jojo-1000:
> you can add an extended option to Duplicati in Main/Options.cs

The issue is that this will show up in all extended options and also CLI help texts. I would prefer not to have any options that have no relevance to CLI in the backup settings. I think the Server UI settings is the right place for this, maybe below the theme.

> I would prefer if the UI of the job would be like what I did.

*If* there is a setting to show the source list, I think it should have its own section.

@Pegoth Can you try to add `white-space: pre-line` to the description field and see if it would be an acceptable compromise? I think it would be positive change either way. This will keep line breaks in the description, but remove extra whitespace and break too long lines.

In my opinion, the way this change is currently done is not bad. The only question is, do enough people need this to justify the added option. Given that at least @Pegoth wanted it bad enough to do the work, I would argue why not. Since this is only a server setting, we could also remove it in the future, if we decide it is not useful.

gpatel-fr:
>>I would prefer not to have any options that have no relevance to CLI in the backup settings

that's your preference. I don't see why there should be a rule forbidding having a job setting impacting the UI, and making this choice an extended setting would also turn it into a job setting (per-job configurable). The Settings page is a page for Server settings, not especially UI settings.

>>Can you try to add white-space: pre-line to the description field and see if it would be an acceptable compromise?

not for me.
Screenshot_20231210_141501

The extra white line before the description would be a regression. And asking for fixing this particular problem without having any idea of the complexity seems unfair. It could be easy or it could turn around to be a monster of a hellish problem involving different browsers implementations or standard conflicts or whatever.

Jojo-1000:
You might have it at the wrong element:

```html
<span ng-show="item.Backup.Description">
    <dt translate>Description:</dt>
    <dd style="white-space: pre-line;">{{item.Backup.Description}}</dd>
</span>
```
If the `dd` tag has the content on one line without whitespace, it is also possible to have the user completely control whitespace:

Before:
description before change

With `pre-line`:
description with pre-line whitespace

With `pre-wrap`:
description with pre-wrap whitespace



The CSS seems to be [supported since 2009-2013 in all major browsers](https://developer.mozilla.org/en-US/docs/Web/CSS/white-space#browser_compatibility)

gpatel-fr:
At the moment I'm unable to repro this. With pre-line the superflous white line still displays. With pre-wrap it's worse:

image

ts678:
I'm surprised it's so hard to put text in a box, but I don't code GUI. Ideally, any method would be portable beyond AngularJS. 
Would browser web developer tools be able to see what's on the page, maybe to give some clue if things aren't looking nice?

If we return to the source list plan, first I'll note that there's a request to sort it, instead of displaying it in original entry order, however a tempting (though I don't know what code or HTML magic it needs) path to source details is to click on Source line.
We're doing something like that on lines in About --> Show log

gpatel-fr:
>>You might have it at the wrong

oh yes, I started on a wrong hypothesis all right, now it's working.
Given that, I'd say that this PR is not needed after all, fixing this problem in the Description display allow to solve it without adding any specific code. Do you want to do a PR for that ?

gpatel-fr:
@Pegoth 
thanks again for the effort, as the idea did not agree with everyone, it appears that the better solution is to fix the description field bug, something I merged just now into Duplicati. I agree that it needs more manual work especially when the sources change, but on the other hand I also prefer not to load the UI with non essential features.

PR #5066, opened 2023-12-12
change query for GetBlocklists in LocalDatabase.cs

gpatel-fr:
to avoid duplicates caused by versions using the same block(s)

credits (https://forum.duplicati.com/t/any-way-to-recover-backup-repair-and-purge-broken-files-dont-help/17048): ts678 for the use case, @jojo-1000 for the fix

### DIFF
--- a/Duplicati/Library/Main/Database/LocalDatabase.cs
+++ b/Duplicati/Library/Main/Database/LocalDatabase.cs
@@ -1294,16 +1294,38 @@ ORDER BY
             m_insertIndexBlockLink.ExecuteNonQuery();
         }
 
+        /// <summary>
+        /// Creates a list of triplets (blocklist hash, block content, size)
+        /// for a given remote block volume
+        /// the block content is itself a concatenation of the hashes of blocks
+        /// the size is the number of hashes in the block content times the hash size
+        /// </summary>
+        /// <param name="volumeid">The ID of the block volume to use</param>
+        /// <param name="blocksize">The block size for the backup</param>
+        /// <param name="hashsize">The hash size for the backup</param>
+        /// <param name="transaction">An optional external transaction</param>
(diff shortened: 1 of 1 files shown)


Duplicati/Library/Main/Database/LocalDatabase.cs
```diff
@@ -1294,16 +1294,38 @@ public void AddIndexBlockLink(long indexVolumeID, long blockVolumeID, System.Dat
...
             {
-                var sql = string.Format(@"SELECT ""A"".""Hash"", ""C"".""Hash"" FROM " +
-                    @"(SELECT ""BlocklistHash"".""BlocksetID"", ""Block"".""Hash"", * FROM  ""BlocklistHash"",""Block"" WHERE  ""BlocklistHash"".""Hash"" = ""Block"".""Hash"" AND ""Block"".""VolumeID"" = ?) A, " +
-                    @" ""BlocksetEntry"" B, ""Block"" C WHERE ""B"".""BlocksetID"" = ""A"".""BlocksetID"" AND " +
-                    @" ""B"".""Index"" >= (""A"".""Index"" * {0}) AND ""B"".""Index"" < ((""A"".""Index"" + 1) * {0}) AND ""C"".""ID"" = ""B"".""BlockID"" " +
-                    @" ORDER BY ""A"".""BlocksetID"", ""B"".""Index""",
-                    blocksize / hashsize
+                // DISTINCT is used to remove duplicates that can happen when several backup versions uses the same block
+                var sql = string.Format(
+                   @"SELECT DISTINCT A.Hash AS BlocklistHash, C.Hash AS BlockHash, B.""Index"" " +
```
Jojo-1000, line 1313: Unfortunately, I don't think a select distinct is going to work here. The "Index" is not guaranteed to always be the same:

Example: fill file with 32769 "a" characters (one full blocklist + "a"), then change to 65536 "b" (two full blocklists) + "a" at the end (total 65537), the second "a" blocklist will have index 2 and your query will fail (I tested this just now). This is why I resorted to `GROUP BY` in my fix.

PR #5067, opened 2023-12-12
Keep user-defined whitespace in description field

Jojo-1000:
This allows users to use the description better, for example by making a list.
It uses the CSS property `white-space: pre-wrap`, which keeps all whitespace but inserts line breaks if necessary.

## Screenshot
description with pre-wrap whitespace


gpatel-fr:
Thanks.

PR #5069, opened 2023-12-13
do a vacuum of bug report database

gpatel-fr:


ts678:
Thanks! Getting this out will make it simpler to get bug reports again.
On alternative code, is below a missed opportunity from 6 years ago?

https://github.com/duplicati/duplicati/blame/e2c77b5310293fba655ce3bd8191df8e9fc87599/Duplicati/Library/Main/Operation/CreateBugReportHandler.cs#L68-L71

PR #5094, opened 2024-02-04
Fixes connect to server not working when a password is used for the UI.

### LINKED ISSUE #3091: Connect to server not working
- [ ] I have searched open and closed issues for duplicates.

## Environment info

- **Duplicati version**: 2.0.2.1_beta_2017-08-01
- **Operating system**: Debian Stretch
- **Backend**: Chromium browser

## Description

The "Reconnect to server..." does not work anymore. Neither the automatic one after some delay nor by pressing the button. Only reloading the site will help

## Steps to reproduce

1. Open Duplicati and log in on the web interface
2. Quit Duplicati, wait a while and restart it

- **Actual result**:
The application won't reconnect to Duplicati without reloading the page

- **Expected result**:
The application should connect automatically or after pressing the button after Duplicati has started again.

kellycampbe:
Fixes #3091 

This issue occurred when a password was set on the web UI, and the web client lost a connection to the server.  The web client would try to display a message to click to log in again, but it would not display for two reasons. First, the 401 unauthorized check in the appservice was not returning a promise so it was causing the Angular app to stop. Second, the connection message was appearing on top of the DialogController message, so when a 401 occurs my commit updates it to change the status to connected so it will then display the "Not Logged In" dialog.

### DIFF
--- a/Duplicati/Server/webroot/ngax/scripts/services/AppService.js
+++ b/Duplicati/Server/webroot/ngax/scripts/services/AppService.js
@@ -53,7 +53,6 @@ backupApp.service('AppService', function($http, $cookies, $q, $cookies, DialogSe
                 DialogService.accept('Not logged in', function () {
                     window.location = appConfig.login_url;
                 });
-                return;
             }
             deferred.reject(response);
         });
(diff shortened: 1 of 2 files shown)


gpatel-fr:
Thanks for your contribution, seems to work just fine.
Looks good to me, planned for next Canary unless objections.

Issue #4810, opened 2022-08-20
Stuck on "Waiting for upload to finish" if remote SSH storage is full


wonx:
- [ x ] I have searched open and closed issues for duplicates.
- [ x ] I have searched the [forum](https://forum.duplicati.com) for related topics.

----------------------------------------

## Environment info

- **Duplicati version**: 2.0.6.3_beta_2021-06-17
- **Operating system**: Debian Buster
- **Backend**: SFTP

## Description

I save my backups to a hard drive in a remote location, using ssh (sftp). Until now I have used rsync, but I wanted to try some alternatives. I tried Duplicati today, created a new backup settings trying to backup a folder with less than 100MB containing a few pdf files and setting my SFTP server as the backend. Tested the connection and seemed to work fine.

However, I start the backup process, and soon it gets stuck at "Waiting for upload to finish". I even left it overnight and it didn't go away. I could see that duplicati created the backup folder and the backup files, but with 0 byte size. After much examining, I noticed that the backup drive was almost full, with only the small reserved system space left (less than 5%). I tried with another folder in another drive, and the backup completed successfully.

Couldn't be a way to detect that the remote destination is full and warn the user, instead of showing "Waiting for upload to finish" indefinitely?

## Steps to reproduce

1. Set Duplicati to save the backup on a SFTP backend
2. Use a backend without enough free space
3. Start backup

- **Actual result**:
Duplicati shows a warning that the backup couldn't complete due to not enough free space available.
- **Expected result**:
"Waiting for upload to finish" indefinitely.

gpatel-fr:
Hello

I can't repro with 2.0.6.104_canary. 
Beware that aborting a job can be very long if your backup if of any significant size and you have the default retry count of 5. 

Duplicati error handling is not the best that can be in this case, because it's trying to send the first block created (that should be up to max block size, so 50 MB by default and as such could take some time if your upload speed is not great) and wait until it's all sent to check for errors, and then retry 5 times.

While I agree that 1) when there is 0 bytes left and there is 100 MB to upload the SSH library returns (probably) an error very quickly and that Duplicati should abort the block retry immediately, and 2) after a block fails Duplicati should check the error reason and if it's a permanent error such as disk full, should skip the remaining retries, the facts of life is that to be performant Duplicati has implemented asynchronous transfers, and async programming is hard and is often the root cause of decidedly non optimal error handling.

So the final answer is that no, Duplicati is not stuck.

wonx:
Hi,

[...]

PR #5101, opened 2024-02-16
update sshnet to 2023.0.1

gpatel-fr:
Should fix https://github.com/duplicati/duplicati/issues/4810


gpatel-fr:
Should fix #4810

PR #5107, opened 2024-02-26
Upgrade to sdk style projects

kenkendk:
This PR is a starting point for further upgrade to .Net8.

This PR updates all .csproj files to use the new SDK style format, removes the `AssemblyInfo.cs` and `packages.config` files.

The only complication is the `mozroots` project which no longer compiles as the dependent libraries have been removed.
Since the inclusion of this tool was a workaround for broken SSL certificates in earlier Mono installs, the impact should be limited. Once the .Net8 migration is completed, the tool will not have any impact anyway.

Since the .Net8 based projects are now platform dependent in the published state, each of the tray icons needs to be a separate project, which is not handled by the current PR.


kenkendk:
Handled with #5111

PR #5123, opened 2024-03-08
Standardize build product names for releases

DeflateAwning:
Currently, build products/installers on the GitHub Release tags aren't named in a standard way, and the `deb` release is missing the notice that it's a canary release:

image

This update fixes this naming inconsistency. 

It's possible that there's a reason the `rpm` and `deb` package names were different than the rest, but I can't think of it, so I think this is a reasonable change.


kenkendk:
The reason for the inconsistent naming was a requirement by RPM and DEB respectively.
See this for .deb: https://www.debian.org/doc/manuals/debian-faq/pkg-basics.en.html#pkgname
(arch was added later)
See this for RPM: https://docs.oracle.com/en/database/oracle/oracle-database/19/ladbi/rpm-packages-naming-convention.html#GUID-04FBD99C-77A8-4E31-9C8D-5B6B2EAE68DB

I would love to have consistent naming, but I don't agree with rpm/deb, and I think arguing for either repo to change their convetion is futile :D

DeflateAwning:
Hmm okay, well I guess that's the answer then. Thanks!

PR #5130, opened 2024-03-15
Feature/change updater for multi arch

kenkendk:
This PR removes the support for launching a sub-process with updates.
The new logic is that the user will visit the download page and install the update manually.

At a later point, we can automate the download and execution of the installer package.
Most of the logic for downloading is maintained, so once we have a good understanding of what packages can update without (too much) user intervention, we can re-activate this part.

The update manifest format needs to update to support the multi-arch packages, so there is no longer a single file entry, and it no longer makes sense to list the files inside the archive, as they will change over time.


Duplicati/Library/AutoUpdater/InstallerEntry.cs
```diff
@@ -1,35 +1,66 @@
...
+        public string PackageTypeId;    
+
+        /// <summary>
+        /// Gets the name of the package file
+        /// </summary>
+        /// <returns>The filename of the package</returns>
+        public string GetFilename()
+        {
+            var guess = Path.GetFileName(new Uri(RemoteUrls[0]).LocalPath);
+            if (string.IsNullOrWhiteSpace(guess))
```
npodbielski, line 60: I would put this code in some service or other code that is using this this data, IMHO data entry should not be 'guessing' about itself either data is correct or not.

kenkendk: I agree that the data should be correct. 

The guessing here is to guard against cases where the URL does not translate to a valid local path, which depends on the host system and may change. Since this is part of the updater, I would like it to be as robust as possible, so future updates or flawed packages does not crash the client.

PR #5151, opened 2024-04-22
Minor updates

achims311:
I noticed ther are 2 options in the systemd unit file, which would enhance the idle setting.

AS well the documentation of the new release part missed a parameter


kenkendk:
Looks good!

Issue #418, opened 2014-08-05
If the backup is stopped due to shutdown or exit, it should be resumed on startup


kenkendk:
_From [gbaeurle@gmail.com](https://code.google.com/u/gbaeurle@gmail.com/) on May 23, 2011 08:12:55_

What steps will reproduce the problem? 1. Restart Duplicati (system reboot) after the previous backup was not completed.
2. Click on the icon "Duplicati is currently paused" and choose "Resume" but instead of resuming the previously unfinished backup Duplicati says "Waiting for next backup"

What is the expected output? 
I would expect that the previously unfinished backup is continued. What version of the product are you using? On what operating system? 1.2 What backend (destination) are you using? WebDAV Please provide any additional information below. In addition to this unintuitive behaviour I wonder why Duplicati does not show the size of the initial backup (which was stopped before being completed because the application was closed and the PC rebooted.

**Attachment:** [screenshot.jpg](http://code.google.com/p/duplicati/issues/detail?id=418)

_Original issue: http://code.google.com/p/duplicati/issues/detail?id=418_

PR #5157, opened 2024-04-24
Removed code for supporting .NET < 4.8

kenkendk:
Defensive code for not supporting TLS1.3 removed

### DIFF
--- a/Duplicati.Library.RestAPI/WebServer/Server.cs
+++ b/Duplicati.Library.RestAPI/WebServer/Server.cs
@@ -160,21 +160,14 @@ namespace Duplicati.Server.WebServer
                 
                     var server = CreateServer(options);
 
-                    if (!certValid)
-                        server.Start(listenInterface, p);
+                    if (certValid)
+                    {
+                        var secProtocols = System.Security.Authentication.SslProtocols.Tls12 | System.Security.Authentication.SslProtocols.Tls13;
+                        server.Start(listenInterface, p, cert, secProtocols, null, false);
+                    }
                     else
                     {
-                        var secProtocols = System.Security.Authentication.SslProtocols.Tls12;
-
-                        try
-                        { 
-                            //try TLS 1.3 (type not available on .NET < 4.8)
-                            secProtocols = System.Security.Authentication.SslProtocols.Tls12 | (System.Security.Authentication.SslProtocols)12288;
-                        }
-                        catch (NotSupportedException)
-                        {
-                        }
-                        server.Start(listenInterface, p, cert, secProtocols, null, false);
+                        server.Start(listenInterface, p);
                     }
 
                     m_server = server;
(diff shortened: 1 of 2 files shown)


ts678:
What happens if the OS doesn't support TLS 1.3?

ts678:
[Protocols in TLS/SSL (Schannel SSP)](https://learn.microsoft.com/en-us/windows/win32/secauthn/protocols-in-tls-ssl--schannel-ssp-) says Windows 10 has TLS 1.3 as "Not Supported" for client or server.
I'd just like to make sure that Duplicati still lands nicely at TLS 1.2 on any systems that don't support 1.3.
I don't know the answer, so I thought I'd ask.

kenkendk:
I had not considered that any recent systems would lack this, so thanks for pointing it out. 

I think I will set it to "system default" which is what MS recommends. This should pick whatever is best, given the system.

ts678:
[SslProtocols Enum](https://learn.microsoft.com/de-de/dotnet/api/system.security.authentication.sslprotocols?view=net-8.0) suggests `None` rather than `Default` (which has an outdated meaning). I don't know which you meant.

ts678:
The MS recommended None (poor choice of words by MS) verssus SystemDefault (deprecated or forbidden) came up again.

I'm not looking at FluentFTP, but I thought I'd report what 2.0.8.1 and 2.0.9.101 (looks like change is in 2.0.9.100) say in UI.

2.0.8.1

```
C:\Duplicati\duplicati-2.0.8.1_beta_2024-05-07>Duplicati.CommandLine.exe help allowed-ssl-versions
  --allowed-ssl-versions (Flags): Sets allowed SSL versions
    This option changes the default SSL versions allowed. This is an advanced
    option and should only be used if you want to enhance security or work
    around an issue with a particular SSL protocol.
    * values: Ssl3, Tls, Tls11, Tls12, SystemDefault, Tls13
    * default value: SystemDefault
```

> --allowed-ssl-versions=Tls12,SystemDefault,Tls13

image

2.0.9.101

```
C:\Duplicati\duplicati-2.0.9.101_canary_2024-06-27-win-x64-gui>Duplicati.CommandLine.exe help allowed-ssl-versions
  --allowed-ssl-versions (Flags): Sets allowed SSL versions
    This option changes the default SSL versions allowed. This is an advanced option and should only be used if you want to enhance security or work around an issue with a particular SSL protocol.
    * values: SystemDefault, Ssl3, Tls, Tls11, Tls12, Tls13
    * default value: SystemDefault
```

> --allowed-ssl-versions=SystemDefault

image

[Changes to Kestrel's default supported TLS protocol versions #418](https://github.com/dotnet/aspnetcore/issues/22563) announced Kestrel's 2020 move to [SslProtocols.None](https://docs.microsoft.com/en-us/dotnet/api/system.security.authentication.sslprotocols), and the Duplicati move to SystemDefault concerns me. Would you like me to test more (including Wireshark) to see what it's negotiating?

Issue #5167, opened 2024-04-30
Fix build warnings for platform specific items


kenkendk:
The codebase is using a custom platform detection in `Duplicati.Library.Utility` which can be updated to use the standard .NET [`OperatingSystem` class](https://learn.microsoft.com/en-us/dotnet/api/system.operatingsystem?view=net-8.0).

Methods and classes that are OS specific should be marked with `[SupportedOSPlatform("windows")]` attributes, and any warnings related to calling operating system specific code should be fixed.

Jojo-1000:
There are 167 warnings about incorrect calls. Most of them can be resolved by using the attribute and correct platform check except for Captcha/CaptchaUtil: This will need to be reworked, because it uses System.Drawing, which needs GDI+ and is [only supported on Windows](https://learn.microsoft.com/en-us/dotnet/api/system.drawing?view=net-8.0#remarks).
> If you can't use System.Drawing with your application, recommended alternatives include [ImageSharp](https://sixlabors.com/products/imagesharp), [SkiaSharp](https://github.com/mono/SkiaSharp), [Windows Imaging Components](https://learn.microsoft.com/en-us/windows/desktop/wic/-wic-about-windows-imaging-codec), and [Microsoft.Maui.Graphics](https://github.com/dotnet/Microsoft.Maui.Graphics).

kenkendk:
I have fixed the Captcha stuff as part of dropping `HttpServer`. I am using ImageSharp for now.

kenkendk:
This is fixed in source

PR #5172, opened 2024-05-03
Add SupportedOSPlatform attribute to Windows-specific classes

SimonNyvall:
I hope this is correct, and that I understood the task at hand.
#5167 

Added the marker 
`[SupportedOSPlatform("windows")]`

On the Win32.cs and WinTools.cs classes.


SimonNyvall:
Just let me know if I missed anything. :)

kenkendk:
@SimonNyvall I think there are more places where this is needed, but the I see no reason for not including the fixes you added.

PR #5207, opened 2024-05-29
Feature/nullable support utility

bigboybamo:
Made changes to files under `Utility` Library

### DIFF
--- a/Duplicati/Library/Utility/AsyncHttpRequest.cs
+++ b/Duplicati/Library/Utility/AsyncHttpRequest.cs
@@ -158,7 +158,7 @@ namespace Duplicati.Library.Utility
             }
 
             if (m_state == RequestStates.GetRequest)
-                return (Stream)m_asyncRequest!.GetResponseOrStream();
+                return (Stream)(m_asyncRequest?.GetResponseOrStream() ?? throw new InvalidOperationException("The async request is null."));
 
             if (m_state != RequestStates.Created)
                 throw new InvalidOperationException();
@@ -176,7 +176,7 @@ namespace Duplicati.Library.Utility
         public WebResponse GetResponse()
         {
             if (m_state == RequestStates.GetResponse)
-                return (WebResponse)m_asyncResponse!.GetResponseOrStream();
+                return (WebResponse)(m_asyncResponse?.GetResponseOrStream() ?? throw new InvalidOperationException("The async response is null."));
 
             if (m_state == RequestStates.Done)
                 throw new InvalidOperationException();
--- a/Duplicati/Library/Utility/AsyncHttpRequest.cs
+++ b/Duplicati/Library/Utility/AsyncHttpRequest.cs
@@ -45,11 +45,11 @@ namespace Duplicati.Library.Utility
         /// <summary>
         /// The request async wrapper
         /// </summary>
-        private AsyncWrapper m_asyncRequest = null;
+        private AsyncWrapper? m_asyncRequest = null;
         /// <summary>
         /// The response async wrapper
         /// </summary>
-        private AsyncWrapper m_asyncResponse = null;
+        private AsyncWrapper? m_asyncResponse = null;
         /// <summary>
         /// The request/response timeout value
         /// </summary>
@@ -158,7 +158,7 @@ namespace Duplicati.Library.Utility
             }
 
             if (m_state == RequestStates.GetRequest)
-                return (Stream)m_asyncRequest.GetResponseOrStream();
+                return (Stream)m_asyncRequest!.GetResponseOrStream();
 
             if (m_state != RequestStates.Created)
                 throw new InvalidOperationException();
@@ -176,7 +176,7 @@ namespace Duplicati.Library.Utility
         public WebResponse GetResponse()
         {
             if (m_state == RequestStates.GetResponse)
-                return (WebResponse)m_asyncResponse.GetResponseOrStream();
+                return (WebResponse)m_asyncResponse!.GetResponseOrStream();
 
             if (m_state == RequestStates.Done)
                 throw new InvalidOperationException();
@@ -206,11 +206,11 @@ namespace Duplicati.Library.Utility
         /// </summary>
         private class AsyncWrapper
         {
-            private readonly IAsyncResult m_async = null;
-            private Stream m_stream = null;
-            private WebResponse m_response = null;
+            private readonly IAsyncResult? m_async = null;
+            private Stream? m_stream = null;
+            private WebResponse? m_response = null;
             private readonly AsyncHttpRequest m_owner;
-            private Exception m_exception = null;
+            private Exception? m_exception = null;
             private readonly ManualResetEvent m_event = new ManualResetEvent(false);
             private readonly bool m_isRequest;
             private bool m_timedout = false;
@@ -248,7 +248,7 @@ namespace Duplicati.Library.Utility
                         var wex = ex;
                         if (ex is WebException exception && exception.Response == null)
                         {
-                            WebResponse resp = null;
+                            WebResponse? resp = null;
 
                             try { resp = r.GetType().GetProperty("Response", System.Reflection.BindingFlags.Instance | System.Reflection.BindingFlags.NonPublic)?.GetValue(r) as WebResponse; }
                             catch {}
@@ -304,9 +304,9 @@ namespace Duplicati.Library.Utility
                     throw m_exception;
 
                 if (m_isRequest)
-                    return m_stream;
+                    return m_stream!;
                 else
-                    return m_response;
+                    return m_response!;
             }
         }
     }
--- a/Duplicati/Library/Utility/BlockingQueue.cs
+++ b/Duplicati/Library/Utility/BlockingQueue.cs
@@ -30,7 +30,7 @@ namespace Duplicati.Library.Utility
     /// </summary>
     public class BlockingQueueAsEnumerable<T> : IEnumerable<T>
     {
-        private IEnumerator<T> m_enumerator;
+        private IEnumerator<T>? m_enumerator;
 
         public BlockingQueueAsEnumerable(BlockingQueue<T> queue)
         {
@@ -60,7 +60,7 @@ namespace Duplicati.Library.Utility
         {
             private readonly BlockingQueue<T> m_queue;
             private readonly bool m_first = true;
-            private T m_current;
+            private T? m_current;
 
             public BlockingQueueEnumerator(BlockingQueue<T> queue)
             {
@@ -91,7 +91,8 @@ namespace Duplicati.Library.Utility
             {
                 get
                 {
-                    return m_current;
+
+                    return m_current!;
                 }
             }
             #endregion
@@ -101,7 +102,7 @@ namespace Duplicati.Library.Utility
             {
                 get
                 {
-                    return m_current;
+                    return m_current!;
                 }
             }
             #endregion
@@ -217,7 +218,7 @@ namespace Duplicati.Library.Utility
                     }
 
                     if (m_completed)
-                        return default(T);
+                        return default(T)!;
 
                     m_itemsProduced.Reset();
                 }
--- a/Duplicati/Library/Utility/BlockingQueue.cs
+++ b/Duplicati/Library/Utility/BlockingQueue.cs
@@ -92,7 +92,7 @@ namespace Duplicati.Library.Utility
                 get
                 {
 
-                    return m_current ?? throw new InvalidOperationException("The current value is null.");
+                    return m_current;
                 }
             }
             #endregion
--- a/Duplicati/Library/Utility/BlockingQueue.cs
+++ b/Duplicati/Library/Utility/BlockingQueue.cs
@@ -92,7 +92,7 @@ namespace Duplicati.Library.Utility
                 get
                 {
 
-                    return m_current!;
+                    return m_current ?? throw new InvalidOperationException("The current value is null.");
                 }
             }
             #endregion
--- a/Duplicati/Library/Utility/CallContextSettings.cs
+++ b/Duplicati/Library/Utility/CallContextSettings.cs
@@ -32,7 +32,7 @@ namespace Duplicati.Library.Utility
 {
     public static class SystemContextSettings
     {
-        private static string defaultTempPath = null;
+        private static string? defaultTempPath = null;
 
         private struct SystemSettings
         {
@@ -54,7 +54,7 @@ namespace Duplicati.Library.Utility
             }
         }
 
-        public static IDisposable StartSession(string tempdir = null, long buffersize = 0)
+        public static IDisposable StartSession(string? tempdir = null, long buffersize = 0)
         {
             if (buffersize < 1024)
                 buffersize = 64 * 1024;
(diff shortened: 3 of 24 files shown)


Duplicati/Library/Utility/AsyncHttpRequest.cs
```diff
@@ -158,7 +158,7 @@ public Stream GetRequestStream(long contentlength = -1)
             }
 
             if (m_state == RequestStates.GetRequest)
-                return (Stream)m_asyncRequest.GetResponseOrStream();
+                return (Stream)m_asyncRequest!.GetResponseOrStream();
```
kenkendk, line 161: Can we check if this is null and throw instead of suppress?

Duplicati/Library/Utility/AsyncHttpRequest.cs
```diff
@@ -176,7 +176,7 @@ public Stream GetRequestStream(long contentlength = -1)
         public WebResponse GetResponse()
         {
             if (m_state == RequestStates.GetResponse)
-                return (WebResponse)m_asyncResponse.GetResponseOrStream();
+                return (WebResponse)m_asyncResponse!.GetResponseOrStream();
```
kenkendk, line 179: Can we check if this is null and throw instead of suppress?

Duplicati/Library/Utility/BlockingQueue.cs
```diff
@@ -91,7 +91,8 @@ public object Current
             {
                 get
                 {
-                    return m_current;
+
+                    return m_current!;
```
kenkendk, line 95: Are we sure this cannot be null?

Duplicati/Library/Utility/CallContextSettings.cs
```diff
@@ -253,13 +253,13 @@ public static T Settings
             {
                 var key = ContextID;
                 if (string.IsNullOrWhiteSpace(key))
-                    return default(T);
+                    return default(T)!;
```
kenkendk, line 256: Can this not be null?

Duplicati/Library/Utility/CallContextSettings.cs
```diff
@@ -277,11 +277,11 @@ public static T Settings
         /// </summary>
         /// <returns>The disposable handle for the context.</returns>
         /// <param name="initial">The initial value.</param>
-        public static IDisposable StartContext(string settingsKey, T initial = default(T))
+        public static IDisposable StartContext(string settingsKey, T? initial = default(T))
         {
             contextSettingsType = settingsKey;
             var res = new ContextGuard();
-            Settings = initial;
+            Settings = initial!;
```
kenkendk, line 284: Can this be done better? If we allow it to be null, then `Settings` should be nullable as well?

Duplicati/Library/Utility/CallContextSettings.cs
```diff
@@ -253,13 +253,13 @@ public static T Settings
...
-                    return default(T);
+                    return default(T)!;
 
 
-                if (!_settings.TryGetValue(key, out T res))
+                if (!_settings.TryGetValue(key, out T? res))
                     lock (_lock) // if not present but context id is not null, wait
                         if (!_settings.TryGetValue(key, out res))
-                            return default(T);
+                            return default(T)!;
```
kenkendk, line 262: This one looks like it could be null

Duplicati/Library/Utility/BlockingQueue.cs
```diff
@@ -217,7 +218,7 @@ public T Dequeue()
                     }
 
                     if (m_completed)
-                        return default(T);
+                        return default(T)!;
```
kenkendk, line 221: This could technically return `null`?

Duplicati/Library/Utility/DirectStreamLink.cs
```diff
@@ -252,7 +252,7 @@ private void readerClosed()
                 if (m_readerClosed) return;
                 m_readerClosed = true;
                 m_signalBufferAvailable.Set(); // unblock potentially waiting writer
-                m_readerStream = null;
+                m_readerStream = null!;
```
kenkendk, line 255: It guess it should be nullable if we set it to null?

Duplicati/Library/Utility/DirectStreamLink.cs
```diff
@@ -266,7 +266,7 @@ private void writerClosed()
                 if (m_writerClosed) return;
                 m_writerClosed = true;
                 m_signalDataAvailable.Set(); // unblock potentially waiting reader
-                m_writerStream = null;
+                m_writerStream = null!;
```
kenkendk, line 269: It guess it should be nullable if we set it to null?

Duplicati/Library/Utility/Power/LinuxPowerSupplyState.cs
```diff
@@ -57,12 +57,12 @@ private bool IsAC()
                 {
                     if (!reply)
                     {
-                        string sourceType = File.ReadLines(Path.Combine(source, "type")).FirstOrDefault();
+                        string? sourceType = File.ReadLines(Path.Combine(source, "type")).FirstOrDefault();
```
kenkendk, line 60: Feel free to use `var` :)

Duplicati/Library/Utility/Power/MacOSPowerSupplyState.cs
```diff
@@ -52,7 +52,7 @@ private PowerSupply.Source GetSourcePmset()
                 };
 
                 var pi = System.Diagnostics.Process.Start(psi);
-                pi.WaitForExit(1000);
+                pi!.WaitForExit(1000);
```
kenkendk, line 55: maybe add:
```
var pi = System.Diagnostics.Process.Start(psi) ?? throw new Exception("Failed to start process");
```

Duplicati/Library/Utility/Power/MacOSPowerSupplyState.cs
```diff
@@ -89,7 +89,7 @@ private PowerSupply.Source GetSourceIoreg()
...
```
kenkendk, line 92: maybe add:
```
var pi = System.Diagnostics.Process.Start(psi) ?? throw new Exception("Failed to start process");
```

bigboybamo:
Hi @kenkendk I updated the PR again

Duplicati/Library/Utility/CallContextSettings.cs
```diff
@@ -277,11 +276,11 @@ public static T Settings
         /// </summary>
         /// <returns>The disposable handle for the context.</returns>
         /// <param name="initial">The initial value.</param>
-        public static IDisposable StartContext(string settingsKey, T initial = default(T))
+        public static IDisposable StartContext(string settingsKey, T? initial = default(T))
         {
             contextSettingsType = settingsKey;
             var res = new ContextGuard();
-            Settings = initial;
+            Settings = initial ?? throw new ArgumentNullException(nameof(initial), "The initial value cannot be null.");
```
Jojo-1000, line 283: This is a contradiction. `default(T)` is null for any reference type.

Duplicati/Library/Utility/CallContextSettings.cs
```diff
@@ -236,7 +236,7 @@ public static class CallContextSettings<T>
         /// The context setting types that are used and need their private setting key
         /// to prevent overwriting each others settings.
         /// </summary>
-        private static string contextSettingsType = null;
+        private static string? contextSettingsType = null;
```
Jojo-1000, line 239: This should not be nullable, because it can always be initialized in the constructor.

bigboybamo: Fixed this

Duplicati/Library/Utility/CallContextSettings.cs
bigboybamo: Made a fox for this as well

Duplicati/Library/Utility/FileBackedList.cs
```diff
@@ -36,13 +36,13 @@ public abstract class FileBackedList<T> : IEnumerable<T>, IDisposable
     {
         private class StreamEnumerator : IEnumerator<T>, System.Collections.IEnumerator
         {
-            private Stream m_stream;
+            private Stream? m_stream;
```
kenkendk, line 39: The stream can actually not be null and should be readonly.
The reason why it appears nullable is that it is being used as a `m_isDisposed` flag when null, which is bad practice.

Duplicati/Library/Utility/FileBackedList.cs
```diff
@@ -101,8 +107,8 @@ public void Reset()
             }
         }
 
-        private Library.Utility.TempFile m_file;
-        private Stream m_stream;
+        private Library.Utility.TempFile? m_file;
+        private Stream? m_stream;
```
kenkendk, line 111: Sam as above, it cannot be null and should be readonly.

Duplicati/Library/Utility/JoinedFilterExpression.cs
```diff
@@ -73,9 +75,9 @@ public static IFilter Join(IFilter first, IFilter second)
         public override string ToString()
         {
             if (this.First.Empty)
-                return this.Second.ToString();
+                return this.Second.ToString() ?? throw new InvalidOperationException("Second is null");
```
kenkendk, line 78: This looks off. Can `.ToString()` ever be null? The `First` and `Second` properties are never null.

Duplicati/Library/Utility/OverrideableStream.cs
```diff
@@ -33,7 +33,7 @@ namespace Duplicati.Library.Utility
     /// </summary>
     public class OverrideableStream : Stream
     {
-        protected System.IO.Stream m_basestream;
+        protected System.IO.Stream? m_basestream;
```
kenkendk, line 36: This one can never be null and should be `readonly`, it is being used as an `m_isDisposed` flag.
This removes all the other null checks.

Duplicati/Library/Utility/SslCertificateValidator.cs
```diff
@@ -31,10 +31,16 @@ public class SslCertificateValidator
...
-            private readonly string m_certificate = null;
+            private readonly string? m_certificate = null;
             private readonly SslPolicyErrors m_errors = SslPolicyErrors.None;
 
-            public string Certificate { get { return m_certificate; } }
+            public string Certificate 
+            {
+                get
+                {
+                    return m_certificate ?? throw new InvalidOperationException("Certificate is null");
```
kenkendk, line 41: There is a check for this property in `RemoteOperation.cs`, I think this property should allow null to be returned.

Duplicati/Library/Utility/TempFile.cs
```diff
@@ -35,7 +35,7 @@ public class TempFile : IDisposable
         /// </summary>
         public static string APPLICATION_PREFIX = Utility.getEntryAssembly().FullName.Substring(0, 3).ToLower(System.Globalization.CultureInfo.InvariantCulture) + "-";
 
-        private string m_path;
+        private string? m_path;
```
kenkendk, line 38: This is being set to null to avoid it being read after being disposed. It is better to make it `readonly` and add a `m_isDisposed` flag.

Duplicati/Library/Utility/TempFolder.cs
```diff
@@ -29,7 +29,7 @@ namespace Duplicati.Library.Utility
     /// </summary>
     public class TempFolder : IDisposable
     {
-        private string m_folder;
+        private string? m_folder;
```
kenkendk, line 32: This should be `readonly` and a `m_isDisposed` flag should be added.

Duplicati/Library/Utility/ThrottledStream.cs
```diff
@@ -135,6 +135,7 @@ public override int Read(byte[] buffer, int offset, int count)
 				var chunksize = (int)Math.Min(remaining, m_readspeed <= 0 ? remaining : m_readspeed * 2);
 				DelayIfRequired(ref m_readspeed, chunksize, ref m_last_read_sample, ref m_current_read_counter, ref m_current_read_speed);
 
+				if(m_basestream == null) throw new InvalidOperationException("m_basestream is null");
```
kenkendk, line 138: This will go away once the stream is not nullable in `OverrideableStream`

Duplicati/Library/Utility/ThrottledStream.cs
```diff
@@ -164,6 +165,7 @@ public override void Write(byte[] buffer, int offset, int count)
                 UpdateLimits();
 				var chunksize = (int)Math.Min(count, m_writespeed <= 0 ? count : m_writespeed * 2);
 				DelayIfRequired(ref m_writespeed, chunksize, ref m_last_write_sample, ref m_current_write_counter, ref m_current_write_speed);
+				if (m_basestream == null) throw new InvalidOperationException("m_basestream is null");
```
kenkendk, line 168: This will go away once the stream is not nullable in `OverrideableStream`

Duplicati/Library/Utility/WorkerThread.cs
```diff
@@ -254,7 +254,7 @@ public Tx CurrentTask
         {
             get
             {
-                return m_currentTask;
+                return m_currentTask ?? throw new InvalidOperationException("m_currentTask is null");
```
kenkendk, line 257: The return value can be null. As far as I can read the code, it starts out being null.

kenkendk (changes requested):
Sorry for the delay in reviewing, it is a much harder change in this project than what I anticipated.

I think some of the changes may need to be pushed to another PR if you feel they are too involved.

Duplicati/Library/Utility/FileBackedList.cs
bigboybamo: marking this stream as readonly means it cannot be reassigned in this Dispose methosd
image

Duplicati/Library/Utility/FileBackedList.cs
bigboybamo: Same thing here
image

So I'm not entirely sure how to proceed

Duplicati/Library/Utility/OverrideableStream.cs
bigboybamo: Changing that stream to readonly means the assignment here causes an error
image

Duplicati/Library/Utility/TempFolder.cs
bigboybamo: Not sure where to add the m_isDisposed flag?

Duplicati/Library/Utility/SslCertificateValidator.cs
```diff
@@ -31,10 +31,16 @@ public class SslCertificateValidator
         [Serializable]
         public class InvalidCertificateException : Exception
         {
-            private readonly string m_certificate = null;
+            private readonly string? m_certificate = null;
             private readonly SslPolicyErrors m_errors = SslPolicyErrors.None;
 
-            public string Certificate { get { return m_certificate; } }
+            public string Certificate
```
kenkendk, line 37: Should this not be `string?` as it returns `m_certificate` ?

Duplicati/Library/Utility/OverrideableStream.cs
kenkendk, line 1: Can we make `protected readonly Stream m_basestream;`  ?
This requires only that the dispose call does not set it to null, and removes all the extra code for handling nullable.

Duplicati/Library/Utility/FileBackedList.cs
kenkendk: I see, it is lazy initialized, so we cannot make it readonly :'(

kenkendk:
Sorry @bigboybamo, I know you put in a lot of effort, but I think we will have to handle nullability in a more bit-by-bit manner. The current PR is not suited for recovering.

PR #5208, opened 2024-05-29
Fix SQL string quotes

### LINKED ISSUE #5202: SQL statements using double quotes fail on FreeBSD

## Environment info

- **Duplicati version**: 2.0.8.1
- **Operating system**: FreeBSD
- **Backend**: -

## Description
Some SQL statements use double quotes for string literals, which is incorrect. It is supported by SQLite for legacy reasons, but it is encouraged to compile without this compatibility mode on by default. FreeBSD disables this in the new sqlite packages:
https://www.reddit.com/r/freebsd/comments/1chb82b/sqlite3_pkg_just_became_stricter_with_quoting/

For example, this uses the wrong quotes for `Temporary`:
https://github.com/duplicati/duplicati/blob/54b9063f1ba807c78c1f31f412d36c0a803aff87/Duplicati/Library/Main/Database/LocalRecreateDatabase.cs#L708

All statements should be looked at and changed to the correct quotes. Also, the legacy mode should be disabled on all platforms in the SQLite config to prevent hidden bugs in the future.

Jojo-1000:
Closes #5202 

Changed all occurrences of double-quoted strings to single quotes.

Also, the compatibility options to allow these strings are disabled in SQLite to make incorrect statements easier to detect in the future. If that is not desired, we could alternatively change the options to always enable the compatibility mode on all systems, even if disabled by default. I am not sure why the SQLiteLoader only accesses the connection with reflection (is that still relevant with the .NET build?), so please check whether my method introduces some undesired effects.

### DIFF
--- a/Duplicati/Library/Main/Database/LocalBugReportDatabase.cs
+++ b/Duplicati/Library/Main/Database/LocalBugReportDatabase.cs
@@ -66,10 +66,10 @@ namespace Duplicati.Library.Main.Database
                         */
                 }
 
-                cmd.ExecuteNonQuery(@"UPDATE ""LogData"" SET ""Message"" = ""ERASED!"" WHERE ""Message"" LIKE ""%/%"" OR ""Message"" LIKE ""%:\%"" ");                
-                cmd.ExecuteNonQuery(@"UPDATE ""LogData"" SET ""Exception"" = ""ERASED!"" WHERE ""Exception"" LIKE ""%/%"" OR ""Exception"" LIKE ""%:\%"" ");                
+                cmd.ExecuteNonQuery(@"UPDATE ""LogData"" SET ""Message"" = 'ERASED!' WHERE ""Message"" LIKE '%/%' OR ""Message"" LIKE '%:\%' ");                
+                cmd.ExecuteNonQuery(@"UPDATE ""LogData"" SET ""Exception"" = 'ERASED!' WHERE ""Exception"" LIKE '%/%' OR ""Exception"" LIKE '%:\%' ");                
 
-                cmd.ExecuteNonQuery(@"UPDATE ""Configuration"" SET ""Value"" = ""ERASED!"" WHERE ""Key"" = ""passphrase"" ");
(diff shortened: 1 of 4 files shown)


kenkendk:
> I am not sure why the SQLiteLoader only accesses the connection with reflection (is that still relevant with the .NET build?)

No, I don't think it is required anymore. The reason was to support both the Mono and the .NET types and map them to `System.Data.IDbConnection` from one place. With the single shared type, there is no need anymore.

kenkendk:
This one is suspiciously failing on MacOS only when running the tests.
I have seen another place that MacOS would just exit the process on SQL errors/warnings (something with logging not working).

kenkendk:
Arh, it is the `SetConfigurationOption` that crashes it.
You can set it like this:
```
            if (con is System.Data.SQLite.SQLiteConnection sqlitecon && !Platform.IsClientWindows)
            {
                sqlitecon.SetConfigurationOption(System.Data.SQLite.SQLiteConfigDbOpsEnum.SQLITE_DBCONFIG_DQS_DDL, false);
                sqlitecon.SetConfigurationOption(System.Data.SQLite.SQLiteConfigDbOpsEnum.SQLITE_DBCONFIG_DQS_DML, false);
            }

```

Jojo-1000:
The question is whether that is caused because the configuration is not supported, or is there a mac-only statement that wants to print an error message?

kenkendk:
The test setup I am using on MacOS is the M3 (Arm64) version.

The build for SQLite on Arm64 has a bug that any logged message causes it to terminate the process.
I spent some time trying to figure out why, but gave up and just disabled the logging feature for the SQLite binary on MacOS Arm64.

For that reason I would not expect to get any error messages.
What I *can* see is that calling `sqlitecon.SetConfigurationOption` causes the process to exit immediately.

From my perspective, we have a relatively low user-base for MacOS, and setting the options mostly increases the security of statements, so they will fail if we have embedded double quotes around values. If the code is correct, this should never happen and we should catch it on all other platforms. With that I think it is reasonable to just not activate the check here.

PR #5209, opened 2024-05-30
Fix supported platform warnings (except Captcha)

### LINKED ISSUE #5167: Fix build warnings for platform specific items

Jojo-1000:
Partially completes #5167 

This fixes most of the platform warnings and adds attributes to more platform-specific classes. There are two things still remaining:
- Refactor the `SystemIO` concept, because it currently needs platform-specific classes on all platforms
- Replace the Captcha (System.Drawing library or completely)

The windows server projects were changed to only target windows plaforms.

### DIFF
--- a/Duplicati.Library.RestAPI/Database/ServerSettings.cs
+++ b/Duplicati.Library.RestAPI/Database/ServerSettings.cs
@@ -493,7 +493,7 @@ namespace Duplicati.Server.Database
                 if (String.IsNullOrEmpty(settings[CONST.SERVER_SSL_CERTIFICATE]))
                     return null;
 
-                if (Platform.IsClientWindows)
+                if (OperatingSystem.IsWindows())
                     return new X509Certificate2(Convert.FromBase64String(settings[CONST.SERVER_SSL_CERTIFICATE]));
                 else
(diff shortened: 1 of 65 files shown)


kenkendk:
I am good with this change, but it breaks the CI builds because it attempts to build the WindowsService on non-Windows.
I have not investigated how to fix that part.

We can do `System.Drawning` in another PR.
I would recommend using something like https://github.com/SixLabors/ImageSharp

Jojo-1000:
> I am good with this change, but it breaks the CI builds because it attempts to build the WindowsService on non-Windows.
> I have not investigated how to fix that part.

I tried to look into this, but there doesn't seem to be an easy solution to exclude the project depending on the OS. It is probably easiest to allow any target and just check when running it.

Issue #5210, opened 2024-05-30
Incompatible NuGet packages for .NET 8


Jojo-1000:
- [x] I have searched open and closed issues for duplicates.
- [x] I have searched the [forum](https://forum.duplicati.com) for related topics.

----------------------------------------

## Description
These packages only support .NET Framework and need to be updated or replaced:
- AlphaFS: Only used for symlinks on windows, which seems to be broken and may be supported by the standard file API, no longer maintained
- Sharp.Xmpp: Used for SendJabberMessage, looking for maintainer
- ARSoft.Tools: Prerequisite for Sharp.Xmpp

It also seems that symlink support for windows is broken, so the AlphaFS methods are never called.

kenkendk:
AlphaFS was also originally used to support long paths, but this is also added to .NET now.
It was also a dependency 

I have made PR #5211 which removes it.

For Sharp.Xmpp, it also drags in BouncyCastle, so it would be nice to get rid of it.
I think it was included as it was the only non-GPL open-source library we could find.
Maybe it is better to just remove it and no longer support Xmpp?

Jojo-1000:
There is this one with MIT license:
https://www.nuget.org/packages/Artalk.Xmpp

The last update was 2 years ago, but it would probably work for our purposes.

PR #5211, opened 2024-05-30
Removing unsupported AlphaFS.

kenkendk:
This updates #5210

This update removes the unsupported AlphaFS package, in favor of some built-in .NET features.
The update needs testing, and impacts only Windows.
Sadly, there has been many quirks related to Windows support for file-links, so this update has the potential to break in some specific situation that is not easily detected in testing.


Jojo-1000:
I tried to backup a symlink on Windows (created with `mklink`) to test whether the library is necessary. A directory symlink is recognized as a normal (empty) directory by the `IsSymlink()` method. This was already with the last released version, because the file attributes doesn't have ReparsePoint.

kenkendk:
> Edit: Backup and restore of file symlinks seems to work with this update, but there is strange behavior with the reporting: It only works if it is not the only selected file, and it does not show up in the log statistics. I don't think that is a priority now.

Excellent!

But, does this mean that this update does not work with directories either? Or did I misunderstand?


Re. the "only item" bug:
There is a specific rule for having the entry-items not evaluated as symlinks. It was originally added to allow choosing a symlinked folder as a source, and not ending up with getting just the symlink (and not the contents) stored.
in other words: it works as intended, but maybe it is not logical.

Jojo-1000:
> But, does this mean that this update does not work with directories either? Or did I misunderstand?

The behavior is the same with these changes and in the last beta release. The symlink detection fails for some reason:
- create a symlink directory with `mklink /d`
- select parent folder as backup source
- backup with symlink policy store
- the restore view has an empty directory, when restored it is also empty and not a symlink

In the debugger, I can see that  the directory is not detected as a symlink, but as a normal directory in SnapshotUtility.IsSymlink.

https://github.com/duplicati/duplicati/blob/531249910794b70c02d0f85103987dd34ca58ce0/Duplicati/Library/Snapshots/SnapshotUtility.cs#L95-L102

The file attributes only have directory and not ReparsePoint.

kenkendk:
There is a [SO thread](https://stackoverflow.com/questions/1485155/check-if-a-file-is-real-or-a-symbolic-link) discussing it here.

According to that it says ReparsePoint should be on both directory and file :/
But it also mentions `.LinkTarget != null` on `FileInfo`/`DirectoryInfo`.

kenkendk:
I cannot reproduce the issue, I made a folder with symlinks inside and then restored:
Screenshot 2024-06-04 at 11 51 58

I am testing on Windows 11 22H2, @Jojo-1000 is this comparable to your setup?

Jojo-1000:
@kenkendk I tried to reproduce this again, but now it is working correctly the same as for you (on Win 10). Maybe I was looking at the attributes of the wrong file before.

PR #5213, opened 2024-06-01
Change XMPP library to Artalk.Xmpp

Jojo-1000:
The old library was not compatible with .NET 8. It was also required to update System.Drawing.Common and Microsoft.Win32.SystemEvents to 7.0.

I tested that sending XMPP messages works on both Windows and Linux.


kenkendk:
Awesome find!

PR #5231, opened 2024-06-10
Increase the default test coverage to be 0.1%

kenkendk:


Duplicati/Library/Main/Options.cs
```diff
@@ -364,7 +364,7 @@ public IList<ICommandLineArgument> SupportedCommands
                     new CommandLineArgument("index-file-policy", CommandLineArgument.ArgumentType.Enumeration, Strings.Options.IndexfilepolicyShort, Strings.Options.IndexfilepolicyLong, IndexFileStrategy.Full.ToString(), null, Enum.GetNames(typeof(IndexFileStrategy))),
                     new CommandLineArgument("no-backend-verification", CommandLineArgument.ArgumentType.Boolean, Strings.Options.NobackendverificationShort, Strings.Options.NobackendverificationLong, "false"),
                     new CommandLineArgument("backup-test-samples", CommandLineArgument.ArgumentType.Integer, Strings.Options.BackendtestsamplesShort, Strings.Options.BackendtestsamplesLong("no-backend-verification"), "1"),
-                    new CommandLineArgument("backup-test-percentage", CommandLineArgument.ArgumentType.Integer, Strings.Options.BackendtestpercentageShort, Strings.Options.BackendtestpercentageLong, "0"),
+                    new CommandLineArgument("backup-test-percentage", CommandLineArgument.ArgumentType.Integer, Strings.Options.BackendtestpercentageShort, Strings.Options.BackendtestpercentageLong, "0.1"),
```
Jojo-1000, line 367: Could this be a problem, since the argument type implies integer, but the value is not? I assume at least the UI will struggle with this default value, because it uses `parseInt` for this type.

kenkendk: Good point. 

I have also made the parsing in the code use invariant culture (`en-US` similar), to avoid parsing issues with `,` as I think the CLI is technical enough to warrant this.

I will add a note about the parsing of commas in the help file and introduce a new data type for this.

The UI on the other hand should use the browser locale for display/edit but send invariant numbers to the backend.

kenkendk: For now, I have added `decimal` as `text` in the advanced input.
If decimal is used more, we can revisit adding a proper parser with regional support.

PR #5232, opened 2024-06-10
Fix compare on linux showing 0 bytes when exclude filters are used

### LINKED ISSUE #5229: Compare backup returns zero when a folder is excluded
Details in the forum post: https://forum.duplicati.com/t/compare-feature-not-working/14792

Jojo-1000:
Closes #5229

The SQL statement that is used when the filters are simple on a case sensitive filesystem assumed that the filter is an include filter. I added a proper distinction between include and exclude filters.

### DIFF
--- a/Duplicati/Library/Main/Database/LocalListChangesDatabase.cs
+++ b/Duplicati/Library/Main/Database/LocalListChangesDatabase.cs
@@ -164,8 +164,19 @@ namespace Duplicati.Library.Main.Database
                             cmd.SetParameterValue(0, s);
                             cmd.ExecuteNonQuery();
                         }
-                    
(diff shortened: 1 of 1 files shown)


ts678:
> assumed that the filter is an include filter

I had previously found Linux `--include=*` fine. I now try `--exclude=*`, it fails. Isn't this contrary to above?
Regardless, nice find on noticing the difference in the way the compare is done between Windows and Linux.

Jojo-1000:
With exclude=* the expected behavior is to exclude all the files from the comparison, so nothing is left. It is still annoying that the excludes from the backup job are active by default, but that is just a shortcoming of the web interface.

Also I believe that any glob or regex filter disable this filtering method and fall back to the C# implementation, so you should see no difference with more complex filters.

ts678:
Second part might be the key. If I `--exclude=/tmp/short.txt` maybe that got treated like `--include`.
I just started with path from the `Commandline arguments` box, and prepended it into the option to test.
Both cases did the compare.

Issue #4290, opened 2020-08-21
Support Powershell Pre-Script


jwoodard80:
## Environment info

- **Duplicati version**: 0.0.0.0
- **Operating system**: Windows 10
- **Backend**: SFTP

## Description

Cannot run ps1 (Powershell) files as a pre/post script. 

I am ultimately trying to make flat files of my SQL DB's. Since I cannot do that VIA native functions I have to use a small PS script to generate them and then backup.

## Steps to reproduce

1. Select pre/post script and use a .ps1 file 
2. Run Task

- **Actual result**:
The specified executable is not a valid application for this OS platform.

- **Expected result**:
Run the PS script

Shlamy:
A powershell script generally needs to be run by `PowerShell.exe` with a `-File` argument that points to the location of the `scriptFile.ps1` file itself. 

Let's say I have a powershell script on my Desktop called `TestThings.ps1` which should do some important thing prior to the backup running.

If `run-script-before-required` is set to `powershell.exe -File "C:\Users\Shlamy\Desktop\TestThings.ps1"`
only `powershell.exe` is handed off to Duplicati, none of the arguments are included. Tested with various single/double quote variations.

Looks like ts678 was probably on the right track in that it looks like changing ProcessStartInfo(String) to ProcessStartInfo(String, String) would be required to include the arguments natively.

 **Workaround:** You can specify a batch file in `run-script-before-required`. For example `C:\Users\Shlamy\Desktop\RunPS1.bat` which contains the line `powershell.exe -File "C:\Users\Shlamy\Desktop\TestThings.ps1"`, Duplicati will now run the `TestThings.ps1` script as expected prior to the backup job running.

If the batch file doesn't seem to run or complains about a `non-zero exit code`, add another line to the batch file `exit /b 0` this will force a 0 exit code on the batch which will allow Duplicati to run anyways. Upon completion Duplicati should issue a nice warning at the end about what's wrong, in my case I needed to enable set-execution for the local machine, even gave me a link to MS's site on the subject. Without the `exit /b 0` the batch file only gave an error about the non-zero batch error and not the warning from the powershell script. Once you've discovered what warning or error is being sent, you can remove the `exit /b 0` line to regain proper error level checking.

duplicatibot:
This issue has been mentioned on **Duplicati**. There might be relevant details there:

https://forum.duplicati.com/t/how-to-use-scripts-on-duplicati/16094/3

PR #5234, opened 2024-06-12
Added support for passing arguments to scripts

### LINKED ISSUE #2926: Arguments for scripts before/after options

## Environment info

- **Duplicati version**: 2.0.2.1-1
- **Operating system**: Ubuntu 16.04.3 LTS
- **Backend**: local filesystem

## Description

I need to call a script with arguments but when I call the task I get:

`ApplicationName='/home/foo/bar.sh jhon doe', CommandLine='', CurrentDirectory='', Native error= Cannot find the specified file`

## Steps to reproduce

1. Install duplicati
2. Configure a backup with the option `run-script-before-required`
3. Give to your scripts arguments

- **Actual result**:
The backup fails with this log `ApplicationName='/home/foo/bar.sh jhon doe', CommandLine='', CurrentDirectory='', Native error= Cannot find the specified file`
- **Expected result**:
The backup calls the script `'/home/foo/bar.sh` sending `jhon` and `doe' as arguments.

Thanks

kenkendk:
This closes #2926

### DIFF
--- a/Duplicati/Library/Modules/Builtin/RunScript.cs
+++ b/Duplicati/Library/Modules/Builtin/RunScript.cs
@@ -55,6 +55,7 @@ namespace Duplicati.Library.Modules.Builtin
         private const string FINISH_OPTION = "run-script-after";
         private const string REQUIRED_OPTION = "run-script-before-required";
         private const string TIMEOUT_OPTION = "run-script-timeout";
+        private const string ENABLE_ARGUMENTS_OPTION = "run-script-with-arguments";
         /// <summary>
         /// Option used to set the log level for mail reports
         /// </summary>
@@ -72,6 +73,7 @@ namespace Duplicati.Library.Modules.Builtin
         private string m_startScript = null;
         private string m_finishScript = null;
         private int m_timeout = 0;
+        private bool m_enableArguments = false;
 
         private string m_operationName;
         private string m_remoteurl;
@@ -86,7 +88,7 @@ namespace Duplicati.Library.Modules.Builtin
         /// <summary>
         /// The log storage
         /// </summary>
-        private Utility.FileBackedStringList m_logstorage;
+        private FileBackedStringList m_logstorage;
 
 
         #region IGenericModule implementation
@@ -95,10 +97,10 @@ namespace Duplicati.Library.Modules.Builtin
             commandlineOptions.TryGetValue(STARTUP_OPTION, out m_startScript);
             commandlineOptions.TryGetValue(REQUIRED_OPTION, out m_requiredScript);
             commandlineOptions.TryGetValue(FINISH_OPTION, out m_finishScript);
+            m_enableArguments = Utility.Utility.ParseBoolOption(commandlineOptions, ENABLE_ARGUMENTS_OPTION);
 
-            string tmpResultFormat;
             ResultExportFormat resultFormat;
-            if (!commandlineOptions.TryGetValue(RESULT_FORMAT_OPTION, out tmpResultFormat))
+            if (!commandlineOptions.TryGetValue(RESULT_FORMAT_OPTION, out var tmpResultFormat))
             {
                 resultFormat = ResultExportFormat.Duplicati;
             }
@@ -109,8 +111,7 @@ namespace Duplicati.Library.Modules.Builtin
 
             resultFormatSerializer = ResultFormatSerializerProvider.GetSerializer(resultFormat);
 
-            string t;
-            if (!commandlineOptions.TryGetValue(TIMEOUT_OPTION, out t))
+            if (!commandlineOptions.TryGetValue(TIMEOUT_OPTION, out var t))
                 t = DEFAULT_TIMEOUT;
 
             m_timeout = (int)Utility.Timeparser.ParseTimeSpan(t).TotalMilliseconds;
@@ -118,7 +119,7 @@ namespace Duplicati.Library.Modules.Builtin
             m_options = commandlineOptions;
 
             m_options.TryGetValue(OPTION_LOG_FILTER, out var logfilterstring);
-            var filter = Utility.FilterExpression.ParseLogFilter(logfilterstring);
+            var filter = FilterExpression.ParseLogFilter(logfilterstring);
             var logLevel = Utility.Utility.ParseEnumOption(m_options, OPTION_LOG_LEVEL, DEFAULT_LOG_LEVEL);
 
             m_logstorage = new FileBackedStringList();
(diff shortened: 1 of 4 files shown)


energywave:
Wonderful that you took time to do that, thank you so much just for that!
To discsuss the implementation (constructive criticism):
1.  did you test those even in Windows?
2. Would not have been easier to use the standard syntax like: `"c:\path with spaces\mybatch.cmd" arg1 "arg 2 with spaces"` ?
3. Or maybe, even better, just pass them to the shell to execute so that it's up to the user to write a correct syntax for the OS in use? I don't get why to separate arguments from the commandline.
4. I'm sceptical about the json syntax. I appreciate the possibility and welcome it but thinking about Duplicati that's used by all kind of people and must keep it simple, that will be not so understandable by many persons.

However thank you for the implementation!

kenkendk:
Thanks for the feedback @energywave !

1. No, I do not see anything that is Windows specific here. 
Do you see anything that would break on Windows?

2. My primary concern is backwards compatibility. 
If users have set it up to run a script with spaces in the path, it will break if we start to parse the path.
But I like the syntax! I am used to escaping spaces, so I had not thought of that.

3. I have not seen the option to pass arguments directly to the shell from C#:
https://learn.microsoft.com/en-us/dotnet/api/system.diagnostics.process.start?view=net-8.0
It all requires at least parsing the executable name.
I will experiment with passing the rest of the string without modification.

4. It is overkill for most users, but when dealing with tricky escape rules, I find it easier to just use a standard JSON encoding.
I do not think the separation of arguments and commandline is useful either, it was just an easy way to preserve compatibility with the previous non-parsing of the argument.

kenkendk:
I have made a new approach here, based on your suggestion.

The logic is now that nothing extra is needed.
You just add the arguments to the script as suggested, optionally using single or double quotes to group arguments.

There is a small possibility that this will break existing setups because the script path was previously treated as a single path, but is now treated as a commandline. I think I will address this in the release notes.

Jojo-1000:
Have you checked how this affects command line users? I would imagine they need to double escape the strings, which is annoying but there isn't really a way around it.

For space splitting, does the Linux way of escaping spaces with a backslash before them also work?

I have to say I liked the backwards compatible version. I quite often have spaces in my paths, but on the other hand the syntax is unusual and more complexity is not really what we need.

Jojo-1000:
> 3\. Or maybe, even better, just pass them to the shell to execute so that it's up to the user to write a correct syntax for the OS in use? I don't get why to separate arguments from the commandline.

This could also fix the existing issue that Powershell scripts are not executed automatically, if such a way exists.

energywave:
> Thanks for the feedback @energywave !

Thanks to you that implemented it!

> 1. No, I do not see anything that is Windows specific here.
>    Do you see anything that would break on Windows?

No, I don't, I took a look at the code, seems ok to me (but real life tests are a good option :) )

> 3. I have not seen the option to pass arguments directly to the shell from C#:

Yes sorry, I didn't remember that, you're right. We need to detect where executable path ends and arguments starts. I've checked in my softwares where I do such things and I implemented a function to split entire commandline by iterating characters and keeping tracks of "in apix" or "outside apix" state while progressing, until I find a space outside apixes to split the commandline in executable path and arguments. Does not need to split arguments, however, just passing then "as is" to the `Process.Start` function

> I have made a new approach here, based on your suggestion.

Hey, I'm honored you liked my suggestion and implemented it, thank you!

> There is a small possibility that this will break existing setups because the script path was previously treated as a single path, but is now treated as a commandline. I think I will address this in the release notes.

To avoid that (just an idea) you could add a new parameter, something like `--run-script-syntax` or maybe `--run-script-with-arguments` to switch the new behavior on or off with default off, so that existing users will have no change and will not have any malfunctions.
I'm with you that keeping backward compatible changes is a good practice as a strict minority of users really reads changelogs... :_(

> For space splitting, does the Linux way of escaping spaces with a backslash before them also work?

It could be theorically possible to detect both ways but the author implemented this using regex and I don't know if the regex can be modified to handle that (I personally hate regexes... :D )
With the function I've made (talked about that above) it could be easy to add also that.
But let's hear what @kenkendk thinks about it.

> I have to say I liked the backwards compatible version. I quite often have spaces in my paths, but on the other hand the syntax is unusual and more complexity is not really what we need.

Could adding a switch option with the old default, like proposed above, be a good choice for you?

ts678:
[System.Diagnostics.ProcessStartInfo.UseShellExecute property](https://learn.microsoft.com/en-us/dotnet/fundamentals/runtime-libraries/system-diagnostics-processstartinfo-useshellexecute) says it's not a command shell (e.g. specifically disclaiming bash).

energywave:
> [System.Diagnostics.ProcessStartInfo.UseShellExecute property](https://learn.microsoft.com/en-us/dotnet/fundamentals/runtime-libraries/system-diagnostics-processstartinfo-useshellexecute) says it's not a command shell (e.g. specifically disclaiming bash).

From what I know UseShellExecute is used in Windows to "execute" a document, like an Excel spreadsheet file, to let Windows open it with the corresponding executable assigned in Windows Settings.
The article claims that "Shell" is not to be intended as "Linux shell". I'll cite:

`The word "shell" in this context (UseShellExecute) refers to a graphical shell (similar to the Windows shell) rather than command shells (for example, bash or sh) and lets users launch graphical applications or open documents.`

If set to False you can just execute "real" executables.
So I don't think that should be relevant in this discussion. Hoping to not miss something, in that case I'm sorry.

kenkendk:
> Have you checked how this affects command line users? I would imagine they need to double escape the strings, which is annoying but there isn't really a way around it.
> 
> I have to say I liked the backwards compatible version. I quite often have spaces in my paths, but on the other hand the syntax is unusual and more complexity is not really what we need.

Yes, that was my motivation for the original approach. I was thinking of how to support spaces in paths, and I wanted to avoid inventing something new, so I went for JSON, which handles all sorts of edge cases and is well known.

I was considering adding a compatibility switch like `--run-script-no-argument-parsing` (or similar), but I think it adds complexity with very little gain.

> For space splitting, does the Linux way of escaping spaces with a backslash before them also work?

Not right now, but it can be added. Not sure how to correctly support that on Windows which has slightly complex rules because backslash is both a valid path character and an escape character.

> Does not need to split arguments, however, just passing then "as is" to the `Process.Start` function

That is an option, but I think that is slightly inconsistent, because you then have different ways to escape the arguments and the executable.

> This could also fix the existing issue that Powershell scripts are not executed automatically, if such a way exists.

@Jojo-1000 Do you know why this happens? Is it because `.ps` files need to be invoked with `UseShellExecute=true` ?

If this is the case, we could either autodetect it or add `--run-script-use-shell-execute`.

> maybe `--run-script-with-arguments`

Yes, that would maintain backwards compatibility. @Jojo-1000 would it work for your use as well?

Jojo-1000:
> Do you know why this happens? Is it because `.ps` files need to be invoked with `UseShellExecute=true` ?

The reason why powershell scripts don't work by default is that, unlike .bat, the .ps1 extension is not registered with the executable in Windows (for double click). So you either need to register it, which has some security concern, or invoke it with `powershell.exe -file script.ps1`. That would be possible now with this support for arguments.
I don't know if UseShellExecute makes any difference.
https://github.com/duplicati/duplicati/issues/4290

energywave:
> I would assume this would actually enable the `.ps1` extension to execute, but unfortunately we cannot use it because launching with `UseShellExecute = true` is not possible when also redirecting stdout or stderr.

But I think it's not needed, as stated by @Jojo-1000 : by using arguments you can now specify `powershell.exe -file script.ps1` that will handle powershell scripts by using the powershell version between the installed ones of your choice.

About the baskslash escaping you're right, it would be needed to enable it only on Linux, not on Windows as there is the path separator.

PR #5243, opened 2024-06-21
Use sqlite for recovery lookup

kenkendk:
This PR is a follow up to #5241 
It changes the low-memory implementation to use a single SQLite database to keep track of the block/file mapping, instead of relying on reading the plain-text index file and seeking in that.

Testing with a 2GiB backup shows that:
- Plain-text-based restore takes ~10s
- SQLite-based restore takes ~6s
- In-memory-based restore takes ~4s (from #5241)

(Tested on M3 with an SSD drive)


kenkendk:
Closing this as there are no immediate plans to merge it

PR #5245, opened 2024-06-24
Changed logic so all modules are loaded before any are started

kenkendk:


Jojo-1000:
I think it would be helpful to document this module lifecycle in the interface:
- Configure is called when modules are loaded
- OnStart is called after all modules are configured. If there is an exception, any later modules don't execute OnStart
- OnFinish is executed after an operation, even if OnStart was not called for this module

This change might also screw up the script module, because the options are changed in OnStart. Other modules will still have the old options for configure, so I think this will be a problem.

kenkendk:
> This change might also screw up the script module, because the options are changed in OnStart. Other modules will still have the old options for configure, so I think this will be a problem.

Good catch!

If we want modules to be configured with options from scripts, but scripts can abort as well, we need to define the load order. Right now the load order is dictated by the GenericLoader which does not enforce a particular order.

kenkendk:
Closing for now, need to reconsider the right order to load things in

PR #5260, opened 2024-07-08
Added tool to generate test data

kenkendk:


ts678:
It's nice to see something like this. Will there be executable versions laying around somewhere?

I guess that could be more complicated than before, due to platform-specific builds these days.

Originally I wondered how well it could update existing data, but I think I see ways to persuade.

FWIW my most-used similar tool has been a Python data initializer and changer. Old copy [here](https://forum.duplicati.com/t/duplicati-stuck-on-compact-when-getting-putting-from-onedrive-and-results-in-missing-files/16641/16?u=ts678).

These both have subtleties to potential application which are not obvious from reading the help.

kenkendk:
I was not planning on providing binaries for this, because it is not intended to have any value for end users.
Developers can (should?) be able to compile it, and would perhaps like to make changes to the code, beyond what the options allow.

In case someone wants to try it, you need the .NET SDK installed and then you can run:
```
> cd duplicati/Tools/TestDataGenerator
> dotnet run help
> dotnet run create ./data
```

My goal with this tool was to create a 1 TiB dataset that I can use for performance testing.
I have created the 1 TiB random set and am almost done creating a backup of the data, so I can look into the restore/recreate speed issues.

As with your script, it is not a perfect replication of actual backup data, but I think either tool can make a decent replica for at least performance testing.

Issue #5226, opened 2024-06-05
Duplicati crashes when backup starts, with no logs


DeflateAwning:
- [x] I have searched open and closed issues for duplicates.
- [x] I have searched the [forum](https://forum.duplicati.com) for related topics.

----------------------------------------

## Environment info

- **Duplicati version**: v2.0.9.100_canary_2024-05-30
- **Operating system**: Linux Mint 21.3
- **Backend**: Storj

## Description
As of the latest update, when a backup starts after unpausing the paused-on-start state, the tray icon disappears. It also appears that they backups do not actually start/work, also.

## Steps to reproduce
1. Access Web UI.
2. Start backup.
3. It says "Starting backup" for a minute or so
4. Cannot connect via web interface anymore (and icon has disappeared, per #5224)

kenkendk:
@DeflateAwning Is this contained to Storj based backups, or does it happen with others as well?

luixxiul:
I experience the issue with Storj as well, on Ubuntu LTS. The issue is intermittently experienced.

TopperDEL:
Just found this issue as I got tagged. Which architecture are you on @luixxiul ? I would assume that there is no binary present for that type and therefore it fails (my assumption).

luixxiul:
> Which architecture are you on @luixxiul ? I would assume that there is no binary present for that type and therefore it fails

I'm on x64 and have installed `duplicati-2.0.9.100_canary_2024-05-30-linux-x64-gui.deb` on Ubuntu 22.04.4 LTS before reverting to `Duplicati - 2.0.7.103_canary_2024-04-19`.

TopperDEL:
Ok, so as soon as Duplicati uses the latest uplink.NET (Version 2.13.*), this should work then. The 2.13-release includes binaries for Linux x64, Linux ARM and Linux ARM64.
@kenkendk Could you update Duplicati to the latest uplink.NET? Or shall I create a PR for that?

luixxiul:
Unfortunately the issue does not seem to have been fixed on [v2.0.9.102_canary_2024-08-02](https://github.com/duplicati/duplicati/releases/tag/v2.0.9.102_canary_2024-08-02). Duplicati still crashs on conducting a backup task to Storj.

[...]

PR #5262, opened 2024-07-09
Prepared for new Uplink packages that correctly package the Storj binaries

kenkendk:
For now there is still an issue with the linux versions: https://github.com/TopperDEL/uplink.net/issues/39#issuecomment-2217243464

### DIFF
--- a/Duplicati/Library/Backend/Storj/Duplicati.Library.Backend.Storj.csproj
+++ b/Duplicati/Library/Backend/Storj/Duplicati.Library.Backend.Storj.csproj
@@ -9,15 +9,10 @@
 
 	<ItemGroup>
 		<PackageReference Include="Newtonsoft.Json" Version="13.0.3" />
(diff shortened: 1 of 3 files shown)


luixxiul:
Will this close https://github.com/duplicati/duplicati/issues/5226 maybe?

kenkendk:
@luixxiul I assume it will. Merging this PR should make Storj work on all platforms except Linux non-x64. I am waiting a bit to see if @TopperDEL can make an updated Linux package, so we can close it all in one go.

TopperDEL:
Please use version 2.13.3484 for uplink.NET. It now includes the correct binaries.
I missed the fact that during build the target-architecture was not respected and therefore was always the one from the host (x64).

And yes @luixxiul: this should close #5226 - at least from the uplink.NET-perspective.

kenkendk:
@TopperDEL Great! I can confirm it now has the correct library for all builds.

PR #5278, opened 2024-07-19
Make a div on throttle.html translatable

luixxiul:
This PR intends to make a div on `throttle.html` translatable.

throttle

Since I am not quite sure if adding `translate` to the element would be sufficient, please let me know what I should do additionally if any.


kenkendk:
I can see this is done in other places as well

Issue #5303, opened 2024-07-26
Translation of description on warning dialog on a backup task with Microsoft Office 365 Group is only partially applied


luixxiul:
- [x] I have searched open and closed issues for duplicates.
- [ ] I have searched the [forum](https://forum.duplicati.com) for related topics.

----------------------------------------

## Environment info

- **Duplicati version**: Debug
- **Operating system**: irrelevant
- **Backend**: Microsoft Office 365 Group

## Description

As title says, translation of description on warning dialog on a backup task with Microsoft Office 365 Group is only partially applied.

## Steps to reproduce

1. Set up a backup task with Microsoft Office 365 Group
2. Change the UI language into Dutch
3. Select the task
4. Click "Bewerken..."
5. Select "Doel"
6. Select "Microsoft Office 365 Groep"
7. Input something to the AuthID field
8. Click "Test verbindung"
9. Confirm the warning message is displayed

- **Actual result**: the warning message is partially displayed in Dutch.

- **Expected result**: it should be completely displayed in Dutch, as translated here: https://github.com/duplicati/duplicati/blob/2569f94de8b3d1427554c8aeb381579560945603/Localizations/webroot/localization_webroot-nl_NL.po#L3260

## Screenshots

![1](https://github.com/user-attachments/assets/8a13e4fd-3df8-4aca-b2b6-29f72b9de0b3)

PR #5304, opened 2024-07-26
Fix the issue that strings on the warning dialog of a backup task with Microsoft Office 365 Group are partially displayed in English

luixxiul:
Fixes https://github.com/duplicati/duplicati/issues/5303

This PR intends to fix the issue that strings on the warning dialog of a backup task with Microsoft Office 365 Group are partially displayed in English.

The issue apparently is caused by Transifex's auto removal of a whitespace character added as a first character of a string, which therefore removes the whitespace character from PO files (see [here](https://github.com/duplicati/duplicati/blob/2569f94de8b3d1427554c8aeb381579560945603/Localizations/webroot/localization_webroot-nl_NL.po#L3301) for an example), causing the strings are not properly recognized. This PR should fix the issue by moving the whitespace character inside the string on EditUriBackendConfig.js. This also makes sense in that it lets translators decide how to handle the whitespace.

This PR also removes the duplicate block from EditUriBuiltins.js.

|Before|After|
|---------|------|
|||


kenkendk:
Nice that you removed the duplicated code and fixed the unused return value

PR #5307, opened 2024-07-26
Remove periods from *Short strings

luixxiul:
This PR intends to remove period from `*Short` strings.


kenkendk:
Great work! Unifying that has been on much wish list for a long time :)

Issue #5293, opened 2024-07-25
Sort strings with suffix `...Long` before ones with `...Short` for consistency


luixxiul:
- [x] I have searched open and closed issues for duplicates.
- [ ] I have searched the [forum](https://forum.duplicati.com) for related topics.

----------------------------------------

## Environment info

- **Duplicati version**: Debug
- **Operating system**: Debian
- **Backend**: irrelevant

## Description

There are some cases where strings with suffixes `...Long` and `...Short` are inverted, causing a bug which is hard to detect. Here is an example:

https://github.com/duplicati/duplicati/blob/2569f94de8b3d1427554c8aeb381579560945603/Duplicati/Library/Main/Strings.cs#L179-L180

It seems the strings which end with `...Long` have been traditionally placed *before* ones end with `...Short` alphabetically, and all of the strings should be ordered so.

Of course, the former could be placed *after* the latter, but the whole point is that we should have a consistent pattern for ordering lines to avoid confusion.

## Steps to reproduce

1. Look for lines like the one above

- **Actual result**: `UseblockcacheShort` returns the **long** description and `UseblockcacheLong` returns the **short** description.

- **Expected result**: They should be vice-versa ie. `UseblockcacheShort` should return the short description.

## Screenshots

![1](https://github.com/user-attachments/assets/90699e86-82c7-4e24-b18b-edf05ce4b13e)
![2](https://github.com/user-attachments/assets/876db880-b5fa-4223-b7c2-0bd1a3c77e5f)

luixxiul:
For reviewer's sanity the issue should be fixed with multiple PRs to make the review experience as smooth as possible.

kenkendk:
@luixxiul I think you have the overview on when this issue can be closed?

luixxiul:
I expect it will be able to be closed in a week as I've got the most work done in my own local repository :smile:

luixxiul:
Hopefully #5331 will be the last PR which should close this issue...

PR #5331, opened 2024-07-27
Sort strings on Main/Strings.cs

luixxiul:
Closes https://github.com/duplicati/duplicati/issues/5293

This PR intends to update Main/Strings.cs by sorting strings. It consists of multiple commits for easier review by trying to creating diffs so that they should explicitly show that the changes are only related to sorting the lines, not changing the contents of them.

This PR will be blocked until other PRs to be reviewed and hopefully merged, in order to take care of the unfixed strings altogether with this one.


luixxiul:
I checked manually that all strings with `...Long` on all Strings.cs files have been sorted before ones with `...Short`.

PR #5338, opened 2024-07-28
Fix "select" and "input" elements style rules on the step two of adding/editing a backup task

luixxiul:
This PR intends to fix `select` and `input` elements style rules on the step two of adding/editing a backup task. 

By using `grid` value for `display` property, it is possible to align elements without tweaking the layout manually with declarations such as `position: relative` and `left: xxx px`.

Since the grid layout has not been adopted anywhere and it can break the whole layout, this PR intentionally does not add the rules to form.less to avoid unexpected regressions.

| |Before|After|
|-|---------|------|
|OpenStack|||
|S3 compatible|||
|Google Cloud Storage|||


kenkendk:
I think grid is preferable in most cases, but it was not well supported in browsers back when the UI was developed.

luixxiul:
> it was not well supported in browsers back when the UI was developed.

I've got tons of idea to update CSS (LESS) codebase, let me see how it could be updated :smile:

PR #5344, opened 2024-07-29
Add support for using the faster built-in .NET zip.

kenkendk:
This PR adds the option for using the built-in in Zip library in .NET8.

The library appears substantially faster across systems, but lacks many features only found in SharpCompress.

Since both are essentially "zip" and would use this extension, the PR splits the current zip compression module, so it is possible to switch implementations via options. This strategy is the same as is used for the S3 backend that can use either Minio or AWS's library for the same destination.

The default setting is "auto" which means that most users will automatically use the built-in zip library, but users who require a specific compression algorithm will be using SharpCompress. If the built-in zip library fails to load an existing zip file, the implementation will automatically switch to using SharpCompress as it supports more compression algorithms and fail-safe modes for extracting information.


kenkendk:
To justify this PR I should add a performance comparison to show that there is actually some performance benefits.

kenkendk:
A simple timing test:
```
rm -rf db.sqlite target
mkdir target
time ./Duplicati.CommandLine backup ./target ./source --dbpath=./db.sqlite --blocksize=1mb --no-encryption --zip-compression-level=0 --no-backend-verification --zip-compression-library=???
```

Source dir is 10 files with 5.2GiB of data. Tested on MacOS with M3 chip:


No compression gives:
- SharpCompress: 33.74s user 11.14s system 423% cpu 10.594 total
- .NET Zip: 12.59s user 7.16s system 278% cpu 7.100 total

Setting default compression level
- SharpCompress: 158.91s user 9.59s system 604% cpu 27.867 total
- .NET Zip: 107.02s user 11.56s system 589% cpu 20.110 total

Issue #5337, opened 2024-07-28
Add a native Japanese storage provider to the destination


luixxiul:
- [x] I have searched open and closed issues for duplicates.
- [x] I have searched the [forum](https://forum.duplicati.com) for related topics.

----------------------------------------

## Environment info

- **Duplicati version**: Debug
- **Operating system**: irrelevant
- **Backend**: "Sakura Cloud"

## Description

Working on the translation into Japanese and almost having completed it (~95%), I would like to propose to add at lease one native Japanese storage provider to a backup destination for Japanese users' sake.

There are not many providers available, but "[Sakura Cloud](https://manual.sakura.ad.jp/cloud/objectstorage/about.html#objectstorage-spec)" (in Japanese) by Sakura Internet provides a S3 compatible storage service and I think it is a good option, considering the scale of the corporation.

- https://www.sakura.ad.jp/corporate/en/profile/
- https://finance.yahoo.com/quote/3778.T/

Here is the basic information to be provided and configured (please see the link above):

- endpoint: isk01.sakurastorage.jp
- region: jp-north-1
- signature_version: s3v4
- AWS SDK

I think it would be able to be added in the same way as https://github.com/duplicati/duplicati/pull/4314, and it would be ideal if the value of the region was applied by default since it is not available in the pulldown menu of the bucket regions.

![1](https://github.com/user-attachments/assets/bcd2cf6e-5525-4922-aad5-1886338667ae)

If needed, I could test the integration as I have an account on Sakura Internet.

## Steps to reproduce

1. Add a backup task
2. Select "S3 compatible" on the step two

- **Actual result**: there should be a Japanese object storage provider.

- **Expected result**: there is not a Japanese provider.

luixxiul:
Implemented by https://github.com/duplicati/duplicati/pull/5345.

PR #5345, opened 2024-07-29
Added Sakura Cloud to list of S3 providers

kenkendk:


kenkendk:
This fixes #5337
@luixxiul Can you test that this works as expected?

luixxiul:
> This fixes #5337 @luixxiul Can you test that this works as expected?

Sure, I have not used Sakura Cloud but it should not be difficult to do so as I have the account of the company. Let me see how it work :smile:

Duplicati/Library/Backend/S3/S3Backend.cs
```diff
@@ -66,6 +66,7 @@ public class S3 : IBackend, IStreamingBackend, IRenameEnabledBackend
             { "Infomaniak Swiss Backup cluster 2", "s3.swiss-backup02.infomaniak.com" },
             { "Infomaniak Swiss Backup cluster 3", "s3.swiss-backup03.infomaniak.com" },
             { "Infomaniak Public Cloud 1", "s3.pub1.infomaniak.cloud" },
+            { "さくらのクラウド (Sakura Cloud)", "isk01.sakurastorage.jp" }
```
luixxiul, line 69: According to example commands on the page https://manual.sakura.ad.jp/cloud/objectstorage/about.html, the endpoint URL seems to actually be `s3.isk01.sakurastorage.jp`.

It claims so on the actual UI as well:

PR #5353, opened 2024-07-30
Updating the default protocol tol TLS 1.3 if the system supports it

kenkendk:
This PR changes the default protocol to be TLS1.2+TLS1.3 if the system supports TLS 1.3, otherwise it uses FluentFTP defaults which seems to be TLS+TLS1.1+TLS1.2


kenkendk:
This should be updated to use `SslProtocols.None` as per this issue: https://github.com/robinrodricks/FluentFTP/issues/1623

PR #5360, opened 2024-07-30
Fix a word puzzle on localdatabase.html

luixxiul:
This PR intends to fix a word puzzle on localdatabase.html.

For "word puzzle", please see: https://techbase.kde.org/Development/Tutorials/Localization/i18n_Mistakes#Pitfall_#2:_Word_Puzzles.

### DIFF
--- a/Duplicati/Server/webroot/ngax/templates/localdatabase.html
+++ b/Duplicati/Server/webroot/ngax/templates/localdatabase.html
@@ -1,8 +1,5 @@
 <div ng-controller="LocalDatabaseController" class="localdb" >
-    <h1>{{'Local database for' | translate}} 
-        <b ng-show="Backup.Backup.Name">{{Backup.Backup.Name}}</b>
-        <b ng-hide="Backup.Backup.Name" translate>…loading…</b>
-    </h1>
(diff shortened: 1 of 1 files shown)


luixxiul:
I am not perfectly sure if this works as expected; please feel free to fix it if it is not a right solution.

luixxiul:
> Not sure the translation works correctly when there are html tags inside the translated string?

It should work fine. Here is an example:

https://github.com/duplicati/duplicati/blob/f444382e5f6c4d730bbe433cfe6cde02e2c96fcb/Duplicati/Server/webroot/ngax/templates/about.html#L12-L24

And the translation was properly handled:




HTML tags are apparently converted into placeholders on Transifex, and re-converted into HTML tags for PO files.

https://github.com/duplicati/duplicati/blob/f444382e5f6c4d730bbe433cfe6cde02e2c96fcb/Localizations/webroot/localization_webroot.pot#L3372-L3374

I do not see the reason why <b> tag would not be converted into a placeholder, so it should work fine...

kenkendk:
Yeah, you are right, if it works in that instance, it will also work for `<b>`

luixxiul:
Let's just revert the PR if it was found it indeed did not work :+1:

PR #5365, opened 2024-08-01
Run extract.sh for Localizations/webroot/localization_webroot.pot

luixxiul:
I am afraid of not following the official procedure, but this PR intends to be a follow-up to https://github.com/duplicati/duplicati/pull/5360.


kenkendk:
The .pot files are a bit special. They are generated, so they should not really be in the repo.
I usually just run the extract+push_to_transifex when I notice changes to translate-able strings.

Issue #5359, opened 2024-07-30
WebSocketException error on Microsoft.AspNetCore.Server.Kestrel


luixxiul:
- [x] I have searched open and closed issues for duplicates.
- [x] I have searched the [forum](https://forum.duplicati.com) for related topics.

----------------------------------------

## Environment info

- **Duplicati version**: Debug
- **Operating system**: Debian
- **Backend**: irrelevant

## Description

I am not quite sure how to reproduce the error consistently, but loading the same page several times quickly seems to return the error below related to websocket:

````
fail: Microsoft.AspNetCore.Diagnostics.ExceptionHandlerMiddleware[1]
      An unhandled exception has occurred while executing the request.
      System.Net.WebSockets.WebSocketException (0x80004005): The remote party closed the WebSocket connection without completing the close handshake.
       ---> Microsoft.AspNetCore.Connections.ConnectionResetException: Connection reset by peer
       ---> System.Net.Sockets.SocketException (104): Connection reset by peer
         --- End of inner exception stack trace ---
         at System.IO.Pipelines.Pipe.GetReadResult(ReadResult& result)
         at System.IO.Pipelines.Pipe.GetReadAsyncResult()
         at Microsoft.AspNetCore.Server.Kestrel.Core.Internal.Http.Http1UpgradeMessageBody.ReadAsyncInternalAwaited(ValueTask`1 readTask, CancellationToken cancellationToken)
         at System.Runtime.CompilerServices.PoolingAsyncValueTaskMethodBuilder`1.StateMachineBox`1.System.Threading.Tasks.Sources.IValueTaskSource<TResult>.GetResult(Int16 token)
         at Microsoft.AspNetCore.Server.Kestrel.Core.Internal.Http.HttpRequestStream.ReadAsyncInternal(Memory`1 destination, CancellationToken cancellationToken)
         at System.Runtime.CompilerServices.PoolingAsyncValueTaskMethodBuilder`1.StateMachineBox`1.System.Threading.Tasks.Sources.IValueTaskSource<TResult>.GetResult(Int16 token)
         at System.IO.Stream.ReadAtLeastAsyncCore(Memory`1 buffer, Int32 minimumBytes, Boolean throwOnEndOfStream, CancellationToken cancellationToken)
[...]

PR #5367, opened 2024-08-01
bug(5359): swallow WebSocketException exception when client aborted

npodbielski:
Fixes https://github.com/duplicati/duplicati/issues/5359#issuecomment-2258082674

### DIFF
--- a/Duplicati/WebserverCore/Middlewares/WebsocketExtensions.cs
+++ b/Duplicati/WebserverCore/Middlewares/WebsocketExtensions.cs
@@ -6,7 +6,8 @@ namespace Duplicati.WebserverCore.Middlewares;
 
 public static class WebsocketExtensions
 {
-    public static IApplicationBuilder UseNotifications(this IApplicationBuilder app, IEnumerable<string> allowedOrigins, string notificationPath)
+    public static IApplicationBuilder UseNotifications(this IApplicationBuilder app, IEnumerable<string> allowedOrigins,
+        string notificationPath)
     {
         var opts = new WebSocketOptions();
         if (allowedOrigins.Any())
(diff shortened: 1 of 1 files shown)


Duplicati/WebserverCore/Middlewares/WebsocketExtensions.cs
```diff
@@ -52,24 +53,40 @@ public static IApplicationBuilder UseNotifications(this IApplicationBuilder app,
...
+            WebSocketReceiveResult? receiveResult;
+            try
+            {
+                receiveResult = await webSocket.ReceiveAsync(new ArraySegment<byte>(buffer), cancellationToken);
+            }
+            catch (WebSocketException e)
+                when (e is
+                      {
+                          Message:
+                          "The remote party closed the WebSocket connection without completing the close handshake."
```
kenkendk, line 79: Seems a little brittle to rely on the exception message? Can it appear translated for instance?

npodbielski: Ok it should be better now. Though I still not sure how to solve this one: 

> Duplicati.WebserverCore.Exceptions.UnauthorizedException: Token family not found
   at Duplicati.WebserverCore.Services.TokenFamilyStore.GetTokenFamily(String userId, String familyId, CancellationToken ct) in /home/natan/git/FOOS/duplicati/Duplicati/WebserverCore/Services/TokenFamilyStore.cs:line 41
   at Duplicati.WebserverCore.Services.LoginProvider.PerformLoginWithRefreshToken(String refreshTokenString, CancellationToken ct) in /home/natan/git/FOOS/duplicati/Duplicati/WebserverCore/Services/LoginProvider.cs:line 29

npodbielski: I think while you refreshing browser constantly this is true `if (tokenFamily.Counter != refreshToken.Counter)` and token is removed from DB.

kenkendk: Yes, I think you can trigger it with a race where you reload the page after the request has been sent, but before the response is stored in the browser. In this case, the browser will have the previous token. Next request will send the old token which will mismatch with the database and it will clear the tokens.

I think it is a different PR that should solve that, the fix you made looks good for the original task.

PR #5369, opened 2024-08-01
Replace <br> tag with <p> tags to create paragraphs on modal-dialog, adding declarations for the tags

luixxiul:
This PR intends to replace `<br>` tag for paragraphs on `.modal-dialog`. It also should add declarations for `p` tags to override the default block margin values.

Before:

After:


Localizations/webroot/localization_webroot.pot
```diff
@@ -28,6 +28,12 @@ msgstr ""
 msgid "...loading..."
 msgstr ""
 
+#: index.html:294
+msgid ""
+"<p>Connection to server was rejected due to invalid authentication. Reload browser window to reconnect to server properly. </p>\n"
+"                <p>If this problem persist open this page from the TrayIcon instead.</p>"
```
luixxiul, line 34: I do not like the indents and `\n` here since it affects the localization UX on Transifex, but it looks it should be taken care of by our script to update pot files.

kenkendk:
I agree, `<br/>` is not easy to work with from a CSS perspective either.

luixxiul:
I believe the layout of the dialog, ie the spacing around sentences can be improved. I will work on it later :smile:

PR #5371, opened 2024-08-01
Update a string on EditUriBuiltins.js to reflect that it is not always expected to adjust a bucket name

luixxiul:
This PR intends to update a string on EditUriBuiltins.js to reflect that it is not always expected to adjust a bucket name.

Reviewing https://github.com/duplicati/duplicati/pull/5345, S3 providers do not always request the bucket to start with the username ([source](https://manual.sakura.ad.jp/cloud/objectstorage/about.html#objectstrage-bucket-create)), so I think it should be sensible to regard providers which request to do so are rather exceptions.


kenkendk:
It is not required for any provider that I know of. But for most services, the bucket names are "global" and not tied to the user. This creates problems when two different users want to create a bucket with the same name (e.g., "Backup" or "Photos"). 
For that reason, AWS recommends that you prefix the bucket name with your accountId to guarantee uniqueness. As an added benefit it reduces guess-ability as well.

I have just applied that logic to all S3 connections, regardless of the host.

It is a recommendation so I would prefer the wording to reflect this, could we use something like:
```
To prevent bucket naming conflicts, it is recommended to prepend your account id to the bucket name. Prepend automatically?
```

luixxiul:
I get it what you mean; checking Sakura Cloud S3 integration, I noticed there was a notification that the bucket name should be unique among the users.

luixxiul:
Updated the PR.

PR #5372, opened 2024-08-01
Update the description styling on sia.html

luixxiul:
This PR intends to update the styling on sia.html by replacing `<br>` and add a whitespace character after `Note:`.

Before:

before

After:

after


Duplicati/Server/webroot/ngax/less/base.less
```diff
@@ -561,6 +561,21 @@ ul.notification {
     }
 }
 
+.note {
+    & p {
+        /* Override default block margin */
+        margin-block: 0.5rem; /* Magic number */
```
luixxiul, line 567: The value can be configured later when we improve the whole UI.

PR #5377, opened 2024-08-01
Update elements styling in advanced options editor once more

luixxiul:
This PR intends to update elements in advanced options editor again by:

- Replacing div with p tags to wrap elements inside `longdescription` on the advanced options editor, so that there is gap between description and default value lines as p tag has block margin by default
- Using custom property for the block margin value to ensure consistency

Before:

before

After:

after


kenkendk (approved):

luixxiul:
I'll resolve the conflicts soon.

PR #5391, opened 2024-08-03
Fix and sort entries on Duplicati.License.csproj to fix license tab on about page

luixxiul:
This PR intends to fix the issue that some of the third party libraries are not listed on the license tab on about page.

Due to an error that licensedata.json for `SshNet.Security.Cryptography` was not included on `Duplicati.License.csproj`, these three libraries are not listed on the tab (on master branch):

- `SshNet.Security.Cryptography`
- `qcloud-sdk-dotnet`
- `uplink.NET`

This PR also should remove the hard-coded period from about.html for empty descriptions (please see `qcloud-sdk-dotnet`) and fix the license information for `SharePointPnP-Sites-Core` not being correctly specified. Duplicated periods are also going to be removed (please see `MimeKit` for example).

In order to tidy the list up, this PR also should sort entries.

Before:
before

After:
after

### DIFF
--- a/Duplicati/License/Duplicati.License.csproj
+++ b/Duplicati/License/Duplicati.License.csproj
@@ -27,6 +27,10 @@
       <Link>licenses\alphavss\License.txt</Link>
       <CopyToOutputDirectory>PreserveNewest</CopyToOutputDirectory>
     </Content>
+    <Content Include="..\..\thirdparty\alphavss\licensedata.json">
+      <Link>licenses\alphavss\licensedata.json</Link>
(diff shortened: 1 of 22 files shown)


luixxiul:
Note that the issue is not reproduced on `Duplicati - 2.0.7.103_canary_2024-04-19` and the issue was introduced since then at some point.

kenkendk:
> Note that the issue is not reproduced on Duplicati - 2.0.7.103_canary_2024-04-19 and the issue was introduced since then at some point.

There was some heavy merging going on to update the `.csproj` files from .NET4, so that is likely the reason.

Issue #5289, opened 2024-07-24
Pause and throttle icons should be able to be selected with tab key


luixxiul:
- [x] I have searched open and closed issues for duplicates.
- [ ] I have searched the [forum](https://forum.duplicati.com) for related topics.

----------------------------------------

## Environment info

- **Duplicati version**: Debug
- **Operating system**: Debian
- **Backend**: irrelevant

## Description

From the perspective of accessibility, the action icons to pause or throttle should be able to be selected with <kbd>tab</kbd> key:

https://github.com/duplicati/duplicati/blob/d3071e95bb363e70e4699b1180e02fd29e645e67/Duplicati/Server/webroot/ngax/index.html#L147-L148 

https://github.com/duplicati/duplicati/blob/d3071e95bb363e70e4699b1180e02fd29e645e67/Duplicati/Server/webroot/ngax/index.html#L198-L199

## Steps to reproduce

1. Open http://localhost:8200/ngax/index.html
2. Set focus to the page
3. Push <kbd>tab</kbd> key to select the pause or throttle icon

- **Actual result**: they cannot be selected.

- **Expected result**: they should be selected.

Jojo-1000:
See also this issue for other accessibility concerns: #5098

luixxiul:
Adopting WAI-ARIA should be conducted although it should require intensive research and take a lot of manpower.

PR #5395, opened 2024-08-03
Use button HTML elements for buttons on the header

luixxiul:
Fixes https://github.com/duplicati/duplicati/issues/5289

This PR intends to make buttons (pause, start, and throttle ones) selectable with <kbd>tab</kbd> key by changing them to native HTML `<button>` element.

[1.webm](https://github.com/user-attachments/assets/8f941ac4-90e8-4cd4-8bd0-2bbb2ccad0da)

For the reasoning of this change, please see: https://developer.mozilla.org/en-US/docs/Learn/Accessibility/WAI-ARIA_basics#when_should_you_use_wai-aria

> Ideally, you should always use [native HTML features](https://developer.mozilla.org/en-US/docs/Learn/Accessibility/HTML) to provide the semantics required by screen readers to tell their users what is going on.


kenkendk (approved):

luixxiul:
The conflict was fixed.

PR #5402, opened 2024-08-04
Replace PGP keyserver with OpenPGP.org

luixxiul:
This PR intends to replace PGP keyserver with OpenPGP.org.

This can be of just my preference; if we will stick to pgp.mit.edu, then it is fine and I am going to close this PR.


kenkendk:
Good catch. We have changed it in the signature file as well, because the MIT once seems to be unmaintained.

Issue #5393, opened 2024-08-03
Improve accessibility: implement WAI-ARIA specification


luixxiul:
- [x] I have searched open and closed issues for duplicates.
- [ ] I have searched the [forum](https://forum.duplicati.com) for related topics.

----------------------------------------

## Environment info

- **Duplicati version**: Debug
- **Operating system**: irrelevant
- **Backend**: irrelevant

## Description

Looking at [here](https://github.com/matrix-org/matrix-react-sdk/pulls?q=is%3Apr+accessibility), I noticed there was a lot of room to improve Duplicati when it comes to accessibility. It should make it easier to use Duplicati with accessibility software such as screen reader.

- https://developer.mozilla.org/en-US/docs/Learn/Accessibility/WAI-ARIA_basics

Please note accessibility can be improved with HTML itself:

- https://developer.mozilla.org/en-US/docs/Learn/Accessibility/HTML

## Steps to reproduce
Irrelevant

## Screenshots
Irrelevant

## Debug log
Irrelevant

Related issues: https://github.com/duplicati/duplicati/issues/1483 https://github.com/duplicati/duplicati/issues/5289

luixxiul:
This is blocked on https://github.com/duplicati/duplicati/pull/5039 (migration to Angular from AngularJS)

luixxiul:
Closing as stale. The UI has been greatly improved since this was opened.

PR #5405, opened 2024-08-05
Make steps header on adding/editing backup task accessible with keyboard

luixxiul:
For https://github.com/duplicati/duplicati/issues/5393

This PR intends to replace `div` elements for steps with `buttons` with `aria-labelledby` attributes to improve accessibility.

I am not perfectly satisfied with how the buttons are styled, but addressing it will be done in another PR.

## Before:

Accessibility checker on Firefox devtool:

before

## After:

Accessibility checker on Firefox devtool:

after

after1

With the changes the steps on the header can be selected with the keyboard:

[after.webm](https://github.com/user-attachments/assets/f2278ac9-b6ce-4021-8dc2-85c16ab4db5a)


Duplicati/Server/webroot/ngax/templates/addoredit.html
```diff
@@ -2,30 +2,30 @@
...
     </div>
     <!-- .steps -->
 
     <ol class="steps-legend">
-        <li ng-class="{active: CurrentStep == 0}" class="step1" ng-click="CurrentStep = 0" translate>General</li>
-        <li ng-class="{active: CurrentStep == 1}" class="step2" ng-click="CurrentStep = 1" translate>Destination</li>
-        <li ng-class="{active: CurrentStep == 2}" class="step3" ng-click="CurrentStep = 2" translate>Source Data</li>
-        <li ng-class="{active: CurrentStep == 3}" class="step4" ng-click="CurrentStep = 3" translate>Schedule</li>
-        <li ng-class="{active: CurrentStep == 4}" class="step5" ng-click="CurrentStep = 4" translate>Options</li>
+        <li id="steps-legend__1" ng-class="{active: CurrentStep == 0}" class="step1" ng-click="CurrentStep = 0" translate>General</li>
```
luixxiul, line 24: Should `step1` be rather `step0` as `CurrentStep` is 0 ?

kenkendk: Code-wise the index is zero, but for the user, the first step is "step 1".
Since these are not directly visible, I would suggest `step0` but I can see that the styling is inconsistent here and using `step1`. 

Lets keep it as it is now.

kenkendk (approved):

luixxiul:
The conflicts were fixed.

PR #5406, opened 2024-08-05
Update style rules for select element

luixxiul:
This PR intends to fix alignment of select elements and tidy up the style rules on form.less.

Before:
before

After:
after


Duplicati/Server/webroot/ngax/less/form.less
```diff
@@ -1,14 +1,16 @@
 // form.css for duplicati 2.0 | Alex Franzelin 2015
 
 form.styled {
-    div.leftflush input {
-        width: auto;
-        margin-top: 10px;
-    }
+    div.leftflush {
+        & input {
```
luixxiul, line 5: `&` is used here to make it easier to migrate LESS into vanilla CSS; on CSS when you nest type selectors you are expected to add `&` before them (cf.: https://www.w3.org/TR/css-nesting-1/#nest-selector).

kenkendk (approved):

luixxiul:
The conflicts were fixed.

PR #5408, opened 2024-08-05
Make strings for retention policy consistent

luixxiul:
This PR intends to update the strings about the retention policy on the options tab of adding/editing a backup task.

I picked up the string displayed under `Backup retention` when you select "custom backup retention"  on the options tab as the standard, since it is displayed to the user more easily and frequently, and the one displayed on the advanced options editor does not reflect the actual usage in a perfectly correct sense (For example, the option is not always used to reduce the number of versions as it accepts the unlimited time interval).

Still, the part of the sentence `and deletes every backup older than this` is going to be picked up from it as it is important information for users, who might misunderstand that the backup would be stored forever even after the specified time interval.


ts678:
TL;DR At least please change "an unlimited time interval" to "unlimited", but there's much here that can potentially be clearer.

This is a tough thing to describe, harder when trying to be brief. Code has changed a bit, with varying documentation editing.
I'm fairly familiar with it, and with the typical points of confusion, however there's only so much to say given the limited space.

Duplicati code changes seem to usually (but not always) hit the code help text. The jump to the user manual is usually missing.
Most of the text there is from program outputs many years back, but there have been many tweaks to the messages recently. 

Although this is probably a high-level organizational and resource question, I hope the user manual can get updates someday.

[Easier syntax for unlimited timeframes and intervals #3059](https://github.com/duplicati/duplicati/pull/3059)

> This allows the user to specify U for both the timeframe and interval to hopefully make some edge cases easier to understand

added the support for "U" in the time frame (to the left of the colon), but the help text was not changed. The time frame side (documentation removed here, perhaps to save space) became more important when a change was done to implicitly delete:

[Feature/issue 3008 retention optimizations #3026](https://github.com/duplicati/duplicati/pull/3026)

> --retention-policy now deletes backups that are outside of any time frame. No need to specify --keep-time as well. Help text was changed accordingly.

The time frame concept itself is a little confusing. Some people think they occur sequentially. Actually they overlap. They're ages.

I wonder if we need to document h m s as hours minutes and seconds? Hours are sometimes needed due to collision problems because (for example) 1 day backup schedule and 1 day interval may wind up with backups not exactly 1 day apart, so maybe a backup gets deleted (if too near), or maybe it doesn't. One can make spacing more as expected by using hours, e.g. 23h not 1D.

luixxiul:
Thanks for the comment!

> Duplicati code changes seem to usually (but not always) hit the code help text

I am wondering where the help text exist- I could not find it in this repository.

> I wonder if we need to document h m s as hours minutes and seconds? Hours are sometimes needed due to collision problems because (for example) 1 day backup schedule and 1 day interval may wind up with backups not exactly 1 day apart, so maybe a backup gets deleted (if too near),

Is it not taken care of by default? I thought so because there is `disable-time-tolerance` option which seems to me to be protection against that issue.

ts678:
> I am wondering where the help text exists- I could not find it in this repository.

I'm speaking mostly about what you've been editing (and fixing mixups between long and short). Compare to the manual.

`https://github.com/duplicati/duplicati/blob/master/Duplicati/CommandLine/CLI/help.txt`

is another chunk of help text which was recently changed, and in fact I can now see "help decimal" and I couldn't in 2.0.8.1.
This suggests it's not just a dead file, but I can't explain to you the exact code-level mechanics behind all the help systems.

The user manual is a separate repository created by the volunteer who wrote it originally (but has been low on time lately).
I'm not sure if all the Duplicati code devs know about this, but one might claim whoever takes a PR in should remind them.
The home page of the manual says:

> Corrections/changes/additions may be submitted as a pull request at Github: https://github.com/kees-z/DuplicatiDocs

and to me, this is one extra step to miss. I don't know if any rearranging is planned, especially given Duplicati Inc. changes.
I'm not told everything that's being planned there, but for some reason there's a big focus on text and translations lately...

luixxiul:
I'm a bit confused / nit-picky, since the file says that it is the manual for command line client and I am not sure if updating it is required / recommended when the change is related to the web UI as well... Apparently updating https://github.com/duplicati/duplicati/wiki/Developer-Documentation about updating the documentation is necessary :smiley: 

> for some reason there's a big focus on text and translations lately...

One of the reasons is that I have participated to the Japanese translation and almost completed it (to introduce Duplicati to friends) from ~30% as the existing one was not great if I'm honest.

ts678:
> disable-time-tolerance

I "think" this only applies to the `--keep-time` option, for example in the code below. It has been proposed to be general.

https://github.com/duplicati/duplicati/blob/b4b4cf281ca5333297bb68d82d503acbc8ab9a1d/Duplicati/Library/Main/Options.cs#L696-L699

2.0.8.1 help text also mentions `keep-time` (plus lots of dashes):

```
  --disable-time-tolerance (Boolean): Deactivates tolerance when comparing
    times
    When matching timestamps, Duplicati will adjust the times by a small
    fraction to ensure that minor time differences do not cause unexpected
    updates. If the option ----keep-time is set to keep a week of backups,
    and the backup is made the same time each week, it is possible that the
    clock drifts slightly, such that full week has just passed, causing
    Duplicati to delete the older backup earlier than expected. To avoid
    this, Duplicati inserts a 1% tolerance (max 1 hour). Use this option to
    disable the tolerance, and use strict time checking
    * default value: false
  ```
> I have always found the description confusing. Not sure if the new explanation works for everyone, but at least for me, it is easier to read.

As someone used to seeing reference documents specifying syntax, removing it removes both confusion and clarification. Internally it's thought of as a time frame and interval. or one can think of it as an age and a proximity to the previous one.
Messaging also uses terms like that. Below is one, and there are others later in the code (but some need heavier logging).

https://github.com/duplicati/duplicati/blob/b4b4cf281ca5333297bb68d82d503acbc8ab9a1d/Duplicati/Library/Main/Operation/DeleteHandler.cs#L318-L318

Trying to fit this all into a short text is hard, no doubt about it. I suppose I can try, but for the time being I'll just comment.

> Placeholders are D/W/Y for days/weeks/years and U for unlimited

I'm not sure D/W/Y are technically placeholders. They're sort of units. They only work on the left side of the removed syntax.
U works on both sides, except when speaking about the interval side, it means the backups can never be too close together. 
When used as a time frame, it means the time frame never ends, so this is unlimited in the large direction, rather than small.

The removal of "age" in favor of a consistent "next" is another explicit to implicit move, and may beg "next relative to what". Previous description with "age" also said "by deleting most of the old backups" which may be alarming, even if may be true. Removing such alarm potential is fine with me, but I'm still not sure I like removing the concept of a backup having an age.

```
C:\Duplicati\duplicati-2.0.8.1_beta_2024-05-07>Duplicati.CommandLine.exe help keep-time
  --keep-time (Timespan): Keep all versions within a timespan
    Use this option to set the timespan in which backups are kept.
```

might already be rewritten (the attention to messages is much appreciated BTW even if I'm struggling with this tough one).
IMO how retention is described should use consistent language in all of the different ways that one can specify a retention.

"The syntax is: 7D:1D,4W:1W,36M:1M. This example keeps one backup for each of the next 7 days" and subsequent "one" is begging questions like "what if I want two backups -- do I change the 1 to a 2"? Well no, the "1" is the unexplained interval.

Explaining in a specific simplified example can be easy to read standalone, but leaves it to the reader to model a framework.

PR #5412, opened 2024-08-06
Update SSHv2/Strings.cs

luixxiul:
This PR intends to update SSHv2/Strings.cs by editing its strings.

Before:
before

After:
after


Duplicati/Library/Backend/SSHv2/Strings.cs
```diff
@@ -53,11 +53,11 @@ internal static class SSHv2Backend
...
+        public static string DescriptionSshkeyfileLong { get { return LC.L(@"Points to a valid OpenSSH keyfile. If the file is encrypted, the password supplied is used to decrypt it. If the keyfile is specified, the password is not used to authenticate."); } }
         public static string DescriptionSshkeyfileShort { get { return LC.L(@"Uses a SSH private key to authenticate"); } }
-        public static string DescriptionSshkeyLong(string urlprefix) { return LC.L(@"An url-encoded SSH private key. The private key must be prefixed with {0}. If the file is encrypted, the password supplied is used to decrypt the keyfile. If this option is supplied, the password is not used to authenticate.", urlprefix); }
+        public static string DescriptionSshkeyLong(string urlprefix) { return LC.L(@"An url-encoded SSH private key. The private key must be prefixed with {0}. If the key is encrypted, the password supplied is used to decrypt it. If the private key is specified, the password is not used to authenticate.", urlprefix); }
```
luixxiul, line 58: Here "the file" was replaced with "the key", since the option expects the encoded key, not the keyfile.

PR #5413, opened 2024-08-06
rework certificates to support EC keys and preparation for GUI options

mnaiman:
added EC certificate support - dropped by kestrel introduction
tested on Windows & Linux
backwards compatible for those who have already certificates setup


Duplicati/Server/WebServerLoader.cs
```diff
@@ -57,6 +64,16 @@ public static class WebServerLoader
...
 
+    /// <summary>
+    /// The default certificate file for https
+    /// </summary>
+    public const string DEFAULT_OPTION_CERTIFICATEFILE = "server-webui.pfx";
+
+    /// <summary>
+    /// Option for setting if to use HTTPS
+    /// </summary>
+    public const string OPTION_USEHTTPS = "webservice-usehttps";
```
kenkendk, line 75: This is slightly confusing. Is this meant to support disabling HTTPS without deleting the certificate? If so, perhaps we can do something else. Maybe just have it inverted to `webservice-disablehttps`? 

Or is it meant to be a safguard the prevents the server to start without a valid SSL configuration?

I think there are a few places that check for the certificate and they all need to be updated with this logic.

Duplicati.Library.RestAPI/Database/ServerSettings.cs
```diff
@@ -563,17 +565,28 @@ public string ServerListenInterface
             }
         }
 
-        public X509Certificate2 ServerSSLCertificate
+        public bool ServerUseHTTPS
```
kenkendk, line 568: I think it makes sense to have a general property that explains if HTTPS should be enabled.
I think this should replace the different checks for empty strings.

Duplicati.Library.RestAPI/Strings.cs
```diff
@@ -29,8 +29,9 @@ public static class Program
...
         public static string WebserverPortDescription { get { return LC.L(@"The port the webserver listens on. Multiple values may be supplied with a comma in between."); } }
-        public static string WebserverCertificateFileDescription { get { return LC.L(@"The certificate and key file in PKCS #12 format the webserver use for SSL. Only RSA/DSA keys are supported."); } }
-        public static string WebserverCertificatePasswordDescription { get { return LC.L(@"The password for decryption of certificate PKCS #12 file."); } }
+        public static string WebserverUseHTTPSDescription { get { return LC.L(@"If provided, determine whether to use HTTPS. If not provided, compatibility mode determine based on certificate presence about usage of HTTPS."); } }
+        public static string WebserverCertificateFileDescription { get { return LC.L(@"The certificate and key file in PKCS #12 format the webserver use for SSL."); } }
+        public static string WebserverCertificatePasswordDescription { get { return LC.L(@"The password for decryption of certificate PKCS #12 file. USE STRONG PASSWORD, CERTIFICATE FILE WILL BE SAVED TO DRIVE ENCRYPTED BY CERTIFICATE PASSWORD!"); } }
```
kenkendk, line 34: Could we use a random generated passphrase for encrypting the certificate?
If the user has placed a file, they can use as sloppy a password as they like.
If the user uploads the certificate, we only need the password for decrypting, but we could save it with a strong randomly generated password.

Also, why do we need to store it as a file? Would it be sufficient to store it in the database?

Duplicati/GUI/Duplicati.GUI.TrayIcon/Program.cs
```diff
@@ -114,10 +114,9 @@ public static int Main(string[] _args)
                 // Tell the hosted server it was started by the TrayIcon
                 Server.Program.Origin = "Tray icon";
 
-                var cert = Server.Program.DataConnection.ApplicationSettings.ServerSSLCertificate;
                 var scheme = "http";
 
-                if (cert != null && cert.HasPrivateKey)
+                if (!string.IsNullOrEmpty(databaseConnection.ApplicationSettings.ServerSSLCertificate))
```
kenkendk, line 119: Would this fail with `--webservice-usehttps=false` ?

Duplicati/Server/WebServerLoader.cs
mnaiman: Yes it was meant to be support for disable HTTPS without deleting certificate.

Duplicati.Library.RestAPI/Database/ServerSettings.cs
mnaiman: Checks for empty strings are there for backwards compatibility.

Duplicati.Library.RestAPI/Strings.cs
mnaiman: user will provide certificate file in pfx form, which has to have password by definition (could be empty). I like proposed idea (was thinking in same way) but that would need to reencrypt private key with generated password. I didnt find way how to import, reecrypt, export certificate with EC keys without new libraries. :/ I will look again, there was some bug in past with openssl(linux version).

mnaiman: Storing certificate to file is only was to pass certificate to kestrel, because kestrel fail to load certificate with EC key from X509Certificate2 instance.

Duplicati/GUI/Duplicati.GUI.TrayIcon/Program.cs
mnaiman: again for backward compatibility with users that have now certificate set, but dont use new ServerUseHTTPS parameter

mnaiman: without that they would fall back to http with new version.
its true, that without backward compatibility, it would simplify code a little bit

Duplicati.Library.RestAPI/Strings.cs
mnaiman: found a way to reencrypt, that would simplify code

mnaiman: will be fixed in next commit, guid is generated as password for storing in DB

Duplicati/GUI/Duplicati.GUI.TrayIcon/Program.cs
kenkendk: I see, I was not reading it with a good understanding of upgrade for existing certificate users.

Duplicati.Library.RestAPI/Strings.cs
mnaiman: fixed

mnaiman: password is unique generated guid

mnaiman:
Should I adjust something? If there will be canary or debug build (container) I can update all my instances for upgrade test.

kenkendk:
This PR is extended in #5608

PR #5417, opened 2024-08-08
Fix inconsistent capitalization of 'ZIP'

luixxiul:
Based on https://english.stackexchange.com/questions/497639/what-is-the-correct-capitalization-for-zip-file/497641#497641 (and http://www.iana.org/assignments/media-types/application/zip), this PR intends to apply consistent capitalization to "ZIP".

Please note that the intention of this PR is to address the issue that both "Zip" and "zip" are used on UI, and as there does not seem to be a consensus about its capitalization in general, it should be fine to use "Zip" instead as well, as long as the inconsistency is fixed, ie. mixing "Zip" and "zip" is avoided.

Please let me know if we will prefer "Zip" to "ZIP", so that I will replace them.


kenkendk (approved):

luixxiul:
The conflicts were fixed.

Issue #5452, opened 2024-08-14
Slow loading for v2.0.9.102


kenkendk:
The update to ASP.NET Core with Kestrel has added some significant loading delays, up to several seconds.

The problem seems to be the massive number of individual requests being performed to load all the scripts during loading.

We can investigate using defered loading or some kind of pre-processing (webpack, etc).

The real fix is to update to a modern Angular that has all this baked in, but for now we may be able to do the preprocessing just for the release builds.

It should significantly reduce load time if all scripts are combined into a single file, and the includes in `index.html` are replaced with a single entry.

kenkendk:
This will be fixed with ngclient

PR #5461, opened 2024-08-15
Feature/mini webpack for release

kenkendk:
This could be a partial fix for #5452


kenkendk:
Not implementing this as we should focus on the new Angular FE

Issue #5420, opened 2024-08-12
Improve security of stored credentials


kenkendk:
When using the UI a number of sensitive credentials are stored in the database `Duplicati-server.sqlite`.
Previously, these credentials were partially obfuscated with SQLite RC4 scrambing, but after .NET8 this is no longer possible.

Since Duplicati is used in many different setups the storage of credentials should support different storage options. This can be done using a modular design similar to how compression, encryption, etc is already done.

kenkendk:
Closed with https://github.com/duplicati/duplicati/pull/5472

PR #5466, opened 2024-08-16
Feature/encryptsettings

marceloduplicati:
This is a proposed change to address [issue 5420](https://github.com/duplicati/duplicati/issues/5420)

Secure storage of credentials is a conundrum faced in IT and it is mostly tackled with mitigation, since no matter where the setting is stored and how it is encrypted, even with HSM hardware devices, there will always be a point of access, after all, the legitimate access of the secret is a necessity

In the spirit of mitigation, the following changes are proposed to further secure the credential storage within Duplicati's database.

At the root level, it begins with defining a Settings Encryption Key, it can be set via environment variable, namely SETTINGS_ENCRYPTION_KEY or if undefined Duplicati will use a self computed key derived from the motherboard serial number.

This change was made in such way that existing unencrypted records are supported, but whenever a change is made in configuration, the sensitive fields are automatically encrypted. Of course this means that older versions would not recognize the data as a valid input.


Duplicati.Library.RestAPI/Database/Connection.cs
```diff
@@ -36,6 +37,7 @@ public class Connection : IDisposable
         public const int ANY_BACKUP_ID = -1;
         public const int SERVER_SETTINGS_ID = -2;
         private readonly Dictionary<string, Backup> m_temporaryBackups = new Dictionary<string, Backup>();
+        private static string[] _encryptedFields = { ServerSettings.CONST.JWT_CONFIG, ServerSettings.CONST.PASSPHRASE_FIELDNAME };
```
kenkendk, line 40: I was searching for `ArgumentType.Password` in the codebase and found quite a lot of unexpected strings.
Should we look for these dynamically (so we pick up new entries automatically) or compile a list of known password arguments? For instance `auth-password` was used in some places.

Duplicati/WebserverCore/Endpoints/V1/Backups.cs
```diff
@@ -79,7 +80,7 @@ public static void Map(RouteGroupBuilder group)
                     Include = y.Include,
                     Expression = y.Expression,
                 }),
-                TargetURL = x.Backup.TargetURL,
+                TargetURL = EncryptedFieldHelper.Decrypt(x.Backup.TargetURL),
```
kenkendk, line 83: This looks like it will be decrypted automatically when leaving the `Connection` class?

Duplicati/WebserverCore/Endpoints/V1/Backups.cs
```diff
@@ -161,7 +162,7 @@ private static Dto.CreateBackupDto ExecuteAdd(Connection connection, Dto.BackupA
                 Name = data.Backup.Name,
                 Description = data.Backup.Description,
                 Tags = data.Backup.Tags,
-                TargetURL = data.Backup.TargetURL,
+                TargetURL = EncryptedFieldHelper.Encrypt( data.Backup.TargetURL),
```
kenkendk, line 165: Is this required? Would the `Connection` helper not encrypt this when saving?
I think it reduces complexity if only the database interaction is aware of the encryption, so anything in/out of `Connection` appears unencrypted.

kenkendk (approved):

Duplicati/WebserverCore/Endpoints/V1/Backups.cs
marceloduplicati: Yes, it is required. As explained [here](https://github.com/duplicati/duplicati/pull/5466#discussion_r1721817614) in the way it was implemented the TargetURL field of backup table is handled outside of Connection.

Duplicati/Library/Utility/Duplicati.Library.Utility.csproj
```diff
@@ -8,7 +8,12 @@
   </PropertyGroup>
 
   <ItemGroup>
-    <PackageReference Include="Microsoft.Win32.Registry" Version="4.7.0" />
+    <PackageReference Include="DeviceId" Version="6.7.0" />
+    <PackageReference Include="DeviceId.Linux" Version="6.4.0" />
+    <PackageReference Include="DeviceId.Mac" Version="6.3.0" />
+    <PackageReference Include="DeviceId.Windows" Version="6.6.0" />
+    <PackageReference Include="DeviceId.Windows.Wmi" Version="6.6.0" />
+    <PackageReference Include="Microsoft.Win32.Registry" Version="5.0.0" />
```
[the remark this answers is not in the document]
marceloduplicati: Yes. `<PackageReference Include="Microsoft.Win32.Registry>` can be safely removed.

Duplicati/Library/Utility/Duplicati.Library.Utility.csproj
```diff
@@ -8,7 +8,12 @@
   </PropertyGroup>
 
   <ItemGroup>
-    <PackageReference Include="Microsoft.Win32.Registry" Version="4.7.0" />
+    <PackageReference Include="DeviceId" Version="6.7.0" />
+    <PackageReference Include="DeviceId.Linux" Version="6.4.0" />
+    <PackageReference Include="DeviceId.Mac" Version="6.3.0" />
+    <PackageReference Include="DeviceId.Windows" Version="6.6.0" />
+    <PackageReference Include="DeviceId.Windows.Wmi" Version="6.6.0" />
```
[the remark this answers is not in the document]
kenkendk: Looks like it is required after all...

kenkendk:
Replaced by #5472

PR #5507, opened 2024-08-30
Supplement the Chinese translation for zh_CN.

nadirvishun:
The current translation for Chinese is quite lacking, so I have organized and added translations for `zh_CN`, including both frontend and backend. Initially, I wanted to handle translations for `zh_HK` and `zh_TW` as well, but after reviewing the existing translations, I found that they differ in language style from `zh_CN`, so I will not proceed with them. Let's wait for someone more suitable to handle those.


kenkendk:
Hi @nadirvishun thanks for the translation work.

However, it looks like you have modified the po/mo files directly?

We are using Transifex to manage the translations: https://transifex.com/duplicati
The logic is then to extract files from the project and send to Transifex.
Once they are translated, they are pulled back from Transifex.

You can see the scripts here:
https://github.com/duplicati/duplicati/tree/master/Localizations

With our current setup, we cannot handle changes directly to the mo/po files as they are overwritten when we pull from Transifex.

I can see all three language variants are already registered in Transifex, so it should be a matter of updating the strings in Transifex.

Can you somehow get your changes extracted and imported to Transifex?

I will add a README to the localization folder explaining the setup for future translators.

PR #5514, opened 2024-08-30
Added README files to explain the localization process

kenkendk:


luixxiul:
For confirmation, it it only possible for someone with proper permissions to update the strings on Transifex by running `extract_all.sh` and `push_source_files_to_transifex.sh` on this repository, meaning it is not possible for random users to update strings, right?

PR #5540, opened 2024-09-06
Replacing deprecated WebClient for HttpClient

marceloduplicati:
This is a proposal for replacing the deprecated WebClient with the HttpClient.

This commit implements the core functionality following the HttpClientFactory pattern as suggested by the framework. It also supports an alternative approach in areas where the HttpClient pattern is not desirable.

The first backend component to be migrated is WebDav, and the HttpClient pattern is also applied to the update downloader.

Once merged, other backend components can be migrated, eventually removing the deprecated WebClient from the codebase.

Additionally, the debug-profind-file option from webdav, which was a remnant from early development, is removed as it is no longer useful for debugging different webdav server responses.


Duplicati/Library/AutoUpdater/UpdaterManager.cs
```diff
@@ -316,10 +318,16 @@ public static UpdateInfo CheckForUpdate(ReleaseType channel = ReleaseType.Unknow
...
-                        wc.Headers.Add("X-Install-ID", InstallID);
-                        wc.DownloadFile(url, tmpfile);
+
+                        using var request = new HttpRequestMessage(HttpMethod.Get, url);
+
+                        request.Headers.Add(System.Net.HttpRequestHeader.UserAgent.ToString(), string.Format("{0} v{1}{2}", APPNAME, SelfVersion.Version, string.IsNullOrWhiteSpace(InstallID) ? "" : " -" + InstallID));
+                        request.Headers.Add("X-Install-ID", InstallID);
+
+                        using var timeoutToken = new CancellationTokenSource();
+                        timeoutToken.CancelAfter(HttpContextSettings.ReadWriteTimeout);
```
kenkendk, line 328: This should be `HttpContextSettings.OperationTimeout` as that is the overall operation timeout.
I think `HttpContextSettings.ReadWriteTimeout` will have to be deprecated as this concept is not supported with `HttpClient`.

Duplicati/Library/AutoUpdater/UpdaterManager.cs
```diff
@@ -439,15 +447,15 @@ public static bool DownloadUpdate(UpdateInfo version, PackageEntry package, stri
...
-                            var wreq = (System.Net.HttpWebRequest)System.Net.WebRequest.Create(url);
-                            wreq.UserAgent = string.Format("{0} v{1}", APPNAME, SelfVersion.Version);
-                            wreq.Headers.Add("X-Install-ID", InstallID);
+                            using var request = new HttpRequestMessage(HttpMethod.Get, url);
+
+                            request.Headers.Add(System.Net.HttpRequestHeader.UserAgent.ToString(), string.Format("{0} v{1}", APPNAME, SelfVersion.Version));
+                            request.Headers.Add("X-Install-ID", InstallID);
+
+                            using var timeoutToken = new CancellationTokenSource();
+                            timeoutToken.CancelAfter(HttpContextSettings.ReadWriteTimeout);
```
kenkendk, line 456: This should be `HttpContextSettings.OperationTimeout` as that is the overall operation timeout.

Duplicati/Library/Backend/WEBDAV/WEBDAV.cs
```diff
@@ -59,6 +84,16 @@ public class WEBDAV : IBackend, IStreamingBackend
         //private static readonly byte[] PROPFIND_BODY = System.Text.Encoding.UTF8.GetBytes("<?xml version=\"1.0\"?><D:propfind xmlns:D=\"DAV:\"><D:allprop/></D:propfind>");
         private static readonly byte[] PROPFIND_BODY = new byte[0];
 
+        /// <summary>
+        /// The default timeout in seconds for PUT/GET file operations
+        /// </summary>
+        private const int LONG_OPERATION_TIMEOUT_SECONDS = 30000;
```
kenkendk, line 90: I like this distinction between long/short. Should it be a general option?

Duplicati/Library/UsageReporter/ReportSetUploader.cs
```diff
@@ -70,23 +73,20 @@ public static Tuple<Task, IWriteChannel<string>> Run()
...
-                                        using (var rs = areq.GetRequestStream())
-                                            Library.Utility.Utility.CopyStream(fs, rs);
+                                        using var request = new HttpRequestMessage(HttpMethod.Post, UPLOAD_URL);
+                  
+                                        request.Content = new StreamContent(fs);
 
-                                        using (var resp = (HttpWebResponse)areq.GetResponse())
-                                            rc = (int)resp.StatusCode;
+                                        using var timeoutToken = new CancellationTokenSource();
+                                        timeoutToken.CancelAfter(HttpContextSettings.ReadWriteTimeout);
```
kenkendk, line 86: This should be `HttpContextSettings.OperationTimeout` as that is the overall operation timeout.

kenkendk (approved):
Only the timeout property needs to be changed

Duplicati/Library/Backend/WEBDAV/WEBDAV.cs
marceloduplicati: I believe so, because for operations such as fetching lists, or deleting, it is expected to be much shorter than a 50mb+/- block upload.

I am wondering if they should be constants or a parameter with a set default, allowing the user to override via configuration if they deem necessary.

kenkendk (approved):

PR #5547, opened 2024-09-09
Unify password detection

kenkendk:
Moved the password option lookup into a shared place so the same logic can be used from multiple places when guarding password output.


kenkendk:
Already implemented, but slightly different

PR #5558, opened 2024-09-13
Add dynamic timeout to WebDAV upload/download

kenkendk:
Enabled in WebDAV with HttpClient


kenkendk:
Closing in favor of #5931

PR #5561, opened 2024-09-17
Added healt checks to server.

### LINKED ISSUE #5560: Add a `/health` endpoint
To support setups in Docker better the server should expose `/health` as a simple mapper that just returns `200 - OK` to signal that the system is responding.

Also, since cURL is not part of the Docker image, the `server-util` could be adapted to include a `health` command, like:
```bash
> server-util health
```

Which would return 0 if the server responds, and 1 if it fails.

kenkendk:
Added health command to server-util.
This fixes #5560

### DIFF
--- a/Duplicati/CommandLine/ServerUtil/Commands/Health.cs
+++ b/Duplicati/CommandLine/ServerUtil/Commands/Health.cs
@@ -1,0 +1,38 @@
+using System.CommandLine;
+using System.CommandLine.NamingConventionBinder;
+
+namespace Duplicati.CommandLine.ServerUtil.Commands;
+
+public static class Health
+{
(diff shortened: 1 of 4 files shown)


tygill:
Is it also possible to add a healthcheck statement to the default Dockerfile? I think something like the following would work:

```
HEALTHCHECK --interval=5m --timeout=3s \
  CMD /opt/duplicati/duplicati-server-util health
```

Reference:
https://docs.docker.com/reference/dockerfile/#healthcheck

kenkendk:
@tygill Yes, I think that should be possible as well. 
If we add it to the base image, there needs to be some logic that checks the port, so it also works if the user has a different port on the image.

PR #5574, opened 2024-09-29
Implemented async version of `Get` for all backends.

kenkendk:
This is the first step of moving towards fully `async` backends with improved cancellation support.
Most updates are simple. SSH and Aliyun backends are now actually async with TPL wrapping the `BeginInvoke` / `EndInvoke` methods.
This adds `ConfigureAwait(false)` to all backend async methods.
For async methods that are still executed synchronously, the `Await()` extension method is now used everywhere to ensure the same handling.

Added `DeleteAsync`, `RenameAsync`, `TestAsync` and `CreateFolderAsync` for completion


kenkendk:
This change touches ALL backends.

Although the changes are quite systematic, there is a possibility of one or more backends breaking.

The tested backends are:
- file
- ssh
- s3
- webdav
- mega
- dropbox
- b2
- googledrive

PR #5589, opened 2024-10-14
Removing the legacy FTP Backend

marceloduplicati:
For this change, what was known as the alternate FTP client which uses FluentFTP library now officially replaces the FTP Backend which still used the deprecated FtpWebrequest.

To maintain compatibility with existing configurations, both ftp and aftp targets and configuration prefixes are accepted.

Tests were performed to ensure that existing configured backups with aftp targets and configurations are accepted and processed by the unified FTP backend.


ts678:
> Tests were performed to ensure that existing configured backups with aftp targets and configurations are accepted and processed by the unified FTP backend.

There are probably statistics in the usage reporter, but I suspect `ftp` might be more commonly picked. What of options?

```
Options:
  --ftp-passive = false
    Activate this option to make the FTP connection in passive mode, which works better with some firewalls. If the option --ftp-regular is set, this option is ignored.
  --ftp-regular = true
    Activate this option to make the FTP connection in active mode. Even if the option --ftp-passive is also set, the connection will be made in active mode.
  --auth-password
    The password used to connect to the server. This may also be supplied as the environment variable "AUTH_PASSWORD".
  --auth-username
    The username used to connect to the server. This may also be supplied as the environment variable "AUTH_USERNAME".
  --use-ssl
    Use this option to communicate using Secure Socket Layer (SSL) over ftp (ftps).
  --disable-upload-verify
    To protect against network failures, every upload will be attempted verified. Use this option to disable this verification to make the upload faster but less reliable.
```

Sorry I haven't looked through the code to see if everything is already mapped transparently, so I'm picking on the test note.

> To maintain compatibility with existing configurations

seems the goal, and that includes `ftp`, all its options, and its behavior without any options (does that default to active FTP?).

marceloduplicati:
@ts678 That is a hawks eye, thanks. The test was made to ensure the unified backend would process both ftp and aftp options, but indeed there are some incompatible settings.

Both : 
ftp-passive  boolean
ftp-regular  boolean   

From the legacy FTP Backend will be demuxed into **ftp-data-connection-type**

And **use-ssl** into **ftp-ssl-protocols**

Lets get that sorted and run specific tests forcing all the options/modes.

ts678:
Active FTP is probably more troublesome than passive, so I think Fluent FTP made the better choice (if I read our `help` right).
While I'm pushing for not breaking existing `ftp` configs, I'm also hoping to maximize chance that a simple new config works.

marceloduplicati:
With this change, any existing FTP backend configuration that had the flags set will work with the same behavior and issue warnings about the deprecated flags.

The precedence of the parameters matches the documentation.

kenkendk (approved):

PR #5600, opened 2024-10-25
Implemented explict support for IPv6 listener in the webserver.

### LINKED ISSUE #4177: Web interface not accessible via IPv6

## Environment info
- **Duplicati version**: 2.0.5.104, 2.0.5.1
- **Operating system**: Windows Server 2012 R2
- **Backend**: OneDrive for Business (Microsoft Graph API)

## Description
The Web server doesn't seem to listen on any IPv6 address on any interface, making remote access impossible for machines on IPv6-only networks.

## Steps to reproduce
1. Install Duplicati on Windows (as a service).
2. Allow remote access in the Duplicati settings and restart.
3. Try accessing the Web interface using any of the machine's IPv6 addresses (locally or from a remote computer, doesn't matter).
- **Actual result**:
Connection timeout.
- **Expected result**:
Successful connection to the Web interface.

kenkendk:
Using `any`/`all`/`*` will now listen on both IPv4 and IPv6 if available. 
To force IPv4-any, use `0.0.0.0` and similar `[::]` for IPv6. 
~The `loopback` value listens on IPv4 and the new `ipv6loopback` value selects IPv6 loopback. 
It is not possible to listen on both loopback interfaces at the same time.~

The `loopback` value now listens on both IPv4 and IPv6 if available (thanks @ts678  and @marceloduplicati for pointing to the solution).

This fixes #4177


ts678:
> It is not possible to listen on both loopback interfaces at the same time.

[KestrelServerOptions.ListenLocalhost Method](https://learn.microsoft.com/en-us/dotnet/api/microsoft.aspnetcore.server.kestrel.core.kestrelserveroptions.listenlocalhost?view=aspnetcore-8.0) (Microsoft) seems to say otherwise FWIW. There might be other factors involved.

marceloduplicati:
Indeed ListenLocalhost uses LocalhostListenOptions which tries to bind to both ipv4 and ipv6 loopback.

https://github.com/dotnet/aspnetcore/blob/v8.0.0/src/Servers/Kestrel/Core/src/LocalhostListenOptions.cs

I made a quick test just replacing the listen for **ListenLocalHost** and it did bind and listen to both (and served correctly):

> tcp6       0      0  ::1.8200               *.*                    LISTEN
> 
> tcp4       0      0  127.0.0.1.8200         *.*                    LISTEN

Maybe the new parameter can be dropped and just adopt this call.

kenkendk:
I Implemented the fixes using `ListenLocalHost` as suggested.

PR #5608, opened 2024-10-29
Feature/support ec certificates

kenkendk:
This PR is an update to #5413
The update changes the flag from enabling HTTPS to support disabling HTTPS so it is less work to maintain backwards compatibility.
It also adds a way to update the certificate via the API so a future UI feature can support updating the certificate and restarting the server.

This PR also adds automatic detection of certificate hashes for the given certificates, such that the trayicon and serverutil will now pick up and trust the specific certificate. 
This makes it a little easier to work with self-signed certificates, but still requires system-level trust if shown in the browser.


kenkendk:
@mnaiman Could you try this on your system(s) as well?

mnaiman:
Yes, for windows I can build myself, can you build for linux or docker? Or is there any debug build produced by CI/CD?

kenkendk:
I can try on Linux, we do not yet have automated builds :/

kenkendk:
Tested on Ubuntu 24.04.1 with no issues.

Commands used to generate certificate:
```bash
> openssl ecparam -name prime256v1 -genkey -noout -out private-key-ec.pem
> openssl req -new -x509 -key private-key-ec.pem -out cert-ec.pem -days 360
> openssl pkcs12 -export -inkey private-key-ec.pem -in cert-ec.pem -out cert-ec.pfx
```

The starting server/trayicon with:
```
--webservice-sslcertificatefile=./cert-ec.pfx
--webservice-sslcertificatepassword=<password>
```

mnaiman:
Tested on Windows:
1) Took my personal database from main machine, started without any ssl related params and it failed as described in screenshot.

image

2) Started same db with 2 ssl params and EC certificate and started ok - working.

3) Stopped, started again without 2 ssl params and started without, started ok - working.

So working ok, probably just needed to be rewritten to db again.

kenkendk:
I tried the upgrade path and the issue here is that the certificate is actually stored unencrypted in the database prior to this PR. So when trying to load the certificate it fails due to an empty password.

I have updated the PR to handle this, and then it works on non-Windows.
Steps to reproduce:
1. Generate key
```bash
> openssl genrsa -out private-key-rsa.pem 3072
> openssl req -new -x509 -key private-key-rsa.pem -out cert-rsa.pem -days 360
> openssl pkcs12 -export -inkey private-key-rsa.pem -in cert-rsa.pem -out cert-rsa.pfx
```

2. Run TrayIcon on master branch with:
```bash
--webservice-sslcertificatefile=./cert-rsa.pfx
--webservice-sslcertificatepassword=<password>
```

3. Stop the application (certificate is now in database)
4. Checkout this branch
5. Run TrayIcon without parameters

kenkendk:
@mnaiman Tested again on Windows (Arm64) and it works without any issues.
Is your certificate originally stored with a .NET4-based build?
If so, I think there is an incompatibility between the stored format and the way .NET8 reads it.

mnaiman:
Yes, it was stored in origin .NET4 build (this issue is blocker for me to update).
You are magician! Now it works in both cases on Windows.
Looking forward to new canary to upgrade.

PR #5617, opened 2024-11-01
Updated rclone backend to handle empty options 

kenkendk:
Empty option values are now treated the same as missing options (apply default values).
Made the local repo optional in the UI because it has a default value.


ts678:
> Made the local repo optional in the UI because it has a default value.

Is there a reason why it's necessary at all? Rclone [appears to think not](https://rclone.org/local/#configuration), except to add consistency.
Their approach is very reminiscent of Duplicati [Local folder or drive](https://docs.duplicati.com/en/latest/05-storage-providers/#local-folder-or-drive) where you can just type  path.

PR #5658, opened 2024-11-08
Bumped Avalonia to match latest Tmds package

kenkendk:


kenkendk:
Confirmed working on Ubuntu 23.10, Debian 12.5, MacOS and Windows.

PR #5685, opened 2024-11-18
Added option to pass in pre-parsed filter to reduce static memory usage.

kenkendk:
Added option to let the filter pre-calculate the default result.
If this is applied in all places, we can remove the static analyzer cache.


kenkendk:
Closing for now, the use of static caches should be revisited

PR #5761, opened 2024-12-11
Changed to ip address instead of hostname for better support without DNS

kenkendk:
On some systems `localhost` will resolve to something different than `127.0.0.1`.
This changes a few strings to use the assigned IP instead of `localhost`.
This needs to be tested with IPv6 as well.


marceloduplicati:
I ran a test with a linux with ipv6 only stack and you can't bind a socket listen to 127.0.0.1 

This could be a reliable way to get the loopback IP without DNS.

```
using System.Net;
using System.Net.NetworkInformation;
using System.Net.Sockets;

bool hasIPv4Loopback = NetworkInterface.GetAllNetworkInterfaces()
    .Where(ni => ni.OperationalStatus == OperationalStatus.Up)
    .SelectMany(ni => ni.GetIPProperties().UnicastAddresses)
    .Any(addr => addr.Address.AddressFamily == AddressFamily.InterNetwork 
                 && addr.Address.Equals(IPAddress.Loopback));

Console.WriteLine($"Has Loopback: {hasIPv4Loopback}");
Console.WriteLine($"Loopback IP: {(hasIPv4Loopback ? IPAddress.Loopback : IPAddress.IPv6Loopback)}");
```

It default's to ipv4 loopback if both stacks are available or **::1** on ipv6 only machines.

On the ipv6 only linux it behaved as expected, and on ipv4/v6 macos as well.

PR #5780, opened 2024-12-17
Primary functionality implemented for CIFS

marceloduplicati:
This will add CIFS support to Duplicati.

One of the advantages is that the resource does not need to be mounted as the backend connects and authenticates directly to the resource via TCP/or Netbios over TCP.

The authentication supported is domain/username/password. At present integrated Windows authentication is not supported.

This implementation relies on SMBLibrary library (https://github.com/TalAloni/SMBLibrary).

Summited as draft for review & Testing


kenkendk (approved):
Looks good!

Should we consider keeping a single `SMBShareConnection` for a single `CIFS` instance?

Some backends, like FTP+SSH, keeps the connection and handles all requests over a single connection. Others are semi-stateless, like S3, and create an new connection for each request.

For CIFS, could there be some authentication overhead that we can save by keeping a single instance alive?

marceloduplicati:
> Should we consider keeping a single SMBShareConnection for a single CIFS instance?

Indeed we can. SMBShareConnection already has semaphores on every method.

The savings in connection/authentication are likely to be minimal, but nevertheless a resource optimization (perhaps also avoiding many TIME_WAIT sockets) so I believe its worthwhile. 

I will implement and run the tests and on the next commit I will also push the the CI Tests.

PR #5781, opened 2024-12-17
Feature/improve ci testing

kenkendk:
This PR adds a bit of delay between some operations to test if the operations are less error prone when the storage has a chance to process requests.

Also added a bit of debugging output to further pinpoint the issues.

Fixed a problem with AzureBlob not reporting the correct exceptions.


marceloduplicati (commented):
Complementing on CI changes, the CI tests were designed with the max-parallel set to 1 to avoid using the same provider with the same credentials at the same time therefore avoiding conflicts during the test execution, however some backends still give an error that is symptomatic of concurrent execution. (You can pick a case where its failed on a previous workflow run, if you run again, it will succeed).

Example case (There are many others on the workflow history):

` Verifying file list ...
 *** File with name tDDVz_M9WtRl2Hrc9gZ4T76uZurHW1ow97vVPPKjk-tNdvRKE41k1SdC was found on server but not uploaded!
 *** File with name gz2nCLsmu7qi4VJuePe8OfTQIkOeoZvxsA7dBCBGItxeByfLBw9vDIao was found on server but not uploaded!
 *** File with name yM88lES4qDudu3PK4fvmD583J5eoLV1O4grpVvDWItxtZsqIc1oSY was found on server but not uploaded!
 *** File with name AE_8YJ8yY_M5w1JkWSQ4wlMag8DCR2zoMmseBGToRRR7f9HmQmj was found on server but not uploaded!
 *** File with name 5xJhW7FoTqGCZUstkjPReleEJXobHWXVEzcmhXvJ7jrSXHokDoyuj_XSB4IrR7GXYjZASWb was found on server but not uploaded!
 *** File with name Gk5mldlH1ZnMhueajR1PSULDhBM6qpIJA1jGIl5kn2TEyvrPKSfeGuQw9Vm3egh1uf-XlT9 was found on server but not uploaded!
 *** File with name BbgJekkwMQdCwRZ6bh7wSLQiovz7qjXqRrVkC was found on server but not uploaded!
 *** File with name iULScjQc7t6u4 was found on server but not uploaded!`
 
 [On this execution](https://github.com/duplicati/duplicati/actions/runs/12350908617/job/34464944904)
 
The tests have max-parallel: 1 on [action yml](https://github.com/duplicati/duplicati/blob/master/.github/workflows/backendtests.yml)

I will raise an issue on this to investigate the concurrency on the actions and experiment.

kenkendk:
I think there can still be an issue if we have 2 unrelated PRs running at the same time.
Not sure what the right fix is for this, as we probably do not want a folder for each run, as the destinations would be bloated quite quickly.

marceloduplicati:
Yes it could be, I want to test if we are able to solve that with [concurrency control](https://docs.github.com/en/actions/writing-workflows/choosing-what-your-workflow-does/control-the-concurrency-of-workflows-and-jobs).

kenkendk:
Great find! I think that would fix cross-talk between the tests

PR #5860, opened 2025-01-08
Improve ServerUtil

kenkendk:
Added `status` command.
Added `--wait` to the `run` command.

### DIFF
--- a/Duplicati/CommandLine/ServerUtil/Commands/RunBackup.cs
+++ b/Duplicati/CommandLine/ServerUtil/Commands/RunBackup.cs
@@ -31,9 +31,21 @@ public static class RunBackup
             new Argument<string>("backup", "The backup to run, either ID or exact name (case-insensitive)") {
                 Arity = ArgumentArity.ExactlyOne
             },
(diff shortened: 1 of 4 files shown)


SokoFromNZ:
Hi.
I'm anxiously waiting for this as duplicatiy_client.exe (https://github.com/Pectojin/duplicati-client/issues/38) does not work anymore which I've used since years now.
Any plans when this will be released officially?
thanks
Soko

kenkendk:
Hi @SokoFromNZ, in principle, the `ServerUtil.exe` is "stand alone", but it is built with references to the server, so you may need a few dll's for it to start. If you just download the zip package and run from there, it should work.

There is no change to the API so the `ServerUtil.exe` will work with Duplicati `2.0.9.106` and newer.

SokoFromNZ:
Hi @kenkendk .
I've downloaded the gui-zip-file from then canary build, unzipped it into a folder and run `Duplicati.CommandLine.ServerUtil.exe run 1 --wait` out of there.
Works like a charm as the backup itself uses the installed "stable" version.
thanks

PR #5863, opened 2025-01-10
Simplified logging

kenkendk:
This simplifies the logging by only keeping the object instance, not the key-dictionary that was needed in .NET4.

It also allows writing logs directly to the desired scope.

There appears to be some quirk with the logging and call context, when invoked by multiple threads in the same area.

This commit fixes it, but most likely just masks the underlying issue.


kenkendk:
Fixed using AsyncLocal

PR #5869, opened 2025-01-13
Better `wix.exe` support

kenkendk:


sanoj-v:
@kenkendk 
FYI
Some of the elements are deprecated in in wix v5 
image

`fileIdGenerator ??= (x) => ConvertToIdentifier(Path.GetRelativePath(sourceFolder, x));`

if fileIdGenerator function returns more than 72 character that it also gives an build failed for MSI.
Error: `warning WIX1026: Standard identifiers are 72 characters long or less. `

Do you think can we create separate file for wix v5 ??

Any suggestion?

kenkendk:
@sanoj-v I have updated the id generator to cap the fileIds at 72 bytes.

For Wix v5, we can create additional files, but they will need to be maintained, which I think will be problematic.

Do you need v5?

It looks like it works fine for now, the warnings should not prevent the MSI from being built.
Can it run with just generating a few warnings?

sanoj-v:
For me its creating a build failed @kenkendk 
image

Wix v5 doestn't have Product element anymore. 
image

Could you please provide updated version of wix v5 file if you have one to resolve error ?

kenkendk:
Hi @sanoj-v I have registered WiX v5 support as issue #5885

It is not a priority for me at this time to work on it, but feel free to contribute a v5 version.

PR #5913, opened 2025-01-27
Added UI warning for unencrypted database

kenkendk:
This adds a warning to the UI that would otherwise only be shown in various logs.
To avoid excessive warning, it shows up once per start per version (i.e., shows when an update has been installed).


ts678:
> "The database is not encrypted. This is a security risk and should be fixed as soon as possible."

Please make sure that the user knows which database this means, and can easily find directions.
Preferably the directions provide actual protection (including protection for the encryption key).

kenkendk:
@ts678 Added a "Show help" button that links to:
https://docs.duplicati.com/detailed-descriptions/the-server-database
(Added in #5914 )

PR #5914, opened 2025-01-27
Improve data folder assignment

### LINKED ISSUE #5847: Prevent storing files in Windows subfolder
When running Duplicati as a service under the local system account, it maps `%LOCALAPPDATA%` to `C:\Windows\System32\config\systemprofile\AppData\Local\Duplicati`.

This folder is problematic as it is under `C:\Windows` which will be wiped when operating system updates are installed.

Duplicati should check for this and give a warning message in the logs. If possible another location should be used for this setup.

### LINKED ISSUE #5902: AutoUpdater broke in 2.0.9.108
```
C:\Duplicati\duplicati-2.1.0.107_canary_2025-01-17-win-x64-gui>Duplicati.CommandLine.AutoUpdater check
Error detected: System.NullReferenceException: Object reference not set to an instance of an object.
   at Duplicati.Library.Utility.HttpClientExtensions.DownloadFile(HttpClient client, HttpRequestMessage request, String filename, Action`1 progressReportingAction, CancellationToken cancellationToken)
   at Duplicati.Library.AutoUpdater.UpdaterManager.CheckForUpdate(ReleaseType channel)
Error detected: System.NullReferenceException: Object reference not set to an instance of an object.
   at Duplicati.Library.Utility.HttpClientExtensions.DownloadFile(HttpClient client, HttpRequestMessage request, String filename, Action`1 progressReportingAction, CancellationToken cancellationToken)
   at Duplicati.Library.AutoUpdater.UpdaterManager.CheckForUpdate(ReleaseType channel)
No updates found
```

See forum thread for details:
https://forum.duplicati.com/t/release-2-1-0-107-canary-2025-01-17/19958/8

kenkendk:
This PR adds a check to warn the user if the database is stored inside the Windows folder.

If the database is found to be within the Windows folder, an attempt is made to use `{SystemRoot}\Users\LocalService\Duplicati` instead.

If the path needs to be within Windows (because it is already contains data), a warning is emitted to the logs and a warning is shown in the UI.

This fixes #5847 

This PR also updates logic for placement of the storage folder on Linux, in cases where `LOCALAPPDATA` returns `/`.
Before this commit, the default storage folder would be `/Duplicati` but with this change it will be `/var/lib/Duplicati`.

For most users, the behavior is unchanged, resolving the default path to `~/.config/Duplicati`.

For MacOS there are no changes.

# Additional update to files other than `Duplicati-server.sqlite`
Because there are additional files, besides the database, this PR also updates the storage of other files.
Applying either `--server-datafolder` or `--portable-mode` will ensure that configuration files an statistics files are stored in the designated storage folder

To get feature parity, the options `--server-datafolder` and `--portable-mode` have now been added to CLI, Agent, and ServerUtil.

This impacts some existing installations because the use of `--server-datafolder` or `--portable-mode` would previously **only** affect the `Duplicati-server.sqlite` file. In cases where the user had chosen a different folder in a previous version, the usual storage folder would still contain: `machineid.txt`, `installation.txt`, `dbconfig.json`, and `preload.json`.

With the update, these files will now be read from the same folder as `Duplicati-server.sqlite`.
- If the `installation.txt` is missing, a new will be generated, causing inflated statistics on the usagereporter.
- If the `machineid.txt` is missing, a new will be generated, causing the machine to be registered as a new machine in the Duplicati console.
- If `dbconfig.json` is missing (CLI only) it will fail to locate local databases (see below why this won't happen)
- If `preload.json` is missing, some parameters will not be applied.

If an overriden folder is not supplied, everything works as before. This should make it have no impact on the Agent and CLI as they did not support the custom folder before and would complain if called with an override.

ServerUtil could be impacted because it stored a refresh token in the default path, but since `--server-datafolder` points to the `Duplicati-server.sqlite` folder it will now use that folder. The result is that a re-login is required in this case.

Also fixed a crash in AutoUpdater. This fixes #5902 

# For developers
This PR changes the logic for debug builds, so having the DEBUG flag defined will make the builds default to use "portable mode". In portable mode the files are stored in a `data` subfolder from the executable.

Before this PR the DEBUG builds would use the folder itself, mixing executables and data. Due to this change, developers will need to move `Duplicati-server.sqlite` from the build output folder and into the `data` subfolder. If you do not mind starting with a clean database, you can ignore this detail.


ts678:
> "The configuration is stored in the Windows folder, which will be deleted on Windows updates."

Please make sure that there is easily available information for the user to know what to do for this.

It's also probably overstating the worry. Most Windows updates are security updates, not versions.

kenkendk:
@ts678 I have added a link to this page on the warnings:
https://docs.duplicati.com/detailed-descriptions/the-server-database

Also, I updated the wording a bit on the Windows folder warning, based on your input.

Duplicati/CommandLine/CLI/Strings.cs
```diff
@@ -54,6 +54,12 @@ internal static class Program
         public static string QuietConsoleOptionLong { get { return LC.L(@"If this option is set, progress reports and other messages that would normally go to the console will be redirected to the log."); } }
         public static string QuietConsoleOptionShort { get { return LC.L(@"Disable console output"); } }
         public static string SkippingSourceArgumentsOnNonBackupOperation { get { return @"The --source argument was specified in the parameter file, but the current operation is not a backup operation, so the argument is ignored."; } }
+        public static string AutoUpdateOptionShort { get { return LC.L(@"Toggle automatic updates"); } }
+        public static string AutoUpdateOptionLong { get { return LC.L(@"Set this option if you prefer to have the commandline version automatically update"); } }
```
carljohnsen, line 58: The text of the PR didn't mention this feature; that the CLI also supports auto updating Duplicati.

Duplicati/Library/AutoUpdater/DataFolderLocator.cs
```diff
@@ -0,0 +1,133 @@
...
+            // Use the most recent location found with content
+            // If none are found, use the most recent location
+            folder = matches.FirstOrDefault() ?? folderOrder.Last();
+        }
+
+        if (OperatingSystem.IsMacOS())
+        {
+            // Special handling for MacOS:
+            //   - Older versions use ~/.config/
+            //   - but new versions use ~/Library/Application\ Support/
```
carljohnsen, line 101: The PR text states that "For MacOS there are no changes.", but this comment makes it sound like it has changed? Or is the "old/new" relating to a previous folder change?

Duplicati/CommandLine/CLI/Strings.cs
kenkendk: These strings are actually just moved from `Program.cs`. For some reason those two strings were placed in the code file instead of the string file.

Duplicati/Library/AutoUpdater/DataFolderLocator.cs
kenkendk: The MacOS placement changed with the switch from Mono to .NET8. Due to refactoring the `DatabaseLocator.cs` into two files this shows up as "new" code, but it has not been changed.

PR #5928, opened 2025-01-29
add docker tag $version-$channel

### LINKED ISSUE #5927: Add channel suffix/prefix in versioned docker tags.
**Is your feature request related to a problem? Please describe.**

Currently versioned docker tags cannot be distinguished between beta/canary/etc.
So automation tools like https://github.com/renovatebot/renovate will think that 2.1.0.107 is greater than 2.1.0.3
Which technically is, but its not possible currently to stay on the beta channel.

**Describe the solution you'd like**
When releasing a tag in dockerhub, for a version provide an additional tag with channel pre/suffix. eg
`2.1.0.3`, `2.1.0.3-beta`
`2.1.0.107`, `2.1.0.107-canary` 

This way someone can configure renovate or other tools to regex max versions like
`^(?<major>\\d+)\\.(?<minor>\\d+)\\.(?<patch>\\d+)-beta$`

**Describe alternatives you've considered**

Using the `beta` tag, but if there is any changes when the tag is mutated it will not be catch-ed in time and users might experience issues until we released a new compose template to address such issue. 
Also docker does not always re-pull the image if the tag already exists on the host, which causes sometimes confusion when users end up with different versions.

**Additional context**

I'm the maintainer of TrueNAS App Catalog and I'm looking into adding Duplicati (https://github.com/truenas/apps/issues/147).
We use renovate for automatically creating PRs / notifying us of updated container images so we can push the updates down to the users.

stavros-k:
Closes #5927

### DIFF
--- a/ReleaseBuilder/Build/Command.CreatePackage.cs
+++ b/ReleaseBuilder/Build/Command.CreatePackage.cs
@@ -1,22 +1,22 @@
 // Copyright (C) 2025, The Duplicati Team
 // https://duplicati.com, hello@duplicati.com
-// 
-// Permission is hereby granted, free of charge, to any person obtaining a 
-// copy of this software and associated documentation files (the "Software"), 
-// to deal in the Software without restriction, including without limitation 
-// the rights to use, copy, modify, merge, publish, distribute, sublicense, 
-// and/or sell copies of the Software, and to permit persons to whom the 
+//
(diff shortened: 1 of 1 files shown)


stavros-k:
Saving on my editor removed some white space on the end of the lines, if you want I can revert those.

kenkendk:
Looks good, not sure why the newlines are weird though.

PR #5936, opened 2025-02-01
Fixed not creating releases on master when building from a branch

kenkendk:


ts678:
Does this explain why `https://github.com/duplicati/duplicati/archive/refs/tags/v2.1.0.3_beta_2025-01-22.zip` source release doesn't seem to match the built release? For example, source shows the B2 backend after the move to `HttpClient`.

Somehow 2.1.0.4 Stable source escaped this issue, and (for example) B2AuthHelper.cs is identical to the 2.1.0.2 Beta version.

kenkendk:
> Does this explain why https://github.com/duplicati/duplicati/archive/refs/tags/v2.1.0.3_beta_2025-01-22.zip source release doesn't seem to match the built release? For example, source shows the B2 backend after the move to HttpClient.

Yes. The short tag in the form `version_type_date` would point to the current master commit, instead of the build commit. The branch and the long-form tag points to the right source.

> Somehow 2.1.0.4 Stable source escaped this issue, and (for example) B2AuthHelper.cs is identical to the 2.1.0.2 Beta version.

Yes, I noticed the problem with the stable release and hand-fixed it, which required re-uploading all binaries.

Issue #3875, opened 2019-08-28
“Google Drive (full access) login” restricted in early 2020


ts678:
- [x] I have searched open and closed issues for duplicates.

----------------------------------------

## Environment info

- **Duplicati version**: N/A
- **Operating system**: N/A
- **Backend**: Google Drive

## Description

This is a placeholder issue for an issue that needs preparation, maybe issue tracking by [milestone](https://github.com/duplicati/duplicati/milestones).
For initial text, I'm just going to post the original content I once did to try to get planning started:

This will mean some people will lose access to backups, without a workaround or slow and costly work.

[Enhancing security controls for Google Drive third-party apps ](https://cloud.google.com/blog/products/identity-security/enhancing-security-controls-for-google-drive-third-party-apps)

[OAuth API Verification FAQ](https://support.google.com/cloud/answer/9110914?hl=en)

[Duplicati OAuth Handler](https://duplicati-oauth-handler.appspot.com/) looks like its [code](https://github.com/duplicati/oauth-handler/blob/582bb95ddf9e3413da64e4e59a7212cf26e1e469/settings.py#L229) for “limited” uses [scope](https://developers.google.com/drive/api/v2/about-auth) `drive.file` which is less impacted, but “full access” uses scope `drive` which would require tens of thousands of dollars for a security audit.

I’m not sure if usage reporting can tell us how much it’s used, and polling suffers from limited readership. Price tag may require us to just give this up, and seek workarounds, but below posts aren’t encouraging:

[The remote server returned an error: (403) Forbidden with google drive](https://forum.duplicati.com/t/the-remote-server-returned-an-error-403-forbidden-with-google-drive/723/4)
[Google Drive 403 errors during file verification process](https://forum.duplicati.com/t/google-drive-403-errors-during-file-verification-process/3528/2)

There might be time to add some sort of notifier in the next beta for affected users, but what could it say?

[OneDrive failing : “The remote server returned an error: (410) Gone.”](https://forum.duplicati.com/t/onedrive-failing-the-remote-server-returned-an-error-410-gone/5936) is how some people learned of *that* , however at least [Old OneDrive API being deprecated - needs to be rewritten to use the Graph API #3024](https://github.com/duplicati/duplicati/issues/3024) was prepared in advance, forced 2.0.4.5 beta out, and so people had a reasonable fix for the shutdown…

ts678:
> It looks like open access using Drive.file will be costly.

[...]

PR #5937, opened 2025-02-03
Feature/Remote synchronization tool

carljohnsen:
This PR adds a tool for performing remote synchronization between two (potentially different) backends. The tool assumes that the intent is to have the destination match the source.

If the destination has files that are not in the source, they will be deleted (or renamed if the `--retention` option is set).

If the destination has files that are in the source but are different in size or has a newer timestamp, they will be overwritten by the source files.

If the `--force` option is set, the destination will be overwritten by the source, regardless of the state of the files. It will also skip the initial comparison, and delete (or rename) all files in the destination.

If the verify option is set, the files will be downloaded and compared after uploading to ensure that the files are correct. Files that already exist in the destination will be verified before being overwritten (if they seemingly match).

It has the following signature:
```
  RemoteSynchronization <backend_src> <backend_dst> [options]
```
Each backend string follows the same format as the regular `Duplicati.CommandLine`, where options can be passed as follows:
```
file://somefile?optionkey=optionvalue&anotherkey=anothervalue
```
It has the following options:
```
  -d, --dry-run                      Do not actually write or delete files. If not set here, the global options will be checked [default: False]
  --src-options <src-options>        Options for the source backend. Each option is a key-value pair separated by an equals sign, e.g. --src-options key1=value1 key2=value2
  --dst-options <dst-options>        Options for the destination backend. Each option is a key-value pair separated by an equals sign, e.g. --dst-options key1=value1 key2=value2
  --verify-contents                  Verify the contents of the files to decide whether the pre-existing destination files should be overwritten [default: False]
  --verify-get-after-put             Verify the files after uploading them to ensure that they were uploaded correctly [default: False]
  --retry <retry>                    Number of times to retry on errors [default: 3]
  --force                            Force the synchronization [default: False]
  --retention                        Toggles whether to keep old files. Any deletes will be renames instead [default: False]
  --confirm                          Automatically confirm the operation [default: False]
  --global-options <global-options>  Global options all backends. May be overridden by backend specific options (src-options, dst-options). Each option is a key-value pair separated by an   
                                     equals sign, e.g. --global-options key1=value1 key2=value2
  --log-level <log-level>            The log level to use. If not set here, global options will be checked [default: Information]
  --log-file <log-file>              The log file to write to. If not set here, global options will be checked []
  --progress                         Print progress to STDOUT [default: False]
  --parse-arguments-only             Only parse the arguments and then exit [default: False]
  --version                          Show version information
  -?, -h, --help                     Show help and usage information
```


ts678:
[“Google Drive (full access) login” restricted in early 2020 #3875](https://github.com/duplicati/duplicati/issues/3875)
could probably make use of this if Google someday disables Duplicati "full access" login, and leaves it with "limited" only.

Tools/RemoteSynchronization/Program.cs
```diff
@@ -0,0 +1,815 @@
...
+
+        /// <summary>
+        /// Global configuration for the tool. Should be set after parsing the commandline arguments.
+        /// </summary>
+        private class GlobalConfig
+        {
+            /// <summary>
+            /// Whether the tool should not actually write or delete files. Defaults to false.
+            /// </summary>
+            internal static bool DryRun = DEFAULT_DRY_RUN;
```
kenkendk, line 62: Instead of these, consider binding to a single record. If you add the `getDefault: () => false` to the options, you can have a single record with all the options, and only define the default value inline with the options.

Example:
```
private sealed record Config(
  bool DryRun,
  bool Force
);

 var dry_run_opt = new Option<bool>(aliases: ["--dry-run", "-d"], description: "Do not actually write or delete files. If not set here, the global options will be checked", getDefaultValue: () => false);
var force_opt = new Option<bool>(aliases: ["--force"], description: "Force the synchronization", getDefaultValue: () => false);

root_cmd.SetHandler = CommandHandler.Create<Config>(Run);

public static void Run(Config config) {
  Console.WriteLine($"DryRun: {config.DryRun}");
}

```

The binding is a bit hidden in the explanation, but you can find it here:
https://github.com/dotnet/command-line-api/blob/main/docs/model-binding.md#more-complex-types

Tools/RemoteSynchronization/Program.cs
```diff
@@ -0,0 +1,815 @@
...
+        }
+
+        /// <summary>
+        /// The main logic of the tool.
+        /// </summary>
+        /// <param name="src">The connection string for the source backend.</param>
+        /// <param name="dst">The connection string for the destination backend.</param>
+        /// <param name="options">Various options for the tool</param>
+        /// <returns>The return code for the main entry; 0 on success.</returns>
+        private static async Task<int> Run(string src, string dst, Dictionary<string, object?> options)
```
kenkendk, line 177: Is it for compatibility that you want the options dictionary?

Tools/RemoteSynchronization/Program.cs
```diff
@@ -0,0 +1,815 @@
...
+            if (parse_only)
+            {
+                Duplicati.Library.Logging.Log.WriteInformationMessage(LOGTAG, "rsync", "Arguments parsed successfully; exiting.");
+                return 0;
+            }
+
+            // Load the backends
+            using var b1 = Duplicati.Library.DynamicLoader.BackendLoader.GetBackend(src, src_opts);
+            var b1s = b1 as IStreamingBackend;
+            System.Diagnostics.Debug.Assert(b1s != null);
```
kenkendk, line 244: Consider throwing an exception or writing to console, so the user gets a nicer message.
In the Agent I am catching `UserInformationException` and printing the message, otherwise printing a stack trace as well.

Tools/RemoteSynchronization/Program.cs
```diff
@@ -0,0 +1,815 @@
...
+            // If there are still errors, retry a few times
+            if (copy_errors.Any())
+            {
+                Duplicati.Library.Logging.Log.WriteWarningMessage(LOGTAG, "rsync", null, "Could not copy {0} files.", copy_errors.Count());
+                if (retries > 0)
+                {
+                    Duplicati.Library.Logging.Log.WriteInformationMessage(LOGTAG, "rsync", "Retrying {0} more times to copy the {1} files that failed", retries, copy_errors.Count());
+                    for (int i = 0; i < retries; i++)
+                    {
+                        Thread.Sleep(5000); // Wait 5 seconds before retrying
```
kenkendk, line 327: Nitpicky, but since this is async code, it should be `await Task.Delay(5000)`

Tools/RemoteSynchronization/Program.cs
```diff
@@ -0,0 +1,815 @@
...
+                Duplicati.Library.Logging.Log.WriteVerboseMessage(LOGTAG, "rsync", "Copying {0} from {1} to {2}", f.Name, b_src.DisplayName, b_dst.DisplayName);
+
+                try
+                {
+                    sw_get_src.Start();
+                    await b_src.GetAsync(f.Name, s_src, CancellationToken.None);
+                    sw_get_src.Stop();
+                    if (GlobalConfig.DryRun)
+                    {
+                        Duplicati.Library.Logging.Log.WriteDryrunMessage(LOGTAG, "rsync", "Would write {0} bytes of {1} to {2}", s_src.Length, f.Name, b_dst.DisplayName);
```
kenkendk, line 401: There is a human-size formatter in `Duplicati.Library.Utility.Utility.FormatSizeString`

Tools/RemoteSynchronization/Program.cs
```diff
@@ -0,0 +1,815 @@
...
+                    sw_put_dst.Stop();
+                    sw_get_dst.Stop();
+                    sw_get_cmp.Stop();
+                }
+            }
+
+            if (GlobalConfig.Progress)
+                Console.WriteLine($"\rCopying: {n}/{n}");
+
+            Duplicati.Library.Logging.Log.WriteProfilingMessage(LOGTAG, "rsync", "Copy | Get source: {0} ms, Put destination: {1} ms, Get destination: {2} ms, Get compare: {3} ms", sw_get_src.ElapsedMilliseconds, sw_put_dst.ElapsedMilliseconds, sw_get_dst.ElapsedMilliseconds, sw_get_cmp.ElapsedMilliseconds);
```
kenkendk, line 458: If you use `TimeSpan.FromMilliseconds(...)` the logged values will be easier to read than the raw millisecond values.

Tools/RemoteSynchronization/Program.cs
```diff
@@ -0,0 +1,815 @@
...
+        /// <returns>A tuple of Lists each holding the files to copy, delete and verify.</returns>
+        private static (IEnumerable<IFileEntry>, IEnumerable<IFileEntry>, IEnumerable<IFileEntry>) PrepareFileLists(IStreamingBackend b_src, IStreamingBackend b_dst)
+        {
+            var sw = new System.Diagnostics.Stopwatch();
+
+            sw.Restart();
+            var files_src = b_src.List();
+            sw.Stop();
+            Duplicati.Library.Logging.Log.WriteProfilingMessage(LOGTAG, "rsync", "Prepare | List source: {0} ms", sw.ElapsedMilliseconds);
+            sw.Restart();
```
kenkendk, line 547: There is also a `Duplicati.Library.Logging.Timer` class you can use as:
```csharp
using (new Logging.Timer(LOGTAG, "rsync", "Prepare | List source"))
  var files_dst = b_dst.List();
```

It is not as precise as the stopwatch, but it has nanosecond resolution.

Tools/RemoteSynchronization/Program.cs
```diff
@@ -0,0 +1,815 @@
...
+        /// The renaming is done by deleting the file and re-uploading it with a new name.
+        /// </summary>
+        /// <param name="b">The backend to rename the files in.</param>
+        /// <param name="files">The files to rename.</param>
+        /// <param name="dry_run">Flag for whether the renaming should be printed rather than performed.</param>
+        /// <returns>The number of successful renames.</returns>
+        private static async Task<long> RenameAsync(IBackend b, IEnumerable<IFileEntry> files)
+        {
+            long successful_renames = 0;
+            string prefix = $"{System.DateTime.Now:yyyyMMddHHmmss}.old";
```
kenkendk, line 635: Consider using `UtcNow` ?

kenkendk (approved):

Tools/RemoteSynchronization/Program.cs
carljohnsen: Nice, makes for a cleaner definition. It's been implemented!

Tools/RemoteSynchronization/Program.cs
carljohnsen: Yes, that was to pass the parsed options in easier. However, with the new way, this has become the config record.

Tools/RemoteSynchronization/Program.cs
carljohnsen: Right, this has been changed to log an error message with an exception, which is then thrown.

Tools/RemoteSynchronization/Program.cs
carljohnsen: Cool! I've implemented it where applicable. For some of the places, the timings are internal and accumulative, so there the Logging can't be used. But for all of the places where it was "start, run, stop, log" it makes a lot of sense! Only annoying thing is that the scope cannot be used to declare variables. Would be nice to have something like the `out` syntax, but that's for some other time!

PR #5947, opened 2025-02-06
Removing TimeoutObservingStream from backend manager

marceloduplicati:
@kenkendk Can you confirm, this is where we would remove the TimeoutObservingStream.

I have specifically left it in the BackendTester as I believe its useful to control the timeout for CI tests.


kenkendk:
@marceloduplicati yes, that looks correct. Then we need to push it into the backends for the actual copy

marceloduplicati:
Ok, I will implement it first on the B2 backend where I will also add the 429 resiliency, then we can cascade it in individual PR for each backends migrated.

PR #5975, opened 2025-02-14
This commit adds relaxed name parsing.

kenkendk:
In the case either the prefix or the actual content names start with forward slashes, this is considered an error in the input and the slashes are removed.

The option `--s3-strict-name-parsing` can be used to disable the lenient parsing.


kenkendk:
This is fixed with the alternate source PR #5955

PR #5978, opened 2025-02-17
Final conversion to async backends

kenkendk:
This PR is the final change, where all backends are now purely implemented with async interfaces.

The code changes here replace the LIST call with an async version. This affect all backends (and then some) but the changes are mostly mechanical rewrites of the code.

In many places, the backeds were already prepared for async output, in other some glue was needed.

To reduce the scope of this change, some backends simply report a synchronous result as an async list.


Duplicati/Library/Backend/AzureBlob/AzureBlobWrapper.cs
```diff
@@ -118,61 +118,61 @@ public Task DeleteObjectAsync(string keyName, CancellationToken cancelToken)
...
+                throw;
+            }
 
-            list.AddRange(segment.Results);
+            foreach (var item in segment.Results)
+                yield return item;
 
             while (segment.ContinuationToken != null)
             {
                 segment = await _container.ListBlobsSegmentedAsync(null, false, ListDetails, null, segment.ContinuationToken, null, _operationContext);
```
marceloduplicati, line 143: ListBlobsSegmentedAsync has a cancellationToken overload as well.

marceloduplicati (commented):
Just a second set of eyes through everything since there all backends were touched.

PR #5979, opened 2025-02-18
Prevent Double Click opening multiple windows

Alex-Heard:
Prevents double clicks from opening two windows by ignoring click events that happen within 1 second of last click


kenkendk:
Looks good! I would like to remove the extra `.sln` file, do you need it for something?

kenkendk:
No need to re-submit, I can fix it in this PR.

PR #5983, opened 2025-02-19
Performance/restore parallel download

carljohnsen:
This PR enables parallel downloads, which greatly benefits the performance of the reworked restore flow (preliminary tests show that it's several times faster than before). Before, even though multiple processes handling the downloads were spawned, the `BackendManager` would block, resulting in serialized downloads.

- The get operation of the `BackendManager` is now non-blocking in the same fashion as the put operation.
- A get operation can now be triggered without decryption by using `BackendManager.GetDirectAsync()`.

Additionally, this PR makes the following changes:

- Fixed a bug in `BlockManager` where the formatting of the logging strings would throw an exception.
- Added additional profiling logging to the legacy restore flow.
- Added additional `.ConfigureAwait(false)` to the async calls in the reworked restore flow.
- Renamed "pending" to active when referring to the parallel uploads in the BackendManager as this better reflects that the uploads are in progress.
- Exposed the decryption methods of `BackendManager.GetOperation` to allow the restore processes to access the same decryption methods.

### DIFF
--- a/Duplicati/Library/Main/Backend/BackendManager.Handler.cs
+++ b/Duplicati/Library/Main/Backend/BackendManager.Handler.cs
@@ -89,6 +89,10 @@ partial class BackendManager
         private static readonly string LOGTAG = Logging.Log.LogTagFromType<Handler>();
 
         /// <summary>
+        /// The list of active downloads
+        /// </summary>
+        private readonly List<Task> activeDownloads = [];
+        /// <summary>
         /// The list of active uploads
         /// </summary>
         private readonly List<Task> activeUploads = [];
@@ -104,6 +108,10 @@ partial class BackendManager
         /// The context for the handler
         /// </summary>
         private readonly ExecuteContext context;
+        /// <summary>
+        /// The maximum number of parallel downloads
+        /// </summary>
+        private readonly int maxParallelDownloads;
         /// <summary>
         /// The maximum number of parallel uploads
         /// </summary>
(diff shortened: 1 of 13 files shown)


Duplicati/Library/Main/Backend/BackendManager.Handler.cs
```diff
@@ -234,16 +275,23 @@ private async Task Run(IReadChannel<PendingOperationBase> requestChannel)
...
+
+                            // Operation is accepted into queue, so we can signal completion
+                            activeDownloads.Add(ExecuteWithRetry(getOp, tcs.Token));
+                        }
                         else
                         {
-                            // Wait for all pending uploads to complete
-                            await EnsureAtMostNPendingUploads(1).ConfigureAwait(false);
+                            // Wait for all of the active uploads to complete
+                            await EnsureAtMostNActiveUploads(1).ConfigureAwait(false);
```
kenkendk, line 294: Do we also need to ensure that there are no pending downloads at this stage? 
The way the code is constructed this will not currently happen, but maybe it is logical that we have flushed active operations when we do something "synchronous" ?

Duplicati/Library/Main/Backend/BackendManager.Handler.cs
```diff
@@ -259,25 +307,25 @@ private async Task Run(IReadChannel<PendingOperationBase> requestChannel)
             }
             finally
             {
-                // Terminate any pending uploads
+                // Terminate any active uploads
                 tcs.Cancel();
 
                 if (activeUploads.Count > 0)
```
kenkendk, line 313: The teardown should also cancel any active downloads

kenkendk (dismissed):

PR #6008, opened 2025-03-05
Azure Library Upgrade and Fixes

marceloduplicati:
Removed obsolete Microsoft.Azure.* references which were deprecated and replaced with Azure.Storage.*
Cleaned up unused references
Changes to code due to obsolete classes
Patched some missing ConfigureAwait(false) on calls that could freeze the UI

### DIFF
--- a/Duplicati/Library/Backend/AzureBlob/AzureBlobWrapper.cs
+++ b/Duplicati/Library/Backend/AzureBlob/AzureBlobWrapper.cs
@@ -19,16 +19,12 @@
 // FROM, OUT OF OR IN CONNECTION WITH THE SOFTWARE OR THE USE OR OTHER 
 // DEALINGS IN THE SOFTWARE.
 
-using System.Collections.Generic;
-using System.IO;
+using Azure.Storage.Blobs;
+using Azure.Storage.Blobs.Models;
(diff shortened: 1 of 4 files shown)


Duplicati/Library/Backend/AzureBlob/AzureBlobWrapper.cs
```diff
@@ -37,141 +33,126 @@ namespace Duplicati.Library.Backend.AzureBlob
...
                 {
-                    if (x is CloudBlockBlob cb)
-                    {
-                        var lastModified = new System.DateTime();
-                        if (cb.Properties.LastModified != null)
-                            lastModified = new System.DateTime(cb.Properties.LastModified.Value.Ticks, System.DateTimeKind.Utc);
-                        res = new FileEntry(res.Name, cb.Properties.Length, lastModified, lastModified);
-                    }
+                    // Note: AsyncPageable<BlobItem> return internally uses .ConfigureAwait(false).
+                    return _container.GetBlobsAsync(cancellationToken: cancelToken);
```
kenkendk, line 150: If this is not awaited here, the exception handler will not catch the error, because the continuation will not be invoked until the task is awaited.

kenkendk (approved):

PR #6011, opened 2025-03-06
Improved failure resilience

kenkendk:
This update adds a new flag to the database that keeps track of the potential for partial files to be on the remote storage.
If no files are supposed to be partial, the initial validation will be more strict and not automatically clean up dangling uploads.

Any operation that modifies the database will attempt to clean up any leftover files.
Any operation that does not modify the database will not attempt to clean up the remote destination.

Added more logging to alert if downloads or uploads are in-flight when terminating.

Changed logic so non-modifying operations do not clean up (list, restore, test), and others do not automatically clean up. The backup command will automatically clean up if the previous backup left partial files, and the repair command will clean up regardless. This change makes it harder for errors to cause silent or unexpected removal of files.

Added tests and fixed code to support cases where the initial upload fails and there are no dlists on the remote storage.
Added additional database consistency checks to avoid undetected problems with the database.
Added repair code that can handle cases where a filest or remote dlist entry is missing, and then recreates the contents.


Duplicati/Library/Main/Backend/BackendManager.Handler.cs
```diff
@@ -325,19 +316,30 @@ private async Task Run(IReadChannel<PendingOperationBase> requestChannel)
                     for (int i = activeUploads.Count - 1; i >= 0; i--)
                     {
                         var t = activeUploads[i];
-                        activeUploads.RemoveAt(i);
-                        if (t.IsCompleted && !t.IsCanceled && t.Exception != null)
-                            Logging.Log.WriteWarningMessage(LOGTAG, "BackendManagerDisposeError", t.Exception, "Error in active upload: {0}", t.Exception.Message);
+                        if (t.IsCompleted)
```
carljohnsen, line 319: I know that this wasn't present before, but shouldn't this logic also be applied to the downloads?

Duplicati/Library/Main/Database/LocalDatabase.cs
```diff
@@ -737,42 +737,34 @@ public IDictionary<string, string> GetDbOptions(System.Data.IDbTransaction trans
...
-                    opts.Remove("partially-recreated");
+        public bool PartiallyRecreated
+        {
+            get => GetDbOptions().ContainsKey("partially-recreated");
+            set => UpdateDbOption("partially-recreated", value);
+        }
 
-                SetDbOptions(opts);
-            }
+        public bool TerminatedWithActiveUploads
```
carljohnsen, line 764: Same as before, shouldn't there be a `TerminatedWithActiveDownloads`?

Duplicati/Library/Main/Operation/FilelistProcessor.cs
```diff
@@ -78,13 +102,13 @@ public static async Task VerifyLocalList(IBackendManager backendManager, LocalDa
             }
         }
 
-        public static async Task VerifyRemoteList(IBackendManager backend, Options options, LocalDatabase database, IBackendWriter backendWriter, bool latestVolumesOnly, IDbTransaction transaction)
+        public static async Task VerifyRemoteList(IBackendManager backend, Options options, LocalDatabase database, IBackendWriter backendWriter, bool latestVolumesOnly, VerifyMode verifyMode, IDbTransaction transaction)
```
carljohnsen, line 105: Missing docstring?

Duplicati/Library/Main/Database/LocalDatabase.cs
```diff
@@ -737,42 +737,34 @@ public IDictionary<string, string> GetDbOptions(System.Data.IDbTransaction trans
             return GetDbOptionList(transaction).ToDictionary(x => x.Key, x => x.Value);
         }
 
-        public bool RepairInProgress
+        private void UpdateDbOption(string key, bool value)
```
carljohnsen, line 740: Missing docstring

carljohnsen: Also true for the next 3 Properties

Duplicati/Library/Main/Operation/FilelistProcessor.cs
```diff
@@ -281,7 +319,10 @@ orderby n.Time descending
                         if (remoteFound)
                         {
                             Logging.Log.WriteInformationMessage(LOGTAG, "RemoveUnwantedRemoteFile", "removing remote file listed as {0}: {1}", i.State, i.Name);
-                            await backendManager.DeleteAsync(i.Name, i.Size, true, CancellationToken.None).ConfigureAwait(false);
+                            if (options.Dryrun)
+                                Logging.Log.WriteDryrunMessage(LOGTAG, "WouldDeleteRemoteFile", "Would delete remote file: {0}", i.Name);
+                            else if (verifyMode != VerifyMode.VerifyOnly)
+                                await backendManager.DeleteAsync(i.Name, i.Size, true, CancellationToken.None).ConfigureAwait(false);
```
carljohnsen, line 325: Shouldn't `VerifyOnly` warn that the volume is marked for deletion? Or is that handled by another operation at a later point? I'm thinking that the verify only should report if anything is not as it should be.

Duplicati/Library/Main/Operation/FilelistProcessor.cs
```diff
@@ -329,10 +375,18 @@ orderby n.Time descending
...
+                                throw new RemoteListVerificationException($"The remote volume {i.Name} is in uploading state, partially uploaded, and strict mode is on", "UnexpectedUploadingFile");
+                            }
                             else
                             {
                                 Logging.Log.WriteInformationMessage(LOGTAG, "Remove incomplete file", "removing incomplete remote file listed as {0}: {1}", i.State, i.Name);
-                                await backendManager.DeleteAsync(i.Name, i.Size, true, CancellationToken.None).ConfigureAwait(false);
+                                if (options.Dryrun)
+                                    Logging.Log.WriteDryrunMessage(LOGTAG, "WouldDeleteRemoteFile", "Would delete remote file: {0}", i.Name);
+                                else if (verifyMode != VerifyMode.VerifyOnly)
+                                    await backendManager.DeleteAsync(i.Name, i.Size, true, CancellationToken.None).ConfigureAwait(false);
```
carljohnsen, line 389: Same comment as before

Duplicati/Library/Main/Operation/FilelistProcessor.cs
```diff
@@ -364,7 +418,15 @@ orderby n.Time descending
             }
 
             // cleanup deleted volumes in DB en block
```
carljohnsen, line 420: What's "DB en block" ?

Duplicati/Library/Main/Operation/FilelistProcessor.cs
```diff
@@ -364,7 +418,15 @@ orderby n.Time descending
...
-            database.RemoveRemoteVolumes(cleanupRemovedRemoteVolumes, null);
+            if (verifyMode == VerifyMode.VerifyStrict && missingUploadingVolumes.Count > 0)
+                throw new RemoteListVerificationException($"The remote volumes {string.Join(", ", missingUploadingVolumes)} are supposed to be deleted, but strict mode is on. Try running the \"repair\" command", "DeleteDuringStrictMode");
+            else if (verifyMode == VerifyMode.VerifyAndClean)
+            {
+                database.RemoveRemoteVolumes(missingUploadingVolumes.Concat(temporaryAndDeletedVolumes), null);
+                // Clear the flag after we have cleaned up
+                if (!options.Dryrun)
+                    database.TerminatedWithActiveUploads = false;
+            }
```
carljohnsen, line 429: Should a warning / message be emitted here with `VerifyOnly`?

carljohnsen (approved):

Duplicati/Library/Main/Database/LocalDatabase.cs
kenkendk: The flag can be renamed, but the intention is to record if the remote storage is being altered and the process terminates, as we can then expect the remote storage to be partially updated. Since downloads do not alter the remote storage state, they are not considered for this flag.

Duplicati/Library/Main/Backend/BackendManager.Handler.cs
kenkendk: I have refactored this to a method so it is called on both uploads and downloads.
The reason why it was not really there before, is that it does not hurt anything if the downloads are not completed, but for consistency we now also log if these are hanging.

Duplicati/Library/Main/Operation/FilelistProcessor.cs
kenkendk: My logic for this is that the operations that send in `VerifyOnly` only needs to read the data. Even if there are partial uploads, it seems weird if a "list" or "restore" operation can delete remote files.

Since the files are "accepted partial files" they will be cleaned up on the next run that is not `VerifyOnly`.

Duplicati/Library/Main/Operation/FilelistProcessor.cs
kenkendk: Not my comment, but it should be "en bloc" meaning "all at the same time"

Duplicati/Library/Main/Operation/FilelistProcessor.cs
kenkendk: I think this follows the reasoning for for the other `VerifyOnly` that the operation is only interested to know if the data can be read safely. Partial stuff is not really important for listing for instance.

Duplicati/Library/Main/Backend/BackendManager.Handler.cs
carljohnsen: Cool, that was what I had in mind - notifying that things didn't go as expected.

Duplicati/Library/Main/Database/LocalDatabase.cs
carljohnsen: I see. My thought was like the download failed reporting, that this could be used to indicate that something went wrong.

PR #6025, opened 2025-03-09
Implemented single connection setup.

kenkendk:
With this commit, the database connection is performed once, and the same connection is used for sub-operations.

This makes it much cleaner when a connection is passed to a sub-operation, and allows the backend manager to directly write log data to the database without needing to pass the connection around.

For future updates, this also makes it possible to move the transaction into the manager, instead of having to pass around the transaction as it is done now.


kenkendk:
Closing because most of the changes here have been implemented in a different way.
The solution here also introduces a deadlock somewhere and permission issues on Windows.

PR #6060, opened 2025-03-19
Improve auth parsing helper and add timeouts

### LINKED ISSUE #5931: Rewrite the timeout support
The current timeout support is built into the BackendHandler with the idea that timeouts are handled from a single location.

Unfortunately this may not work in all cases, as the backend may use chunked uploads or unexpected extra calls.
If the backend is doing something else than transfering, e.g., listing the folder, the BackendManager is unaware and will enfore the timeout on the entire operation, instead of only the transfer as intended.

For this reason, we need to remove the timeout code from the BackendHandler and place it inside each of the backends.
Since most of the backends end up copying a stream, we could add a method like `CopyStreamWithTimeoutGuards()` that can be used from within the backends.

Special care is needed for things like the S3 backend, where we call a library and do not know the internals of the call.

Edit: We should also look into a "first-package-timeout" so we can allow a longer processing time for the call, but once data starts flowing we can use a smaller timeout.

kenkendk:
Revisited all existing backends and added in timeout support.
Unlike the previous timeout functionality, this is embedded in each backend and wraps only the relevant calls, making it less likely that it will cause issues.

To support this, a set of new options helpers were introduced. These options helpers ensure that all backends parse the options the same way. This was applied to any backend that has something similar to username/password.

Some backend bugs were fixed, and cases where the async propagation was not ideal were fixed.

This affects all backends.

This fixes #5931


LiveTests/Duplicati.Backend.Tests/CIFS/CIFSTests.cs
```diff
@@ -92,7 +92,7 @@ chmod 700 /shares/testshare1
         var exitCode = CommandLine.BackendTester.Program.Main(
             new[]
             {
-                $"cifs://localhost/testshare1/new/?transport=directtcp&auth-domain&auth-username=smbuser1&auth-password={randomPassword}",
+                $"sMB://localhost/testshare1/new/?transport=directtcp&auth-domain&auth-username=smbuser1&auth-password={randomPassword}",
```
marceloduplicati, line 95: mixed case typo, not critical

marceloduplicati (approved):

LiveTests/Duplicati.Backend.Tests/CIFS/CIFSTests.cs
kenkendk: Good catch! The file itself should also be renamed.

PR #6062, opened 2025-03-20
Improve SQLite loading

kenkendk:
This commit changes the way the new restore flow loads the extra database connections so these will also apply the temporary folder settings and custom pragmas.

This also changes the logic around the SQLite temporary folder. Previously, this folder would unconditionally be set to the general temporary folder, allowing only an override with a custom pragma.

Now the setup checks if the system has the environment variable `SQLITE_TMPDIR` set already. If this variable is already set, it will NOT be overwritten. Actual effects of this change depends on how SQLite internally resolves the temporary folder.

### DIFF
--- a/Duplicati/Library/SQLiteHelper/SQLiteLoader.cs
+++ b/Duplicati/Library/SQLiteHelper/SQLiteLoader.cs
@@ -100,6 +100,37 @@ namespace Duplicati.Library.SQLiteHelper
         }
 
         /// <summary>
+        /// Applies user-supplied custom pragmas to the SQLite connection
+        /// </summary>
+        /// <param name="con">The connection to apply the pragmas to.</param>
+        /// <returns>The connection with the pragmas applied.</returns>
+        public static System.Data.IDbConnection ApplyCustomPragmas(System.Data.IDbConnection con)
+        {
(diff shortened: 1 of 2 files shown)


Duplicati/Library/SQLiteHelper/SQLiteLoader.cs
```diff
@@ -99,6 +99,37 @@ public static System.Data.IDbConnection LoadConnection()
...
+        /// <param name="con">The connection to apply the pragmas to.</param>
+        /// <returns>The connection with the pragmas applied.</returns>
+        public static System.Data.IDbConnection ApplyCustomPragmas(System.Data.IDbConnection con)
+        {
+            var opts = Environment.GetEnvironmentVariable("CUSTOMSQLITEOPTIONS_DUPLICATI");
+            if (opts == null)
+                return con;
+            var topts = opts.Split(new char[] { ';' }, StringSplitOptions.RemoveEmptyEntries);
+            if (topts.Length == 0)
+                return con;
```
carljohnsen, line 114: Maybe a warning indicating that whatever they put in the variable didn't parse correctly? I guess that the only case where this is hit is where the variable has been set to 0 or more semicolons, and nothing else.

Issue #6068, opened 2025-03-21
new restore code doesn't restore folder timestamps


ts678:
- [x] I have searched open and closed issues for duplicates.
- [x] I have searched the [forum](https://forum.duplicati.com) for related topics.

----------------------------------------

## Environment info

- **Duplicati version**: 2.1.0.111 Canary but first noticed on 2.1.0.107 Canary
- **Operating system**: Windows 10
- **Backend**: N/A

## Description

Restore code before the rewrite restored folder "Date modified" and "Date created". New code doesn't.

## Steps to reproduce

1. Backup at least a folder with a file in it.
2. Restore it some other folder. No special options, especially no inadvertent `restore-legacy` enabled.
3. Browse there using File Explorer. Right click header bar to turn on `Date created`, at least for a look.

- **Actual result**:
Folder seems to have created and modified times from its time of restore.
- **Expected result**:
Folder has original created and modified times as captured in the backup.

PR #6079, opened 2025-03-24
Bugfix/restore folder metadata

carljohnsen:
This PR addresses Issue #6068 by:
- Adding a unit test to capture restoring a folders metadata correctly.
- Fixing the issue in the new restore flow, first by having the `FileLister` also request all of the target folders after having requested all of the files. Then the `FileProcessor` is reworked to operate in three phases:
  1. Restore all of the files, like before, until a folder request is received.
  2. Rendezvous all of the `FileProcessor`'s to ensure no more writes to the target folders. 
  3. Restore all of the target folders metadata, unless the `--skip-metadata=true` option has been specified.
- Updating the CLI restore help text, as it did not reflect the changes introduced by the new restore flow.

### DIFF
--- a/Duplicati/CommandLine/CLI/help.txt
+++ b/Duplicati/CommandLine/CLI/help.txt
@@ -118,13 +118,33 @@ Usage: %CLI_EXE% find <storage-URL> ["<filename>"] [<options>]
 > duplicati.commandline.exe help restore
 
 Usage: %CLI_EXE% restore <storage-URL> ["<filename>"] [<options>]
-  
-  Restores <filename> to its original destination. If <filename> exists already, <filename> is changed to <filename-timestamp.extension>. To restore all files use "*" or leave empty.
+
+  Restores <filename> to its original destination. If <filename> exists already and --overwrite=false is set, then <filename> is changed to <filename-timestamp.extension>. To restore all files use "*" or leave empty.
 
   --overwrite=<boolean>
     Overwrites existing files.
+  --restore-cache-max=<size>
+    Use this option to set the maximum size of the cache used for restoring files. The cache is used to store the data blocks that are downloaded from the remote storage. It assumes that the value is divisable by the block size, except for when it is 0, which disables the block cache. The default value is 4gb. This option is ignored if --restore-legacy=true is set.
+  --restore-cache-evict=<int>
+    Use this option to set the eviction ratio of the data block cache during restore. The eviction ratio is the percentage of the cache that is evicted when the cache is full. The default value is 50, which means that 50% of the cache is evicted when the cache is full. This option is ignored if --restore-legacy=true is set.
+  --restore-channel-buffer-size=<int>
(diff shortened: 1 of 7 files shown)


ts678:
FWIW in the issue, I didn't check the GUI option for what I think is another `restore` option not listed under `help restore`:

```
  --restore-permissions = false
    By default permissions are not restored as they might prevent you from accessing your files. Use this option to restore the permissions as well.
```

but there's a question I haven't explored in either the old or new code, which is whether permissions can block file restores.
IIRC old code did all that at the end, so during restore you'd have a weird blend of partial files with at least bad timestamps.

carljohnsen:
Good catch! I think that the new solution should handle this, as it's calling the same function as the legacy restore when applying metadata to files (and now also folders!). I'll add another unit test for this to make sure that is the case.

Duplicati/CommandLine/CLI/help.txt
```diff
@@ -118,27 +118,47 @@ Usage: %CLI_EXE% find <storage-URL> ["<filename>"] [<options>]
...
-  
-  Restores <filename> to its original destination. If <filename> exists already, <filename> is changed to <filename-timestamp.extension>. To restore all files use "*" or leave empty.
+
+  Restores <filename> to its original destination. If <filename> exists already and --overwrite=false is set, then <filename> is changed to <filename-timestamp.extension>. To restore all files use "*" or leave empty.
 
   --overwrite=<boolean>
     Overwrites existing files.
+  --restore-cache-max=<size>
+    Use this option to set the maximum size of the cache used for restoring files. The cache is used to store the data blocks that are downloaded from the remote storage. It assumes that the value is divisable by the block size, except for when it is 0, which disables the block cache. The default value is 4gb. This option is ignored if --restore-legacy=true is set.
+  --restore-cache-evict=<int>
```
kenkendk, line 128: Not sure if these options should be listed here?
The help text is intended for am overview of the most relevant options, and these options are more performance tuning, so I think they clutter the help text.

carljohnsen: Right, so you'd like these that would change the behaviour of the restore operation:
```
  --overwrite=<boolean>
    Overwrites existing files.
  --restore-legacy=<bool>
    Use this option to use the legacy restore method. The legacy restore method is slower than the new method, but may be more reliable in some cases. The default is false, which means that the new restore method is used.
  --restore-path=<string>
    Restores files to <restore-path> instead of their original destination. Top folders are removed if possible.
  --restore-permissions=<bool>
    By default permissions are not restored as they might prevent you from accessing your files. Use this option to restore the permissions as well.
  --skip-metadata=<bool>
    Use this option to skip the metadata restore. The default is false, which means that metadata is restored.
  --time=<time>
    Restore files that are older than the specified time.
  --version=<int>
    Restore files from a specific backup.
```

kenkendk: Yes, those are "common" I would say

Issue #6035, opened 2025-03-13
[FR] Reorder job based on time


nodiaque:
When looking at the home page, would be nice to be able to sort by something like backup name, next schedule run, etc. Right now, I'm trying to make my schedule for various backup and since they were created at different time, the order in they are showed make it difficult.

kenkendk:
Fixed with #6090

Issue #5719, opened 2024-11-29
Add (optional) alphabetical sorting to the home page list of backups


Oylex:
**Is your feature request related to a problem? Please describe.**

I name my backup with a prefix system to categorize them and right now, when I add a new backup, it goes to the end of the list instead of next to the existing backups of that "category".

**Describe the solution you'd like**

I'd like an option to sort the backups by name on the home page, this would be visual only, it wouldn't affect the order of the backups internally. Maybe the option could allow choosing between: Name and Backup ID. I only need ASC, but you could consider DESC as well.

**Describe alternatives you've considered**

One option would be to re-import the backup configurations in the order that I want. Another one could be making a little chrome extension that would reorder the listing, I haven't checked yet how feasible it is.

**Additional context**

Here's an example of how I named my backups (the last 2 items show my issue):

- Games - Game1 - Settings
- Games - Game1 - Saves
- Games - Game2 - Full
- Softwares - Software1
- Desktop
- Games - Game1 - Screenshots
- Softwares - Software2

marceloduplicati:
Another [issue](https://github.com/duplicati/duplicati/issues/6035) open about ordering.

I am also linking the ngclient [issue](https://github.com/duplicati/ngclient/issues/40) about re-ordering.

Filtering would be only client side as non persistent and ordering is persisted on the backend when toggled.

Even though a manual order feature could be interesting, the minimal approach would be a dropdown that allows to order by "last executed", "created on", "name", "backend type", "faulted first" which would be simple to implement by just one backend route to persist the sort method and applying the sorting on the existing route that returns the backup.

The "faulted first" would be good to bring to attention to failed backups and backups with warnings, in that order.

kenkendk:
Fixed with #6090

PR #6090, opened 2025-03-26
Adding OrderBy support on API + NGAX interface

marceloduplicati:
This allows the user to sort the backups by Create Order (ID), Name, LastRun or Destination Type.

The NGClient implementation will be committed on its own repository.

This is a no breaking change, both legacy client on the new api, or new client on legacy API will work.

The order is not persisted across sessions, defaulting to order by Create Order.

Note: Less CSS re-compiled.


marceloduplicati:
This will close this issues:

https://github.com/duplicati/duplicati/issues/5719

https://github.com/duplicati/duplicati/issues/6035

And clears path for: https://github.com/duplicati/ngclient/issues/40

kenkendk (approved):

Issue #6114, opened 2025-03-31
Don't set permissions when using ServerUtil


kenkendk:
When using ServerUtil, it will use the same code as the Server to locate the database.
This has the side-effect of creating the folder, and also setting permissions.
The read-only lookup should not set the permissions.

kenkendk:
Fixed with #6121

PR #6121, opened 2025-04-01
Refactoring of how permissions are set whilst obtaining DataFolder

marceloduplicati:
This addresses [Issue](https://github.com/duplicati/duplicati/issues/6114).

Whenever in non-write context, the permissions of the folder are not set.

Changing pattern from static constructor execution.


kenkendk (approved):

ts678:
Duplicati/Server/Program.cs ServerdatafolderLong changed DataFolderManager.DATAFOLDER_ENV_NAME into DataFolderManager.GetDataFolder call.

I think this is why 2.1.0.120 `Duplicati.Server help` looks like

> `--server-datafolder: Duplicati needs to store a small database with all settings. Use this option to choose where the settings are stored. This option can also be set with the environment variable C:\Users\<redacted>\AppData\Local\Duplicati\.`

2.1.0.112 looks like

> `--server-datafolder: Duplicati needs to store a small database with all settings. Use this option to choose where the settings are stored. This option can also be set with the environment variable DUPLICATI_HOME.`

PR #6123, opened 2025-04-02
Adding auth/validate route for ngclient

marceloduplicati:
This route would be used to test authentication (currently being done with a /systeminfo route call which is resource expensive)


kenkendk (approved):

marceloduplicati:
Yes, indeed, that is why I believe the client migration for this should be held back as well. 

How about we merge this route now on the backend, not for use yet, and give it time to propagate on the canary and stable builds then further ahead we can make the change?

Issue #6103, opened 2025-03-28
FTP backend appears to not close connection


kenkendk:
Reported on the forum:
https://forum.duplicati.com/t/release-2-1-0-111-canary-2025-03-15/20341/12

kenkendk:
Fixed with #6141

PR #6141, opened 2025-04-08
Fix Resource Leak in BackendManager Connection Handling

marceloduplicati:
While investigating issue [#6103](https://github.com/duplicati/duplicati/issues/6103), found that connections were not disposed due to the backend pool in BackendManager's handler. Set up a FileZilla server to test the issue before and after the change, confirming that connections now close as expected with this fix.


kenkendk (approved):
Nice find!

Issue #6125, opened 2025-04-04
Box backend can time out before the set timeout


kenkendk:
The HttpClient used in the box backend will time out after 100s, even if the requested timeout is larger.

Stack trace:
```
Unittest failed: System.Threading.Tasks.TaskCanceledException: The request was canceled due to the configured HttpClient.Timeout of 100 seconds elapsing.
  ---> System.TimeoutException: The operation was canceled.
  ---> System.Threading.Tasks.TaskCanceledException: The operation was canceled.
  ---> System.IO.IOException: Unable to read data from the transport connection: Operation canceled.
  ---> System.Net.Sockets.SocketException (89): Operation canceled
    --- End of inner exception stack trace ---
    at System.Net.Sockets.Socket.AwaitableSocketAsyncEventArgs.ThrowException(SocketError error, CancellationToken cancellationToken)
    at System.Net.Sockets.Socket.AwaitableSocketAsyncEventArgs.System.Threading.Tasks.Sources.IValueTaskSource<System.Int32>.GetResult(Int16 token)
    at System.Net.Security.SslStream.EnsureFullTlsFrameAsync[TIOAdapter](CancellationToken cancellationToken, Int32 estimatedSize)
    at System.Runtime.CompilerServices.PoolingAsyncValueTaskMethodBuilder`1.StateMachineBox`1.System.Threading.Tasks.Sources.IValueTaskSource<TResult>.GetResult(Int16 token)
    at System.Net.Security.SslStream.ReadAsyncInternal[TIOAdapter](Memory`1 buffer, CancellationToken cancellationToken)
    at System.Runtime.CompilerServices.PoolingAsyncValueTaskMethodBuilder`1.StateMachineBox`1.System.Threading.Tasks.Sources.IValueTaskSource<TResult>.GetResult(Int16 token)
    at System.Net.Http.HttpConnection.SendAsync(HttpRequestMessage request, Boolean async, CancellationToken cancellationToken)
    --- End of inner exception stack trace ---
    at System.Net.Http.HttpConnection.SendAsync(HttpRequestMessage request, Boolean async, CancellationToken cancellationToken)
    at System.Net.Http.HttpConnectionPool.SendWithVersionDetectionAndRetryAsync(HttpRequestMessage request, Boolean async, Boolean doRequestAuth, CancellationToken cancellationToken)
    at System.Net.Http.RedirectHandler.SendAsync(HttpRequestMessage request, Boolean async, CancellationToken cancellationToken)
[...]

PR #6142, opened 2025-04-08
Fix to Box backend - overwrite default HttpClient 100 second timeout

marceloduplicati:
If timeouts are specified, they are applied, otherwise it will default to infinite timeout rather than the 100 second timeout HttpClient defaults to.

### DIFF
--- a/Duplicati/Library/Backend/Box/BoxBackend.cs
+++ b/Duplicati/Library/Backend/Box/BoxBackend.cs
@@ -54,12 +54,11 @@ namespace Duplicati.Library.Backend.Box
         {
             private readonly TimeoutOptionsHelper.Timeouts _timeouts;
             public BoxHelper(string authid, TimeoutOptionsHelper.Timeouts timeouts)
-                : base(authid, "box.com")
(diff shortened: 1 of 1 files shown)


marceloduplicati:
Resolves (#6125)(https://github.com/duplicati/duplicati/issues/6125)

Duplicati/Library/Backend/Box/BoxBackend.cs
```diff
@@ -54,12 +54,11 @@ private class BoxHelper : OAuthHelperHttpClient
         {
             private readonly TimeoutOptionsHelper.Timeouts _timeouts;
             public BoxHelper(string authid, TimeoutOptionsHelper.Timeouts timeouts)
-                : base(authid, "box.com")
+                : base(authid, "box.com", new HttpClient { Timeout = timeouts.ReadWriteTimeout != TimeSpan.Zero ? timeouts.ReadWriteTimeout : Timeout.InfiniteTimeSpan })
```
kenkendk, line 57: I think this should just always use infinite?
If the user has set a timeout, that will be observed (outside of HttpClient), and the task will be cancelled on timeout.
If the user has set infinite timeout, it should be infinite here as well.

marceloduplicati: Indeed with the new timeoutobserving streams everywhere, yes.

PR #6169, opened 2025-04-16
Ensure commit before upload

kenkendk:
This makes the commits a bit more explicit.
The underlying problem is that prior to the backend performing any operation, it should write and commit the changes to the database.

This is problematic because the other tables in the database may be in an inconsistent state when the commit should be made.

Since SQLite only supports a single transaction this needs some manual transaction handling, where temporary tables are copied in before a commit, or multiple databases are in use.

The fix here does not address the underlying issue.


kenkendk:
Closing in favor of #6173

PR #6171, opened 2025-04-16
Fixed not uploading dlist on failure.

kenkendk:
If any of the uploads fail, do not upload the dlist in parallel with the failing upload.

Previously, it was possible to upload the dlist if the very last dblock was slow in failing.

This also fixes some logic around getting the size of the last uploaded volume, which is used for compating decisions.

There are still some quirks with the transactions. Flushing the queue with a transaction active causes partial updates to the database.


kenkendk:
Closed in favor of #6195

PR #6234, opened 2025-05-06
Tahoe LAFS migration to HttpClient with new timeouts & SSL options

marceloduplicati:
Migrated to HttpClient.

### DIFF
--- a/Duplicati/Library/Backend/TahoeLAFS/TahoeBackend.cs
+++ b/Duplicati/Library/Backend/TahoeLAFS/TahoeBackend.cs
@@ -19,287 +19,243 @@
 // FROM, OUT OF OR IN CONNECTION WITH THE SOFTWARE OR THE USE OR OTHER 
 // DEALINGS IN THE SOFTWARE.
 
+using System.Net;
+using System.Reflection;
+using System.Runtime.CompilerServices;
 using Duplicati.Library.Common.IO;
 using Duplicati.Library.Interface;
 using Duplicati.Library.Utility;
 using Duplicati.Library.Utility.Options;
 using Newtonsoft.Json;
-using Newtonsoft.Json.Linq;
-using System.Net;
-using System.Runtime.CompilerServices;
+using Uri = Duplicati.Library.Utility.Uri;
 
-namespace Duplicati.Library.Backend
+namespace Duplicati.Library.Backend;
+
+public class TahoeBackend : IStreamingBackend
 {
-    public class TahoeBackend : IBackend, IStreamingBackend
+    /// <summary>
+    /// Base URL for the Tahoe-LAFS backend
+    /// </summary>
+    private readonly string? _url;
+        
+    /// <summary>
+    /// The timeout options for API requests.
+    /// </summary>
+    private readonly TimeoutOptionsHelper.Timeouts? _timeouts ;
+
+    /// <summary>
+    /// The options for the SSL certificate validation
+    /// </summary>
+    private readonly SslOptionsHelper.SslCertificateOptions _certificateOptions = null!;
+
+    /// <summary>
+    /// Cached instance of HttpClient to be used
+    /// </summary>
+    private HttpClient? _httpClient;
+
+    public TahoeBackend()
     {
-        private readonly string m_url;
-        private readonly bool m_useSSL = false;
-        private readonly TimeoutOptionsHelper.Timeouts m_timeouts;
+            
+    }
 
-        private class TahoeEl
+    public TahoeBackend(string url, Dictionary<string, string?> options)
+    {
+        //Validate URL
+        var u = new Uri(url);
+        u.RequireHost();
+
+        if (!u.Path.StartsWith("uri/URI:DIR2:", StringComparison.Ordinal) && !u.Path.StartsWith("uri/URI%3ADIR2%3A", StringComparison.Ordinal))
+            throw new UserInformationException(Strings.TahoeBackend.UnrecognizedUriError, "TahoeInvalidUri");
+
+        _certificateOptions = SslOptionsHelper.Parse(options);
+
+        _url = u.SetScheme(_certificateOptions.UseSSL ? "https" : "http").SetQuery(null).SetCredentials(null, null).ToString();
+        _url = Util.AppendDirSeparator(_url, "/");
+        _timeouts = TimeoutOptionsHelper.Parse(options);
+    }
+        
+    /// <inheritdoc />
+    public Task TestAsync(CancellationToken cancelToken)
+        => this.TestListAsync(cancelToken);
+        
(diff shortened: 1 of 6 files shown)


Duplicati/Library/Backend/TahoeLAFS/TahoeBackend.cs
```diff
@@ -19,287 +19,243 @@
...
+        
+    /// <inheritdoc />
+    public Task TestAsync(CancellationToken cancelToken)
+        => this.TestListAsync(cancelToken);
+        
+    /// <inheritdoc />
+    public async Task CreateFolderAsync(CancellationToken cancelToken)
+    {
+        using var resp = await Utility.Utility.WithTimeout(_timeouts!.ListTimeout, cancelToken,
+            _ => GetHttpClient().SendAsync(CreateRequest(string.Empty, "t=mkdir", HttpMethod.Post),
```
kenkendk, line 85: The inner cancellation token here needs to be passed to `SendAsync` instead of the outer one, to be able to cancel the request on timeout.

Duplicati/Library/Backend/TahoeLAFS/TahoeBackend.cs
```diff
@@ -19,287 +19,243 @@
...
-
-            // TODO: When upgrading to HttpClient, also support the certificate options
-            var certOptions = SslOptionsHelper.Parse(options);
-            m_useSSL = certOptions.UseSSL;
-
-            m_url = u.SetScheme(m_useSSL ? "https" : "http").SetQuery(null).SetCredentials(null, null).ToString();
-            m_url = Util.AppendDirSeparator(m_url, "/");
-            m_timeouts = TimeoutOptionsHelper.Parse(options);
+            using var resp = await Utility.Utility.WithTimeout(_timeouts!.ListTimeout, cancelToken,
+                _ => GetHttpClient().SendAsync(CreateRequest(string.Empty, "t=json", HttpMethod.Get),
```
kenkendk, line 105: Same here with the inner/outer cancelToken

Duplicati/Library/Backend/TahoeLAFS/TahoeBackend.cs
```diff
@@ -19,287 +19,243 @@
...
+
+    /// <inheritdoc />
+    public async Task DeleteAsync(string remotename, CancellationToken cancelToken)
+    {
+        try
         {
-            using (var fs = File.Create(filename))
-                await GetAsync(remotename, fs, cancelToken).ConfigureAwait(false);
+            using (await Utility.Utility.WithTimeout(_timeouts!.ShortTimeout, cancelToken,
+                       _ => GetHttpClient().SendAsync(CreateRequest(remotename, string.Empty, HttpMethod.Delete),
```
kenkendk, line 172: Same here with the inner/outer cancelToken

Duplicati/Library/Backend/TahoeLAFS/TahoeBackend.cs
```diff
@@ -19,287 +19,243 @@
...
+    }
 
-        public IList<ICommandLineArgument> SupportedCommands => [
-            .. SslOptionsHelper.GetSslOnlyOption(),
-        ];
-
-        public string Description => Strings.TahoeBackend.Description;
+    /// <inheritdoc />
+    public IList<ICommandLineArgument> SupportedCommands => [
+        .. SslOptionsHelper.GetSslOnlyOption(),
```
kenkendk, line 185: Needs the timeout options as well

Duplicati/Library/Backend/TahoeLAFS/TahoeBackend.cs
```diff
@@ -19,287 +19,243 @@
...
+    }
 
-        public void Dispose()
+    /// <inheritdoc />
+    public async Task PutAsync(string remotename, Stream stream, CancellationToken cancelToken)
+    {
+        try
         {
-        }
+            var request = CreateRequest(remotename, string.Empty, HttpMethod.Put);
```
kenkendk, line 204: Needs a `using` ?

Duplicati/Library/Backend/TahoeLAFS/TahoeBackend.cs
```diff
@@ -19,287 +19,243 @@
...
-    public class TahoeBackend : IBackend, IStreamingBackend
+    /// <summary>
+    /// Base URL for the Tahoe-LAFS backend
+    /// </summary>
+    private readonly string? _url;
+        
+    /// <summary>
+    /// The timeout options for API requests.
+    /// </summary>
+    private readonly TimeoutOptionsHelper.Timeouts? _timeouts ;
```
kenkendk, line 44: Instead of making it nullable, you can make the default constructor with:
```
{
_timeouts = null!;
_certificateOptions = null!;
}
```

That way we avoid using `!` overrides in the rest of the code, and once we get around to removing the default constructor, everything works without `!` overrides.

Duplicati/Library/Backend/TahoeLAFS/TahoeBackend.cs
```diff
@@ -19,287 +19,243 @@
...
+    }
 
-        public async Task GetAsync(string remotename, System.IO.Stream stream, CancellationToken cancelToken)
-        {
-            var req = CreateRequest(remotename, "");
-            req.Method = WebRequestMethods.Http.Get;
+    /// <inheritdoc />
+    public async Task GetAsync(string remotename, Stream stream, CancellationToken cancelToken)
+    {
+        using var resp = await Utility.Utility.WithTimeout(_timeouts!.ListTimeout, cancelToken, _ => GetHttpClient().SendAsync(CreateRequest(remotename, string.Empty, HttpMethod.Get),cancelToken)).ConfigureAwait(false);
```
kenkendk, line 227: Same here with the inner/outer cancelToken

kenkendk, line 227: Do we need to pull out the call to `CreateRequest` so it can be properly disposed with a `using` statement?

kenkendk (approved):

Duplicati/Library/Backend/TahoeLAFS/TahoeBackend.cs
marceloduplicati: Well noted, fixed on all calls.

Duplicati/Library/Backend/TahoeLAFS/TahoeBackend.cs
marceloduplicati: I changed all of the implementations and got that covered.

marceloduplicati:
I will create a TestContainers based test of Tahoe in a separate PR so we can get that on CI as well.

PR #6277, opened 2025-05-20
Set VSS to Auto if user has permissions

### LINKED ISSUE #2833: Default of snapshot-policy should be auto

kenkendk:
If th user has the SeBackupPrivilege, set the default snapshot policy to `Auto` and otherwise use `Off` (the current default).

This ensures that if the process is running with the correct permissions, there will be fewer warnings.

This only changes the default value on Windows.

This fixes #2833

### DIFF
--- a/Duplicati/Library/Main/Options.cs
+++ b/Duplicati/Library/Main/Options.cs
@@ -110,6 +110,13 @@ namespace Duplicati.Library.Main
         private const long DEFAULT_BACKUP_TEST_SAMPLES = 1;
 
         /// <summary>
+        /// The default snapshot policy
+        /// </summary>
+        private static readonly OptimizationStrategy DEFAULT_SNAPSHOT_POLICY = PermissionHelper.HasSeBackupPrivilege()
+            ? OptimizationStrategy.Auto
+            : OptimizationStrategy.Off;
+
+        /// <summary>
         /// The default number of compressor instances
         /// </summary>
         private static readonly int DEFAULT_COMPRESSORS = Math.Max(1, Environment.ProcessorCount / 2);
@@ -372,7 +379,7 @@ namespace Duplicati.Library.Main
 
             new CommandLineArgument("upload-unchanged-backups", CommandLineArgument.ArgumentType.Boolean, Strings.Options.UploadUnchangedBackupsShort, Strings.Options.UploadUnchangedBackupsLong, "false"),
 
-            new CommandLineArgument("snapshot-policy", CommandLineArgument.ArgumentType.Enumeration, Strings.Options.SnapshotpolicyShort, Strings.Options.SnapshotpolicyLong, "off", null, Enum.GetNames(typeof(OptimizationStrategy))),
+            new CommandLineArgument("snapshot-policy", CommandLineArgument.ArgumentType.Enumeration, Strings.Options.SnapshotpolicyShort, Strings.Options.SnapshotpolicyLong, DEFAULT_SNAPSHOT_POLICY.ToString(), null, Enum.GetNames(typeof(OptimizationStrategy))),
             new CommandLineArgument("snapshot-provider", CommandLineArgument.ArgumentType.Enumeration, Strings.Options.SnapshotproviderShort, Strings.Options.SnapshotproviderLong, OperatingSystem.IsWindows() ? Snapshots.SnapshotProvider.AlphaVSS.ToString() : Snapshots.SnapshotProvider.LVM.ToString(), null, (OperatingSystem.IsWindows() ? [Snapshots.SnapshotProvider.AlphaVSS, Snapshots.SnapshotProvider.Wmic] : new [] { Snapshots.SnapshotProvider.LVM }).Select(x => x.ToString()).ToArray()),
             new CommandLineArgument("vss-exclude-writers", CommandLineArgument.ArgumentType.String, Strings.Options.VssexcludewritersShort, Strings.Options.VssexcludewritersLong, "{e8132975-6f93-4464-a53e-1050253ae220}"),
             new CommandLineArgument("vss-use-mapping", CommandLineArgument.ArgumentType.Boolean, Strings.Options.VssusemappingShort, Strings.Options.VssusemappingLong, "false"),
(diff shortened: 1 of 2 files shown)


ts678:
> if the process is running with the correct permissions

What are those, and are they easy to have?

I think the Issue wants VSS to usually work. Beyond question of what can enable SeBackupPrivilege, does VSS decide from that?

kenkendk:
> What are those, and are they easy to have?

That is how Windows decides if you are allowed to do snapshots. It is granted for Administrators and LocalService, but you can fiddle with the group policies if you want to control who has it.

> Beyond question of what can enable SeBackupPrivilege, does VSS decide from that?

More the other way around. To create a snapshot with VSS the current process needs to call a Windows API. If the SeBackupPrivilege is not enabled on that process, the call will fail.

So, you cannot do snapshots if you do not have that privilege enabled.

ts678:
> So, you cannot do snapshots if you do not have that privilege enabled.

This is not what Microsoft documentation says. It also doesn't match actual results for me.

[Security Considerations for Requesters](https://learn.microsoft.com/en-us/windows/win32/vss/security-considerations-for-requestors) does not explicitly talk privileges (maybe implicit?).

`whoami /priv` for an elevated administrator comes up like:

`SeBackupPrivilege                         Back up files and directories                                      Disabled`

and yet current Duplicati can do snapshots (provided I ask for them). But I have to elevate.
That's the part where Duplicati offers no built-in support, but if it did, it'd need a UAC use.

The directly documented (linked above) reason VSS is allowed is visible in `whoami /groups`

`BUILTIN\Administrators                                        Alias            S-1-5-32-544 Mandatory group, Enabled by default, Enabled group, Group owner`

Without elevation, an "administrator" account is a lot like a non-administrator account per

`BUILTIN\Administrators                                        Alias            S-1-5-32-544 Group used for deny only`

[How User Account Control works](https://learn.microsoft.com/en-us/windows/security/application-security/application-control/user-account-control/how-it-works) describes two tokens that an administrator account gets.

One question is whether one needs true administrator privileges to enable the permission.

The robocopy `/B :: copy files in Backup mode` is said (web search) to require elevation.

[Managing privileges in a file system](https://learn.microsoft.com/en-us/windows-hardware/drivers/ifs/privileges) describes SeBackupPrivilege. It bypasses ACL checking.

So I'm wondering if VSS ability and ACL bypass are getting mixed together inappropriately.

My unelevated administrator account doesn't even have `SeBackupPrivilege` available at all.

I'm guessing that this PR will not give anything to unelevated administrator, but if one has gone to the trouble to elevate for some reason (snapshots being potentially one of them), it can turn snapshots on (which worked before), and privilege adds ACL bypass.

Testing could disprove my guess. What I think people would like is unelevated administrator can VSS, but will Windows allow that?

kenkendk:
> It also doesn't match actual results for me.

There is a distinction between *having* a claim and an *enabled* claim.
The administrator has the claim, but it is not enabled by default.
The updated code will try to enable the claim, and only if that works, will it change the default to `Auto`.

From what I understand, it is possible to use group policies that prevent enabling the claim, so the only true test I could figure out was to simply try and enable it.

What you see with `whoami /privilege` is that your Administrator account has the claim, but it is not enabled.
This is the same I see. On a default Windows install, enabling the claim works.

Non-administrators do not have the claim, so they cannot enable it.

> and yet current Duplicati can do snapshots (provided I ask for them). But I have to elevate.

Yes, that is because the AlphaVSS component (or maybe just VSS itself) enables the claim somewhere in the process.
This PR has the side-effect that it enables the claim for the process, even if you do not use it.

> I'm guessing that this PR will not give anything to unelevated administrator, but if one has gone to the trouble to elevate for some reason (snapshots being potentially one of them), it can turn snapshots on (which worked before), and privilege adds ACL bypass.
> 
> Testing could disprove my guess. What I think people would like is unelevated administrator can VSS, but will Windows allow that?

From what I have read that is not possible, and I think that is logical. If you are granted `SeBackupPrivilege` somehow, you can essentially read any file, regardless of permissions and locking. This can then be used to bypass the entire security model.

A more advanced approach, that we have not started on, is to isolate the code that accesses files into a separate process. This process can then run as a service with elevated permissions, and the rest of Duplicati can use the service to gain access as if it was an elevated process. However, since this bypasses the security model, some careful design is needed to avoid letting any process bypass the security of the operating system.

> The robocopy /B :: copy files in Backup mode is said (web search) to require elevation.

That is a slightly different approach, which I would like to also add to Duplicati.
This works without snapshots, but reads files with `[BackupRead](https://learn.microsoft.com/en-us/windows/win32/api/winbase/nf-winbase-backupread)` that returns the contents of files in a [packed format](https://learn.microsoft.com/en-us/windows/win32/api/winbase/ns-winbase-win32_stream_id).

It still requires the `SeBackupPrivilege` though.

ts678:
> There is a distinction between having a claim and an enabled claim.

Agree with this paragraph. I got concerned when seeing "has the SeBackupPrivilege", as "has" was unclear until I read code.
Looking in [MS doc](https://learn.microsoft.com/en-us/windows-hardware/drivers/ifs/privileges), I see they do use the word "held", but while there, also notice `SeRestorePrivilege` and what that does.

> Yes, that is because the AlphaVSS component (or maybe just VSS itself) enables the claim somewhere in the process.

I guess this theory is hard to prove or disprove. The Microsoft documentation on VSS doesn't talk about `SeBackupPrivilege`.
Have you found any documentation supporting the theory? One not-for-sure point is Process Explorer during a VSS backup:

image

Logically, `SeBackupPrivilege` has related `SeRestorePrivilege` with similar super-access, yet VSS is tied to SeBackupPrivilege?

So if AlphaVSS enabled it, it looks like it disabled it later, which would be good if it's only needed at a certain spot. Any docs?

> This PR has the side-effect that it enables the claim for the process, even if you do not use it.

That worried me too, because

> If you are granted SeBackupPrivilege somehow, you can essentially read any file, regardless of permissions and locking. This can then be used to bypass the entire security model.

Since Duplicati enables it, does every run-script have it? You can test that on the new code with Process Explorer on the script.

PR #6320, opened 2025-06-03
Add ssh-agent support for SSH backend

### LINKED ISSUE #1557: Support authentication agents for SSH, like Pageant
There should be a full implementation here, just a matter of adding it:
https://github.com/dimov-cz/win-sshfs/commit/8eb20b34ec27c430cf3d235fb588ed4aa71a937f

kenkendk:
This adds the `SshNet.Agent` library to support Pageant on Windows and `ssh-agent` on Windows/Linux/MacOS.

For MacOS, this has some integration to the Keychain where on-disk private keys are added if their password is available through Keychain. If the private key itself is stored in Keychain (or on a hardware token), it will not be added as there is no API to extract such keys.

This fixes #1557


kenkendk:
Package mismatch

Issue #6340, opened 2025-06-12
Ensure that all httpClients have infinite timeout


kenkendk:
Since we manage request timeouts manually, we need to ensure that all `HttpClient` instances are configured with an infinite timeout.

duplicatibot:
This issue has been mentioned on **Duplicati**. There might be relevant details there:

https://forum.duplicati.com/t/release-2-1-0-119-canary-2025-05-29/20715/43

PR #6341, opened 2025-06-12
Sets infinite timeout for HTTP clients

marceloduplicati:
Ensures that long-running HTTP requests, such as downloads and uploads, do not timeout prematurely.

This change configures specific HttpClient instances with an infinite timeout to accommodate operations that may take an extended period. The default 100s timeout is retained for some operations like registration, where a long delay may indicate a problem.

Response to issue #6340

### DIFF
--- a/Duplicati/Library/AutoUpdater/UpdaterManager.cs
+++ b/Duplicati/Library/AutoUpdater/UpdaterManager.cs
@@ -231,7 +231,10 @@ namespace Duplicati.Library.AutoUpdater
                         using var timeoutToken = new CancellationTokenSource();
                         timeoutToken.CancelAfter(TimeSpan.FromSeconds(SHORT_OPERATION_TIMEOUT_SECONDS));
                         using (var client = HttpClientHelper.CreateClient())
+                        {
+                            client.Timeout = Timeout.InfiniteTimeSpan;
                             client.DownloadFile(request, tmpfile, null, timeoutToken.Token).Await();
+                        }
 
                         using (var fs = File.OpenRead(tmpfile))
                         {
@@ -372,7 +375,10 @@ namespace Duplicati.Library.AutoUpdater
                             timeoutToken.CancelAfter(TimeSpan.FromSeconds(DOWNLOAD_OPERATION_TIMEOUT_SECONDS));
 
                             using (var client = HttpClientHelper.CreateClient())
+                            {
+                                client.Timeout = Timeout.InfiniteTimeSpan;
                                 client.DownloadFile(request, tempfile, cb, timeoutToken.Token).Await();
+                            }
 
                             var sha256 = System.Security.Cryptography.SHA256.Create();
                             var md5 = System.Security.Cryptography.MD5.Create();
--- a/Duplicati/Library/Backend/Filen/FilenBackend.cs
+++ b/Duplicati/Library/Backend/Filen/FilenBackend.cs
@@ -107,7 +107,11 @@ public class FilenBackend : IStreamingBackend
         {
             _client?.Dispose();
             _client = null;
-            _client = await FilenClient.CreateClientAsync(HttpClientHelper.CreateClient(), _auth.Username!, _auth.Password!, _twoFactorCode, cancellationToken).ConfigureAwait(false);
+            _client = await FilenClient.CreateClientAsync(((Func<HttpClient>)(() => {
+                var httpClient = HttpClientHelper.CreateClient();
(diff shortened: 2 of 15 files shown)


Duplicati/Library/AutoUpdater/UpdaterManager.cs
```diff
@@ -231,7 +231,10 @@ public static Version TryParseVersion(string? str)
                         using var timeoutToken = new CancellationTokenSource();
                         timeoutToken.CancelAfter(TimeSpan.FromSeconds(SHORT_OPERATION_TIMEOUT_SECONDS));
                         using (var client = HttpClientHelper.CreateClient())
+                        {
+                            client.Timeout = Timeout.InfiniteTimeSpan;
                             client.DownloadFile(request, tmpfile, null, timeoutToken.Token).Await();
```
kenkendk, line 236: The client.DownloadFile does not have a timeout observing stream. Should we add one or just let this part retain the 100s timeout?

Duplicati/Library/AutoUpdater/UpdaterManager.cs
```diff
@@ -372,7 +375,10 @@ public static bool DownloadUpdate(UpdateInfo version, PackageEntry package, stri
                             timeoutToken.CancelAfter(TimeSpan.FromSeconds(DOWNLOAD_OPERATION_TIMEOUT_SECONDS));
 
                             using (var client = HttpClientHelper.CreateClient())
+                            {
+                                client.Timeout = Timeout.InfiniteTimeSpan;
                                 client.DownloadFile(request, tempfile, cb, timeoutToken.Token).Await();
```
kenkendk, line 380: The client.DownloadFile does not have a timeout observing stream. Should we add one or just let this part retain the 100s timeout?

Duplicati/Library/Backend/Filen/FilenBackend.cs
```diff
@@ -107,7 +107,11 @@ private async Task<FilenClient> GetClientAsync(CancellationToken cancellationTok
         {
             _client?.Dispose();
             _client = null;
-            _client = await FilenClient.CreateClientAsync(HttpClientHelper.CreateClient(), _auth.Username!, _auth.Password!, _twoFactorCode, cancellationToken).ConfigureAwait(false);
+            _client = await FilenClient.CreateClientAsync(((Func<HttpClient>)(() => {
```
kenkendk, line 110: Do we need a callback function here? Could we just create the client before calling the `CreateClientAsync()` method?

Duplicati/Library/Modules/Builtin/SendHttpMessage.cs
```diff
@@ -393,7 +393,7 @@ protected override void SendMessage(string subject, string body)
             using HttpClientHandler httpHandler = new HttpClientHandler();
             HttpClientHelper.ConfigureHandlerCertificateValidator(httpHandler, m_acceptAnyCertificate, m_acceptSpecificCertificates);
 
-            using var client = new HttpClient(httpHandler);
+            using var client = HttpClientHelper.CreateClient(httpHandler);
```
kenkendk, line 396: This does not set the timeout, which I think is correct.
Do we need a comment here?

Duplicati/Library/UsageReporter/ReportSetUploader.cs
```diff
@@ -89,6 +90,7 @@ public static Tuple<Task, IWriteChannel<string>> Run()
                                         timeoutToken.CancelAfter(TimeSpan.FromSeconds(UPLOAD_OPERATION_TIMEOUT_SECONDS));
 
                                         using var client = HttpClientHelper.CreateClient();
+                                        client.Timeout = Timeout.InfiniteTimeSpan;
```
kenkendk, line 93: I think this one should keep the 100s timeout here

kenkendk (approved):

Duplicati/Library/AutoUpdater/UpdaterManager.cs
marceloduplicati: We do have timeoutToken above ˆ

Duplicati/Library/AutoUpdater/UpdaterManager.cs
marceloduplicati: We do have timeoutToken above ˆ

Duplicati/Library/Backend/Filen/FilenBackend.cs
marceloduplicati: Yes, it was just an abuse of inline packing.

Issue #6359, opened 2025-06-18
Delayed termination of duplicati process by 30 to 90 seconds


WalterFaber:
- [x] I have searched open and closed issues for duplicates.
- [x] I have searched the [forum](https://forum.duplicati.com) for related topics.

----------------------------------------

## Environment info
- **Duplicati version**: 2.1.0.5_stable_2025-03-04 (deb package)
- **Operating system**: Fresh Ubuntu 24.04.2 LTS installation, Gnome 46, Wayland,
- **Backend**: Onedrive v2

## Description
Either in “idle” state or "paused" state duplicati won’t quit immediately after I send sigterm via kill.
Via kill, it takes about 30 to 90 seconds to stop.
 If I select “quit” from the tray options, it stops within 3 seconds.

This is an issue, because it delays regular system shutdown by the same amount of time i.e. I have to wait 30 to 90seconds to shut down my computer because it waits for duplicati to shut down (even if it was in idle state).

Remark: Scheduled backups work as expected.

## Steps to reproduce
1. Install duplicati deb package on Ubuntu 24.04.2
2. Run duplicati via Tray Icon (Tray Icon appears)
3. Shut down the system (Ubuntu via "Power Off" menu entry in the upper right corner) resp. send "pkill duplicati" via console (= sig term)

- **Actual result**:
The tray icon disappears immediately.
It takes 30 to 90 seconds until the Ubuntu shuts down resp. it takes 30 to 90 seconds until the duplicati process stops.

See Logs A)

- **Expected result**:
Ubuntu starts shutdown process within 3 seconds resp. duplicati process stops within 3 seconds.

REMARK 1:
 If I select “quit” from the tray options, it stops within 3 seconds.
See Logs B)

REMARK 2:
User ts678 reported a very similar issue already in March 2025 in the forum (however in a different context, so it was not recognized as an issue): https://forum.duplicati.com/t/tray-icon-wont-launch-after-update/19756/20 
He kindly tested it today again and also confirmed it (as far as I understand on a linux system and even a windows system. See his logs and comments here: https://forum.duplicati.com/t/tray-icon-wont-launch-after-update/19756/62
Maybe it is the same, maybe it is different.

## Debug log
Logs A) - Delayed shutdown of system resp. pkill duplicati command (sigterm)
`
Jun 18 12:11:17 hostname duplicati[6816]: Request error: One or more errors occurred. (An error occurred while sending the request.)
[...]

PR #6372, opened 2025-06-25
Addresses shutdown and thread safety issues

marceloduplicati:
Improvement on AvaloniaRunner, HttpServerConnection, and HostedInstanceKeeper to prevent race conditions during shutdown.

Adds cancellation tokens and locking mechanisms to ensure that operations are properly stopped and resources are released in a controlled manner.

Also handles a NullReferenceException during Avalonia shutdown, and avoids invoking `InstanceShutdown` or `ConnectionClosed` multiple times.

This fixes solve issue #6359


Duplicati/Server/Program.cs
```diff
@@ -143,6 +143,11 @@ public static int Main(IApplicationSettings applicationSettings, string[] _args)
             PreloadSettingsLoader.ConfigurePreloadSettings(ref _args, PackageHelper.NamedExecutable.Server, out var preloadDbSettings);
 
             applicationSettings ??= new ApplicationSettings();
+            
+            System.Runtime.Loader.AssemblyLoadContext.Default.Unloading += context =>
```
kenkendk, line 147: Is this needed? Could we end up triggering this if some random assembly is unloaded?

Duplicati/GUI/Duplicati.GUI.TrayIcon/HttpServerConnection.cs
```diff
@@ -370,17 +442,37 @@ private async Task<T> PerformRequestInternalAsync<T>(string method, string endpo
             // Set up response timeout, use 100s which is the .NET default
             using var cts = new CancellationTokenSource();
             cts.CancelAfter(timeout.HasValue
-                ? timeout.Value + TimeSpan.FromSeconds(5)
+                ? timeout.Value + TimeSpan.FromSeconds(30)
```
kenkendk, line 445: The intention with +5s was to account for any jitter, so it is more likely that the server will trigger the timeout and return a long-poll response, instead of having a cancellation.

Do we need 30s to cover jitter?

kenkendk (approved):

Duplicati/Server/Program.cs
marceloduplicati: I will test, but indeed likely to be unnecessary.

Duplicati/GUI/Duplicati.GUI.TrayIcon/HttpServerConnection.cs
marceloduplicati: No, I was trying different values to pinpoint the issue and ended up not reverting after it was solved elsewhere. I will re-test on 5 and ammend.

PR #6386, opened 2025-07-01
Update support for VSS

### LINKED ISSUE #3094: UnauthorizedAccessException for Cygwin files with restrictive access rights

## Environment info
- **Duplicati version**: v2.0.2.21-2.0.2.21_canary_2018-03-06
- **Operating system**: Windows 7 x64
- **Backend**: od4b, local
- **Parameters**: --usn-policy=Required --snapshot-policy=Required

## Description
Duplicati is running as a service on Win7x64. The user running the service is SYSTEM (S-1-5-18).
Duplicati cannot make a backup of some of my Cygwin files & folders.

```
System.UnauthorizedAccessException: (5) Zugriff verweigert: [\\?\GLOBALROOT\Device\HarddiskVolumeShadowCopy11\cygwin\home\user\.bash_history]
   bei Alphaleonis.Win32.NativeError.ThrowException(UInt32 errorCode, String readPath, String writePath)
   bei Alphaleonis.Win32.Filesystem.NativeMethods.IsValidHandle(SafeHandle handle, Int32 lastError, String path, Boolean throwException)
   bei Alphaleonis.Win32.Filesystem.File.CreateFileCore(KernelTransaction transaction, String path, ExtendedFileAttributes attributes, FileSecurity fileSecurity, FileMode fileMode, FileSystemRights fileSystemRights, FileShare fileShare, Boolean checkPath, Boolean continueOnException, PathFormat pathFormat)
   bei Alphaleonis.Win32.Filesystem.File.OpenCore(KernelTransaction transaction, String path, FileMode mode, FileSystemRights rights, FileShare share, ExtendedFileAttributes attributes, Nullable`1 bufferSize, FileSecurity security, PathFormat pathFormat)
   bei Duplicati.Library.Snapshots.SystemIOWindows.FileOpenRead(String path)
   bei Duplicati.Library.Snapshots.WindowsSnapshot.OpenRead(String file)
   bei Duplicati.Library.Main.Operation.BackupHandler.HandleFilesystemEntry(ISnapshotService snapshot, BackendManager backend, String path, FileAttributes attributes)
```
also similar:
```
System.UnauthorizedAccessException: (5) Zugriff verweigert: [\\?\GLOBALROOT\Device\HarddiskVolumeShadowCopy11\cygwin\home\user\.ssh]
   bei Alphaleonis.Win32.NativeError.ThrowException(UInt32 errorCode, String readPath, String writePath)
   bei Alphaleonis.Win32.Filesystem.FindFileSystemEntryInfo.ThrowPossibleException(UInt32 lastError, String pathLp)
   bei Alphaleonis.Win32.Filesystem.FindFileSystemEntryInfo.FindFirstFile(String pathLp, WIN32_FIND_DATA& win32FindData, Boolean suppressException)
   bei Alphaleonis.Win32.Filesystem.FindFileSystemEntryInfo.<Enumerate>d__84`1.MoveNext()
   bei System.Linq.Buffer`1..ctor(IEnumerable`1 source)
   bei System.Linq.Enumerable.ToArray[TSource](IEnumerable`1 source)
   bei Alphaleonis.Win32.Filesystem.Directory.GetDirectories(String path)
   bei Duplicati.Library.Snapshots.WindowsSnapshot.ListFolders(String folder)
   bei Duplicati.Library.Utility.Utility.<EnumerateFileSystemEntries>d__22.MoveNext()
```

I already traced and debugged the issue pretty much through the whole stack.
It started with this issue: https://github.com/alphaleonis/AlphaVSS/issues/17
But it is not a AlphaVSS issue - so I created this issue: https://github.com/alphaleonis/AlphaFS/issues/428
But it is also not a AlphaFS issue - it's actually a Duplicati issue.
I have documented a working solution in https://github.com/alphaleonis/AlphaFS/issues/428
I would like to see it fixed in Duplicati but it is not that easy to fix.

The lack of the pivilege SE_BACKUP_NAME "SeBackupPrivilege" is the root cause of the issue.
I found `WinNativeMethods.BackupPrivilege` and added it here:

https://github.com/shiek/duplicati/blob/3dbe476da38faf197e234c09a607343dba74faab/Duplicati/Library/Main/Operation/BackupHandler.cs#L961

I know it's kinda dirty to add it to the BackupHandler.cs but I don't know a good place to put it.

Enabling "SeBackupPrivilege" is not enough to fix the issue.
I had to add `AlphaFS.ExtendedFileAttributes.BackupSemantics` to the `Alphaleonis.Win32.Filesystem.File.Open` call:

https://github.com/shiek/duplicati/blob/3dbe476da38faf197e234c09a607343dba74faab/Duplicati/Library/Snapshots/SystemIOWindows.cs#L190

I also added `AlphaFS.PathFormat.LongFullPath` because it seems like the best option.

PS: `AlphaFS.PathFormat.LongFullPath` seems like the better option here anyway because the default is `AlphaFS.PathFormat.RelativePath` - "Using LongFullPath has a slight performance advantage compared to using RelativePath." - http://alphafs.alphaleonis.com/doc/2.1/api/html/5E9EE47E.htm
PPS: I used AlphaFS-dev which will soon be AlphaFS 2.2 (Duplicati currently uses 2.1.3) with Duplicati-master and it didn't work. I didn't test it yet but I think using `AlphaFS.PathFormat.LongFullPath` should solve this issue too.

kenkendk:
This updates VSS to default use Vanara in favor of AlphaVSS which is no longer maintained.

The build for Vanara requires targeting `net8.0-windows7.0`, which will cause significant build overhead and complexity for cross platform builds. To counter this, the setup is to have a single project, `Duplicati.Library.WindowsModules`, that is targeting `net8.0-windows7.0`. The output from this project is then hoisted into the TrayIcon project for Windows builds so the files are available when debugging on Windows. A top-level dummy executable project is added to ensure the project always builds. The built modules are then loaded with reflection when requested.

With the use of Vanara there is now also support for using BackupRead to read files without making a VSS snapshot. With BackupRead, it is possible to read locked files, but it still requires the SeBackupPrivilege as VSS does as well.

Unfortunately, the `vssapi.dll` file is not shipped for Arm64 on Windows, so even with Vanara this will not work, and only WMIC is supported on Arm64. A workaround is to run Duplicati with x64 emulation if more advanced VSS features are needed (HyperV and MSSQL support).

This PR also updates options and filters out unsupported options for each operating system, so options that are not supported by the current OS are not reported and will give warnings if they are used, as opposed to just being ignored.

The release builder project has been updated to exclude the unused project, and purge unwanted outputs.

This fixes #3094


ts678:
While I dislike the time it takes VSS to start (my 1 second backup becomes a 1 minute backup), VSS does many things with an application that is VSS-aware to ensure that files are prepared for backup and in a consistent state. Microsoft describes this at:

[How a shadow copy is created](https://learn.microsoft.com/en-us/windows-server/storage/file-server/volume-shadow-copy-service#how-a-shadow-copy-is-created)

> With the use of Vanara there is now also support for using BackupRead to read files without making a VSS snapshot.

Somewhere in the help and/or docs, it should be explained what one loses by choosing it. It's probably hard to explain quickly.
Looking at this another way, reading open files that are not consistent is probably one's fate on any OS without VSS, e.g. Linux.

kenkendk:
For Linux, you can use LVM to create a disk snapshot, but LVM does not integrate with writers like VSS does.
This makes the snapshots faster, but less consistent.
There are many downsides to VSS (speed and stability for instance), but the logic is far better at creating consistent snapshots than any other solution.

The option to use BackupRead is a middle ground that is more similar to the Linux way. BackupRead is not used when reading from a snapshot, as that is assumed to no be locked.

Further on, the use of BackupRead will allow us to capture more metadata, but this is currently not used.

PR #6387, opened 2025-07-02
Added installer script for Windows installs

kenkendk:
This PR adds installer scripts for automated install and configuration of Duplicati as a service.


kenkendk:
Closed as this if fixed with #6934

PR #6404, opened 2025-07-08
Bugfix/Fixes to the argument parsing of the remote synchronization tool

carljohnsen:
The argument parsing in the remote synchcronization tool was very simple, leading to arguments with multiple equals (`=`) signs being parsed incorrectly. 

This PR fixes this by recombining the remaining tokens into a single string, rather than just taking the first remaining token. The PR also adds additional testing of whether the arguments were parsed correctly, along with extensions to the `TestMainMethodParsesArgumentsCorrectly` unit test to test these new types of options.

Finally, it reduces the sizes of the larger remote synchronization test to speed it up, as the file sizes shouldn't have a large effect on the behaviour. The test went from 42 seconds to 7 seconds.

### DIFF
--- a/Tools/RemoteSynchronization/Program.cs
+++ b/Tools/RemoteSynchronization/Program.cs
@@ -143,7 +143,7 @@ destination will be verified before being overwritten (if they seemingly match).
         private static async Task<int> Run(Config config)
         {
             // Unpack and parse the multi token options
-            Dictionary<string, string> global_options = ParseOptions(config.GlobalOptions);
+            var global_options = ParseOptions(config.GlobalOptions);
 
             // Parse the log level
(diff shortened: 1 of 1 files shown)


Tools/RemoteSynchronization/Program.cs
```diff
@@ -143,9 +143,7 @@ destination will be verified before being overwritten (if they seemingly match).
         private static async Task<int> Run(Config config)
         {
             // Unpack and parse the multi token options
-            Dictionary<string, string> global_options = config.GlobalOptions
-                .Select(x => x.Split("="))
-                .ToDictionary(x => x[0], x => x[1]);
+            Dictionary<string, string> global_options = ParseOptions(config.GlobalOptions);
```
kenkendk, line 146: This should be `var`

Issue #6438, opened 2025-07-29
Backblaze B2 uploads fail, often with timeout. Big read-write-timeout may help.


ts678:
- [x] I have searched open and closed issues for duplicates.
- [x] I have searched the [forum](https://forum.duplicati.com) for related topics.

----------------------------------------

## Environment info

- **Duplicati version**: 2.1.1.0 Experimental and roughly Canary 2.1.0.102-2.1.0.125 so far.
- **Operating system**: Windows or Linux
- **Backend**: Backblaze B2

## Description

Backblaze B2 has had read-write-timeout upload failures since 2.1.0.102 (hard to pinpoint).
Default 30 second timeout is not solid enough, so I'm testing with 15 second timeout now.
See [Release: 2.1.0.125 (Canary) 2025-07-15](https://forum.duplicati.com/t/release-2-1-0-125-canary-2025-07-15/20889), and another user where 5 minute timeout fails.
This seems to affect the B2 backend, not Backblaze S3. Early contrary reports seem in error.

2.1.0.125 with BackendTool:

```
Command failed: The operation has timed out.
System.TimeoutException: The operation has timed out.
   at Duplicati.StreamUtil.TimeoutObservingStream.ReadImplAsync(Byte[] buffer, Int32 offset, Int32 count, CancellationToken cancellationToken)
   at System.IO.Stream.<CopyToAsync>g__Core|27_0(Stream source, Stream destination, Int32 bufferSize, CancellationToken cancellationToken)
   at System.Net.Http.HttpContent.<CopyToAsync>g__WaitAsync|56_0(ValueTask copyTask)
   at System.Net.Http.HttpConnection.SendRequestContentAsync(HttpRequestMessage request, HttpContentWriteStream stream, Boolean async, CancellationToken cancellationToken)
   at System.Net.Http.HttpConnection.SendAsync(HttpRequestMessage request, Boolean async, CancellationToken cancellationToken)
   at System.Net.Http.HttpConnection.SendAsync(HttpRequestMessage request, Boolean async, CancellationToken cancellationToken)
   at System.Net.Http.HttpConnectionPool.SendWithVersionDetectionAndRetryAsync(HttpRequestMessage request, Boolean async, Boolean doRequestAuth, CancellationToken cancellationToken)
   at System.Net.Http.RedirectHandler.SendAsync(HttpRequestMessage request, Boolean async, CancellationToken cancellationToken)
[...]

PR #6439, opened 2025-07-29
Moves timeout output to correct location

marceloduplicati:
Ensures that the timeout value is correctly applied either from environment variables or command line areguments before the output to console.


marceloduplicati:
Minor correction found while testing different values for timeout on issue #6438 

Affects only the output to console, nevertheless fixed.

PR #6463, opened 2025-08-07
PoC for guarding against server stalls

kenkendk:
This is a PoC that illustrates how to guard against a server stall that happens after a stream has been fully uploaded to the server, but where the server then does not respond to the message within a set period.


Duplicati/Library/Backend/Backblaze/B2.cs
```diff
@@ -323,8 +324,8 @@ public async Task PutAsync(string remotename, Stream stream, CancellationToken c
             request.Content.Headers.Add("Content-Type", "application/octet-stream");
             request.Content.Headers.Add("Content-Length", timeoutStream.Length.ToString());
 
-            var response = await _httpClient.UploadStream(request, cancelToken).ConfigureAwait(false);
-            var rdata = await response.Content.ReadAsStreamAsync(cancelToken).ConfigureAwait(false);
+            var response = await _httpClient.SendAsync(request, HttpCompletionOption.ResponseHeadersRead, timeoutCts.Token).ConfigureAwait(false);
```
marceloduplicati, line 327: There should be no significant difference between HttpCompletionOption.ResponseHeadersRead and HttpCompletionOption.ResponseContentRead given the response is always a small json, but I guess by principle or  HttpCompletionOption.ResponseHeadersRead

kenkendk:
Should instead use timeout when reading the response, and not couple it with the upload stream timeout

Issue #6456, opened 2025-08-05
Ability to cap size of temporary space during restore


kenkendk:
When restoring files with a unlucky order, the restore process can "hang-on" to downloaded files for an extended period, potentially exhausting the temporary storage.

It should be possible to set a cap on the maximum size of active downloaded files.

duplicatibot:
This issue has been mentioned on **Duplicati**. There might be relevant details there:

https://forum.duplicati.com/t/disk-full-during-restore/20923/4

PR #6476, opened 2025-08-08
Feature/restore volume cache size parameter

carljohnsen:
This PR addresses #6456 by adding the option `restore-volume-cache-hint` to control the number of volumes that are cached during the restore operation. This provides a trade-off between performance and temporary disk consumption. It is a hint to the client, where if the specified limit is reached, it will try to evict the least recently used volume that is not actively in use in the cache. This means that for small values, the value will be exceeded, but the client will more aggressively try to evict.

The implementation required slight modification to the new restore flow, as the volumes cannot be removed from cache (and thus also disk) until we're sure that all of the `VolumeDecompressors` are done with them. This is implemented by having the `BlockManager.VolumeConsumer` message the `VolumeManager` with acknowledgements stating that the request for said volume has completed, in turn letting the `VolumeManager` keep track of how many references to each volume is "in flight". 

The restore flow is also changed to make the process network have more points of checking `CancellationToken`'s, which should make it cancel-able even during a deadlock.  

Steps remaining until this PR is done: 
- [x] Evaluate whether tuning this option does affect disk consumption. 
- [x] Evaluate the performance difference of this modification to the restore flow. 
- [x] Add unit tests that verify that the eviction logic works as expected.


Duplicati/Library/Main/Operation/Restore/FileProcessor.cs
```diff
@@ -93,6 +101,12 @@ public static Task Run(Channels channels, LocalRestoreDatabase db, IChannel<Bloc
 
                 // Indicates whether this FileProcessor is still restoring files
                 var decremented = false;
+                // ID for this file processor, used for debugging.
+                int my_id = -1;
```
kenkendk, line 105: Consider using this instead:
```csharp
var my_id = System.Threading.Interlocked.Increment(ref id);
```

This saves a lock object

kenkendk (approved):

PR #6480, opened 2025-08-11
Lowercase S3 signature header value

kenkendk:
This PR introduces a custom AWS S3 client that will lower-case the `x-amz-content-sha256` header for improved compatibility with Wasabi S3.


kenkendk:
Closed in favor of #6484

PR #6492, opened 2025-09-02
Ensure that a query string is formatted with invariant string formatting

kenkendk:


carljohnsen (commented):
All of the database methods use `Duplicati.Library.Utility.Utility.FormatInvariant` through their calls to `SetCommandAndParameters` extension method. Does it have to be called on the substrings when they're formatted as well? Because then there might be other places that need it as well.

PR #6498, opened 2025-09-11
fix(S3): validate endpoint does not contain path

smyja:
Add validation to ensure custom S3 endpoint URLs like r2 from cloudflare don't contain paths and throw appropriate error message. This prevents potential issues with malformed endpoint configurations. Previously, all it said for the AWS SDK was "Name or service not known (https:80)" which was confusing, i believe this would make it clearer.


kenkendk:
Thanks, looks good, I will merge when it passes the tests

PR #6532, opened 2025-09-29
Updated README to reduce confusion about incremental/differential

kenkendk:


ts678:
[Backup type is incremental?](https://forum.duplicati.com/t/backup-type-is-incremental/8786/7?u=ts678) is maybe why differential was seen as the better word (discussion back and forth).
Writing it as "differential to save bandwidth and storage" makes less sense, since incremental does that better.
[Checkpoint cumulative updates](https://learn.microsoft.com/en-us/windows/whats-new/whats-new-windows-11-version-24h2#checkpoint-cumulative-updates) show how Windows Update is ending cumulative-since-RTM -- it gets too big.

Neither word is correct for Duplicati, so it's bound to get confusing, so might just be trying to scare users least.
Calling it a "deduplicated full backup" is in some ways accurate, but it sounds scary as soon as one sees "full"...

Issue #363, opened 2014-08-05
Create a german translation


kenkendk:
_From [ernst.m...@gmail.com](https://code.google.com/u/114302124077215675693/) on February 07, 2011 17:42:13_

In Duplicati version 2.1 for Windows XP there is no German translation.

_Original issue: http://code.google.com/p/duplicati/issues/detail?id=363_

PR #6547, opened 2025-10-14
Removed unused option

kenkendk:
The option `asynchronous-upload-limit` was not used, and has been removed.

### DIFF
--- a/Duplicati/Library/Main/Options.cs
+++ b/Duplicati/Library/Main/Options.cs
@@ -90,7 +90,7 @@ namespace Duplicati.Library.Main
         /// <summary>
         /// The default number of concurrent uploads
         /// </summary>
-        private const int DEFAULT_ASYNCHRONOUS_CONCURRENT_UPLOAD_LIMIT = 4;
+        private const int DEFAULT_ASYNCHRONOUS_UPLOAD_LIMIT = 4;
 
         /// <summary>
         /// The backends where throttling is disabled by default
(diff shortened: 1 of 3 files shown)


ts678:
[asynchronous-upload-limit setting ignored #363](https://github.com/duplicati/duplicati/issues/3634) noticed it wasn't working. Will it be hardcoded? Current help thinks it's 4.

kenkendk:
@ts678 Prior to this PR the option `--asynchronous-concurrent-upload-limit` would limit the number of concurrent uploads.

I have now changed this PR so the primary name is now `--asynchronous-upload-limit` and `--asynchronous-concurrent-upload-limit` is an alias for the option.

The net effect is that users who already use `--asynchronous-concurrent-upload-limit` will not have to do anything.
Users who are using `--asynchronous-upload-limit` have so far had no effect, but presumably they *wanted* to get an effect. After this PR the option will have an effect.

ts678:
@kenkendk 

> but presumably they wanted to get an effect. After this PR the option will have an effect.

but not the effect they wanted. This isn't quite as bad as redefining working option though.
Risk will be that a config asking for past documented effect gets surprised by different one.

[asynchronous-upload-limit](https://docs.duplicati.com/community-docs/community-docs-advanced-options#asynchronous-upload-limit) might keep old effect if old options doc keeps getting obsolete.

EDIT 1:

Original issue needs no presumption about what was wanted as an effect. They won't get it.

> Expected result:
I expect 2 backup archives to be pending for upload in the temp directory, up to 3 if you include the current file being uploaded.

PR #6588, opened 2025-11-02
Add api to pause resume single task

kenkendk:
This adds an API to pause/resume the current task


kenkendk:
Leaving this for now, it does not seem like we need it. The server pause will also pause the active task.

PR #6593, opened 2025-11-03
Manually merged restore to match master

carljohnsen:
This PR manually merges the files related to the reworked restore flow from the current state of master (14be3f5).


zieren:
I'm running Duplicati 2.2.0.3 - 2.2.0.3_stable_2026-01-06. Given that the above was merged in beta-2.2.0.2 I'd expect the bug to be fixed for me, but it isn't. In fact I'm seeing about 200% disk space use, e.g.:

```
$ du -sh *
4.1G	test
7.0G	tmp
```

I've pointed --tempdir to a dedicated btrfs subvolume and that fixed it, but I assume this will significantly limit users running with /tmp.

Issue #5870, opened 2025-01-13
ServerUtil report backup status


kenkendk:
When using the serverutil and waiting for the backup to complete, it would be good if the reported status code reflects the status of the backup.
Since the backup has completed, it should be sufficient to check the backup status, examine the `suggestedIcon` property and use that to report the status.

PR #6618, opened 2025-11-07
Change ServerUtil run to return backup status after finishing

greyxr:
This is an attempt to address the issue [here](https://github.com/duplicati/duplicati/issues/5870). When run with the ```--wait``` flag, the ServerUtil run command now displays the status of the most recent specified backup from the log ("Success", "Interrupted", or "Fatal") after finishing the attempt. Previously, the command would always output "Backup finished" instead when run with the ```--wait``` flag.


kenkendk:
Looks good!

PR #6648, opened 2025-11-21
Add support for scheme in url for S3 

kenkendk:
This adds support for supplying a url for the hostname which will also toggle the default SSL setting.

Removed providers MyCloudyPlace and dinCloud as they appear to no longer operate.


aureliandevel:
Hi @kenkendk 

This change introduced a regression, at least for my configuration, for which I've submitted #6765.

I'm wondering if the flip in default behaviour for the `--use-ssl` option was intended to be so broadly applicable?  The descriptions above read like the intention is to enable SSL by default when a URL, presumably with http scheme, is supplied.

I haven't restored the previous default behaviour in my PR, however the current behaviour is in direct conflict with the documentation at https://docs.duplicati.com/backup-destinations/standard-based-destinations/s3-compatible-destination.

Issue #5714, opened 2024-11-28
Allow changing the password with the recovery tool


kenkendk:
The recovery tool supports the chain of operations:
download -> decrypt -> decompress -> compress -> encrypt -> upload

But it is lacking an option to supply two different passphrases: old and new.
The current implementation uses the same passphrase for both decrypt and encrypt.

This can be fixed by simply adding an option to input a secondary passphrase and use that for the encryption operation.

PR #6657, opened 2025-12-02
Allow RecoveryTool decompress command to reencrypt with new passphrase

greyxr:
A change to allow the recovery tool to optionally take a new passphrase parameter for use with reencryption. If no new passphrase is provided, the tool will use the old passphrase as usual. This is intended to fix [this issue](https://github.com/duplicati/duplicati/issues/5714).


kenkendk:
Thanks @greyxr !

It looks great.

PR #6701, opened 2026-01-11
Add Windows folder icon overlay for backup status

### LINKED ISSUE #6450: Folder indicator of backup status
Hey, this is a request from a user, not a programmer, so I hope it really is as easy as I think it is:

It would be cool if on windows, the folder Icon could reflect whether the folder is backed up. This idea came as I was setting duplicati up for others, and since I decide on the folders to back up, they kind of have no way of knowing what folder I chose. 

I know that this could be solved by teaching them how the program works, but also the green checkmark indicator that other drive storage solutions (e.g. google drive) have that shows when the synchronisation is done could be cool if the backup worked as scheduled for that folder.

These are what I mean: 

Image 

Image

AndyBodnar:
Hey, this adds a folder indicator feature that was requested in issue #6450. The idea is to show overlay icons on folders in Windows Explorer, similar to what you see with OneDrive or Google Drive when they show sync status.

I looked at how the existing status system works and created a new API endpoint that the shell extension can query. The shell extension is a COM component that Windows loads to display the overlays.

What this does:

The server side has a new /api/v1/folderstatus endpoint that returns the backup status of source folders. It checks if the folder is in any backup, whether the last backup succeeded, had warnings, failed, or is currently running.

The client side is a Windows Shell Extension with icon overlay handlers. When you browse folders in Explorer, it queries the Duplicati server and shows a green checkmark if the folder is backed up successfully, yellow for warnings, red for errors, and a sync icon if backup is in progress.

I added four different overlay handlers because Windows only allows one overlay per handler class. Each handler has a different priority and checks for a specific status.

The shell extension uses a cache with a 30 second expiration to avoid hammering the server. It also skips system folders like Windows and Program Files.

There is a PowerShell registration script included that can register or unregister the COM components. You need to run it as admin.

I did not include the actual icon files yet since those would need to be created by a designer, but I added a README explaining the requirements.

Testing notes:
1. Build the Duplicati.ShellExtension project
2. Run Register.ps1 as administrator
3. Restart Explorer or reboot
4. Folders in your backup sources should show overlay icons

Let me know if you want me to change anything or if there are issues with the approach.

Closes #6450


kenkendk:
Overall I like this PR and think we should merge it.

There are some follow up tasks related to integrating the feature with the MSI package so it can be smoothly installed, but I don't think this should delay merging it.

I do worry a bit about having an unauthenticated API endpoint exposed by default. The implementation looks clean, but just as a precaution, I think this is something that should be enabled, only if needed.

I can come up with some ideas for how to do this, but @AndyBodnar do you have thoughts on how to do this?

AndyBodnar:
Apologies, I have been in disposed yesterday and today. I will look at everything tonight.

TraGicCode:
Hey @kenkendk and @AndyBodnar  ,

Does this work when running duplicati with docker? Does the tray icon functionality have a way to install it independently and it will communicate with a rest api to get this information?

PR #6702, opened 2026-01-11
Implement templating engine for reports

### LINKED ISSUE #6280: Implement templating engine for reports
For all the reports that are supported, it would be nice if we had a templating system (Handlebars, etc).
Then we can ship with standard templates for each supported destination, and allow users to provide custom templates.

AndyBodnar:
This implements a templating engine for reports as discussed in issue #6280. I went with Handlebars.Net since it's a well-established templating library with a simple syntax that most people will find familiar.

The basic idea is that instead of just the current token replacement approach (like %OPERATIONNAME%), users can now write full Handlebars templates with loops, conditionals, and custom formatting. For example, you can iterate over log lines, conditionally show different content based on status, or format file sizes nicely.

I've included a few built-in templates that cover common use cases: a plain text default, an email-friendly format, an HTML version with some basic styling, and a Slack webhook format. Users can also point to their own .hbs files if they want something completely custom.

The implementation adds some helper functions that I think will be useful: formatSize for human-readable byte counts, formatDuration for time spans, formatDate for date formatting, and a few others for conditional logic. These should cover most of what people need for backup reports.

I added unit tests covering the serializer creation, all the helpers, and file loading. Let me know if you'd like me to add any additional test coverage or if there's anything about the API design you'd want to change.

Closes #6280


kenkendk:
Fantastic work 🤩

kenkendk:
@AndyBodnar The flaky test is something we are investigating. It is related to some timing issue when running in the CI.

Issue #6705, opened 2026-01-12
New restore flow on cross-os restores gives index-out-of-bounds


kenkendk:
Looks like the new restore flow needs some tweaking to work correctly when doing cross-OS restores (likely a `/` vs `\` issue).

See forum post for details:
https://forum.duplicati.com/t/restore-failing-on-every-attempt/21786

PR #6709, opened 2026-01-15
Bugfix/cross OS restore

carljohnsen:
This PR adds tests to cover issue #6705, which highlights that there may be a problem with restoring files from a backup created on a different operating system that the one being restored on. 

While the tests pass with the current implementation, the test still provides value in covering this scenario. 

The tests works by downloading a zip folder containing:
1. The database that was created during the backup. 
2. The original files. 
3. The backup created using [the master branch at the time of writing](https://github.com/duplicati/duplicati/commit/6d6168c2962f54e57fa12b8e4c582c04859d3dc5).

There exists a zip file for each of the major operating systems (Linux, MacOS, and Windows). The tests then downloads and extracts each of these zip folders, followed by a restore from each of them with all combinations of the following configurations:
1. With and without using the packed database.
2. With and without using the legacy restore flow. 
3. With and without skipping metadata during the restore. 
resulting in 8 combinations for each of the 3 operating systems, totaling 24 restores.

### DIFF
--- a/Duplicati/UnitTest/Issue6705.cs
+++ b/Duplicati/UnitTest/Issue6705.cs
@@ -142,16 +142,16 @@ public class Issue6705 : BasicSetupHelper
     [OneTimeSetUp]
     public void DownloadAndExtractBackups()
     {
-        bool remote = false; // Set to false to use local files instead of downloading from S3
+        bool remote = true; // Set to false to use local files instead of downloading from S3
         var localZipPath = "D:\\git\\duplicati-carl";
-        var zipPostfix = "_large"; // Set to "" for small backup, "_large" for large backup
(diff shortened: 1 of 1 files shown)


ts678:
While a test is great, how is it a bugfix? GitHub has closed the cited issue.

carljohnsen:
That's on me: I named the branch in my fork as such and didn't think more of it when creating the PR. The name came from the issue, which was a suspected culprit of the problem from the forum post mentioned in the issue. So I thought when I started working on this, that I would be chasing a bug. This test clarifies that the suspected culprit, that the backup came from another OS than the one performing the restore, wasn't correct. Or at least isn't reproducible in its current form. 

However, maybe one could argue that the lack of this test was the bug being fixed? :)

Issue #4710, opened 2022-04-18
Consider migrating to Google Drive API v3


ts678:
- [x] I have searched open and closed issues for duplicates.
- [x] I have searched the [forum](https://forum.duplicati.com) for related topics.

----------------------------------------

## Environment info

- **Duplicati version**: any
- **Operating system**: any
- **Backend**: Google Drive

## Description

[Drive API v2 & v3 comparison guide](https://developers.google.com/drive/api/guides/v3versusv2) links to resources including [Migrate to Drive API v3](https://developers.google.com/drive/api/guides/migrate-to-v3), and asks people to consider migrating.
There seem to be some performance advantages, however it sounds like v2 is still needed for some uses. Not sure if ours is one.
Quite possibly there's no need to rush, but I'll file this in case somebody will do it or if Google decides to kill version 2 someday.

## Steps to reproduce

1. See https://github.com/duplicati/duplicati/blob/master/Duplicati/Library/Backend/GoogleServices/WebApi.cs
2. See https://github.com/duplicati/duplicati/blob/master/Duplicati/UnitTest/WebApiTests.cs

- **Actual result**:
See `https://www.googleapis.com` URL references to `drive/v2`
- **Expected result**:
Seems like v3 use would say v3, however this is just from sampling the documentation on HTTP calls. Far from the last word.

PR #6750, opened 2026-02-05
Handle duplicate source paths

kenkendk:
This PR adds guards to prevent creating filesets with multiple files that have the same path.

While it should technically be impossible to have multiple entries that have the same path, it could happend either due to glitches or because the source data (manual lists, remote sources, etc) returns duplicates.

This PR adds a simple check for each folder to ensure that on a folder-level, duplicate paths cannot be introduced.

There is also a post-backup check to evict any duplicates, and the recreate process will reject duplicate paths.

Finally, the repair command will remove duplicates if they somehow manage to get into the database anyway.

A database-level prevention is not currently feasible as it needs a cross-table check for uniqueness, which requires more work from the database to check for each added file. Since this is expected to be a very rare event, the added processing was not justified.

### DIFF
--- a/Duplicati/UnitTest/DuplicatePathTests.cs
+++ b/Duplicati/UnitTest/DuplicatePathTests.cs
@@ -76,6 +76,8 @@ namespace Duplicati.UnitTest
                     // Use Path.Combine for cross-platform path handling
                     var searchPath = Path.Combine(DATAFOLDER, "duplicate.txt");
                     // Normalize path separators for the database query
+                    // The database stores paths with the system's directory separator
+                    searchPath = searchPath.Replace('/', Path.DirectorySeparatorChar);
                     cmd.CommandText = @"
                         SELECT ""ID"" FROM ""File""
                         WHERE ""Path"" = @Path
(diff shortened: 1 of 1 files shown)


ts678:
> While it should technically be impossible to have multiple entries that have the same path, it could happend either due to glitches or because the source data (manual lists, remote sources, etc) returns duplicates.

The ending contradicts the opening, e.g. for [Google Drive as remote source](https://github.com/duplicati/duplicati/pull/6732), duplicates are an as-designed function AFAIK.

[Manage file metadata](https://developers.google.com/workspace/drive/api/guides/file-metadata)

[Drive API v2 and v3 comparison guide](https://developers.google.com/workspace/drive/api/guides/v3versusv2)

> The name property replaces title in the files resource.

[Consider migrating to Google Drive API v3 #4710](https://github.com/duplicati/duplicati/issues/4710)
which still doesn't sound urgent, but basically current GoogleDrive.cs uses `title` not `name`.

kenkendk:
@ts678 Yes, the Google Drive duplicates was one of the reasons for this change.
In that particular case, the backend filters duplicates before returning anything, so it *should* not happen.

I was testing a new source provider and messed up so it returned a duplicated path and that made everything fail with no easy way to recover (for an end user).

The changes here are added to ensure that **if** it ever happens in production, at least it would only create a somewhat understandable warning instead of breaking the database and making recreate database fail.

PR #6762, opened 2026-02-13
fix(backblaze): honor b2-download-url for download requests

calves07:
## Summary
- restore `b2-download-url` override behavior in Backblaze `GetAsync`
- use the effective download URL in `GetDNSNamesAsync`
- add regression tests for URL resolution precedence

## Root cause
`b2-download-url` support was introduced in `690b4f9041fd2fc285468a43af1a41504f2b4ae0`.
Regression was introduced in `89612753afd63d78d185bdb57bae31d5563c153d`, where download requests switched to `config.DownloadUrl` and bypassed the option.

## Testing
- added `B2DownloadUrlTests`
- not run locally in this environment (`dotnet` not installed)


calves07:
Opened by mistake against upstream; recreated correctly on my fork.

Issue #6764, opened 2026-02-15
[Bug] S3 Backend: Endpoint port stripped and incorrect path validation in Build 2.2.0.102


aureliandevel:
-   [x] I have searched open and closed issues for duplicates.
-   [x] I have searched the [forum](https://forum.duplicati.com) for related topics.

---

## Environment info

-   **Duplicati version**: 2.2.0.102
-   **Operating system**: Windows 10
-   **Backend**: Minio

## Description
A regression in the S3 backend constructor prevents connection to S3-compatible providers (like MinIO) that use custom ports (e.g., host:9000). The current implementation incorrectly validates the hostname and strips the port number before passing the endpoint to the Minio SDK.

## Steps to reproduce
1. Set up a backup to an S3-compatible destination.
2. In the Server field (or --s3-server-name), enter a host without a scheme and with a port: freenas.local:9000 (as was compatible with pre-canary-102 releases).
3. Attempt to "Test Connection" or run a backup.

## Actual Behavior:
Two failures occur:
- Validation Failure: Duplicati throws S3NoPathInEndpoint ("No path allowed in endpoint").
- Execution Failure: If the validator is bypassed (e.g., via http:// prefix), the port is stripped from the hostname. The client attempts to connect to the host on port 80/443, resulting in "Error in XML document (1, 3)" error because it receives an HTML error page instead of an S3 XML response.

## Root Cause Analysis:
In Duplicati/Library/Backend/S3/S3Backend.cs:

Commit a9f3dc1 introduced code in the S3 class constructor to "Check if hostname is actually an URL" by `new System.Uri(hostname)` on raw host:port strings. Without a scheme, .NET parses the host component as the Scheme and the port as the Path, yet still declares the Uri valid. This causes `hosturi.PathAndQuery.Length > 1` to return true even when no path was intended.

After validation, the code sets `hostname = hosturi.Host`. The .Host property in .NET explicitly excludes the port number.

## Proposed Fix:
1. Normalize the `hostname` string before validation by checking for and, if required, adding a scheme before parsing to a Uri.  
2. Assign the parsed `Uri.Authority` property to `hostname` to ensure the port is preserved for the Minio client.

aureliandevel:
Fixed by #6765

PR #6765, opened 2026-02-15
fix(s3): preserve port in endpoint and fix path validation

aureliandevel:
Summary: Fixes S3 endpoint parsing logic regression that was stripping custom ports and failing path validation.

Testing: Verified with BackendTester and list-broken-files against a local MinIO instance.

Fixes Issue #6764


kenkendk:
Thanks @aureliandevel, looks good!

PR #6766, opened 2026-02-15
Feature/Disk image backup and restore

carljohnsen:
This PR introduces a new proprietary module `Duplicati.Proprietary.DiskImage` that enables backup and restore of raw disks. The module exposes disks, partitions, and file systems as a virtual folder hierarchy. 

The current implementation only supports Windows and does not currently provide any file system aware optimizations (such as skipping unused disk sectors). Due to the nature of the operation, privileged execution is required for raw disk access.  

# Source Provider (Backup)
- Exposes physical disks as a virtual folder for backup. 
- Automatically detects and parses Master Boot Record (MBR) and GUID Partition Table (GPT) partition tables. 
- Enumerates partitions and their file systems as folders. 
- Generates a `geometry.json` metadata file containing:
  - Disk geometry (size, sector size, device path).
  - Partition table information (type, size, etc.)
  - Partition details (number, type, offset, size, file system, GUID)
  - File system information (type, partition association, file system specific details)

To perform a backup of the first drive on the system:
```
Duplicati.CommandLine.exe backup file://backup/path "@diskimage://\\.\PhysicalDrive0"
```
This will create a backup, which looks like this:
```
root/
├── geometry.json                    # Consolidated disk/partition/filesystem metadata
├── part_GPT_1/                      # Partition 1 (GPT)
│   ├── fs_Unknown/                  # Unknown filesystem
│   │   └── 0000000000000000         # First 1 MB block of the first partition
│   │   └── 0000000000100000         # Second 1 MB block of the first partition
│   │   └── ...         
│   └── partition.bin                # Raw partition data (if applicable)
└── part_GPT_2/                      # Partition 2 (GPT)
    └── ...
```
Note that if the disk is online, mounted, and in use, the data read may be stale. Future improvement would be utilizing Volume Shadow copy Service (VSS) for a stable read. 

# Restore Provider (Restore)
- Restores disk images back to physical disks. 
- Reconstructs partition tables from the geometry data. 
- Validates target disk size prior to restoring (can be disabled). 
- Optionally, skip restoring the partition table.
- When restoring a GPT partitioned disk to a target disk that is larger than the source disk from where the backup originated, the secondary GPT header will be correctly written to the end of the disk. 

To restore back to a disk image:

```
Duplicati.CommandLine.exe restore file://backup/path --restore-path="@diskimage://\\.\PhysicalDrive1"
```

Note that the disk has to have all of its volumes unmounted, has to be offline, and has to have its read-only attributes cleared. This can be done using `diskpart` (assuming that disk number is `X`, with volume numbers `Y` and `Z` mounted to `E:\` and `F:\`):
```
select disk X
select volume Y
remove letter E
select volume Z
remove letter F
select disk X
offline disk
attributes disk clear readonly
```

Or in Powershell:
```powershell
$diskNumber = X
Get-Partition -DiskNumber $diskNumber |
Where-Object DriveLetter -ne $null |
ForEach-Object {
    Remove-PartitionAccessPath -DiskNumber $diskNumber `
                               -PartitionNumber $_.PartitionNumber `
                               -AccessPath "$($_.DriveLetter):\"
}
Set-Disk -Number $diskNumber -IsOffline $true
Set-Disk -Number $diskNumber -IsReadOnly $false
```

# New options

- `--diskimage-restore-auto-unmount` : boolean flag (default `false`) for toggling whether to automatically try to unmount the disk for writing (which is a requirement for Windows).
- `--diskimage-restore-skip-partition-table` : boolean flag (default `false`) for toggling whether to write the partition table during disk image restore.
- `--diskimage-restore-validate-size` : boolean flag (default `true`) for toggling whether to validate the target disk size prior to restoring. 

# Testing 

The PR also provides unit tests for the feature, although they are skipped during CI due to the privileged execution requirement. To run these tests on a Windows machine as an Administrator:

```
dotnet test --filter "DiskImage"
```

### DIFF
--- a/proprietary/DiskImage/WebModule.cs
+++ b/proprietary/DiskImage/WebModule.cs
@@ -199,7 +199,9 @@ $diskInfo | ConvertTo-Json -Depth 4
             int.TryParse(drive.Path.AsSpan("\\\\.\\PHYSICALDRIVE".Length), out number);
         }
 
-        var displayPath = Util.AppendDirSeparator(drive.Path);
+        // For now, we only allow picking full disks, not individual partitions 
+        // var resultpath = Util.AppendDirSeparator(drive.Path);   
+        var resultpath = drive.Path.TrimEnd(Path.DirectorySeparatorChar);
         var metadata = new Dictionary<string, string?>
         {
             { "diskimage:Number", number.ToString() },
@@ -209,7 +211,7 @@ $diskInfo | ConvertTo-Json -Depth 4
             { "diskimage:Name", $"{drive.DisplayName} ({Library.Utility.Utility.FormatSizeString(drive.Size)})" },
         };
 
-        result[displayPath] = JsonSerializer.Serialize(metadata);
+        result[resultpath] = JsonSerializer.Serialize(metadata);
     }
 
     private async Task<string> RunPowerShellAsync(string script, CancellationToken cancellationToken)


Duplicati/Library/Main/Operation/BackupHandler.cs
```diff
@@ -156,23 +156,27 @@ private static async Task<List<ISourceProvider>> GetSourceProviders(IEnumerable<
...
+                                    mountpoint = Path.GetFullPath(mountpoint);
+                                    mountpoint = Util.AppendDirSeparator(mountpoint);
+                                }
                                 var backendurl = m.Groups["url"].Value;
 
                                 ISourceProvider provider;
                                 try
                                 {
-                                    provider = await SourceProviderLoader.GetSourceProvider(backendurl, Path.GetFullPath(mountpoint), options.RawOptions, cancellationToken).ConfigureAwait(false);
+                                    // TODO double check that all providers that require a mountpoint throws correctly when mountpoint is not set.
```
kenkendk: This TODO should be resolved (I think it is?)

Duplicati/UnitTest/DiskImage/IDiskImageHelper.cs
```diff
@@ -0,0 +1,135 @@
...
+
+        /// <summary>
+        /// Creates a partition on the specified disk, formats it with the specified filesystem,
+        /// and assigns a drive letter or mount point.
+        /// </summary>
+        /// <param name="diskNumber">The disk number.</param>
+        /// <param name="fsType">The filesystem type ("ntfs", "fat32", "ext4", etc.).</param>
+        /// <param name="sizeMB">The size of the partition in MB (0 for all available space).</param>
+        /// <returns>The assigned drive letter (Windows) or mount point identifier.</returns>
+        char CreateAndFormatPartition(int diskNumber, string fsType, long sizeMB = 0);
```
kenkendk: Should FSType be an enum?

Duplicati/UnitTest/DiskImage/IDiskImageHelper.cs
```diff
@@ -0,0 +1,135 @@
...
+        /// <param name="driveLetter">The drive letter or mount point identifier.</param>
+        void FlushVolume(char driveLetter);
+
+        /// <summary>
+        /// Populates a drive with test data including files of various sizes and directory structures.
+        /// </summary>
+        /// <param name="driveLetter">The drive letter to populate.</param>
+        /// <param name="fileCount">The number of files to create.</param>
+        /// <param name="fileSizeKB">The size of each file in KB.</param>
+        void PopulateTestData(char driveLetter, int fileCount = 10, int fileSizeKB = 10);
```
kenkendk: I don't like that test methods are part of the public interface

proprietary/DiskImage/Disk/Interfaces.cs
```diff
@@ -0,0 +1,120 @@
...
+using System.IO;
+using System.Threading;
+using System.Threading.Tasks;
+
+namespace Duplicati.Proprietary.DiskImage.Disk;
+
+/// <summary>
+/// Low-level disk access interface for raw disk reading and writing.
+/// </summary>
+public interface IRawDisk : IDisposable
```
kenkendk: Should this be internal?

proprietary/DiskImage/SourceItems/GeometrySourceEntry.cs
```diff
@@ -0,0 +1,88 @@
...
+        await Task.CompletedTask;
+        yield break;
+    }
+
+    public override async Task<Dictionary<string, string?>> GetMinorMetadata(CancellationToken cancellationToken)
+    {
+        var metadata = await base.GetMinorMetadata(cancellationToken);
+
+        // Add geometry-specific metadata
+        metadata["diskimage:Type"] = "geometry";
```
kenkendk: Since this is also in the contents, is there any reason to replicate it here?

Duplicati/Library/Main/Operation/RestoreHandler.cs
```diff
@@ -713,6 +726,236 @@ await PatchWithBlocklist(database, blocks, m_options, m_result, m_blockbuffer, m
             m_result.EndTime = DateTime.UtcNow;
         }
 
+        /// <summary>
```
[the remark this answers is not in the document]
carljohnsen: No, because the legacy flow is volume centric and the priority files MUST be restored before any of the other files. I have, however, modified it so that instead of methods that essentially implement the same functionality, the core of the legacy flow is hoisted into its own method, which is then called twice. So the legacy flow now consist of two phases: 1 for the priority files, and 2 for the remainder. This has been resolved in [this commit](https://github.com/duplicati/duplicati/pull/6766/commits/4d1e388e3365cfe366f4e65b2afa66d6979fceac).

Duplicati/UnitTest/DiskImage/IDiskImageHelper.cs
```diff
@@ -0,0 +1,135 @@
...
+        /// </summary>
+        /// <param name="imagePath">The path to the disk image file.</param>
+        /// <returns>The disk number/index, or -1 if not found.</returns>
+        int GetDiskNumber(string imagePath);
+
+        /// <summary>
+        /// Initializes a disk with the specified partition table type (GPT or MBR).
+        /// </summary>
+        /// <param name="diskNumber">The disk number.</param>
+        /// <param name="tableType">The partition table type ("gpt" or "mbr").</param>
```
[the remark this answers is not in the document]
carljohnsen: Yes! Resolved in [this commit](https://github.com/duplicati/duplicati/pull/6766/commits/ca098744a88247875b2ebc5754cd9feb22218beb).

proprietary/DiskImage/SourceItems/GeometrySourceEntry.cs
carljohnsen: Yes, there is some merit: when storing the metadata in the database, the data will also be present there, which is nice for debugging. Compared to a disk, this extra information should be very small, so storing it twice should have very little effect.

proprietary/DiskImage/SourceItems/PartitionTableSourceEntry.cs
```diff
@@ -0,0 +1,82 @@
...
+
+/// <summary>
+/// Represents the partition table as a source entry for backup/restore purposes.
+/// This allows the partition table (MBR or GPT) to be backed up and restored separately.
+/// </summary>
+internal class PartitionTableSourceEntry(string parentPath, IPartitionTable table, IRawDisk disk)
+    : DiskImageEntryBase(System.IO.Path.Combine(parentPath, $"partition_table_{table.TableType}"))
+{
+    public override bool IsFolder => false;
+    public override bool IsMetaEntry => true;
```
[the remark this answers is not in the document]
carljohnsen: Correct, it is stored in the backup, but it's in the `geometry.json` file, so this file is purely for keeping track of the internal classes, hence it should be a meta entry.

Issue #6752, opened 2026-02-06
Slow restore speed from Linux SSH/SFTP to Windows 10 + Severe disk write amplification (10T written for 200GB backup, SSD wear risk)


Rivendeller:
-   [Y] I have searched open and closed issues for duplicates.
-   [Y] I have searched the [forum](https://forum.duplicati.com) for related topics.

---

## Environment info

-   **Duplicati version**: 2.2.0.3
-   **Operating system**: Windows 10 Enterprise 22H2 
-   **Backend**: Replace me

## Description

I have been using Duplicati to back up 200GB of data to a Linux server via SSH/SFTP for a long time. Recently, I tried to restore this backup to a Windows 10 client, but two critical issues occurred:

**1. Extremely slow restore speed**: The speed is far lower than the hardware/network capacity (network/disk write/SSH direct transfer are all normal), no obvious bottlenecks found.

**2. Severe disk write amplification**: During the restore process, Duplicati wrote 10TB of data to my Windows 10 SSD, while the actual backup size is only 200GB (50x write amplification). This abnormal high disk write volume will rapidly consume the SSD's TBW (Total Bytes Written) and directly lead to SSD failure for large backup restores in the future.

I have thoroughly ruled out common bottlenecks (network bandwidth, disk write speed, SSH transfer, system resource usage), but both issues persist. I urgently need to locate the root cause, get optimization solutions, and fix the write amplification problem to avoid SSD hardware damage.

## Steps to reproduce

1. Replace me
2. Replace me
3. ...

-   **Actual result**:
    Replace me
-   **Expected result**:
    Replace me

vasiby:
looks smilar to my issue https://github.com/duplicati/duplicati/issues/6744, I burnt 30TB on a 90gb restore

Rivendeller:
Duplicati 2.1.0.5: Duplicati 2.1.0.5 is fully working — When restoring the same 200GB backup (from the same Linux SSH/SFTP storage to the same Windows 11 client), restore speed is normal (matches SFTP direct transfer speed), and there is no abnormal disk write amplification (SSD write volume is approximately equal to the actual backup size, ~200GB). This version runs stably without any of the aforementioned issues.

ts678:
> This version runs stably without any of the aforementioned issues.

Here's how to keep using that, from CommandLine help for `restore`:

```
  --restore-legacy=<bool>
    Use this option to use the legacy restore method. The legacy
    restore method is slower than the new method, but may be more
[...]

Issue #6744, opened 2026-02-01
RESTORE causes constant writes to the TEMP directory and 100% SSD usage writing dup files.


vasiby:
-   [x] I have searched open and closed issues for duplicates.
-   [x] I have searched the [forum](https://forum.duplicati.com) for related topics.

---

## Environment info

-   **Duplicati version**: 2.2.0.3_stable
-   **Operating system**: windows 11
-   **Backend**: local directory on ssd

## Description

Constant writes to the TEMP directory causing 100% SSD usage performing terabytes of write IO (dup- files being overwritten and evicted), also the CPU is busy constantly decrypting files,

## Steps to reproduce

1. I have a repository of code projects of around 60GB with 1.5 million files backed up (node_modules contribute to the count mostly), and dblock file size is 100MB, about 20 versions since 2023. Backup is encrypted.
2. Restore initiated  on a fresh computer, backup files are on SSD connected via USB.
3. Database recreation successful.
4. During the actual file restoration process the system SSD with the temp directory reports 100% usage doing constaint writes.
5. Resource monitor reports it's writind dup-files constantly
6. After some time 3GB is left to be restored with 1.4million small file
7. Duplicati is in a loop copying the block file from SSD, decrypting into temp (writing constantly), evicting the file
8. Because there's a million small files it seems to download at least 1 block file of 100m for each file that takes insane amount of write wear  and cpu usage decrypting it. Speed was 300mb/s writes to the temp (terabytes), 1 file per second (and there's a million files)
9. After 10TB of temp files has been written and 1.3m files femainig (3GB) I shut down the restore and moved the TEMP into a ramdisk.
10. Does not help, the temp dir on ramdisk (20GB) never grows above 10GB usage and still constandly "downloads" the files from the usb SSD and decrypts them in a loop. Still 1 file per second restore and constant GETs from the source.
11. After 3 days of attempts to reduce the SSD writes and restore speed, the only thing that helped was the `restore-volume-cache-hint` parameter and setting it to 500Gb to be able to hold the entire backup and avoid evictions. TEMP was also moved to a SSD that could hold everything.

-   **Actual result**:
    Small files in backup causes download->decrypt->store in TEMP->restore->evict loop.
-   **Expected result**:
    Downloadeded dblock files are only stored once in TEMP  and not evicted if they are needed for future file restoration. Maybe put the needed file chunks into the Restore directory directly and clean them up at the very end. Just don't evict and  redownload needlessly.

[...]

PR #6785, opened 2026-02-28
Enhance volume caching with disk space pressure eviction logic

aureliandevel:
### Summary

Previously, when `--restore-volume-cache-hint` was not set, the restore volume cache defaulted to 100× the volume size (e.g. 5 GB for 50 MB volumes) — an arbitrary limit that could be far too large or far too small depending on the environment. This PR removes that default, making the cache unlimited by default and enabling disk-space-aware LRU eviction bounded by a new `--restore-volume-cache-min-free` option.

### Behaviour changes

| Scenario | Before | After |
|---|---|---|
| `--restore-volume-cache-hint` not set | Hint at `VolumeSize × 100` bytes | Unlimited; evicts LRU when temp dir free space drops below `--restore-volume-cache-min-free` |
| `--restore-volume-cache-hint 0` | Caching disabled | Unchanged |
| `--restore-volume-cache-hint <size>` | Hint at `<size>` | Unchanged |

### New option

**`--restore-volume-cache-min-free`** (default: `1gb`)

Minimum free disk space to maintain in the temp directory during restore. When the unlimited cache mode is active, LRU eviction triggers whenever available free space drops below this value. A bare number is interpreted as gigabytes. Has no effect when `--restore-volume-cache-hint` is set to a specific size or to `0`.

### Implementation details

- `RestoreVolumeCacheHint` now returns `-1` when unset (sentinel for "unlimited"), cleanly separating unlimited (`< 0`), disabled (`0`), and hinted (`> 0`) modes.
- `VolumeManager` extends the existing cache insertion block to a three-way branch. The unlimited branch queries `DriveInfo.AvailableFreeSpace` on the temp directory on each volume arrival and evicts LRU entries until free space recovers. The volume is then cached unconditionally (it is already on disk; disposing it would not help).
- If a volume has to be re-downloaded after being evicted due to disk pressure, it is counted in `CachePressureRedownloads`.

### Observability

Two new warnings (emitted at most once each per restore operation):

- **`CachePressure`** — emitted after the 5th disk-pressure eviction: _"Restore volume cache has begun evicting cached volumes due to low disk space in '{tempDir}'. Restore performance may be degraded."_
- **`CacheExhausted`** — emitted if the cache is entirely empty but free space is still below the minimum: _"Restore volume cache is empty but disk space in '{tempDir}' is still below the configured minimum. Performance impact is likely."_

Three new fields on `RestoreResults` (omitted from all output when zero via `ShouldSerialize*()`):

- `CachePressureEvictions` — total LRU evictions triggered by disk pressure
- `CachePressureRedownloads` — volumes re-downloaded after disk-pressure eviction
- `TotalVolumesAccessed` — distinct dblock volumes accessed during the restore

The counters and warnings may seem like overkill for typical restore usage — most emergency restores prioritise completion over speed, and disk pressure is unlikely when restoring small datasets. They are primarily aimed at users and automated pipelines that regularly validate their backups and need to know whether volume caching is working efficiently or silently degrading performance. The zero-suppression design (`ShouldSerialize*()`) means these fields cost nothing in the common case and provide actionable signal when it matters.

### Files changed

- `Duplicati/Library/Main/Options.cs` — new constant, updated `RestoreVolumeCacheHint` fallback, new `RestoreVolumeCacheMinFree` property and `SupportedCommands` entry
- `Duplicati/Library/Main/Strings.cs` — updated `RestoreVolumeCacheHintLong`, new `RestoreVolumeCacheMinFreeShort`/`Long`
- `Duplicati/Library/Main/Operation/Restore/VolumeManager.cs` — three-way cache branch, disk-pressure eviction loop, warning logic, result counter writes
- `Duplicati/Library/Main/ResultClasses.cs` — three new properties on `RestoreResults` with `ShouldSerialize*()` zero-suppression
- `Duplicati/UnitTest/OptionsTests.cs` — three new unit tests for the new option properties
- `Duplicati/UnitTest/RestoreHandlerTests.cs` — `null` added to `RestoreVolumeCache` parameterised values; new `RestoreVolumeCacheDiskPressure` integration test

Related to #6752, #6744


aureliandevel:
I'm doing further testing to determine whether this update is even necessary.

I've been looking at ordering of the restore targets to prioritise volume eviction.

My first attempt prioritised high file/volume ratio ordering but surprisingly that yielded almost no improvement.

At this point it appears that highly referenced blocks (i.e. blocks that appear in multiple files as a result of the de-duplication process) is the main culprit of the holding volumes in the cache.  I'm going to try a different sorting algorithm to prioritise high-referenced blocks ... though the test SQL I tried took a pretty long time to complete.

I'll assess whether this PR is worth committing based on the outcome of whether these investigations yield useful results or not.

aureliandevel:
I’m marking this PR as ready for review/check-in.

For those interested, the broader investigation referenced in my previous comment is available here:
- analyser/simulation repo: https://github.com/aureliandevel/duplicati-restore-cache-analyser

Even with that broader work ongoing, I still think this PR is worthwhile on its own merits as a practical improvement to restore cache behaviour under disk-space pressure.

Duplicati/Library/Main/Operation/Restore/VolumeManager.cs
```diff
@@ -230,8 +258,26 @@ void evict_lru()
...
+                                                available_free_space = new DriveInfo(temp_dir).AvailableFreeSpace;
+                                            }
+                                            if (!cache_exhausted_warned && cache_last_touched.Count == 0 && available_free_space < cache_min_free)
+                                            {
+                                                cache_exhausted_warned = true;
+                                                Logging.Log.WriteWarningMessage(LOGTAG, "CacheExhausted", null, "Restore volume cache is empty but disk space in '{0}' is still below the configured minimum. Performance impact is likely.", temp_dir);
+                                            }
+                                            Logging.Log.WriteExplicitMessage(LOGTAG, "VolumeRequest", "Caching volume {0} in unlimited mode (free space: {1})", volume_id, available_free_space);
+                                            cache[volume_id] = volume;
+                                            cache_size += volume.Size;
```
carljohnsen, line 283: This is redundant, since `cache_size` isn't used during unlimited mode? Not too much overhead in keeping the tracking though.

Duplicati/Library/Main/ResultClasses.cs
```diff
@@ -539,6 +539,25 @@ internal class RestoreResults : BasicResults, IRestoreResults
         public long DeletedSymlinks { get; internal set; }
         public string RestorePath { get; internal set; }
 
+        /// <summary>
+        /// Number of LRU evictions triggered by low disk space in the temp directory
+        /// (disk-pressure path, unlimited cache mode only).
+        /// </summary>
+        public long CachePressureEvictions { get; internal set; }
+        public bool ShouldSerializeCachePressureEvictions() => CachePressureEvictions > 0;
```
carljohnsen, line 547: I haven't seen these (`ShouldSerialize...()`) before, and they're not referenced anywhere. Are they required?

carljohnsen (approved):
These changes look good! I like the new default of "if there's space, keep going".

Duplicati/Library/Main/ResultClasses.cs
aureliandevel: (sorry about the comment noise ... really need to do better checking which account I'm using)

ShouldSerialize is basically a callback predicate to check whether the property should be emitted. See https://www.newtonsoft.com/json/help/html/conditionalproperties.htm.

In this case, I didn't want potentially confusing and, in the happy path, values of little use in the general output. These will show up only when there's something of value to report.

Duplicati/Library/Main/Operation/Restore/VolumeManager.cs
aureliandevel: It's still used by explicit messages, even in unlimited mode.

carljohnsen: No, it's only ever used in the `else if (cache_max > 0)` branch. And `cache_max` is based of the options, which doesn't change during an operation. 

Regardless, it's not an issue, I was just thinking maybe some intent was lost. Like how it could be useful as a profiling item, where the maximum cache size (from the volume managers point of view) was reported. So store it in `long cache_size_max_consumed = Max(cache_size_max, cache_size)` and then emit it during the retiring shutdown / cleanup:

```
if (options.InternalProfiling)
{
    Logging.Log.WriteProfilingMessage(LOGTAG, "InternalTimings", $"CacheSet: {sw_cache_set?.ElapsedMilliseconds}ms, CacheEvict: {sw_cache_evict?.ElapsedMilliseconds}ms, Query: {sw_query?.ElapsedMilliseconds}ms, Backend: {sw_backend?.ElapsedMilliseconds}ms, Request: {sw_request?.ElapsedMilliseconds}ms, Wakeup: {sw_wakeup?.ElapsedMilliseconds}ms");
    Logging.Log.WriteProfilingMessage(LOGTAG, "InternalTimings", $"Max used cache size: {Duplicati.Library.Utility.Utility.FormatSizeString(cache_size_max_consumed)}"); 
}
```

Duplicati/Library/Main/ResultClasses.cs
carljohnsen: Ah, I see, I thought it might have been something like that. No problem then :)

Duplicati/Library/Main/Operation/Restore/VolumeManager.cs
aureliandevel: Yes, apologies for that.  You're correct, of course; I was looking at a different branch and, after realising my mistake, didn't feel like spamming you further.

That said, I like your idea of reporting it in the retiring message run; should have thought of it myself, but it immediately also brings to mind the question of how much was downloaded, as distinct from how much was restored.  If my investigation on volume (re)packing pans out, these may simply end up being informative rather than diagnostic anyway.

I'll create a separate PR for these.

aureliandevel:
Hi @carljohnsen ,

Just noting that I added the profile logging discussed above to this PR, instead of raising a new one as mentioned above.

Duplicati/Library/Main/Operation/Restore/VolumeManager.cs
carljohnsen, line 283: Shouldn't this branch also have the 
```
cache_size_max_consumed = Math.Max(cache_size_max_consumed, cache_size);
```
line added? Or maybe it should just be moved outside of the `if | else if | else` branch? `cache_size` isn't being updated in the `cache_max == 0`, so it shouldn't pollute too much.

Duplicati/Library/Main/Controller.cs
```diff
@@ -169,6 +169,14 @@ await restoreDestination.Finalize((pg) =>
                 result.OperationProgressUpdater.UpdatePhase(OperationPhase.Restore_Complete);
                 result.EndTime = DateTime.UtcNow;
 
+                if (m_options.InternalProfiling)
```
carljohnsen: Instead of guarding it with `InternalProfiling`, I'd say just let the profiling log level handle it. 

The information is already part of the results, should it be a log line as well?

carljohnsen (approved):
Looks good! With the two new comments fixed I'll merge it.

Duplicati/Library/Main/Controller.cs
aureliandevel: Right, yes, I thought this is what you were referring to in an earlier comment: "...  and then emit it during the retiring shutdown / cleanup".  My bad.

Duplicati/Library/Main/Operation/Restore/VolumeManager.cs
aureliandevel: Thanks for catching that.

PR #6786, opened 2026-03-01
Add support for managed TLS

kenkendk:
This PR adds support for generating a self-signed CA and then using that CA to generate TLS certificates.

A new tool `duplicati-configure` / `Duplicati.CommandLine.ConfigureTool.exe` is added to manage the certificates.  The tool saves the configuration in the database and is meant to run with elevated privileges for the initial CA installation.

The option `--configure-https` can be added to server/trayicon, and if the process has permissions, this will automatically install the CA certificate.

If a CA is configured, the server will automatically issue, renew and use a TLS certificate.

With this setup, the TLS is opt-in, but once the flow is well tested, we can switch it to opt-out.

### DIFF
--- a/Duplicati/Library/Certificates/README.md
+++ b/Duplicati/Library/Certificates/README.md
@@ -98,3 +98,53 @@ For high-security environments, consider implementing certificate pinning:
 - Extract the CA certificate thumbprint after generation
 - Configure clients or monitoring systems to expect only this specific CA
 - This prevents acceptance of certificates signed by other CAs that might be installed on the system
+
+## Browser Trust (Chrome and Firefox on Linux)
+
+While the CA certificate is installed in the system trust store, Chrome and Firefox on Linux maintain their own certificate stores and may not automatically trust the system CA. This is especially true for sandboxed installations (Snap, Flatpak).
+
+### Exporting the CA Certificate
+
+To manually import the CA into Chrome or Firefox, first export it using the ConfigureTool:
+
+```bash
+# Export to default filename (duplicati-ca.crt)
+duplicati configure-tool https export-ca
+
+# Or specify a custom output path
+duplicati configure-tool https export-ca --file ~/Desktop/duplicati-ca.crt
+```
+
+### Importing to Firefox
+
+1. Open Firefox and go to **Settings** → **Privacy & Security**
+2. Scroll to **Certificates** and click **View Certificates**
+3. Select the **Authorities** tab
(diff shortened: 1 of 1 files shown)


ts678:
Is ConfigureTool or similar going to help with setting up clients? I hope use cases have been thought through.
I'm not that expert in this, but self-signed seems more painful than using a CA that all the clients already trust.

Some people see Caddy as an example of well-done automatic TLS, though sometimes a domain name helps.
I'm not sure what all enterprises do, but I think sometimes the internal network is a subdomain of public DNS.

kenkendk:
@ts678 The idea is that this adds support for getting HTTPS support without needing manual renewal.

See also the README:
https://github.com/duplicati/duplicati/blob/e3155950951deb3abf66b7dc74852b3dc8dc43d6/Duplicati/Library/Certificates/README.md


> I'm not that expert in this, but self-signed seems more painful than using a CA that all the clients already trust.

A self-signed certificate is precisely a pain because you need to manually force the browser to trust it. 

With this PR, the logic is to create a CA and ensure that CA is trusted. 
Since Duplicati created the CA, it knows the private key, and can issue certificates, similar to how this works for LetsEncrypt and other systems.

The CA is self-signed, because we generate it on the client and cannot get an established CA to sign it, but this does not affect the browser or OS trust because we import it into the trust store.

On Windows this will give a dialog asking to approve it when running with user credentials (CA is trusted for current user only). When running with Administrator privileges, this will not ask for permissions and will make the CA trusted for all users on the machine.

> I'm not sure what all enterprises do, but I think sometimes the internal network is a subdomain of public DNS.

Since Duplicati is not normally exposed to the network, it does not matter what domain the host has.
It is possible to generate extra hostnames if you *want* to expose Duplicati to the network, but by default this is not done.
If you expose Duplicati to the network, the CA needs to be manually added as a trusted CA on the other machines that need to access it.


> Is ConfigureTool or similar going to help with setting up clients? I hope use cases have been thought through.
> Some people see Caddy as an example of well-done automatic TLS, though sometimes a domain name helps.

The ConfigureTool is designed to be its own system and does not integrate with other systems.
If you want to use another tool or a certificate signed by an established CA, this is still possible.
But since Duplicati does not have access to the CA private key, it cannot issue renewed certificates and will just blindly serve the certificate, even if it is expired.

There is an API endpoint that allows updating the server certificate, and this PR will also make Duplicati serve the new certificate without requiring a restart.

PR #6802, opened 2026-03-14
Fix restore without local db

### LINKED ISSUE #6637: "volumes with missing filesets" error on no-DB CLI restore or repair with version

---

## Environment info

-   **Duplicati version**: 2.2.0.1_stable_2025-11-09 and likely back to 2.1.0.111_canary_2025-03-15
-   **Operating system**: N/A
-   **Backend**: N/A

## Description

Repair has `--version` to pick versions. CLI restore done in what GUI would call "Direct restore" selects version.
[Added additional consistency checks to local database](https://github.com/duplicati/duplicati/commit/d500097e053e157ff9d29ca1df817e7e2d19ee4e) added a consistency check between `dlist` and Fileset.
Problem can be that anything less than all versions can leave some dlist volumes with no Fileset, thus an error:

```
ErrorID: DatabaseInconsistency
Detected # volumes with missing filesets
```

## Steps to reproduce

1. Make two version backup
1. Delete DB
1. GUI or CLI DB repair one version, or CLI restore. 

-   **Actual result**:
volumes with missing filesets
-   **Expected result**:
works

Especially for restore, logs help because restore seems to probe for dbpath. If missing, it uses temporary file.
Repair when run as recreate (no existing DB) at least leaves DB in a known spot where result can be checked.

[Restore with duplicati-cli fails](https://forum.duplicati.com/t/restore-with-duplicati-cli-fails/21573) noticed CLI/GUI difference. For me, GUI seems to put all dlist versions into DB.
This might be why it works, but there's also mystery of a different topic where restore was from "config file":
[Error Detected 1 volume with missing filesets: VolumeId = 32, Name = duplicati-20251104T211541Z.dlist.zip.aes, State = Uploaded](https://forum.duplicati.com/t/error-detected-1-volume-with-missing-filesets-volumeid-32-name-duplicati-20251104t211541z-dlist-zip-aes-state-uploaded/21548/20?u=ts678)

EDIT 1:

Proposed workaround for CLI restore was DB recreate with all versions, then restore desired version with that.
One might say that's more steps and more DB building, but GUI does the full DB build, not just selected date.
Or so it appears from my testing with logs, but this has been a sprawling test, and I may have made mistakes.

kenkendk:
This fixes a bug where restoring with `--no-local-db` and `--version=0` would fail because the database consistency check sees that data is missing.

The missing data is intentional in this case, as we deliberately recreate a partial database.

The PR adds a test for this scenario and fixes the validation to allow for the partial database. Additionally, the partial database is marked as such, in case it is not correctly deleted after the restore.

This fixes #6637

### DIFF
--- a/Duplicati/Library/Main/Operation/RestoreHandler.cs
+++ b/Duplicati/Library/Main/Operation/RestoreHandler.cs
@@ -103,6 +103,7 @@ namespace Duplicati.Library.Main.Operation
 
         public static RecreateDatabaseHandler.NumberedFilterFilelistDelegate GetNumberedFilelistFilterDelegate(DateTime time, long[] versions, bool singleTimeMatch = false)
         {
+            versions ??= [];
             if (versions.Length == 0 && time.Ticks == 0)
                 return null;
 


ts678:
["volumes with missing filesets" error on no-DB CLI restore or repair with version #6637](https://github.com/duplicati/duplicati/issues/6637) seems related.

PR #6814, opened 2026-03-23
Feature/Full disk FAT32 support

carljohnsen:
# FAT32 File System Support and Incremental Backup Optimizations

This PR extends the `DiskImage` module (introduced in #6766, #6784, and #6799) by adding file system awareness for FAT32 partitions. While previous iterations focused on raw block access, this update enables the module to parse FAT32 boot sectors and directory structures.

This allows for file system aware backups, enabling significant performance improvements during incremental backup strategies and unallocated space optimization.

## Main Changes

- **FAT32 Support:** Added a FAT32 boot sector parser and directory walker to interpret file system metadata.
- **Incremental Backup:** Blocks are now assigned timestamps based on the FAT32 file system metadata. If a block has not been modified since the last backup, it is skipped during subsequent backup runs.
- **Zero-Fill Optimization:** Unallocated blocks (clusters marked as free) are treated as zeros. Instead of reading these sectors from the disk, a pre-allocated zero buffer is used to write them to the backup. This can be circumvented with the `--diskimage-filesystem-unknown` option to force raw block access if desired.

## Implementation Details

- **FAT32 Parser:** Parses the boot sector to determine cluster size and file allocation table (FAT) location.
- **Directory Walker:** Traverses the FAT32 directory structure to map file names to cluster addresses.
- **Block Boundary Checking:** Added rigorous stream boundary checking to ensure unaligned reads and writes are handled correctly across all supported platforms.

## New Options

- `--diskimage-filesystem-unknown`: boolean flag (default `false`) for toggling whether to treat the filesystem as unknown.
    - When enabled, forces the backup process to use raw block-based access instead of filesystem-aware backup.
    - Useful for testing scenarios where the file system type is not detected or supported.
    - Useful for backups that want to back up everything, including soft-deleted files or data not tracked by the file system metadata.

## Bugfixes / improvements

- **Timestamp Handling:** Fixed an issue where unknown filesystems did not handle consecutive backups correctly. Previously, blocks always had the same date, causing the backup process to skip unchanged blocks incorrectly or fail to detect changes. Blocks on an unknown filesystem now report `UtcNow` if no valid timestamp is available, ensuring they are backed up on the next run.
- **Disk Image Unit Test refactoring:** Refactored the disk image unit tests into separate files for better maintainability and readability.

## Testing

The PR includes comprehensive unit and integration tests for FAT32 functionality:

- Tests for FAT32 boot sector parsing.
- Tests for the FAT32 directory walker and file system metadata extraction.
- Incremental backup integration tests to verify timestamp-based skipping logic.
- Tests for unaligned read and write operations on FAT32 volumes.

Some tests have been marked with the `DiskImageLocal` category and are excluded from CI runs to reduce execution time, similar to previous iterations. These should be run locally on Windows, macOS, and Linux for validation:

```bash
dotnet test --filter "DiskImage"
```


kenkendk (approved):

ts678:
> Instead of reading these sectors from the disk

This might help reduce my concern of defragmentation leaving old data behind during file moves.
This would lead to Duplicati block-level deduplication (good), but maybe clog restore cache (bad).

PR #6825, opened 2026-04-04
Fix persisted write-path option restore for maintenance and recompress

### LINKED ISSUE #6824: Persisted backup write-path options are not fully restored for maintenance and recompress operations

---

## Environment info

-   **Duplicati version**: 2.2.0.107 and earlier
-   **Operating system**: Windows
-   **Backend**: local / file backend

## Description

When an operation targets an existing backup with a local database, Duplicati does not consistently restore the backup's persisted write-path configuration before performing maintenance or rewrite work.

This affects settings that determine how new remote volumes are identified or written, specifically:

-   `prefix`
-   `dblock-size`
-   `compression-module`
-   `encryption-module`
-   unencrypted state

The result is that some operations can behave as though default values were intended when the original backup was created with different settings.

There is also a related issue in `RecoveryTool recompress`: when the tool is run with `--reupload=true`, it can update the remote set while leaving the local DB's stored backup options stale. In addition, if a backup was created with a non-default prefix and that prefix is omitted on the command line, the tool can fail to identify the correct remote set even though the local DB already knows the prefix.

## Steps to reproduce

1. Create a backup using a non-default write-path configuration, for example:
   - a custom `prefix`
   - a non-default `dblock-size`
   - explicit compression and/or encryption settings
   - or an unencrypted backup
2. Let the backup create a local database.
3. Run a later operation against the same backup using a minimal option set, for example:
   - `compact` against the existing DB while omitting the original write-path options
   - or `RecoveryTool recompress --reupload=true --dbpath=...` while omitting a custom `--prefix`
4. Inspect the newly written remote files and/or the stored DB options after the operation.

-   **Actual result**:
    The operation does not always reuse the original persisted write-path configuration from the local DB. Newly written remote volumes can use the wrong prefix or other write-path settings, and `RecoveryTool recompress` can leave the local DB out of sync with the rewritten remote set.

-   **Expected result**:
    For an existing backup, Duplicati should restore the persisted write-path configuration from the local DB before writing new remote volumes, unless the user explicitly overrides those values on the command line. `RecoveryTool recompress` should also update the local DB after a successful full rewrite/reupload.

## Screenshots

N/A

## Debug log

No production debug log attached.

This was reproduced and validated from code-level investigation and focused regression coverage. The issue appears to be in how persisted backup options are restored and updated, rather than in a backend-specific implementation.

## Related issues

This appears related to the broader class of bugs where Duplicati does not always use enough persisted state to correctly identify or preserve an existing backup's remote layout:

-   #2678
-   #2952
-   #3416
-   #3845
-   #4244

Issue #5085 does not appear to be the same problem. That one is a RecoveryTool restore path-mapping bug involving `--targetpath`, not persisted backup option restore/update.

## Related forum discussions

There are several forum discussions that appear to be in the same general bug family, where Duplicati acts on incomplete or stale local knowledge of an existing backup:

- https://forum.duplicati.com/t/some-observation-about-database-maintenance/9922
- https://forum.duplicati.com/t/best-way-to-kick-off-new-backups/14747
- https://forum.duplicati.com/t/backup-to-the-same-bucket-folder-should-raise-a-warning/4239
- https://forum.duplicati.com/search?q=change%20prefix%20existing%20backup

These discussions mostly focus on repair/recreate/prefix ownership hazards. The RecoveryTool `recompress` local-database synchronization problem appears to be a related but less explicitly documented variant.

aureliandevel:
## Summary

Fixes #6824.

This change makes Duplicati persist and restore the backup's write-path configuration more consistently when operating on an existing backup.

The fix covers:

- restoring persisted write-path options from the local DB before maintenance operations need them
- preserving CLI-over-DB precedence
- including prefix in the persisted/restored option contract
- updating RecoveryTool recompress to restore DB-backed options early enough to identify the correct remote set
- updating the local DB after a successful full recompress reupload

## Tests

Validated with focused unit coverage for:

- utility-level persisted option restore/update
- compact preserving stored write-path settings when CLI omits them
- RecoveryTool recompress updating local DB state
- RecoveryTool recompress honoring stored custom prefix
- early failure for mixed encryption-module remote sets before backend mutation


kenkendk:
@aureliandevel I am good with the changes here, but I don't understand the reason behind getting `recompress` to update the settings in the database as well.

If you change the encryption mode, you also need to change all filenames in the database for this to work. And in this case there is no need to update the database as it needs to be rebuilt anyway.

Could you explain what the use-case for this change is?

aureliandevel:
@kenkendk,

You’re absolutely right.  I missed the fact that local DB recreation isn't an automatic part of the recompress flow.

The part that is still needed is the early read from the local DB, if it exists, so recompress can pick up the persisted write-path settings before it starts.

I’ve removed the post-reupload DB update logic and pushed the updated code.

PR #6826, opened 2026-04-04
Fix restore volume handling in disabled-cache mode

aureliandevel:
## Summary

This PR fixes restore volume handling in the new restore process when `restore-volume-cache-hint=0b` is used.

In that mode, a downloaded volume could be dropped too aggressively even though pending block work still depended on it. The result was that disabled-cache restores could stall or thrash instead of simply avoiding normal cache retention.

## What this changes

When volume caching is disabled:

- a downloaded volume is still kept alive while pending block work references it
- the volume is still excluded from normal reusable LRU cache retention
- once no remaining blocks depend on that volume, it is explicitly evicted as normal

This preserves the intended meaning of disabled-cache mode while avoiding premature disposal of volumes that are still in active use.

## Additional consistency fix

While validating the disabled-cache fix, test coverage also exposed an inconsistency in bounded-cache eviction bookkeeping.

That adjustment is included here because it is part of the same `VolumeManager` ownership and eviction logic:

- LRU eviction must reduce the tracked cache size consistently
- otherwise bounded-cache restores can stop making forward progress once eviction begins

## How this was found

This was found while validating restore volume cache behaviour with the existing `RestoreVolumeCache` tests.

The failing patterns were cache-mode-specific:

- `restore-volume-cache-hint=0b` exposed the disabled-cache handling bug
- bounded-cache test cases then exposed an eviction-accounting inconsistency during follow-up validation

## Testing

Validated with:

- `RestoreVolumeCache` test matrix
- targeted reruns while verifying disabled-cache and bounded-cache behaviour

## Notes

This PR is intended as a focused fix for restore volume lifecycle and eviction handling in `VolumeManager`.

### DIFF
--- a/Duplicati/Library/Main/Operation/Restore/VolumeManager.cs
+++ b/Duplicati/Library/Main/Operation/Restore/VolumeManager.cs
@@ -133,9 +133,9 @@ namespace Duplicati.Library.Main.Operation.Restore
                         cache.Remove(volume_id, out var volume);
                         if (volume != null)
                         {
-                            cache_size -= volume.Size;
+                            if (cache_last_touched.Remove(volume_id))
+                                cache_size -= volume.Size;
                             volume.Dispose();
-                            cache_last_touched.Remove(volume_id);
                         }
                         sw_cache_evict?.Stop();
                     }
@@ -147,7 +147,6 @@ namespace Duplicati.Library.Main.Operation.Restore
                         {
                             // Pop the last element of cache_last_touched
                             var volume_id = cache_last_touched[0];
-                            cache_last_touched.RemoveAt(0);
                             handle_evict(volume_id);
                         }
                         sw_cache_lru?.Stop();
@@ -180,8 +179,8 @@ namespace Duplicati.Library.Main.Operation.Restore
                                                 sw_request?.Start();
                                                 if (cache.TryGetValue(request.VolumeID, out var volume))
                                                 {
-                                                    cache_last_touched.Remove(request.VolumeID);
-                                                    cache_last_touched.Add(request.VolumeID);
+                                                    if (cache_last_touched.Remove(request.VolumeID))
+                                                        cache_last_touched.Add(request.VolumeID);
                                                     Logging.Log.WriteExplicitMessage(LOGTAG, "VolumeRequest", "Block {0} found in cache", request.BlockID);
                                                     volume.Reference();
                                                     await self.DecompressRequest.WriteAsync((request, volume)).ConfigureAwait(false);
@@ -211,7 +210,13 @@ namespace Duplicati.Library.Main.Operation.Restore
                                     {
                                         sw_cache_set?.Start();
                                         volume.Reference(in_flight_downloads[volume_id].Count);
-                                        if (cache_max > 0)
+                                        if (cache_max == 0)
+                                        {
+                                            // Caching disabled — keep the volume only while blocks still reference it.
+                                            Logging.Log.WriteExplicitMessage(LOGTAG, "VolumeRequest", "Keeping active volume {0} available until all pending blocks are processed (caching disabled)", volume_id);
+                                            cache[volume_id] = volume;
+                                        }
+                                        else if (cache_max > 0)
                                         {
                                             // Check if adding another volume would exceed cache limits.
                                             while (cache_size > 0 && (cache_size + volume.Size) > cache_max)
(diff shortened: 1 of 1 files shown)


Duplicati/Library/Main/Operation/Restore/VolumeManager.cs
```diff
@@ -133,9 +133,9 @@ void handle_evict(long volume_id)
                         cache.Remove(volume_id, out var volume);
                         if (volume != null)
                         {
-                            cache_size -= volume.Size;
+                            if (cache_last_touched.Remove(volume_id))
```
carljohnsen, line 136: I can see how the `cache_size` will drift when `cache_max == 0`. I previously discarded it, because the variable wasn't used in that case. Neither is `handle_evict()` since `cache` is never populated, thus `volume` will always be `null` in this line for `cache_max == 0`. For the other case, the intention is that `cache_size` should follow `cache`, not `cache_last_touched` (although that should also follow `cache`). So I don't think the size should be guarded here. 

I have refactored (in #6838) all of the cases that add to `cache_size` into a method, so that it's performed the same way for both cases that track `cache_size`.

Duplicati/Library/Main/Operation/Restore/VolumeManager.cs
```diff
@@ -147,7 +147,6 @@ void evict_lru()
                         {
                             // Pop the last element of cache_last_touched
                             var volume_id = cache_last_touched[0];
-                            cache_last_touched.RemoveAt(0);
```
carljohnsen, line 150: Eviction should remove the volume from the `cache_last_touched` list.

Duplicati/Library/Main/Operation/Restore/VolumeManager.cs
```diff
@@ -180,8 +179,8 @@ void evict_lru()
                                                 sw_request?.Start();
                                                 if (cache.TryGetValue(request.VolumeID, out var volume))
                                                 {
-                                                    cache_last_touched.Remove(request.VolumeID);
-                                                    cache_last_touched.Add(request.VolumeID);
+                                                    if (cache_last_touched.Remove(request.VolumeID))
```
carljohnsen, line 182: The intention is to touch the volume id, now that it's been accessed from cache. I do however see how this will prevent leaking, which I've added in #6838 for when `cache_max == 0`

Duplicati/Library/Main/Operation/Restore/VolumeManager.cs
```diff
@@ -211,7 +210,13 @@ void evict_lru()
                                     {
                                         sw_cache_set?.Start();
                                         volume.Reference(in_flight_downloads[volume_id].Count);
-                                        if (cache_max > 0)
+                                        if (cache_max == 0)
```
carljohnsen, line 213: This branch was already added in #6785 . 

Also: this is incorrect as your intention is already covered by the `VolumeWrapper`. The wrapper keeps track of how many times the volume has been referenced and only disposes once the reference count == 0. (see line 212 where the count is set to the amount of registered waiters). So in this PR, `cache_max == 0` would mean infinite cache use. So my guess is that it doesn't align with #6785 ?

Duplicati/Library/Main/Operation/Restore/VolumeManager.cs
```diff
@@ -233,7 +238,8 @@ void evict_lru()
                                             Logging.Log.WriteExplicitMessage(LOGTAG, "VolumeRequest", "Not caching volume {0} ({1} + {2} > {3})", volume_id, cache_size, volume.Size, cache_max);
                                             volume.Dispose();
                                         }
-                                        cache_last_touched.Add(volume_id);
+                                        if (cache_max != 0)
```
carljohnsen, line 241: Since refactoring in #6838, this is only ever done when `cache_max = 0`, following your intent.

carljohnsen (commented):
The shortcomings you've found are valid, but doesn't align with your changes in #6785 . I've added #6838 to address the shortcomings and fit together with #6785 .

Duplicati/Library/Main/Operation/Restore/VolumeManager.cs
aureliandevel: Yes, but handle_evict() is doing that ... assuming the volume is cached.

carljohnsen: Ah, right, I see! Nice catch; my comment was a bit too hasty. I've added it to the other PR [in this commit](https://github.com/duplicati/duplicati/pull/6838/changes/9a4a75e0f9dfbfcfa0853b066ca15f53d4de8c5d)

kenkendk:
Closing this as #6838 was merged and covers the changes in this PR

PR #6828, opened 2026-04-06
Update SharpCompress

### LINKED ISSUE #4333: Zstandard as default compression method, replacing deflate
#3708  Environment info
- **Backend**: WebDAV

## Description
The default compression method in Duplicati is deflate (practically gzip).
A modern outperformer is Zstandard/ Zstd.
Zstd is able to get comparable compression ratios at much higher speeds, both for compression and decompression.
But Zstd can be used for any use case: either higher compression at lower speeds or vice versa.
Thus, Zstd is a versatile compression method that vastly outperforms the currently used deflate, with flexibility.
My request is to include Zstandard as the default compression method at a setting that delivers a comparable compression ratio as deflate. For example: zstd level 4 gives a 500% speed improvement and even a minor compression improvement, compared to deflate.

## Steps to reproduce
https://community.centminmod.com/threads/round-3-compression-comparison-benchmarks-zstd-vs-brotli-vs-pigz-vs-bzip2-vs-xz-etc.17259/

- **Actual result**:
Deflate is the default. The bottleneck in backing up is deflate. Or at equal deflate speed, a higher ratio can be achieved with Zstd.
- **Expected result**:
Zstandard is the default compression method at a sane default, e.g. level 4. The bottleneck does not lie with deflate anymore, but with transfer speeds of modern Internet connections, 1Gb networks and conventional HDDs. It will also be possible set a higher Zstd level to increase compression ratio, while still being faster than deflate.

## Screenshots
Have a look at: https://community.centminmod.com/threads/round-3-compression-comparison-benchmarks-zstd-vs-brotli-vs-pigz-vs-bzip2-vs-xz-etc.17259/

## Debug log
None.

kenkendk:
This PR updates SharpCompress to latest version.
Also adds support for choosing ZStandard as a compression method.

This fixes #4333


kenkendk:
This change is currently blocked due to issues with upstream library:
https://github.com/adamhathcock/sharpcompress/pull/1282

Edit: Changed the test to not call `ReadByte()` as that method is not actually called in the code, just in the tests.

Issue #5911, opened 2025-01-27
Add update-backup command to ServerUtil


woutertinusf19:
To manage older versions of Duplicati I was using [duplicati-client](https://github.com/Pectojin/duplicati-client). I want to switch to the recent .NET8 releases, but I'm finding that [ServerUtil](https://docs.duplicati.com/duplicati-programs/command-line-interface-cli-1/serverutil) does not yet offer al the same features. Specifically I'm missing the "update backup" command that allows scripts to update backup settings without creating a new backup (e.g. to rotate credentials). The documentation about "import" notes:

> ...this will create a new backup with the same configuration, so make sure you have removed the previous backup configuration first.

Unfortunately, the tool doesn't offer a command to do this automatically (and it might be bad for performance, if remove/import also means that the local database has to be recreated). Please consider adding a method of updating backup settings automatically.

kenkendk:
@woutertinusf19 How did you use the `update backup` command?
Do you store a full copy of the configuration json and then edit it before calling "update" ?

I would think a "patch" method would be more elegant, so you could have a file like:
```json
{
  "Backup": {
   "Settings": {
      "--volume-size": "80mb",
      "--list-verify-uploads": null
    }
  }
  "Schedule": {
     
  }
}
```

What would fit your use best?

woutertinusf19:
> Do you store a full copy of the configuration json and then edit it before calling "update" ?

Something similar to that, I generate a full configuration file starting from a template which is then modified according to current state (e.g. include specific folders that may not exist on every VM, latest credentials from a vault, alternate target location depending on the source location). In my current script I first check with a `list` command if the backup already exists and based on that either issue either an `update backup` or `create backup` command to apply it, so I can run the same script to configure backups on both new machines and existing ones and they'll all end up with identical settings. 

I think a patch method would do the job for me.

ts678:
> e.g. to rotate credentials

Although it doesn't replace all use cases, and it requires an ability to update files, [Preload settings](https://docs.duplicati.com/detailed-descriptions/preload-settings) might help with some cases.

Although I don't personally use it either, duplicati-client seems to have create, update, and delete. I'd think delete could lead to update, while patch leading to delete (probably that request will happen at some point) seems a more awkward path to follow.

> I think a patch method would do the job for me.

Then another interim path (given an ability to update files) is patching in [run-script-before](https://prev-docs.duplicati.com/en/latest/appendix-g-example-scripts/) to replace things by writing outputs.

[...]

PR #6836, opened 2026-04-10
Add DeleteBackup command to ServerUtil

WouterTinus:
This PR adds a `delete` command to ServerUtil to support basic automation of changes in backup configuration. While update would still be more ideal, I wasn't sure how to approach that in the code base. An export/delete/import cycle will suffice for my and hopefully others needs.

Updates: https://github.com/duplicati/duplicati/issues/5911


kenkendk (approved):

kenkendk:
Thanks!

Looks good and clean.

For the update, this is really just an overwrite of the existing backup, but the server code does not currently support updating from an exported configuration.

The benefit of update would primarily be to keep the backup-id in the database and retain metadata, as opposed to the export/delete/import cycle.

Issue #5527, opened 2024-09-03
Disable time check on a per file basis


bkev:
Further discussion here
https://forum.duplicati.com/t/disable-time-check-on-a-per-file-basis

It would be good to enable the ability to disable the file check on a per file or filter-like expression basis so that the entire backup job doesn't take longer than needed if only some files need a time check to successfully back up.

In "Advanced Options" disable-filetime-check already exists but this is for the complete set.

It would be good to have the option of something like

--disable-filetime-check-include=*.iso

and/or

--disable-filetime-check-include=/path/to/file

bkev:
Would a list of extensions also be able to do a list of files too? Personally, I would find declaring several files for the disable time flag more useful.

For example

--disable-filetime-check-include=/path/to/file
--disable-filetime-check-include=/path/to/file2
--disable-filetime-check-include=/path/to/file3

kenkendk:
Good point @Jojo-1000 

It would essentially create a mini rule engine pr. file. 

I think the direction @bkev points to is better. The default is to use the filetime, and this can be completely disabled, or disabled selectively with match filters (globbing seems usefull here).

duplicatibot:
This issue has been mentioned on **Duplicati**. There might be relevant details there:

https://forum.duplicati.com/t/disable-time-check-on-a-per-file-basis/18883/14

Issue #5523, opened 2024-09-01
Have a list of mountable filetypes that don't change timestamps on edit.


Poikilos:
**Is your feature request related to a problem? Please describe.**
When a VeraCrypt hc file (probably same for TrueCrypt tc with less secure enryption) is mounted, file(s) edited & saved, then unmounted, the timestamp is not changed by VeraCrypt. Duplicati does not make an incremental backup.

**Describe the solution you'd like**
- If such a file type is found, do binary comparison (not just date & size comparison). A fast algorithm is here: https://stackoverflow.com/a/1359947 (exit early on diff, quite fast on 1 GB hc file).
- Do the extra check for hc or tc file (and possibly other types that may have concerns, but I don't now of any others).

**Describe alternatives you've considered**
- use custom program do backup this one file (I do this now, but it is not ideal and is not integrated with Duplicati)
- submitting issue to VeraCrypt (not ideal since some people never update and therefore will always have the issue, but I may do that)

Jojo-1000:
Hi, have you seen the discussion here: https://github.com/duplicati/duplicati/issues/2332

It was also suggested to detect the file extensions. There seems to be an option in Veracrypt to update the timestamp. If that doesn't help, you can use the advanced option `--disable-filetime-check`. This always reads the full files (maybe use a separate backup set for your Veracrypt files, so the rest isn't slowed down), so you don't need a  different program.

Poikilos:
O, my bad. Ok, I commented there.

duplicatibot:
This issue has been mentioned on **Duplicati**. There might be relevant details there:

https://forum.duplicati.com/t/disable-time-check-on-a-per-file-basis/18883/9

Issue #2332, opened 2017-02-21
No incremental backups of VeraCrypt file containers


kees-z:
I have:
- [x] searched open and closed issues for duplicates

----------------------------------------

### Version info

**Duplicati Version:** 2.0.1.45 Experimental
**Operating System:** Windows 10 Pro
**Backend:** WebDAV

### Bug description
I have a VeraCrypt file container of 10 GB included in a backup task. The first time Duplicati runs this task, the complete file container is backed up without problems.
However, when the container is mounted and changes are made to the contents of the container file, Duplicati doesn't detect the change and skips this file.
Reason for this seems to be that the file modification timestamp is not changed by default and the file size stays exactly the same. So Duplicati should be able to detect changes another way, for example:
- Some other mechanism could be introduced to find out if a file has been modified.
- A file list could be added in a backup task (something like the file extensions list for skipping compression) that **always** makes a backup of that file. Deduplication should prevent writing anything to the backend if the file container has not been changed.

### Steps to reproduce
- Create a VeraCrypt container file and make a backup with Duplicati
- Mount the container file and copy some files to the container.
- Re-run the backup
- Nothing is written, even though the source file has been modified.

maciekb:
Hi,

I have not used VeraCrypt, but with TrueCrypt was a similar problem. Backup software not detected changes in volume because of setting "Preserve Modification Timestamp". Often backup software detecting changed files first by changed modification time. If modification time was not changed, a file is skipped. Maybe this is the same issue?

Best regards,
Maciek

kenkendk:
The option in Duplicati would be `--disable-filetime-check`.

kees-z:
Yes, this seems to be the same issue.
Problem is that VeraCrypt (and TrueCrypt) has "Preserve modifications timestamp of file containers" selected by default. If the end user is not aware of this behaviour (Duplicati not detecting that an encrypted file container has been modified), he could be unpleasantly surprised and find out that tons of data have never been backed up, when he needs it and tries to restore the file container.

For me it is not a big deal, I deselected "Preserve modifications timestamp of file containers" in VeraCrypt and Duplicati seems to work well with this setting.

kees-z:
@kenkendk You beat me by 19 seconds!
I'll try that setting soon, but this is probably risky behaviour of a backup solution.
[...]

Issue #189, opened 2014-08-05
Duplicati not copying NTFS Alternate Data Stream


kenkendk:
_From [william.MPLX@gmail.com](https://code.google.com/u/william.MPLX@gmail.com/) on March 21, 2010 17:41:39_

What steps will reproduce the problem? 1. Create file with ADS (from command line "notepad filename.ext:ads1")
2. Save data to ADS stream
3. Backup file using Duplicati
4. Restore file to different directory using Duplicati
5. Attempt to open ADS stream What is the expected output? What do you see instead? Expect ADS stream to be restored along with main stream, instead ADS 
stream is missing Please provide any additional information below. I believe ADS streams are typically enumerated and manipulated in c# by 
using P/invoke and backupread/write

_Original issue: http://code.google.com/p/duplicati/issues/detail?id=189_
##

Yomodo:
AlphaFS now uses `FindFirstStreamW()` / `FindNextStreamW()`

vc:
I am also very interested in implementing the function of backup of alternative streams.
I also want to note that ADS can be present not only in files but also in folders. We are talking about NTFS

PR #6868, opened 2026-04-29
Feature/Full disk NTFS support

carljohnsen:
# NTFS File System Support and Incremental Backup Optimizations
This PR adds NTFS filesystem support for full disk backup. Like #6814, the file system is only being used for detecting which disk sectors has been modified since the last backup. 

For NTFS a block (a collection of sectors) is given the timestamp of the most recent changed file spanning that block to indicate whether it should be backed up. Special care is applied to the MFT, where one change to a cluster propagates to capture the entire MFT, and to the metadata records (such as `$LogFile`) to ensure that they're always backed up. 
# Additional changes
- The full disk unit tests have been refactored to be defined in a subfolder to have a better partitioning of test intent. 
- A partition reader stream has been added to abstractly expose the raw disk through the partition. 

# Bugfix
- Reading from a partition wrongfully read the entire disk into a stream, causing repeated reads to become slow. This wasn't as much of an issue for Unknown and FAT32 file systems, but it became apparent with NTFS.


ts678:
> timestamp of the most recent changed file spanning that block

Note that files can change without changing file's timestamp, e.g.:

[No incremental backups of VeraCrypt file containers #2332](https://github.com/duplicati/duplicati/issues/2332)

> Special care is applied to the MFT

There might be other internal NTFS bits. NTFS.com has some info.

ts678:
Please make sure this will work with FS defragmentation which might move data without timestamp change.

proprietary/DiskImage/Partition/PartitionReadStream.cs
```diff
@@ -0,0 +1,124 @@
...
+using System.Buffers;
+using System.IO;
+using System.Threading;
+using System.Threading.Tasks;
+using Duplicati.Proprietary.DiskImage.Disk;
+
+namespace Duplicati.Proprietary.DiskImage.Partition;
+
+/// <summary>
+/// A stream that buffers reads from a partition.
```
kenkendk, line 32: I don't see any buffering in this class, it seems to just forward calls to the IRawDisk instance?

proprietary/DiskImage/Filesystem/Ntfs/NtfsBitmap.cs
```diff
@@ -0,0 +1,538 @@
...
+    /// </summary>
+    /// <param name="clusterNumber">The cluster number to check.</param>
+    /// <returns>True if the cluster is allocated; otherwise, false.</returns>
+    /// <exception cref="ArgumentOutOfRangeException">Thrown if cluster number is out of range.</exception>
+    public bool IsClusterAllocated(long clusterNumber)
+    {
+        if (clusterNumber < 0 || clusterNumber >= m_totalClusters)
+            throw new ArgumentOutOfRangeException(nameof(clusterNumber), $"Cluster number {clusterNumber} is out of range. Valid range is 0 to {m_totalClusters - 1}.");
+
+        return m_allocationBitmap[(int)clusterNumber];
```
kenkendk, line 153: I realize the int is a requirement for the BitArray, but could we risk overflowing the cluster number when casting here?

kenkendk (approved):

kenkendk:
@ts678 Great that you put a spotlight on this.

I was looking around for a solution, but the only real solution is to add a Windows Kernel Driver that logs which sectors are written. Such a driver is a non-trivial undertaking to write as it needs to be reliable and is loaded before Windows itself starts. But the really hard part is getting a Microsoft approved signing certificate so the driver will load.  For the current PR that is out-of-scope.

It *think* we are good anyway because we check if the timestamp has changed, not just if it is newer. If you defrag the disk, the chances that the timestamps are exactly the same for a sector that has been moved is very close to zero (but not absolute zero).

As a follow up to this PR we can consider purchasing a Windows Kernel Driver that lets us tap into these messages without having to maintain such a feature. This will most likely mean that the feature will not be free though.

ts678:
> If you defrag the disk, the chances that the timestamps are exactly the same for a sector that has been moved is very close to zero (but not absolute zero).

@kenkendk I might misunderstand the plan, but if you mean something like file last modified, I don't think it changes.

There are apparently some subtleties, such as maybe file last accessed changes (which a full metadata read might see).

does windows defrag change timestamp (and similar searches) on Google etc. can likely provide a good overview of it.

> Windows Kernel Driver that logs which sectors are written

This sounds like what UrBackup does with their commercial offering for Change Block Tracking. Other tools use similar.

ts678:
[Shrink a basic volume](https://learn.microsoft.com/en-us/windows-server/storage/disk-management/shrink-a-basic-volume) (Microsoft) and many third-party tools can probably also move files without last modified change:

> When you shrink a partition, any ordinary files are automatically relocated on the disk to create the new unallocated space.

If such low-level file data rearranging actually changed file modified times, tools that trigger on changes would get busy
doing backup, sync, copy-if-newer (e.g. xcopy, robocopy options), etc. and generate lots of user questions. Ever see any?

BTW if there is a FAT32 scheme similar to this NTFS one, then the same sort of questions may apply for low-level moves.

ts678:
For FAT32, Fat32DirectoryWalker might fall into the same timestamp pit. Also note note above on mounts, e.g. Veracrypt.

One peril that NTFS adds is alternate data streams. I saw no mention of those in code. Are they backed up and updated?
[Alternate data streams … redux](https://www.jankyrobotsecurity.com/2019/04/15/alternate-data-streams-redux/) is a claim that those can escape timestamp update, but are they in full disk backup at all?

kenkendk:
@ts678 The way the logic works is that each sector on the disk gets an "aggregated timestamp" which covers all files that "touch" that sector. If you defrag or otherwise move things around, the "aggregated timestamp" for the sector will no longer match the previous stamp (because the files that touch the sector are different).

This does cover alternate data streams, with the caveat that it does not handle cases where the timestamps on files are not updated.

As a side-note, Duplicati does not backup or restore ADS when used in the regular file-mode.

ts678:
@kenkendk 

Thanks for clarifying. The aggregate timestamp sounds like some value that changes if any of its file timestamps change.

If you genuinely mean "sector", that can possibly be boosted to larger cluster size at least for purposes of data allocation.

Beware that small amounts of data can be in the MFT itself.

> This does cover alternate data streams, with the caveat that it does not handle cases where the timestamps on files are not updated.

That caveat seems a general one, applying to default stream too.

This hits the issue with VeraCrypt [linked above](https://github.com/duplicati/duplicati/pull/6868#issuecomment-4352524080) where unchanging timestamp is a configurable but default-on preference.

[Disable time check on a per file basis](https://forum.duplicati.com/t/disable-time-check-on-a-per-file-basis/18883) in forum discusses and links issues which may be relevant to full disk backup as well:

[Disable time check on a per file basis #5527](https://github.com/duplicati/duplicati/issues/5527)

[Have a list of mountable filetypes that don't change timestamps on edit. #5523](https://github.com/duplicati/duplicati/issues/5523)

I also found lots of people on the Internet curious about the old timestamps on VM disks such as those used by Hyper-V.
I have to say I'm finding quite a variety of answers (some for forensic analysis) on how far file timestamps can be trusted.

> As a side-note, Duplicati does not backup or restore ADS when used in the regular file-mode.

[Duplicati not copying NTFS Alternate Data Stream #189](https://github.com/duplicati/duplicati/issues/189)

kenkendk:
After considering the potential issues for a bit, I have decided that the FAT32 and NTFS (in this PR) will be default off, so backups of a partition is by default treated as a stream of sectors. They can be enabled with `--diskimage-filesystem-parsed=true`.

I think the risk of breaking a backup is too large if there are applications that do not update the timestamps with the current approach. By having parsing opt-in, the code can live in the main branch and we can revisit once we have a kernel driver filter working. At that point we should probably run the first backup with parsing enabled and then rely on the filter results for subsequent backups.

PR #6925, opened 2026-05-19
Feature/AutoTune tool

carljohnsen:
This PR adds a tool (`Tools/AutoTune`) for automatic tuning of the concurrency parameters of the restore operation. It is a commandline tool intended to be run as an exploratory step. It can be configured using multiple options, which are documented when providing the `-h` commandline flag: 

```
Description:
  Auto tuning of the Duplicati concurrency parameters. Warning, this program will be accessing the source, destination, and restoretarget a lot, 
  potentially degrading them.

Usage:
  AutoTune [options]

Options:
  --backend-options <backend-options>                  Duplicati options to pass to the backend during backup. Each option is a key-value pair 
                                                       spearated by an equals sign, e.g. --backend-options key1=value1 key2=value2. Default is an 
                                                       empty list. []
  --baseline-params <baseline-params>                  The step value(s) to consider the baseline for the final comparison, 1 or 4 integers. If one 
                                                       value is specified, the same value is used for all parameters. If four values are specified, 
                                                       they are applied individually for file-processors, volume-decompressors, volume-decryptors, 
                                                       and volume-downloaders (in that order). If not specified, the default Duplicati parameters 
                                                       are used. []
  --destination <destination>                          Destination to store the test backup. The destination should be empty (as required by 
                                                       Duplicati). The data will be deleted again after the tuning process. If no argument is 
                                                       specified, then a temporary folder (as optionally specified with the temp_folder argument) 
                                                       will be used. []
  --dont-revisit-parameters                            During tuning, once a new 'better' configuration has been found, all of the tunable 
                                                       parameters become candidates again. Setting this option will disable already visited 
                                                       candidate parameters. This will make the tuning converge faster, but may not find an optimal 
                                                       configuration. [default: False]
  --exponential-steps                                  If specified, the steps taken for next candidate run is to multiply by 2 instead of plus 1. 
                                                       This will make the tuning converge faster, but may not find an optimal configuration. 
                                                       [default: False]
  --restoretarget <restoretarget>                      Target folder to restore a backup to. The folder should be empty beforehand, as it needs to 
                                                       be emptied during measurements. If no argument is specified, then a temporary folder (as 
                                                       optionally specified with the temp_folder argument) will be used. []
  --default-settings                                   Start tuning from the Duplicati default settings instead of a starting step of 1. Ignored if 
                                                       --starting-steps is specified. [default: False]
  --runs <runs>                                        Number of runs to measure. The mean is reported. [default: 3]
  --starting-steps <starting-steps>                    The starting step value(s) for the tunable parameters, 1 or 4 integers. If one value is 
                                                       specified, the same value is used for all parameters. If four values are specified, they are 
                                                       applied individually for file-processors, volume-decompressors, volume-decryptors, and 
                                                       volume-downloaders (in that order). If not specified, the starting step is 1 for all 
                                                       parameters. Cannot be used together with --default-settings. []
  --source-folder <source-folder>                      Source folder to make a backup of. If the folder is empty, then some test data will be 
                                                       generated. If no argument is specified, then a temporary folder (as optionally specified with 
                                                       the temp_folder argument) will be used. []
  --temp-folder <temp-folder>                          Path to where the temporary files should be created. If no argument is specified, then the 
                                                       system default (e.g. /tmp or %TEMP%) will be used. []
  --testdata-max-file-size <testdata-max-file-size>    If no source folder has been specified, this option tunes the maximum size (in bytes) a 
                                                       generated file may have. Default is 1 MB. [default: 1048576]
  --testdata-max-total-size <testdata-max-total-size>  If no source folder has been specified, this option tunes the maximum size (in bytes) the 
                                                       generated files collectively may take up. Default is 512 MB. [default: 536870912]
  --testdata-num-files <testdata-num-files>            If no source folder has been specified, this option tunes how many files are generated as 
                                                       test data. [default: 10000]
  --verbose <verbose>                                  Verbosity level: 0 disables output, 1 prints full progress information during tuning runs. 
                                                       Higher levels reserved for future debug printing. [default: 1]
  --warmup <warmup>                                    Amount of warmup runs to perform before measuring. [default: 1]
  --version                                            Show version information
  -?, -h, --help                                       Show help and usage information
```

The tuning works by first creating some test data (configured using the `--testdata-*` options), unless a directory containing data has been specified in the `--source-folder` option. In its current form, only local folders are considered. 

Then a backup is created from this test data to the specified `--destination`. During the restore benchmark, data will only be written once, but read multiple times from this destination. So take care if reading incurs costs. 

Then the tool will perform a baseline run, which will be used for the final comparison. By default this run uses the default parameters set by Duplicati, but can be configured using `--baseline-parameters`. The baseline benchmark will first run for `--warmup` rounds and measure the final `--runs` runs. 

Then the tool will perform a run with the `--starting-steps` concurrency parameters. These can also be set using `--use-default-settings`, which will take the default concurrency parameters from Duplicati as starting configuration.

After each round of optimization, the tool will look at the wall-clock execution time of the entire operation. If the average of the times are lower than what we've seen before (which is always the case for the first round of optimization) we store the configuration used as the new best.

Then we look at each of the candidate processes, which for the restore operation are `FileProcessor`, `VolumeDecompressor`, `VolumeDecryptor`, and `VolumeDownloader`. As each process is part of a larger pipeline, their execution time can coarsely be abstracted as 

1. Read from input.
2. Execute task.
3. Write to output.

The tuning tool currently only consideres step 2, execute. Given that the processes are running in parallel (with respect to the process instances of the same type) and concurrently (with respect to the processes of other types), the sum of the execution times across the processes are not necessarily equal to the wall-clock execution time. 

However, the individual timings are used for considering the next optimization step. For each process type, we compute the mean of all of the processes. While increasing the amount of instances of a process will lower their internal execution time (as they're now sharing work more equally), it may not necessarily lower the wall-clock execution time, as they may not have been congesting the pipeline. 

We select the most saturated process (i.e. the process with the largest average execution time) as candidate, increase the corresponding concurrency parameter, and run another optimization round. The increase is either incremental (`+ 1`) for the default case, or exponential (`* 2`) if specified by the `--exponential-steps` option. The latter allows for faster convergence at the cost of potentially missing a minima. 

If a new best configuration has been found, all of the parameters are again marked as candidates, unless disabled by the `--dont-revisit-parameters` option. Again, this allows for faster convergence at the cost of potentially missing a minima. 

If the subsequent run didn't provide better wall-clock execution time, that parameter is excluded from the candidate list, and the next in line parameter is selected as a candidate. 

These steps are repeated until the candidate parameter list has been exhausted (i.e. increasing any of the parameters do not provide a better wall-clock result). At the end, the best configuration is reported and reflected against the baseline time:

```
----==== Tuning complete ====----
 Best configuration: FileProcessors: 9, VolumeDecompressors: 5, VolumeDecryptors: 1, VolumeDownloaders: 1
 Best time:         2359 ms
 Default time:      2446 ms
 Speedup:           1,04x
```

Given that the benchmark requires saturation to obtain proper measurements, larger data sizes should be considered, which comes at the cost of execution time. Also do note that one configuration may not transfer well to another setup, as resource characteristics will change. As such, the setup should match that of the target as much as possible. 

# Future improvements
- Rather than considering internal execution time alone, the tool could provide a better insight by looking at the read and write timings as well: if a process spends most of its time waiting for communication, creating more instances may provide little benefit.
- The tool should be extended to look at the backup parameters as well. This could even be further extended to more operations if they can be parameterized as well (e.g. compact, repair).
- The block size and volume size also have a substantial impact, and could be considered for a future iteration of the tool. As it stands now, it can be controlled using the `--backend-options` option. 
- Caching has been largely ignored as this tool is intended to be run on small representative datasets to avoid thrashing resources (e.g. disk or remote quota). However, it does have a performance impact, and could be measured as well. 
- Only a single version backup is currently considered. A more true to the real world configuration would be a version with multiple versions, where only a subset of the files are considered, so that the restore doesn't restore every block from every volume.
- It would be a nice to have if this was integrated as part of the UI, which allowed both storing and loading existing configurations. In its current form, the tool is independant and manually controlled from the commandline. 
- More advanced optimization strategies (such as found in Machine Learning hyper parameter tuning) could be used for faster convergence (e.g. Bayesian optimization).


ts678:
> with multiple versions, where only a subset of the files are considered

I don't think those two items have to go together, but nice to see them.
Deduplication level and caching matter, as well? I didn't find comments.
Multiple versions also invite `compact`. It spreads blocks across volumes.

carljohnsen:
> I don't think those two items have to go together, but nice to see them.

Right, my idea was just construct a backup that contains files spread across volumes to bias the workload a bit more towards the `VolumeDownloader`. My guess is that it would represent some real-world cases better. As the auto tuning tool is meant as a micro benchmarking tool rather than a real benchmarking tool, it's not too important for the first iteration.

> Deduplication level and caching matter, as well? I didn't find comments.

I've added the option `--testdata-sparse-factor` to control deduplication rate of the generated data. Regarding caching, I've added this to the 'Future improvements' list:
```
- Caching has been largely ignored as this tool is intended to be run on small representative datasets to avoid thrashing resources (e.g. disk or remote quota). However, it does have a performance impact, and could be measured as well. 
```

>  Multiple versions also invite compact. It spreads blocks across volumes.

Compaction should benefit restore, if I understand it correctly. While it may spread blocks, it will also compact blocks into less volumes, which could lead to better locality although not guarenteed. 

Reflecting on compaction, I've added this line to the improvements list: 
```
- The tool should be extended to look at the backup parameters as well. This could even be further extended to more operations if they can be parameterized as well (e.g. compact, repair). 
```

ts678:
> Compaction should benefit restore, if I understand it correctly.

This would be an interesting study, but sensitive to write pattern.

By definition it gathers whatever is left from too-empty volumes.
That seems like it's prone to be a quite random bunch of blocks.

Two extreme cases of the backup-only changes of Duplicati are:

* File size doesn't change, but data does. Updates spread out.
Eventually the initial backup of the file also gets compacted.

* File appends at end. Updates spread out. Original stays, but

The initial file backup is likely to be spread over many volumes:

[Channel Pipeline](https://forum.duplicati.com/t/channel-pipeline/11466)

> Multiple processes read blocks and compresses blocks; filled volumes are written as upload requests

is maybe the spot where blocks disassociated from file are read.
Maybe you've seen the effect of a mixed-file volume on restore.

I suppose in some cases a file delete, age-out, and compact can
make volume with less of a file mix, but I'm not sure if it's usual.

Writing from memory, so no guarantees. I can do test if needed
such as backing up two random files to sample volume content.

Issue #6924, opened 2026-05-18
Remote smb source seeing folders as files


realtek2017:
-   [x] I have searched open and closed issues for duplicates.
-   [x] I have searched the [forum](https://forum.duplicati.com) for related topics.

---

## Environment info

-   **Duplicati version**: 2.3.0.1_stable_2026-04-24
-   **Operating system**: Docker Image lscr.io/linuxserver/duplicati
-   **Backend**: lscr.io/linuxserver/duplicati

## Description

I am not sure if this is a server or ui issue (ngclient) but when adding only an smb source via ngclient it is seeing all folders as files and then failing to back any folders (only files) up and is giving the following error:

2026-05-18 17:49:45 +01 - [Warning-Duplicati.Library.Main.Operation.Backup.FileBlockProcessor.FileEntry-FileProcessingFailed]: Failed to process path: /dupl-smb-7eh40cx6/Data
UserInformationException: Failed to open file with error Data with status STATUS_FILE_IS_A_DIRECTORY

## Steps to reproduce

1. Create a new backup
2. Name: Test Backup, Destiation = SFTP
3. Source Data:  SMB
4. Add Remote Source: Server: 192.168.0.x, Share Name: Data, Path on Server: Data, Username, Password
5. Click Test Now (Success)
6. Click Use Remote Source
7. Schedule, Select anything
8. Click continue
9. Click Submit
10. Click Home
11. Click Start on the job
12. Job runs for a very short time ( few seconds) says its Successful
13. Goto logs

-   **Actual result**:
Errors displayed as per above, only files that exist are backed up. No subfolders
- 
-   **Expected result**:
Data backs up

kenkendk:
Fixed with #6938

PR #6937, opened 2026-05-28
Fix SMB source listing misclassifying directories as files

realtek2017:
When using SMB/CIFS share as a backup source, directories are identified as files. This then causes STATUS_FILE_IS_A_DIRECTORY error to be logged and only files in the root of the source are backed up.

SMBShareConnection.ListAsync and IsFolder was set with an exact match:
info.FileAttributes == FileAttributes.Directory

I updated this to match the same check as used in GetEntryAsync and have tested this locally and it is working.

I raised this under issue #6924


realtek2017:
Included additional commits by mistake. I will re-open a new pull request

PR #6938, opened 2026-05-28
Fix SMB source listing misclassifying directories as files

realtek2017:
When using SMB/CIFS share as a backup source, directories are identified as files. This then causes STATUS_FILE_IS_A_DIRECTORY error to be logged and only files in the root of the source are backed up.

SMBShareConnection.ListAsync and IsFolder was set with an exact match:
info.FileAttributes == FileAttributes.Directory

I updated this to match the same check as used in GetEntryAsync and have tested this locally and it is working.

I raised this under issue #6924

### DIFF
--- a/Duplicati/Library/Backend/SMB/SMBShareConnection.cs
+++ b/Duplicati/Library/Backend/SMB/SMBShareConnection.cs
@@ -282,7 +282,7 @@ public class SMBShareConnection : IDisposable, IAsyncDisposable
                             info.LastAccessTime,
                             info.LastWriteTime)
                         {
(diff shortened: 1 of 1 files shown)


kenkendk:
Thanks for finding this!

There was a small typo (`&&` vs `&`), so I updated the PR to use `HasFlag()` to test for the directory bit/flag.

realtek2017:
> Thanks for finding this!
> 
> There was a small typo (`&&` vs `&`), so I updated the PR to use `HasFlag()` to test for the directory bit/flag.

No problem, its a great feature to be able to backup remote smb/cifs. Could not understand why I couldn't get it to work but took some time to understand the code and find that but glad its fixed!

Thanks for the updated PR, that's a much better way of doing it!

PR #6939, opened 2026-05-28
Fix remote source connection test for backend provider

realtek2017:
When using the Test button in the UI on source such as smb, it does not perform a real test.

Fix remote source “Test connection” for SMB and other file sources: `BackendSourceProvider.Test()` now runs read-only `Initialize()` (list share path) instead of `TestAsync()` (write test) like we do on the destination.

This is updated as it was only calling Enumerate().AnyAsync() which was returning the in memory root and not actually doing a connection. So now I updated it to use Initialize() and then performed a read-only operation considering as this is a remote source and not destination.

I have tested it with smb and it is fully working.

### DIFF
--- a/Duplicati/Library/SourceProvider/Builtin/BackendSourceProvider.cs
+++ b/Duplicati/Library/SourceProvider/Builtin/BackendSourceProvider.cs
@@ -84,7 +84,7 @@ public class BackendSourceProvider(IFolderEnabledBackend backend, string mounted
 
     /// <inheritdoc/>
     public Task Test(CancellationToken cancellationToken)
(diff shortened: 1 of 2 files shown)


kenkendk:
Thanks for finding this!

I can see the fix you made would work, but it would essentially put the testing into the WebAPI layer, instead of letting the module perform the testing.

I have made an extended fix in #6945 which updates the backend interface with an option to test read-only or read-write, and then wired that up so the source provider only tests read-only (as we will not write). I also exposed the read-only option to the WebAPI so in the case we are looking to restore, we do not test if we can write the remote.

PR #6965, opened 2026-06-17
Store configuration with backup

### LINKED ISSUE #3073: Automatically Include Configuration in Backup
#1048 mentions: "The option to store settings has not been implemented either."

## Environment info
- **Duplicati version**: 2.0.2.15_experimental_2018-01-03 

## Description
Backup Configuration can be exported to a "duplicati-config.json"-file.
Enhancement: Add an option to include the Configuration of the current Backup in each Snapshot.
This may be by simply done by internally running the "export configuration" command and including the output as "duplicati-config.json" within the dlist.zip.

This would help to restore details of one specific backup when local configs gets corrupted. (Given you have the passphrase at hand.)
There's would be no need for an extra export-and-backup-job besides the actual backup.

### LINKED ISSUE #6256: Provide option to automatically export backup configuration.
**Is your feature request related to a problem? Please describe.**
I use duplicati docker on linux to backup my important files. Recently it came in handy when I had to wipe my linux box and restore all the files. One problem I ran into that, since I had not backed up duplicati folder itself (since sqlite db would have been locked by process) i couldn't restore the backup configs/schedules.
I see that there is option to manually export the config. But given duplicati IS a BACKUP software, shouldn't it have an option to backup it's own config files ?

**Describe the solution you'd like**
Provide an option to include all current backup configs to be backed up along with other files. And when restoring it should give option to restore the configs.

**Describe alternatives you've considered**
Manually exporting and importing config is possible, but then duplicati IS a BACKUP software, shouldn't it have an option to backup it's own config files ?

**Additional context**
None

kenkendk:
This PR revives the `store-task-config` option that was never shown in the user interface and extends the feature to allow more flexibility in exporting the configurations.

The `Auto` setting is now on by default. For encrypted backups, this will store the backup configuration of the current backup (excluding any secrets) with the backup for easy restore of a configuration.

For unencrypted backups, `Auto` means that no settings will be stored by default.

To manually pick the what backup configurations are stored, the following options are also available:
- `None`: no configuration is stored
- `Self`: The configuration of the current backup is stored, without any secrets
- `All`: The configurations of all backups are stored, without any secrets

The options intentionally strip secrets (passphrases, passwords, api-keys, etc) from the configuration. This choice makes it less convenient for the user, as the secrets needs to be re-entered when the configurations are re-created. This is done to guard against a potential cross-exposure that could happen if the backup configuration is leaked. In this case, an attacker would potentially obtain all credentials in the backup and could gain access to the remote storage, additional storage destinations, remote sources, and other endpoints configured. 

With a defensive default setting, the user can choose convenience over security and include the secrets via the options:
- `SelfWithSecrets`
- `AllWithSecrets`

These options will include all secrets in the backup configuration, making the recreate step a one-click operation.

If either of these options are used on a backup without encryption, a warning will be logged and they will revert to Self or All, respectively.

If the user has a carefully managed environment and truly wants to store unencrypted secrets, the following dangerous options are also available:
- `SelfWithUnencryptedSecrets`
- `AllWithUnencryptedSecrets`

If these options are chosen on an encrypted backup, they are effectively the same as `SelfWithSecrets` and `AllWithSecrets`, respectively.

This fixes #6256
This fixes #3073

This PR also introduces new API endpoints that allows the UI to obtain and recreate the stored configurations.

Additionally, this fixes a few issues with importing backup configurations that use additional target urls, and prevents leaking remote source details in exports.


kenkendk:
One concern with the default "Auto" setting is cross-exposure when using multiple destinations.
With the current default, the stored configuration will contain the credentials to connect to additional storage.
If the basic backup is compromised, an attacker can obtain the credentials to secondary storage and disable these.

EDIT: This has been addressed by mapping `Auto` -> `Self`  so secrets are unconditionally cleared.

PR #7017, opened 2026-07-07
Avoid secondary transaction errors after repair failure

### LINKED ISSUE #5827: Repair failure causes exit exception
If the repair is failing due to a non-repairable database, it also triggers a non-existing transaction commit error:

```
The database was attempted repaired, but the repair did not complete. This database may be incomplete and the backup process cannot continue. You may delete the local database and attempt to repair it again.

The operation Repair has failed with error: unknown error
No transaction is active on this connection => unknown error
No transaction is active on this connection

code = Unknown (-1), message = System.Data.SQLite.SQLiteException (0x80004005): unknown error
No transaction is active on this connection
at System.Data.SQLite.SQLiteTransaction.Commit()
at Duplicati.Library.Main.Operation.RecreateDatabaseHandler.DoRun(LocalDatabase dbparent, Boolean updating, IFilter filter, NumberedFilterFilelistDelegate filelistfilter, BlockVolumePostProcessor blockprocessor)
at Duplicati.Library.Main.Operation.RecreateDatabaseHandler.Run(String path, IFilter filter, NumberedFilterFilelistDelegate filelistfilter, BlockVolumePostProcessor blockprocessor)
at Duplicati.Library.Main.Operation.RepairHandler.RunRepairLocal(IFilter filter)
at Duplicati.Library.Main.Operation.RepairHandler.Run(IFilter filter)
at Duplicati.Library.Main.Controller.<>c__DisplayClass21_0.b__0(RepairResults result)
at Duplicati.Library.Main.Controller.RunAction[T](T result, String& paths, IFilter& filter, Action1 method)    at Duplicati.Library.Main.Controller.RunAction[T](T result, IFilter& filter, Action1 method)
at Duplicati.Library.Main.Controller.Repair(IFilter filter)
at Duplicati.CommandLine.Commands.Repair(TextWriter outwriter, Action1 setup, List1 args, Dictionary2 options, IFilter filter)    at Duplicati.CommandLine.Program.ParseCommandLine(TextWriter outwriter, Action1 setup, Boolean& verboseErrors, String args)
at Duplicati.CommandLine.Program.RunCommandLine(TextWriter outwriter, TextWriter errwriter, Action`1 setup, String args)
Return code: 100
```

JamBalaya56562:
## Summary
- Fixes #5827.
- Prevent `ReusableTransaction.DisposeAsync()` cleanup from replacing the original repair/recreate failure with a secondary completed-transaction error.
- Keep explicit `CommitAsync()` and `RollBackAsync()` behavior unchanged; only dispose-time cleanup suppresses already-completed transaction errors.

## Why
When repair/recreate fails for the real underlying reason, disposal may try to roll back or dispose a transaction that has already completed. That can produce a secondary `No transaction is active` / completed transaction error and obscure the actual repair failure.

## Validation
- `dotnet build Duplicati\UnitTest\Duplicati.UnitTest.csproj --no-restore --no-dependencies --verbosity minimal -p:UseSharedCompilation=false -nr:false`
  - Build succeeded.
  - Existing unrelated `Microsoft.OpenApi 2.3.0` `NU1903` warnings remain in `Duplicati.UnitTest`, `Duplicati.Server`, and `Duplicati.WebserverCore`.
- `dotnet test Duplicati\UnitTest\Duplicati.UnitTest.csproj --filter "FullyQualifiedName~Issue5827" --no-restore --no-build --verbosity minimal`
  - Passed: 1, Failed: 0, Skipped: 0.
- `dotnet test Duplicati\UnitTest\Duplicati.UnitTest.csproj --filter "FullyQualifiedName~Issue4988" --no-restore --no-build --verbosity minimal`
  - Passed: 1, Failed: 0, Skipped: 0.

Note: validation used the repo-external temporary .NET SDK 10.0.301 install because the default PATH SDK is older.

### DIFF
--- a/Duplicati/Library/Main/Database/ReusableTransaction.cs
+++ b/Duplicati/Library/Main/Database/ReusableTransaction.cs
@@ -149,9 +149,12 @@ internal class ReusableTransaction(SqliteConnection con, SqliteTransaction? tran
 
     // Microsoft.Data.Sqlite throws InvalidOperationException ("This SqliteTransaction has completed;
     // it is no longer usable.") when a transaction is rolled back or disposed after it has already
-    // been completed. Match on the exception type rather than the (localizable) message text.
+    // been completed. Unfortunately there is nothing specific that can be used to differentiate it
+    // from other exceptions that might occur during disposal, so we rely on the exception type, source and message.
     private static bool IsInactiveTransactionException(Exception ex)
-        => ex is InvalidOperationException;
+        => ex is InvalidOperationException
+            && ex.Source == "Microsoft.Data.Sqlite"
+            && ex.Message.StartsWith("This SqliteTransaction has completed", StringComparison.Ordinal);
 
     /// <summary>
     /// Rolls back the transaction and restarts it.
--- a/Duplicati/Library/Main/Database/ReusableTransaction.cs
+++ b/Duplicati/Library/Main/Database/ReusableTransaction.cs
@@ -153,7 +153,6 @@ internal class ReusableTransaction(SqliteConnection con, SqliteTransaction? tran
     // from other exceptions that might occur during disposal, so we rely on the exception type, source and message.
     private static bool IsInactiveTransactionException(Exception ex)
         => ex is InvalidOperationException
-            && ex.Source == "Microsoft.Data.Sqlite"
             && ex.Message.StartsWith("This SqliteTransaction has completed", StringComparison.Ordinal);
 
     /// <summary>
--- a/Duplicati/Library/Main/Database/ReusableTransaction.cs
+++ b/Duplicati/Library/Main/Database/ReusableTransaction.cs
@@ -121,16 +121,36 @@ internal class ReusableTransaction(SqliteConnection con, SqliteTransaction? tran
             }
             catch (Exception ex)
             {
-                Logging.Log.WriteErrorMessage(LOGTAG, "ReusableTransaction dispose", ex, "Transaction disposed with error: {0}", ex.Message);
-                throw;
+                if (!IsInactiveTransactionException(ex))
+                {
+                    Logging.Log.WriteErrorMessage(LOGTAG, "ReusableTransaction dispose", ex, "Transaction disposed with error: {0}", ex.Message);
+                    throw;
+                }
+
+                Logging.Log.WriteVerboseMessage(LOGTAG, "ReusableTransactionAlreadyCompleted", ex, "Transaction was already completed during dispose: {0}", ex.Message);
             }
             finally
             {
                 m_disposed = true;
-                await m_transaction.DisposeAsync().ConfigureAwait(false);
+                try
+                {
+                    await m_transaction.DisposeAsync().ConfigureAwait(false);
+                }
+                catch (Exception ex)
+                {
+                    if (!IsInactiveTransactionException(ex))
+                        throw;
+
+                    Logging.Log.WriteVerboseMessage(LOGTAG, "ReusableTransactionAlreadyDisposed", ex, "Transaction was already completed before dispose: {0}", ex.Message);
+                }
             }
         }
     }
+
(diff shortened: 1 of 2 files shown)


Duplicati/Library/Main/Database/ReusableTransaction.cs
```diff
@@ -121,17 +121,37 @@ public async ValueTask DisposeAsync()
...
             {
-                Logging.Log.WriteErrorMessage(LOGTAG, "ReusableTransaction dispose", ex, "Transaction disposed with error: {0}", ex.Message);
-                throw;
+                if (!IsInactiveTransactionException(ex))
+                {
+                    Logging.Log.WriteErrorMessage(LOGTAG, "ReusableTransaction dispose", ex, "Transaction disposed with error: {0}", ex.Message);
+                    throw;
+                }
+
+                Logging.Log.WriteVerboseMessage(LOGTAG, "ReusableTransactionAlreadyCompleted", ex, "Transaction was already completed during dispose: {0}", ex.Message);
```
kenkendk: This should be a warning so we capture it in tests etc, as we should fix the underlying issue.

Duplicati/Library/Main/Database/ReusableTransaction.cs
```diff
@@ -121,17 +121,37 @@ public async ValueTask DisposeAsync()
...
+                try
+                {
+                    await m_transaction.DisposeAsync().ConfigureAwait(false);
+                }
+                catch (Exception ex)
+                {
+                    if (!IsInactiveTransactionException(ex))
+                        throw;
+
+                    Logging.Log.WriteVerboseMessage(LOGTAG, "ReusableTransactionAlreadyDisposed", ex, "Transaction was already completed before dispose: {0}", ex.Message);
```
kenkendk: This should be a warning so we capture it in tests etc, as we should fix the underlying issue.

Duplicati/Library/Main/Database/ReusableTransaction.cs
```diff
@@ -121,17 +121,37 @@ public async ValueTask DisposeAsync()
...
+                }
             }
         }
     }
 
+    private static bool IsInactiveTransactionException(Exception ex)
+        => ex.Message.Contains("No transaction is active", StringComparison.OrdinalIgnoreCase)
```
kenkendk: I don't like string matching here; it tends to break under localization, match unintended etc. Can we match on the correct exception type?

kenkendk (changes requested):
Clean, but minor changes are needed/

Duplicati/Library/Main/Database/ReusableTransaction.cs
JamBalaya56562: Done — switched this to `WriteWarningMessage` so it is captured in tests and we can address the underlying cause. Thanks!

Duplicati/Library/Main/Database/ReusableTransaction.cs
JamBalaya56562: Done — changed this one to `WriteWarningMessage` as well.

Duplicati/Library/Main/Database/ReusableTransaction.cs
JamBalaya56562: Good point — replaced the message-string matching with an exception-type check (`ex is InvalidOperationException`). Microsoft.Data.Sqlite throws `InvalidOperationException` ("This SqliteTransaction has completed; it is no longer usable.") for an already-completed transaction; I verified locally (via the Issue5827 test) that this is exactly the type caught, so the type check is reliable and avoids the localization / false-match concerns.

kenkendk: Sorry, tested this now, and the exception is sadly super generic, so we need some string matching. I have fixed it and will merge once the CI completes.

kenkendk (approved):

PR #4574, opened 2021-07-02
[WiP] Add parity module and default implementation based on PAR2

cmpute:
# Summary
This PR adds the functionality to create parity files for the backup (fixes #314). Parity data can be useful when the network connection is not stable or the filenames are been messed in the backend. Part of the implementation is based on #3879. Any suggestions are welcome.

A module interface is defined for parity provider similar to ICompression and IEncrypt interfaces. And a PAR2 module is implemented to utilized the PAR2 standard. Other parity implementations can be integrated through the interface. A requirement for the parity module is that it can produce a single file storing the parity data. It is critical to prevent additional zipping on the parity files, because there's no way to ensure the parity protection on the zipped archive.

# Added options
The options for parity modules:
- `parity-redundancy-level`: redundancy level in percentage, default = 0
- `parity-module`: the module used to create parity, default to par2 implementation, default = `par2`

The options for par2 module:
- `par2-block-count-small-file`: specify the block size for small files, default = 25
- `par2-block-count-large-file`: specify the block size for large files, default = 500
- `par2-file-size-threshold`: specify the file size threshold between small and large files, default = 4MB
- `par2-program-path`: specify the path to the PAR2 program. Windows version is bundled, in Linux par2 can be installed via package manager

# Implementation details

The parity module will act transparently between the main library and backends.

## Upload (Backup, Compact)

The parity file is created after when the data volume is compressed and encrypted and before uploading. The suffix for parity files will be like `+.par2` to make it easier to distinguish encryption suffix and parity suffix.

## Download (Compact, Restore, Find)

When a volume downloaded has a hash mismatch, the backend manager will try to find its parity file on the remote based on file name. If present, the parity file will be downloaded and attempt to repair the volume. The parity file should be able to protect itself from bit rot.

The repaired volume will be immediately re-uploaded without changing file name. *(this behavior is not finalized)*

## Remove (Compact, Purge, Delete)

When a data volume is removed in the remote, its parity file will also be removed.

## Repair

Small amount of bit rot will be automatically repaired during the download process. Meanwhile, it's possible as well to utilize the parity data during disaster recovery.

Par2 files are able to identify file name change, so if the file names are messed up in the remote, the data volumes can still be repaired with all the parity files. However, this operation requires downloading all parity files, so it should be triggered manually.

# Create parity for existing backups

[...]

PR #7020, opened 2026-07-08
Add pluggable parity module with PAR2 error-correction for remote volumes

### LINKED ISSUE #314: Add parity file support
_From [kenneth@hexad.dk](https://code.google.com/u/kenneth@hexad.dk/) on December 07, 2010 22:27:02_

If a volume archive is broken either due to transfer issues, or an issue on the backend, the data will likely not be recoverable, especially if it is encrypted.

One solution to this would be to simply store a parity file together with the volumes.

The most prominent parity application I know of this par2: http://www.par2.net/

_Original issue: http://code.google.com/p/duplicati/issues/detail?id=314_
##

JamBalaya56562:
## Summary

Adds error-correction (**parity**) data for remote data volumes so a volume can be repaired after bit-rot or corruption on the backend. This addresses #314 and revives the intent of the stale draft #4574, **redesigned from scratch** against the current `net10.0` architecture (the old PR targeted the removed mono/monolithic `BackendManager`/`BackendUploader` code, so nothing was ported).

## Design

A pluggable **`IParity`** module system is introduced, mirroring the existing compression/encryption modules:

- `Duplicati/Library/Interface/IParity.cs` + `Duplicati/Library/DynamicLoader/ParityLoader.cs`
- New `Duplicati.Library.Parity` project with the default **`Par2Parity`** implementation, which shells out to the `par2` (par2cmdline) program.

par2cmdline emits an index file **plus** one or more recovery files. To satisfy a single-parity-file contract, these are concatenated into one file — the PAR2 format is a stream of self-describing packets, so a concatenation is a valid PAR2 file. All par2 operations run in a temp working directory under a fixed canonical name so **create and repair stay symmetric** even though Duplicati's temp files are named differently on upload vs download.

## Behavior

- **Upload (Put):** after a data volume (`dblock`/`dlist`) is uploaded, a parity companion `<name>.par2` is generated from the final encrypted bytes and uploaded **unencrypted and untracked**, inline on the held backend. `dindex` is skipped (regenerable). Generated *after* the upload so it always keys off the settled filename (retry/rename safe).
- **Download (Get):** on a hash/size mismatch, the parity companion is fetched and used to **repair the file in place** before decryption. This single choke point covers Restore/Compact/Test/Repair.
- **Delete:** the parity companion is removed (best-effort) together with its data volume.
- **Parsing / listing:** parity files are recognized in `ParsedVolume.Parse` (avoiding the greedy encryption-group trap) and excluded from remote-list reconciliation so they are never flagged as *extra*.

## Options

Parity is **opt-in**:

- `--parity-redundancy-level` (percent, default `0` = off)
- `--parity-module` (default `par2`)
- `--no-parity`
- par2-specific: `--par2-program-path`, `--par2-extra-options`

It is **best-effort and never a hard dependency**: if the `par2` program is absent, the backup still succeeds with a single warning, and repair falls through to the normal hash-mismatch error. Parity is **not tracked in the database** and needs **no schema migration** — the parity name is always derivable from the data volume name.

## Tests

Added NUnit tests under `Duplicati/UnitTest/`:

- **`ParityFilenameTests`** — `GenerateParityFilename` round-trips; `ParseFilename` recognizes a `.par2` suffix as parity (`IsParity`) without the greedy encryption group swallowing it; normal volumes still parse as non-parity; non-volume names return null. *(no external dependency)*
- **`ParityLoaderTests`** — the `par2` module is discoverable via `ParityLoader` (`Keys`/`GetModule`/`GetSupportedCommands`). *(no external dependency)*
- **`Par2ParityTests`** — end-to-end create → verify → corrupt → repair, repair across a differing filename (the real upload/download scenario), and repair failing when damage exceeds redundancy. These require the `par2` program and are **skipped via `Assert.Ignore` when it is unavailable**, so CI without par2 stays green.

Result: `Failed: 0, Passed: 13` with par2 present; `Passed: 10, Skipped: 3` without it.

Also verified manually end-to-end on Linux against a `file://` backend (par2cmdline 1.2.0):

1. Backup creates a `.par2` for each `dblock`/`dlist` (not `dindex`). ✅
2. `test`/`list` do not flag parity files as extra/unknown. ✅
3. A corrupted `dblock` is transparently repaired via parity during `test`/`restore`. ✅
4. Negative control — with the `.par2` removed, the same corruption fails with the original `HashMismatchException`. ✅
5. Full `restore` with **all** dblocks corrupted: every one is repaired via parity and the restored files match the source. ✅
6. Delete cascade — `compact` removes dblocks and their parity companions, leaving no orphans. ✅
7. Absent `par2` binary — backup still succeeds with a single warning and no `.par2` files. ✅

## Scope / follow-ups

This is a focused first cut (generation + repair + delete). Intentionally deferred: retroactive creation during `TEST` for existing backups, durable re-upload of repaired volumes in `RepairHandler`, bundling a Windows `par2.exe`, WebUI options, and an optional managed (dependency-free) PAR2 engine — the `IParity` interface keeps the engine swappable.

Fixes #314. Supersedes the draft #4574.


Duplicati/Library/Main/Backend/BackendManager.GetOperation.cs
```diff
@@ -112,6 +128,61 @@ public GetOperation(string remotefilename, long size, ExecuteContext context, Ca
...
+
+            // Do not attempt to repair a parity file itself, or non-volume files
+            var parsed = Volumes.VolumeBase.ParseFilename(RemoteFilename);
+            if (parsed == null || parsed.IsParity)
+                return false;
+
+            try
+            {
+                using var parity = DynamicLoader.ParityLoader.GetModule(parityModule, Context.Options.RawOptions);
+                if (parity == null || !parity.IsAvailable)
```
kenkendk, line 170: Should we check the type of file here as well? If this is not dblocks or dlist, there _should_ be no parity file to find.

Duplicati/Library/Main/Backend/BackendManager.PutOperation.cs
```diff
@@ -333,9 +333,81 @@ private async Task PerformUploadAsync(IBackend backend, string hash, long size,
...
+                {
+                    parity.Create(LocalFilename, parityTemp);
+                }
+                catch
+                {
+                    parityTemp.Dispose();
+                    throw;
+                }
+
+                // Upload the parity file as an unencrypted, untracked companion,
```
kenkendk, line 386: Add a comment here stating that the parity file is created for the _encrypted_ volume, so no encryption is needed on the parity file (it cannot leak information).

Duplicati/Library/Main/Options.cs
```diff
@@ -460,6 +460,9 @@ private IEnumerable<ICommandLineArgument> GetDebugConditionalCommands()
 
             new CommandLineArgument("encryption-module", CommandLineArgument.ArgumentType.String, Strings.Options.EncryptionmoduleShort, Strings.Options.EncryptionmoduleLong, "aes"),
             new CommandLineArgument("compression-module", CommandLineArgument.ArgumentType.String, Strings.Options.CompressionmoduleShort, Strings.Options.CompressionmoduleLong, "zip"),
+            new CommandLineArgument("parity-module", CommandLineArgument.ArgumentType.String, Strings.Options.ParitymoduleShort, Strings.Options.ParitymoduleLong, "par2"),
```
kenkendk: Default should be empty string, so you opt-in to parity support until it is more tested

Duplicati/Library/Main/Options.cs
```diff
@@ -460,6 +460,9 @@ private IEnumerable<ICommandLineArgument> GetDebugConditionalCommands()
...
+            new CommandLineArgument("parity-redundancy-level", CommandLineArgument.ArgumentType.Integer, Strings.Options.ParityredundancylevelShort, Strings.Options.ParityredundancylevelLong, "0"),
```
kenkendk: The option `parity-redundancy-level` option should be in the parity module's options. The `no-parity` should be removed as that is toggled by an empty string for the parity module.

Duplicati/Library/Main/Backend/BackendManager.GetOperation.cs
```diff
@@ -88,11 +88,27 @@ public GetOperation(string remotefilename, long size, ExecuteContext context, Ca
...
+                    var sizeMismatch = Size >= 0 && dataSizeDownloaded != Size;
+                    var hashMismatch = !string.IsNullOrEmpty(Hash) && fileHash != Hash;
 
-                    if (!string.IsNullOrEmpty(Hash) && fileHash != Hash)
-                        throw new HashMismatchException(Strings.Controller.HashMismatchError(RemoteFilename, Hash, fileHash));
+                    if (sizeMismatch || hashMismatch)
+                    {
+                        // The downloaded file failed verification; attempt a best-effort
+                        // repair using a parity companion file, if one is available.
+                        if (await TryRepairWithParityAsync(backend, tmpfile, cancelToken).ConfigureAwait(false))
```
kenkendk: If this is the test flow, we should not repair the file here as that would mask that files are actually damaged.

Even for normal operations, I would argue that needing to repair the file should emit a warning in any case, so it is visible that the backup data is no longer in prime condition.

Duplicati/Library/Parity/Par2Parity.cs
```diff
@@ -0,0 +1,285 @@
...
+                m_explicitPath = true;
+            }
+            else
+            {
+                m_programpath = DEFAULT_PROGRAM;
+            }
+
+            m_extraOptions = options.TryGetValue(COMMANDLINE_OPTIONS_EXTRA, out var extra) ? (extra ?? string.Empty) : string.Empty;
+
+            m_redundancy = DEFAULT_REDUNDANCY;
```
kenkendk: This parsing should use Library.Utility.Utility.ParseIntOption which ensures we parse all options the same way:
```
m_redundancy = Library.Utility.Utility.ParseIntOption (options, COMMANDLINE_OPTIONS_REDUNDANCY, DEFAULT_REDUNDANCY);
```

Maybe validate it is not zero or less.

Duplicati/Library/Parity/Par2Parity.cs
```diff
@@ -0,0 +1,285 @@
...
+        /// Any extra options supplied to par2 when creating parity
+        /// </summary>
+        private readonly string m_extraOptions;
+
+        /// <summary>
+        /// Process-wide cache of program availability, keyed by the resolved program path.
+        /// A new module instance is created per volume, so this avoids re-probing the
+        /// system (and re-warning) for every upload.
+        /// </summary>
+        private static readonly System.Collections.Concurrent.ConcurrentDictionary<string, bool> _availabilityCache = new();
```
kenkendk: I think this should not be static, so the Par2Parity instance lives for the duration of the operation (backup/restore).

Same for the `_warnedNotFound` value, so each operation reports it. Here we need to ensure that `IsAvailable` is not called on an instance created with the default constructor, but only when fully initialized, as the path could otherwise be different.

Duplicati/Library/Interface/IParity.cs
```diff
@@ -0,0 +1,85 @@
...
+    /// </summary>
+    public interface IParity : IDynamicModule, IDisposable
+    {
+        /// <summary>
+        /// Creates parity data for <paramref name="inputfile"/>, writing a single
+        /// parity file to <paramref name="parityfile"/>.
+        /// </summary>
+        /// <param name="inputfile">The file to protect</param>
+        /// <param name="parityfile">The parity file to create</param>
+        void Create(string inputfile, string parityfile);
```
kenkendk: The `Create`/`Verify`/`Repair` methods should be async, so they can do deferred processing.

Duplicati/Library/Parity/Par2Parity.cs
```diff
@@ -0,0 +1,285 @@
...
+            };
+
+            try
+            {
+                using var p = Process.Start(psi) ?? throw new Exception("Unexpected failure to start par2 process");
+                // Read both streams asynchronously to avoid a deadlock if a buffer fills up
+                var outTask = p.StandardOutput.ReadToEndAsync();
+                var errTask = p.StandardError.ReadToEndAsync();
+                p.WaitForExit();
+                var stderr = errTask.GetAwaiter().GetResult();
```
kenkendk: Once the interface methods are async, this can be regular `await` calls.

Duplicati/Library/Parity/ParityBase.cs
kenkendk, line 1: I suggest removing this file as it only has the dispose logic and is otherwise just a copy of the interface.

Duplicati/UnitTest/Duplicati.UnitTest.csproj
kenkendk, line 1: I would suggest also adding a test that makes a backup, flips a few bits, and then checks the recovery works, so we test the parity module in the full backup/restore flow.

This could also test that the backup correctly detects the corruption, and then the parity files can be exercised to recover the data.

kenkendk (changes requested):

kenkendk:
@JamBalaya56562 Amazing work here, this is a great addition to Duplicati.

Duplicati/Library/Main/Backend/BackendManager.GetOperation.cs
JamBalaya56562: Good catch — added the same `FileType` guard used in the Put/Delete paths, so we only try to repair `Blocks`/`Files` volumes. For anything else there is no parity companion to find.

Duplicati/Library/Main/Backend/BackendManager.PutOperation.cs
JamBalaya56562: Added a comment explaining that the parity is created from the already-encrypted volume on disk, so the parity file itself needs no encryption — it cannot leak information about the source data.

Duplicati/Library/Main/Options.cs
JamBalaya56562: Done — `parity-module` now defaults to an empty string, so parity is opt-in until it is more widely tested. The `ParityModule` accessor returns `null` for an empty value, and all consumers already guard on `IsNullOrEmpty`.

Duplicati/Library/Main/Options.cs
JamBalaya56562: Done. `parity-redundancy-level` is now registered in the par2 module's `SupportedCommands` (default 5), and `no-parity` is removed — parity is toggled solely by an empty vs non-empty `parity-module`.

One note: `parity-redundancy-level` is now par2-specific rather than a shared/general parity option, so a future module would expose its own redundancy option. Happy to make it a shared option again if you'd prefer that.

Duplicati/Library/Main/Backend/BackendManager.GetOperation.cs
JamBalaya56562: Both addressed:

- The test/verification flow no longer repairs. I added an `AllowParityRepair` flag on the download operation, threaded through `GetAsync`/`GetDirectAsync`/`GetWithInfoAsync`/`GetFilesOverlapped*`, and `TestHandler` now passes `allowParityRepair: false` — so verification reports damage instead of silently masking it.
- For normal operations, a successful parity repair now emits a **warning** (previously Information), so it is visible that the remote copy is damaged.

Duplicati/Library/Parity/Par2Parity.cs
JamBalaya56562: Done — switched to `Library.Utility.Utility.ParseIntOption`, and it now rejects a non-positive redundancy with a `UserInformationException`.

Duplicati/Library/Parity/Par2Parity.cs
JamBalaya56562: Done. `_availabilityCache`/`_warnedNotFound` are now per-instance (a lazy `bool?` plus a `bool`). The `BackendManager` holds a single shared parity instance for the whole operation (a `SharedParityProvider` created from the options and disposed with the manager), and Put/Get/Delete reuse it instead of constructing one per volume — so the availability probe and any warning happen once per operation.

`IsAvailable` is only ever called on this fully-initialized (options-constructed) instance, never on the default-constructor metadata instance, so the program path is always the configured one.

Duplicati/Library/Interface/IParity.cs
JamBalaya56562: Done — `Create`/`Verify`/`Repair` are now `CreateAsync`/`VerifyAsync`/`RepairAsync(..., CancellationToken)` returning `Task`/`Task<bool>`. All callers and tests updated.

Duplicati/Library/Parity/Par2Parity.cs
JamBalaya56562: Done — now that the interface is async, `RunPar2Async` uses `await p.WaitForExitAsync(ct)` and awaits the stdout/stderr reads directly, so the `GetAwaiter().GetResult()` calls are gone.

Duplicati/Library/Parity/ParityBase.cs
JamBalaya56562: Done — removed `ParityBase.cs`. `Par2Parity` implements `IParity` directly with its own (no-op) `Dispose()`.

Duplicati/UnitTest/Duplicati.UnitTest.csproj
JamBalaya56562: Added `Par2ParityIntegrationTests.cs`: it makes a backup with parity enabled, asserts a `.par2` companion is created for each data volume, flips bytes in a data volume, checks that `test` **detects** the corruption (repair is disabled in the test flow), and then restores — confirming the data is **recovered** via parity. It is skipped when par2 is unavailable.

Verifying this against real par2 (par2cmdline 0.8.1) in a Linux container also surfaced a bug in the module itself: par2 aborts (`std::out_of_range` / "terminate called without an active exception") when the protected file has a **single-character** base name, which is exactly what the module used (`CANONICAL_NAME = "d"`). I renamed it to `"data"`; with a 2+ character base name par2 works correctly. Without this fix the feature would never actually create or use parity files with the stock par2 binary. After the fix, all parity tests (including the full backup→corrupt→detect→restore→recover flow) pass against real par2.

JamBalaya56562:
Thanks for the detailed review! I've pushed a commit addressing all of the comments; replies are inline on each thread. Summary:

**Parity module**
- `IParity` `Create`/`Verify`/`Repair` are now async (`...Async(..., CancellationToken)`); the par2 runner uses `WaitForExitAsync` and awaits the output reads directly.
- Removed `ParityBase`; `Par2Parity` implements `IParity` directly.
- `parity-redundancy-level` is parsed with `Utility.ParseIntOption` and validated as positive, and lives in the par2 module's own options now.
- Availability probe + warning are per-instance; the module instance lives for the whole operation.

**Options**
- `parity-module` now defaults to empty (opt-in), and is the sole toggle for parity — `no-parity` is removed.

**Backend**
- The download repair path guards on `Blocks`/`Files` volume type; a successful repair emits a **warning**; the verification/test flow no longer repairs (so damage is reported, not masked). `BackendManager` keeps one shared parity instance per operation.

**Tests / verification**
- Added a full backup → corrupt → detect → restore/recover integration test.
- Verified end-to-end against **real par2 (par2cmdline 0.8.1)** in a Linux container — all parity tests pass, and I ran the delete/compact + a backup/restore/repair cycle to check for non-parity regressions.

**One fix worth flagging:** running against real par2 revealed that par2cmdline 0.8.1 **aborts** (`std::out_of_range` on `basic_string::at`) when the protected file has a **single-character** base name — which is what the module used internally (`CANONICAL_NAME = "d"`). I renamed it to `"data"`. Without this, the feature never actually produced/used parity files with the stock par2 binary. It reproduces across Debian 12/13 and Ubuntu 22.04/24.04 par2 builds, and goes away with any 2+ character name.

One open question is on the C4 thread: `parity-redundancy-level` is now par2-specific rather than a shared parity option — let me know if you'd prefer it to remain shared.

kenkendk (approved):

mxxcon:
> @JamBalaya56562 Amazing work here, this is a great addition to Duplicati.

I hope you do realize that it's all AI code...?

PR #7052, opened 2026-07-13
Improve handling of missing metadata

kenkendk:
When repairing a database with zero-length metadata entries, the repair process now respects the `--disable-replace-missing-metadata` option. If enabled, the repair will skip fabrication of new metadata to avoid unintended metadata changes, leaving
`purge-broken-files` command to handle it.

Additionally, the repair logic now attempts to relink existing empty metadata blocks if their `BlocksetEntry` is missing. This ensures that folders and symlinks can pass consistency checks without requiring new remote storage access, provided the backing block still exists in the local database.

Added unit tests to verify that the repair process correctly relinks empty metadata blocks and maintains database consistency.

Additionally, the backup process now proactively stores a 2-byte empty metadata block such that repairs have a reliable metadata entry to point to later. The 2-byte block is also kept during compacts even if it is not needed, because its main purpose is to be available in error scenarios.


kenkendk:
Rewrote this Rewrote with #7212, #7213, #7215, #7220, #7226, #7229, #7234

PR #7097, opened 2026-07-24
Remove relaxed url parsing

kenkendk:
This update removes the support for parsing relaxed urls.

The relaxed url parsing was introduced to allow simpler remote destinations when using the commandline, where escaping things like passwords or folder names could be difficult to do correctly.

Additionally, since there were issues with the `System.Uri` class under Mono, a replacement `Utility.Uri` class was introduced that doubled both as a relaxed url parser and a replacement.

Today we have no issues with Mono, relatively few direct CLI users, and are routinely discovering edge cases where the relaxed parsing is causing issues.

To be more standardized in the future, this PR renames the  `Utility.Uri` class to `Utility.LegacyUri` and defaults to not using it. An intermediate `Utility.CompatUri` is introduced that works by choosing which underlying `Uri` class to use.

By default it will use the .NET default `System.Uri` class, which requires strict url encoding.

For most users, the new UI will already have encoded urls correctly, so the change should have limited impact.

But, since there may be users that rely on the relaxed parsing, and essentially every backend is affected, this change has an escape hatch. A user may either set the `DUPLICATI_LEGACY_URL_PARSING-true` environment variable or place a `legacy-url-parsing.txt` file in the install folder to toggle the use of the `LegacyUri` class.

### DIFF
--- a/Duplicati/CommandLine/BackendTool/Program.cs
+++ b/Duplicati/CommandLine/BackendTool/Program.cs
@@ -104,7 +104,7 @@ namespace Duplicati.CommandLine.BackendTool
                                where n is Library.Interface.IConnectionModule
                                select n).ToArray();
 
-                var uri = new Library.Utility.Uri(args[1]);
+                var uri = new Library.Utility.CompatUri(args[1]);
                 var qp = uri.QueryParameters;
 
                 var backendOpts = new Dictionary<string, string>();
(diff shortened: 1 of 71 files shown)


kenkendk:
Closing this as the blast radius is too large, and the new `CompatUri` essentially just moves the unwanted logic into a new place with new places where it can fail.

PR #7101, opened 2026-07-26
Use the requested path when looking up a Box entry

JamBalaya56562:
## What this fixes

`BoxBackend.GetEntryAsync` builds its lookup from the backend's own configured path instead of the path it was asked for:

```csharp
public async Task<IFileEntry?> GetEntryAsync(string path, CancellationToken cancellationToken)
{
    var parts = GetAbsolutePath(_path).Split(...);   // the argument is never used
```

`GetAbsolutePath` returns `_path` for an empty argument and otherwise appends to `_path`. Since `_path` is set as `Util.AppendDirSeparator(uri.HostAndPath, "/")` it always ends in `/` and is never empty, so feeding it back through the helper **doubles the last segment**. There is no existing issue for this (searched open+closed for `GetEntryAsync`, Box folder browsing, `destination/list`), so this PR doubles as the report.

Two distinct broken behaviors, depending on the url:

| Box url | `_path` | built lookup path | result |
|---|---|---|---|
| `box://folder/sub` | `folder/sub/` | `folder/sub/folder/sub` | always **null** — every entry reports as missing |
| `box://` | `/` | `/` → no segments | **the root entry is returned for every requested path** — no error, wrong contents |

## Where it shows up

`GetEntryAsync` is the single-entry lookup behind remote-source and destination browsing. Backup and restore to a Box *destination* are unaffected — `BackendManager.ListOperation` only uses `ListAsync`.

- `WebserverCore/Endpoints/V2/DestinationList.cs:65` — `POST /destination/list` treats null as `"Folder does not exist"`, and otherwise enumerates the returned entry's children. So browsing into a Box folder either fails, or silently lists the **root's** contents while the user believes they are inside the folder they clicked.
- `Library/Main/Operation/Sync/SyncHandler.cs:174` — a source root that reports as missing is logged as `SourceRootMissing` and skipped.
- `Library/SourceProvider/Builtin/BackendSourceFileEntry.cs:187` — `FileExists` is `entry != null && !entry.IsFolder`, so it is always false and exclusion-marker files on a remote source are never found.

## History — a refactoring slip, not a decision

- Added in [8a08b35546bd](https://github.com/duplicati/duplicati/commit/8a08b35546bd) (2026-01-28, PR #6732), where the argument **was** honoured, combined inline:
  ```csharp
  var fullPath = _path; if (!string.IsNullOrEmpty(path)) { var p = path.TrimStart('/'); … }
  ```
- Broken in [bb78aa57ae15](https://github.com/duplicati/duplicati/commit/bb78aa57ae15) (2026-01-29, commit 6 of the same PR, "More work on Box paths"), which replaced that block with the new `GetAbsolutePath` helper and substituted `_path` for `path`.

So the argument has never been honoured in any merged state, and `sl blame` attributes only that one line to the later commit.

## The fix

```csharp
var parts = GetAbsolutePath(path).Split(new[] { '/' }, StringSplitOptions.RemoveEmptyEntries);
```

The rest of the method already assumes this: `parts.Length == 0` then means "the root was requested", `parentPath`/`name` split the requested path, `FolderMissingException` and a missing child both yield `null`, and `ParseEntry` already sets `IsFolder` and the trailing separator the interface asks for. It also makes the method symmetric with its sibling `ListAsync(string? path, …)`, which already does `GetAbsolutePath(path)`.

For reference, Box was the only one of the nine `IFolderEnabledBackend.GetEntryAsync` implementations that ignored the argument — `GoogleDrive.cs:204` has the identical helper and the identical structure and does `var fullPath = GetAbsolutePath(path);`.

## Verification

New offline tests in `Duplicati/UnitTest/BoxEntryLookupTests.cs` driving the backend against a stubbed `HttpMessageHandler` (the OAuth endpoint is stubbed too and points at an unresolvable host, so no request can escape).

| Test | Before this change |
|---|---|
| `RequestedFolderIsReturned` | fails — `null` |
| `RequestedFileIsReturned` | fails — `null` |
| `ConfiguredFolderIsReturnedForAnEmptyPath` | fails — `null`, which is what makes browsing report "Folder does not exist" |
| `RequestedFolderIsReturnedForABackendAtTheRoot` | fails — returns the root entry: `Expected: "child/" But was: <string.Empty>` |
| `MissingEntryIsReportedAsMissing` | passes before and after (guard: missing stays `null`, not an error) |

`Category=Backend` is green locally on Windows/net10.0. I have no Box account, so the live `Box.com Tests` job is the only end-to-end check; the evidence here is the layer-level red→green plus the paths above.

**Overlap with #7099**: this PR contains the same test seam as [#7099](https://github.com/duplicati/duplicati/pull/7099) — an `internal` `BoxBackend` constructor taking an `HttpClient` (the base `OAuthHelperHttpClient` already accepts one), the `BoxHelper` parameter, and `InternalsVisibleTo`. The edits are byte-identical in both branches so they merge as the same change, the test files have different names, and the one-word fix here is in a different part of the file from #7099's changes. If #7099 lands first, this PR reduces to that one word plus the tests. The public two-argument and parameterless constructors are untouched, since `BackendLoader` creates backends via `Activator.CreateInstance(type, url, options)`.

## Out of scope

`Idrivee2Backend.cs:280` and `SSHv2Backend.cs:649` implement `GetEntryAsync` as `Task.FromResult<IFileEntry?>(null)`, so browsing a remote source on those backends always reports the entry as missing. Left alone here.

### DIFF
--- a/Duplicati/Library/Backend/Box/BoxBackend.cs
+++ b/Duplicati/Library/Backend/Box/BoxBackend.cs
@@ -30,6 +30,8 @@ using Duplicati.Library.Utility;
 using Duplicati.Library.Utility.Options;
 using Newtonsoft.Json;
 
+[assembly: InternalsVisibleTo("Duplicati.UnitTest")]
+
 namespace Duplicati.Library.Backend.Box
 {
     public class BoxBackend : IStreamingBackend, IRenameEnabledBackend, IFolderEnabledBackend
@@ -53,8 +55,8 @@ namespace Duplicati.Library.Backend.Box
         private class BoxHelper : OAuthHelperHttpClient
         {
             private readonly TimeoutOptionsHelper.Timeouts _timeouts;
-            public BoxHelper(AuthIdOptionsHelper.AuthIdOptions authId, TimeoutOptionsHelper.Timeouts timeouts)
-                : base(authId.AuthId, "box.com", authId.OAuthUrl)
+            public BoxHelper(AuthIdOptionsHelper.AuthIdOptions authId, TimeoutOptionsHelper.Timeouts timeouts, HttpClient? httpClient = null)
+                : base(authId.AuthId, "box.com", authId.OAuthUrl, httpClient)
             {
                 AutoAuthHeader = true;
                 _timeouts = timeouts;
@@ -105,6 +107,18 @@ namespace Duplicati.Library.Backend.Box
         }
 
         public BoxBackend(string url, Dictionary<string, string?> options)
+            : this(url, options, null)
+        {
+        }
+
+        /// <summary>
+        /// Creates a backend that uses the supplied <see cref="HttpClient"/>,
+        /// so the Box API responses can be stubbed in tests
+        /// </summary>
+        /// <param name="url">The backend url</param>
+        /// <param name="options">The options to use</param>
+        /// <param name="httpClient">The client to use, or null to create one</param>
+        internal BoxBackend(string url, Dictionary<string, string?> options, HttpClient? httpClient)
         {
             var uri = new RelaxedUri(url);
 
@@ -116,7 +130,7 @@ namespace Duplicati.Library.Backend.Box
             _deleteFromTrash = Utility.Utility.ParseBoolOption(options, REALLY_DELETE_OPTION);
             _timeouts = TimeoutOptionsHelper.Parse(options);
 
-            _oAuthHelper = new BoxHelper(authid, _timeouts);
+            _oAuthHelper = new BoxHelper(authid, _timeouts, httpClient);
 
         }
 
@@ -224,7 +238,7 @@ namespace Duplicati.Library.Backend.Box
 
         public async Task<IFileEntry?> GetEntryAsync(string path, CancellationToken cancellationToken)
         {
-            var parts = GetAbsolutePath(_path).Split(new[] { '/' }, StringSplitOptions.RemoveEmptyEntries);
+            var parts = GetAbsolutePath(path).Split(new[] { '/' }, StringSplitOptions.RemoveEmptyEntries);
             if (parts.Length == 0)
                 return new FileEntry(string.Empty) { IsFolder = true };
 
--- a/Duplicati/UnitTest/BoxEntryLookupTests.cs
+++ b/Duplicati/UnitTest/BoxEntryLookupTests.cs
@@ -1,0 +1,178 @@
+// Copyright (C) 2026, The Duplicati Team
+// https://duplicati.com, hello@duplicati.com
+//
+// Permission is hereby granted, free of charge, to any person obtaining a
+// copy of this software and associated documentation files (the "Software"),
+// to deal in the Software without restriction, including without limitation
+// the rights to use, copy, modify, merge, publish, distribute, sublicense,
+// and/or sell copies of the Software, and to permit persons to whom the
+// Software is furnished to do so, subject to the following conditions:
+//
+// The above copyright notice and this permission notice shall be included in
+// all copies or substantial portions of the Software.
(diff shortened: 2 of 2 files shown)


Duplicati/Library/Backend/Box/BoxBackend.cs
```diff
@@ -225,7 +239,7 @@ public async IAsyncEnumerable<IFileEntry> ListAsync(string? path, [EnumeratorCan
 
         public async Task<IFileEntry?> GetEntryAsync(string path, CancellationToken cancellationToken)
         {
-            var parts = GetAbsolutePath(_path).Split(new[] { '/' }, StringSplitOptions.RemoveEmptyEntries);
+            var parts = GetAbsolutePath(path).Split(new[] { '/' }, StringSplitOptions.RemoveEmptyEntries);
```
kenkendk: This does not look correct. If the backend is configured as `box://folder` we need to prepend the `folder` to the `path` argument to get the full path. Otherwise all paths will be resolved as absolute.

See for instance `GetAbsolutePath` in the Dropbox.cs backend.

JamBalaya56562: Thanks for taking a look.

I believe `GetAbsolutePath` already does the prepending you describe — it turns a *relative* path into the absolute one by putting `_path` in front:

```csharp
private string GetAbsolutePath(string? path)
{
    if (string.IsNullOrWhiteSpace(path))
        return _path;

    var p = path.Replace(Path.DirectorySeparatorChar, '/').Trim('/');
    if (string.IsNullOrWhiteSpace(_path) || _path == "/")
        return "/" + p;
    else
        return Util.AppendDirSeparator(_path, "/") + p;   // <- prepends the configured folder
}
```

So for `box://folder`, `GetAbsolutePath("child")` returns `folder/child`, not `/child`.

The Dropbox backend you pointed at is built the same way, and its `GetEntryAsync` passes the argument through that helper:

```csharp
// Dropbox.cs:129-144
private string GetAbsolutePath(string? path)
{
    ...
    p = Util.AppendDirSeparator(m_path, "/") + p;
    ...
}

// Dropbox.cs:162-164
public async Task<IFileEntry?> GetEntryAsync(string path, CancellationToken cancellationToken)
{
    var fullPath = GetAbsolutePath(path);
```

`ListAsync(string? path, …)` in this same file also already does `GetAbsolutePath(path)`, so this change mainly makes the two methods symmetric.

For what it is worth, here is what the previous `GetAbsolutePath(_path)` produced. `_path` is assigned as `Util.AppendDirSeparator(uri.HostAndPath, "/")`, so it always ends in `/` and is never empty — feeding it back into the helper doubles the last segment:

| url | lookup path that was built | result |
|---|---|---|
| `box://folder/sub` | `folder/sub/folder/sub` | always `null`, every entry reported as missing |
| `box://` | `/` → no segments | the root entry returned for every requested path |

The new tests are arranged so that losing the configured folder would fail them: in `RequestedFolderIsReturned` the stubbed root contains only `target` (the configured folder), and `child` exists **only inside** `target`. If the argument were resolved from the root, the lookup would find nothing and the test would fail on `null`. `RequestedFolderIsReturnedForABackendAtTheRoot` covers the `box://` case, where the requested folder really is at the root.

If you have a caller in mind that passes a path which already contains the configured folder, I am happy to add a test for it and adjust the change. I could not find one — `BackendSourceProvider` passes a backend-relative path, which matches the `ListAsync` documentation on `IFolderEnabledBackend` — but I may well have missed something.

kenkendk: You are correct, I should have read the `GetAbsolutePath` method.

PR #7120, opened 2026-08-01
Decode the %uXXXX escape as a code unit instead of as bytes

JamBalaya56562:
`UrlDecode` recognises the IIS style `%uXXXX` escape but decodes it as bytes, so it produces characters nobody encoded.

### Why

Both alternatives in the regular expression capture into a group named `number`:

```csharp
private static readonly Regex RE_NUMBER = new Regex(
    @"(\%(?<number>([0-9]|[a-f]|[A-F]){2}))|(\+)|(\%u(?<number>([0-9]|[a-f]|[A-F]){4}))");
```

so the handler cannot tell a `%XX` from a `%uXXXX`, and both go down the byte path:

```csharp
var hex = m.Groups["number"].Value;
var bytelen = hex.Length / 2;                            // four digits read as two bytes
Utility.HexStringAsByteArray(hex, inbuf);
var c = decoder.GetChars(inbuf, 0, bytelen, outbuf, 0);  // handed to the UTF-8 decoder
```

The four digits of a `%uXXXX` escape are a **UTF-16 code unit**, not a pair of bytes. `%u00e6` is U+00E6, but it decodes to `"\0"`: the bytes come out as `00 E6`, `0x00` is a NUL, and the trailing `0xE6` opens a three byte sequence that never completes. A surrogate pair fares worse — `%uD83D%uDE00` yields four characters of rubbish instead of the one it encodes.

Whichever way you read the intent, the current output is wrong. If the escape should be recognised, the answer is `æ`. If it should not, the answer is the literal `%u00e6`. It is neither.

### Which reading is right

`UrlDecode` is documented as behaving "like `System.Web.HttpUtility.UrlDecode`", so I measured what that actually does rather than assume. It decodes `%u00e6` to `æ` — the code unit reading. That settles it, and the tests assert against `HttpUtility` directly so the intent is recorded rather than restated in a comment.

### The change

Give the `%uXXXX` alternative its own group and handle it before the byte path:

```csharp
@"(\%(?<number>([0-9]|[a-f]|[A-F]){2}))|(\+)|(\%u(?<unicode>([0-9]|[a-f]|[A-F]){4}))"
```

```csharp
// The four digits of a %uXXXX escape are a UTF-16 code unit,
// not bytes, so they must not go through the byte decoder
var unicode = m.Groups["unicode"];
if (unicode.Success)
    return ((char)Convert.ToUInt16(unicode.Value, 16)).ToString();
```

The byte path is untouched. Surrogate pairs work out because each escape contributes one code unit and they are emitted in order.

### This is not new

The same code is in `Uri.cs` before #7111 renamed it, and in versions before that. I ran into it while writing characterization tests for #7119, which moves this block to another class without changing it. Nothing in the repo documents or tests the `%uXXXX` behaviour, so there was nothing pinning the broken result in place.

### Who reaches it

`UrlDecode` runs over both user input and server output:

- `RelaxedUri`'s own parsing — the host, path, **username and password** of a connection url
- `WEBDAV.cs` — the `href` and file names **the server returns**, in five places
- Dropbox, OneDrive and SSHv2 paths

Duplicati's own `UrlEncode` never emits `%uXXXX`, so a url it built itself round-trips fine either way. What it takes to hit this is a `%u` followed by four hex digits arriving from outside. That is not common, and I am not claiming it is; the WebDAV path is the one that is not under the user's control, since those values come from the server.

### Tests

Added to `Duplicati/UnitTest/UriUtilityTests.cs`.

| Test | Before | After |
| --- | --- | --- |
| `UnicodeEscapeIsDecodedAsCodeUnit` | `%u00e6` → `"\0"` | `"æ"` |
| `UnicodeEscapeMatchesTheFramework` | differs from `HttpUtility` | matches |
| `SurrogatePairIsDecoded` | `%uD83D%uDE00` → 4 chars of rubbish | `"😀"` |
| `ByteEscapesStillDecode` | passes | passes |

What they reported against unmodified code:

```
失敗 UnicodeEscapeIsDecodedAsCodeUnit
  Expected: "æ"
  But was:  "\0"

失敗 UnicodeEscapeMatchesTheFramework
  Expected: "æ"          <- System.Web.HttpUtility.UrlDecode("%u00e6")
  But was:  "\0"

失敗 SurrogatePairIsDecoded
  Expected string length 2 but was 4. Strings differ at index 0.
  Expected: "😀"
  But was:  "�=�\0"
```

`ByteEscapesStillDecode` was green before and after; it is there so the fix cannot quietly break the path that works.

### Not addressed here

The `decoder` is stateful and shared across the whole replacement, so a `%uXXXX` sitting inside an incomplete UTF-8 sequence — `%C3%u0041%A6` — still leaves a byte pending that attaches to the next match. Valid input does not produce that shape, and it was equally broken before, so I left it alone rather than widen the change.

### How this was verified

- The output above is what the tests actually reported, including the `HttpUtility` value that decided the direction
- `dotnet build Duplicati.slnx -c Debug` — clean
- `Category=Utility|Category=UriUtility` — green

**I have no WebDAV server that emits `%uXXXX` and no user reporting it, so I cannot show that the notation is encountered in practice.** The verification is unit tests over string inputs; it covers the mapping from input to output and nothing about how often that input occurs.

### Relation to #7119

#7119 moves this whole block to a new `UrlEncoding` class without changing it. Whichever of the two lands first, the other needs a rebase; on #7119's side the move is scripted, so it is mechanical. This one is deliberately based on master rather than stacked, since a bug fix should not wait on a refactor.

### DIFF
--- a/Duplicati/Library/Utility/UrlEncoding.cs
+++ b/Duplicati/Library/Utility/UrlEncoding.cs
@@ -97,7 +97,7 @@ namespace Duplicati.Library.Utility
         /// <summary>
         /// The regular expression that matches %20 type values in a querystring
         /// </summary>
-        private static readonly System.Text.RegularExpressions.Regex RE_NUMBER = new System.Text.RegularExpressions.Regex(@"(\%(?<number>([0-9]|[a-f]|[A-F]){2}))|(\+)|(\%u(?<number>([0-9]|[a-f]|[A-F]){4}))", System.Text.RegularExpressions.RegexOptions.Compiled);
+        private static readonly System.Text.RegularExpressions.Regex RE_NUMBER = new System.Text.RegularExpressions.Regex(@"(\%(?<number>([0-9]|[a-f]|[A-F]){2}))|(\+)|(\%u(?<unicode>([0-9]|[a-f]|[A-F]){4}))", System.Text.RegularExpressions.RegexOptions.Compiled);
 
         /// <summary>
         /// Decodes a URL, like System.Web.HttpUtility.UrlDecode
@@ -123,6 +123,12 @@ namespace Duplicati.Library.Utility
 
                 try
                 {
+                    // The four digits of a %uXXXX escape are a UTF-16 code unit,
+                    // not bytes, so they must not go through the byte decoder
+                    var unicode = m.Groups["unicode"];
+                    if (unicode.Success)
+                        return ((char)Convert.ToUInt16(unicode.Value, 16)).ToString();
+
                     var hex = m.Groups["number"].Value;
                     var bytelen = hex.Length / 2;
                     Utility.HexStringAsByteArray(hex, inbuf);
--- a/Duplicati/UnitTest/UrlEncodingTests.cs
+++ b/Duplicati/UnitTest/UrlEncodingTests.cs
@@ -65,6 +65,36 @@ namespace Duplicati.UnitTest
 
         [Test]
         [Category("Utility")]
+        public static void UnicodeEscapeIsDecodedAsCodeUnit()
+        {
+            // The four hex digits of a %uXXXX escape are a UTF-16 code unit, not
+            // a pair of bytes to hand to the byte decoder
+            Assert.AreEqual("æ", Library.Utility.UrlEncoding.UrlDecode("%u00e6"));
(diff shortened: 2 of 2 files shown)


kenkendk (approved):

kenkendk:
This needs to be moved after the UrlUtility was introduced, but change itself is good.

PR #7128, opened 2026-08-09
Re-create missing index files after the block volumes are processed

JamBalaya56562:
Takes over #7055, which has been a draft since 2026-07-13 and states:

> This PR singles out a flaky issue in tests where the upload is set to have zero retries. In this case, a (simulated) failing uploaded can cause an in-flight index file to be incorrectly recorded in the database.
>
> **In the current version, no fix is present, but the issue is now reproducible.**

The first commit here is @kenkendk's, unchanged: [`f381c6b8d16e2759d27456567fd8ccd77f83686a`](https://github.com/duplicati/duplicati/commit/f381c6b8d16e2759d27456567fd8ccd77f83686a) — the deterministic reproducer. The second commit is the fix and additional tests.

## The flake is real and not test-local

`TestUploadExceptionWithNoRetriesAsync` has failed on unrelated branches:

- #7101 (ubuntu, 1 of 1493 tests)
- `feature/improve-ms365-userlisting` on 2026-08-01

It fails on a warning rather than an assertion:

```
[Warning-Duplicati.Library.Main.Operation.TestHandler-FaultyIndexFiles]:
Found 3 faulty index files, repairing now
```

## Root cause

The race itself only loses one in-flight dindex. Cleanup and re-creation then work mechanically — the lost index volumes are marked deleted and new ones are created, uploaded and linked. **The re-created index files are empty.**

Instrumenting `IndexVolumeCreator.CreateIndexVolumeAsync` on the reproducer:

```
PROBE recreate ...dindex.zip.aes for block ...dblock.zip.aes (volID=19):
  policy=Full blocks=1017 blocklistsFromDb=0 wrote=0
  :: blocksInVol=1017 blhTotal=21 blhInThisVol=0 bsEntries=6150 blocksets=9
```

The policy is `Full` and the block volume holds 1017 blocks, yet zero blocklists are retrieved and zero are written. Counting the same tables again at verification time, in the same backup:

| | `BlocklistHash` | `BlocksetEntry` | `Blockset` |
|---|---|---|---|
| when the index file is re-created | 21 | 6150 | 9 |
| when the backup is verified | **42** | **12295** | **13** |

Every hash later reported as `Missing` was present in the database by verification time (`inSql=True`), resolving to the linked block volume.

So an interrupted backup leaves a block volume recorded as `Uploaded` while the blocklists it carries are only registered by the *following* backup. `RecreateMissingIndexFiles` ran at the very start of that following backup, before file processing, so it read the database before the registration happened, produced an index file with no blocklists, uploaded it, and added the `IndexBlockLink` as if it were complete. The post-backup verification is then correct to report the index file as missing entries.

Note that this is not only a test problem: whether the run fails or silently passes depends on whether the registration lands before verification. When it does not, an index file that is missing all of its blocklists is left on the remote and recorded as good.

## The fix

Move the `RecreateMissingIndexFiles` call from before file processing to after the block volumes have been flushed, just before the dlist is sent.

At that point the blockset and blocklist registration for the run is complete and committed, and `FlushBackendAsync` has already waited for the block volume uploads to settle, so `GetBlocklistsAsync` can return the correct result.

Nothing between the old and the new position depends on the re-created index file — `FilelistProcessor.VerifyRemoteList` runs earlier, during the pre-backup verification, and the dlist does not reference dindex files. The "make sure the backup does not start before the files are settled" guarantee that motivated the early placement is preserved: `RecreateMissingIndexFiles` still ends with `WaitForEmptyAsync` and a flush when it uploaded anything, and the following `FlushBackendAsync` settles it again before verification.

A side effect worth having: `GetMissingIndexFilesAsync` returns uploaded block volumes with no `IndexBlockLink`, so a block volume that loses its index file to this race is now healed within the same backup instead of the next one.

I also considered the narrower fix of refusing to treat a blocklist-free index file as complete inside `IndexVolumeCreator`. It cannot work: at re-creation time the database is internally consistent — there genuinely are no blocklist rows for that volume yet — so there is nothing local to detect.

## Tests

`RecreateMissingIndexFiles`, `IndexVolumeCreator.CreateIndexVolumeAsync` and `GetMissingIndexFilesAsync` had **no test coverage at all**, which is why an index file with no blocklists could be uploaded unnoticed. New `Duplicati/UnitTest/IndexRecreationTests.cs`:

| Test | What it pins |
|---|---|
| `LostIndexFileIsRecreatedAsync` | A lost index file is re-created by the next backup and passes `full-remote-verification` |
| `RecreatedIndexFileHasTheSameContentAsync` | The re-created index file lists the same blocks **and the same blocklists** as the one it replaces |
| `RecreationHonoursIndexFilePolicyAsync` | `Full` writes blocklists, `Lookup` does not, `None` does not re-create at all |
| `RecreationIsSkippedForDryRunAsync` | A dry-run neither uploads nor registers the re-created index file |

@kenkendk's `TestUploadExceptionWithNoRetriesRaceAsync` is kept unchanged as the regression test for the race itself.

## Verification

Measured locally on Windows, Debug:

| | before the fix | after the fix |
|---|---|---|
| `TestUploadExceptionWithNoRetriesRaceAsync` | **3 of 3 failed** (`Found 3 / 2 faulty index files`) | **3 of 3 passed** |
| `TestUploadExceptionWithNoRetriesAsync` (the CI flake) | — | **10 of 10 passed** |
| `IndexRecreationTests` | — | 6 of 6 passed |
| `Category=Disruption` | — | 43 of 43 passed |
| `RepairHandlerTests`, `RepairWithMissingRemoteFiles`, `BorderTests`, `Issue6504` | — | 70 of 70 passed |

`dotnet build Duplicati.slnx -c Debug` is clean with 0 warnings.

**Limits of this verification.** The original failure is a flake, so I cannot prove it will not recur. What is shown is that the deterministic reproducer went from failing every time to passing every time, and that the flaky test passed 10 consecutive runs. The repeated runs were on Windows only; the reported CI failures were on ubuntu, so the CI run on this PR is the check that matters for the other platforms.

Whether to replace #7055 with this PR or to cherry-pick only the second commit onto it is entirely your call.


kenkendk:
I have applied the fix as a single commit to #7135 so it does not include the test-harness that I applied for #7055

kenkendk:
Closing this as the fixing commit was applied with #7135 and I did not like getting the test-harness from #7055 into the main code.

PR #7132, opened 2026-08-09
Keep the fileset time when replacing a missing dlist file

JamBalaya56562:
Repairing a dlist file that is missing from the destination moves the fileset forward in time, which can renumber the backup versions.

Found while investigating the ubuntu unit test failure on #7128, where `DisruptionTests.FilesetFilesAsync` reported version 1 as a partial backup where it had been a full one. That is a symptom of this; the cause is in `RepairHandler` and is present on master independently of that PR.

> **Updated after review.** The first version kept the time only in the local database, which is lost on a recreate. That gap is real and is now closed by the second commit, which records the time in the dlist instead. No filename is re-used. Details in the second half of this description.

## Cause

`RepairHandler` replaces a missing dlist under a new name, because the old name is still registered — the old registration is only marked `Deleted` [further down the same method](https://github.com/duplicati/duplicati/blob/master/Duplicati/Library/Main/Operation/RepairHandler.cs#L707), so `ProbeUnusedFilenameNameAsync` always has to move at least one second forward. The probed time was then passed on as the time of the *fileset*:

```csharp
var fileTime = await FilesetVolumeWriter.ProbeUnusedFilenameNameAsync(db, m_options, timestamp, ...);
var volumeWriter = newEntry = new FilesetVolumeWriter(m_options, fileTime);
await RunRepairDlistAsync(backendManager, db, volumeWriter, n, fileTime, ...);
// -> CreateFilesetAsync(volumeWriter.VolumeID, filesetTime, ...)
```

The fileset time is what orders the backup versions, so every repaired fileset shifts. Measured on two backups five seconds apart:

```
dlist before: duplicati-20260809T101721Z.dlist.zip, duplicati-20260809T101726Z.dlist.zip
dlist after : duplicati-20260809T101722Z.dlist.zip, duplicati-20260809T101727Z.dlist.zip

filesets before: v0 19:17:26 | v1 19:17:21
filesets after : v0 19:17:27 | v1 19:17:22
```

Both moved by a second. This is not intermittent — the old name is always registered at probe time, so the shift always happens. Reordering follows when two filesets are close together: with a one second gap, repairing the newer one first takes `T+1`, and the older one then finds `T` and `T+1` taken and lands on `T+2`, past the newer one. Which is repaired first depends on the order of `tp.MissingVolumes`, which is why the CI symptom is intermittent while the shift is not.

For a user this means a repair can renumber the versions, so `--version=1` no longer selects the backup it selected before.

## First commit: the fileset keeps its time

The replacement file still gets a free name; the fileset it belongs to keeps the time it had. The other repair path in this file, for a fileset whose remote registration is missing, already works that way — it keeps the existing fileset and only gives the file a new name.

## Second commit: the time survives a recreate

Keeping the time only in the database holds until the database is recreated, because [`RecreateDatabaseHandler`](https://github.com/duplicati/duplicati/blob/master/Duplicati/Library/Main/Operation/RecreateDatabaseHandler.cs#L327) creates the fileset from the time in the filename. So the time is recorded where the dlist can carry it, in the fileset file, next to `IsFullBackup`:

```
FilesetVolumeWriter.CreateFilesetFile(isFullBackup, filesetTime)
  -> FilesetData.GetFilesetInstance writes an optional "Timestamp"

RecreateDatabaseHandler.RecreateFilesetFromRemoteListAsync
  -> reads it right where it already reads IsFullBackup,
     then UpdateFilesetTimestampAsync
```

`IsFullBackup` already travels this way and is applied to the database on recreate, so this reuses that mechanism rather than adding one.

- **No filename is re-used.** The replacement file gets a new name and a new time, exactly as before. Only the fileset's own time is written down.
- **Absent means "use the filename".** Existing dlist files behave exactly as they do today, and an older version reading a newer dlist ignores an unknown field.
- **Only the two repair paths that rename a file pass the time.** The backup, synthetic filelist and purge paths are untouched and keep writing no `Timestamp` at all, so a normal backup's dlist is unchanged.

It also fixes the same divergence in the sibling path: when only the remote *registration* is missing, `RepairHandler` keeps the existing `filesetId` and re-uploads under a new name, so on master the database time and the filename time already disagree there, and a recreate already loses the fileset time. That path now records it too.

## Tests

`RepairFilesetTimestampTests`:

| Test | |
|---|---|
| `RepairKeepsTheFilesetTimeWhenAllDlistFilesAreMissingAsync` | every dlist deleted, then repaired |
| `RepairKeepsTheFilesetTimeAcrossADatabaseRecreateAsync` | repaired, then the local database deleted and rebuilt from the destination |
| `RepairKeepsTheFilesetTimeWhenOneDlistFileIsMissingAsync` | only the older one deleted — the one that can be pushed past the newer |

They assert the fileset times and the full/partial flag per version survive, and the first also asserts that the replacement files *do* get new names, since that part is expected.

## Verification

- The first two tests **fail on master**: `Repair moved the filesets in time`
- The recreate test **fails with only the first commit applied**: `Recreating the database lost the fileset times` — the review comment reproduced
- All three pass with both commits
- Repair, purge and recreate tests alongside them: 55 pass (`Issue6504`, `RepairHandlerTests`, `RepairWithMissingRemoteFiles`, `RecreateDatabase*`, `PurgeTest*`)

**Limits.** I did not reproduce the intermittent `FilesetFilesAsync` failure itself — 25 consecutive local runs on Windows all passed, which is why the tests pin the deterministic cause rather than the flake. The full suite on three platforms is left to CI. This does add a field to the dlist format, so if you would rather it were named differently, or carried somewhere else, say so and I will change it.

### DIFF
--- a/Duplicati/Library/Main/Database/Local/LocalDatabase.cs
+++ b/Duplicati/Library/Main/Database/Local/LocalDatabase.cs
@@ -3149,6 +3149,29 @@ namespace Duplicati.Library.Main.Database.Local
         }
 
         /// <summary>
+        /// Updates the timestamp of a fileset.
+        /// </summary>
+        /// <param name="fileSetId">The ID of the fileset to update.</param>
+        /// <param name="timestamp">The timestamp to record.</param>
+        /// <param name="token">Cancellation token to monitor for cancellation requests.</param>
+        /// <returns>A task that completes when the timestamp has been updated.</returns>
+        public async Task UpdateFilesetTimestampAsync(long fileSetId, DateTime timestamp, CancellationToken token)
+        {
+            await using var cmd = await m_connection.CreateCommandAsync(@"
+                UPDATE ""Fileset""
+                SET ""Timestamp"" = @Timestamp
+                WHERE ""ID"" = @FilesetId
+            ", token)
+                .ConfigureAwait(false);
+
+            await cmd.SetTransaction(m_rtr)
+                .SetParameterValue("@FilesetId", fileSetId)
+                .SetParameterValue("@Timestamp", Library.Utility.Utility.NormalizeDateTimeToEpochSeconds(timestamp))
+                .ExecuteNonQueryAsync(true, token)
+                .ConfigureAwait(false);
+        }
+
+        /// <summary>
         /// Removes all entries in the fileset entry table for a given fileset ID.
         /// </summary>
         /// <param name="filesetId">The fileset ID to clear.</param>
--- a/Duplicati/Library/Main/Operation/RecreateDatabaseHandler.cs
+++ b/Duplicati/Library/Main/Operation/RecreateDatabaseHandler.cs
@@ -993,6 +993,15 @@ namespace Duplicati.Library.Main.Operation
                 .UpdateFullBackupStateInFilesetAsync(filesetid, filesetData.IsFullBackup, cancellationToken)
                 .ConfigureAwait(false);
 
+            // The fileset was created from the time in the filename. A dlist file that
+            // replaced an earlier one has a new name, and therefore a new time, so it
+            // records the time of the fileset it describes. Use that when it is present.
+            var recordedTime = filesetData.GetFilesetTime();
+            if (recordedTime.HasValue)
+                await restoredb
+                    .UpdateFilesetTimestampAsync(filesetid, recordedTime.Value, cancellationToken)
+                    .ConfigureAwait(false);
+
             // clear any existing fileset entries
             await restoredb.ClearFilesetEntriesAsync(filesetid, cancellationToken).ConfigureAwait(false);
 
--- a/Duplicati/Library/Main/Operation/RepairHandler.cs
+++ b/Duplicati/Library/Main/Operation/RepairHandler.cs
@@ -440,7 +440,9 @@ namespace Duplicati.Library.Main.Operation
                             foreach (var p in m_options.ControlFiles.Split(new char[] { System.IO.Path.PathSeparator }, StringSplitOptions.RemoveEmptyEntries))
                                 fsw.AddControlFile(p, m_options.GetCompressionHintFromFilename(p));
 
-                        fsw.CreateFilesetFile(isfull);
+                        // The replacement file has a new name, and therefore a new time,
+                        // but it still describes the fileset at its original time.
+                        fsw.CreateFilesetFile(isfull, timestamp);
                         await db
                             .WriteFilesetAsync(fsw, filesetId, cancellationToken)
                             .ConfigureAwait(false);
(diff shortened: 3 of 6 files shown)


kenkendk:
Duplicati has a basic rule that it does not re-use a filename. This rule prevents a number of partial-write, read-before-write, and other failures.

Because of that we need to choose a new filename when creating a new fileset. The idea here was to use +1s as we expect regular backups to not run every second. This does not hold in tests of course, because we run many backups back-to-back.

The fix here updates the local database so the timestamp recorded in the local database no longer matches the filename. This works, but it only holds until the database is recreated, in which case the filename timestamp is used. 

I think the fix has a good intention, but it makes it so the database cannot be recreated in the same state.

I would like one of two paths:
1. We ignore the error, saying it only happens in tests.
2. We figure out how to encode the correct time, either in the filename or the metadata.

JamBalaya56562:
Thank you, that is a fair objection and you are right about the recreate. I had noticed that the timestamp would come back from the filename on a recreate and then failed to put it in the description, which I should have done — sorry about that.

I went with your option 2, in a second commit. No filename is re-used: the replacement file still gets a free name and a new time. What is written down is the time of the *fileset* it describes, and it goes in the fileset file inside the dlist, next to `IsFullBackup`:

```
FilesetVolumeWriter.CreateFilesetFile(isFullBackup, filesetTime)
  -> FilesetData.GetFilesetInstance writes an optional "Timestamp"

RecreateDatabaseHandler.RecreateFilesetFromRemoteListAsync
  -> reads it right where it already reads IsFullBackup,
     then UpdateFilesetTimestampAsync
```

`IsFullBackup` already travels that way and is applied to the database on recreate, so this reuses the existing mechanism rather than adding one. When the field is absent the filename is used, exactly as today, so existing dlist files are unaffected and an older version reading a newer dlist just ignores it. Only the two repair paths that rename a file pass the time — backup, synthetic filelist and purge are untouched, so a normal backup's dlist has no `Timestamp` at all.

One thing I ran into while doing this: the divergence you describe already exists on master. In the sibling path, where only the remote *registration* is missing, `RepairHandler` keeps the existing `filesetId` and re-uploads under a new name:

```csharp
var fsw = new FilesetVolumeWriter(m_options, fileTime);   // probed, moved forward
fsw.CreateFilesetFile(isfull);
await db.WriteFilesetAsync(fsw, filesetId, ...);          // fileset kept as-is
```

so the database time and the filename time already disagree there, and a recreate already loses the fileset time. My first commit made the missing-*file* path behave like that sibling rather than inventing a new state, and the second commit now records the time in both, so that path is fixed too.

To make sure I was answering the actual objection rather than my reading of it, the new test does what you described: repair the missing dlist files, then delete the local database and rebuild it from the destination, and check the fileset times and the full/partial flag are the original ones. It fails with only the first commit applied and passes with both. Repair, purge and recreate tests around it still pass (55).

This does add a field to the dlist, so if you would rather it were named differently, or carried in the filename, or you would rather take option 1 after all, I am happy to follow — the second commit is separate so it can be dropped on its own.

PR #7143, opened 2026-08-11
Add check for misconfigured filters

kenkendk:
This PR adds a small sanity check on filters where it looks to see if the filters are configured to unconditionally exclude everything. If this happens, a warning is emitted.

Since it is not possible to statically evaluate if a given filter will exclude everything, there is also a back-stop detection after the backup completes that looks at examined files. If no files were examined, this is also treated as a warning.

Overall, the goal is to prevent cases where the backups run sucessfully but do not contain any files. With this change, a warning is logged so it is visible that something is not right.

### DIFF
--- a/Duplicati/Library/Main/Operation/BackupHandler.cs
+++ b/Duplicati/Library/Main/Operation/BackupHandler.cs
@@ -631,6 +631,11 @@ namespace Duplicati.Library.Main.Operation
             // If there is no filter, we set an empty filter to simplify the code
             m_filter = filter ?? new Library.Utility.FilterExpression();
 
+            // Warn about a likely misconfigured filter that excludes everything but has no includes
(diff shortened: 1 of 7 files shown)


ts678:
@kenkendk 

> see if the filters are configured to unconditionally exclude everything

I'd note that this is the default action when you add a filter and leave it.
Although this check may help, a safer default could also be considered.

image

image

(user better remember to click the x to delete that before they forget it)

kenkendk:
@ts678: Thanks for bringing it up. I have also added a small warning triangle to the UI, so if the UI detects that the filters will exclude everything, it will show a warning triangle.

Not sure what a better default would be? I like that we show the `*` so (most) users will recognize the ability to use globbing expressions. But I can also see the problem that the default value excludes everything. A default like `filename*` could perhaps work?

EDIT: Changed ngclient to add `-*.lock` as a new filter, which I think makes some sense and is unlikely to exclude critical files.

PR #7145, opened 2026-08-11
Parse the ssh and smb urls with System.Uri

JamBalaya56562:
Continues the work towards deleting `RelaxedUri`, after #7133, #7134, #7136, #7137 and #7138.

## Why only these two backends

I went through every `RelaxedUri` use in the tree first, because the answer decides how far this can go.

`System.Uri` lower-cases the authority. That is fine when the authority is a server name, and fatal when it is a remote folder name — which is what #7097 ran into and what the TODO in `RelaxedUri.cs` records.

- **24 sites read a case sensitive remote name out of the authority** (`box://`, `dropbox://`, `openstack://`, plus B2, S3, Idrivee2, Jottacloud, Mega, GoogleDrive, and others). These have to keep the current parser.
- **About 30 sites take an arbitrary backend url**, change the query and round-trip it with `ToString()` — `BackendLoader`, `Controller`, `SecretProviderHelper`, `QuerystringMasking`, `Database/Backup.cs`. They see the folder-name urls too, so they have to keep it as well.
- **Five backends read a real server name**: SMB, SSHv2, WEBDAV, SharePoint, Tahoe-LAFS.

This PR converts **SSHv2 and SMB**, which are the two with the same shape: authority is a server name, path is a remote path. A server name is a DNS name, a NetBIOS name or an IP literal, and none of those are case sensitive; both backends hand it straight to a resolver rather than using it to name anything remote.

`FTPBackend.cs` and `pCloudBackend.cs` are deliberately left alone in this PR so it would not collide with #7144.

## Shape

Two commits, so the extraction can be read separately from the parser change.

1. **Extract and pin.** Each backend gets an `internal record struct` and an `internal static` parse method, in the same form as `ParseXmppTarget` from #7138, and the tests record what the current parser answers for each url. No behaviour change.
2. **Swap the parser.** Only the body of the two parse methods changes.

The seam sits in a different place in the two backends, on purpose: **normalization driven by an option stays in the constructor, normalization decided by the url alone moves into the parse method.** So SSH keeps its `ssh-relative-path` handling in the constructor — pulling it in would force the parse method to take the options dictionary — while SMB takes the share out of the first path segment inside the parse method, because that split is a pure function of the url and is the thing worth asserting.

`RequireHost` and the new `RequireNoFragment` went into `Duplicati.Library.Utility` as extension methods on `System.Uri`. The missing-host message is built from `Strings.Uri.NoHostname` and `Utility.StripUrlUserInfo`, **both internal to that assembly**, so a backend cannot reproduce it — putting the check there is what keeps the message and the exception type identical to `RelaxedUri.RequireHost()`. The other five backends can use the same two methods when they convert.

`RequireNoFragment` throws `UserInformationException` rather than matching `RequireHost`, because there is no earlier behaviour to preserve there and the user has something to fix.

`IsDefaultPort` is not needed here, unlike in #7138: `ssh`, `smb` and `cifs` have no registered default port, so `Port` is `-1` when the url names none — the same answer the old parser gave. There is a test pinning that, so if a future runtime registers one it fails in the unit job instead of silently moving everyone to a different port.

## What changes for users

Every one of these is covered by a test, and each expectation was measured before and after rather than predicted.

| | before | after |
| --- | --- | --- |
| `ssh://Host.Example.COM/p` | host `Host.Example.COM` | host `host.example.com` |
| `ssh://host/a+b` | path `a b` | path `a+b` |
| `ssh://host/backup#1/` | path `backup#1/` | **reported to the user** |
| `ssh://host/a/../b` | path `a/../b` | path `b` |
| `ssh://user:p@ss@host/p` | host `ss@host`, password `p` | **reported to the user** |
| `ssh://host:99999/p` | port 99999 | **reported to the user** |
| `ssh://:pass@host/p` | rejected as having no hostname | empty user, password `pass` |
| `smb://host/share\sub` | share `share\sub`, path empty | share `share`, path `sub` |

Three of these deserve a note.

**The `+`** was a query-string rule (`UrlEncoding.UrlDecode` maps `+` to a space) applied to paths, where it does not belong. A url that has been through `RelaxedUri.ToString()` carries `%2B`, so this only reaches someone who hand-wrote a `+` on the command line.

**The `#`** is the one I was least comfortable leaving alone. `System.Uri` reads it as a fragment, so `ssh://host/backup#1/` would silently become `backup` — a different folder, with no error. Rather than accept that, `RequireNoFragment` reports it and tells the user to write `%23`. `ssh://host/%23hash/` is unaffected and is pinned by a test. If you would rather keep the old meaning, dropping that call and building the path from `GetComponents(UriComponents.Path | UriComponents.Fragment, UriFormat.UriEscaped)` restores it in one line.

**The unencoded `@`** was already broken: the old regex cut the password at the first `@` and let the rest into the host name, so `ssh://user:p@ss@host/p` connected to `ss@host` and failed to resolve. It is now reported at construction. The `UriFormatException` message is "Invalid URI: The hostname could not be parsed." and does not echo the url, so no credentials leak; there is a test asserting that.

`smb://host/share\sub` starting to work is a fix — `PATH_SEPARATORS` in that same file already treats `\` as a separator everywhere downstream, so the constructor was the odd one out. An encoded `%5C` stays part of the share name, with its own test.

## Testing

- `--filter "Category=UriUtility"`: **151 passed** after the change, and 151 passed before it as well
- after swapping the parser body, exactly **11 tests failed, all of them rows in the table above** — nothing unpredicted moved, which is what the two-commit shape is for
- `dotnet build Duplicati.slnx`: 0 errors

**Limits of what I checked:** the credentialed backend jobs (`SshTests`, `CIFSTests`) were not run locally. I followed their urls by hand — `ssh://testuser:testpassword@127.0.0.1:3000` and `smb://localhost/testshare1/new/?...` — and both parse identically before and after, but that is reasoning, not a measurement.

### DIFF
--- a/Duplicati/Library/Backend/SMB/SMBBackend.cs
+++ b/Duplicati/Library/Backend/SMB/SMBBackend.cs
@@ -136,18 +136,31 @@ public class SMBBackend : IStreamingBackend, IFolderEnabledBackend, IRenameEnabl
     /// <returns>The parts the backend needs.</returns>
     internal static SmbUrl ParseSmbUrl(string url)
     {
-        var uri = new Utility.RelaxedUri(url);
-        uri.RequireHost();
+        var uri = new Uri(url);
+        uri.RequireHost(url);
+        uri.RequireNoFragment(url);
 
-        var input = uri.Path.TrimEnd('/');
+        // The path arrives escaped and with its leading separator, where the previous
+        // parser handed it over decoded and without one. Decoding first keeps an
+        // encoded separator in the place it used to land in.
+        var path = Uri.UnescapeDataString(uri.AbsolutePath);
+        if (path.StartsWith("/", StringComparison.Ordinal))
+            path = path.Substring(1);
+
+        // The user info is not decoded, so it is decoded here.
+        var userinfo = uri.UserInfo.Split(new[] { ':' }, 2);
+        var username = userinfo.Length > 0 && userinfo[0].Length > 0 ? Uri.UnescapeDataString(userinfo[0]) : null;
+        var password = userinfo.Length > 1 ? Uri.UnescapeDataString(userinfo[1]) : null;
+
+        var input = path.TrimEnd('/');
         var slashIndex = input.IndexOf('/');  // Find first slash to separate share and path if present.
 
         return new SmbUrl(
-            uri.Host!,
+            uri.Host,
             slashIndex >= 0 ? input[..slashIndex] : input,
             slashIndex >= 0 ? input[(slashIndex + 1)..] : "",
-            uri.Username,
-            uri.Password);
+            username,
+            password);
     }
 
     /// <summary>
--- a/Duplicati/Library/Backend/SMB/SMBBackend.cs
+++ b/Duplicati/Library/Backend/SMB/SMBBackend.cs
@@ -29,6 +29,8 @@ using Duplicati.Library.Common.IO;
 using Duplicati.Library.Utility.Options;
 using Duplicati.Library.Utility;
 
+[assembly: InternalsVisibleTo("Duplicati.UnitTest")]
+
 namespace Duplicati.Library.Backend;
 
 /// <summary>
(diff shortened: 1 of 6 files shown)


kenkendk:
I like this approach, let us move towards real urls with the pain it may cause for odd targets (like paths with a `#` in the name).

Later on, I think we should consider what to do with box, dropbox etc. One solution is to keep the authority empty and then have the path as a string, like: `box://?path=SomePath` so we can rely on a regular url and avoid casing problems. Such a change is a bit destructive, so we may have to first add the `?path=` support and make the code prefer the query param. Then update the FE and other places that emit the URL to emit the new format, and then later remove the old support.

PR #7152, opened 2026-08-12
Better classification of MS365 sites

kenkendk:
This PR adds the option to classify "Personal" sites it is now possible to specify either "PersonalEnabledUser" or "PersonalDisabledUser". The "Personal" option is still included for backwards compatibility.

With this it is possible to exclude any personal site from a disabled user.


kenkendk:
Fixed with #7163

PR #7174, opened 2026-08-15
Name the two steps inside UploadNewFileset that had no timer

### LINKED ISSUE #7127: "Waiting for upload to finish" stuck in SQLite on large backup after small first

JamBalaya56562:
> **Correction (and thanks to @ts678 for catching it).** The first version of this description said the reporter "cannot say more than" `UploadNewFileset`. That was wrong. The forum thread's original post identified the suspected query and where it sat in the code, and the slow query monitor later confirmed it by name. I had not read the thread carefully enough before writing that, and the sentence has been removed rather than softened.

#7127 reports that a backup stops with the status bar at "Waiting for upload to finish", and that the server profiling log puts it at `UploadNewFileset`. That timer covers five things, and two of them were never measured.

## What the phase contains, and what could be measured

`UploadRealFilelist.cs:54-80`:

| step | timed before this change |
|---|---|
| `AddControlFile` | no — but empty unless `--control-files` is set |
| **`db.WriteFilesetAsync`** | **no** |
| **`filesetvolume.Close()`** | **no** |
| `db.UpdateRemoteVolumeAsync` | no — a single UPDATE |
| `db.CommitTransactionAsync` | yes — `CommitTransaction: CommitUpdateRemoteVolume` |
| `backendManager.PutAsync` | yes — `RemoteOperationPut` |

Two timers close the gap, and then the phase divides into named parts.

## What it looks like

On a backup of 111,110 files in 11,110 folders, the profiling log now reads:

```
Starting - Uploading a new fileset
  Starting - Writing the fileset
  Writing the fileset took 0:00:00:01.602
  Starting - Closing the fileset volume
  Closing the fileset volume took 0:00:00:00.878
  CommitTransaction: CommitUpdateRemoteVolume took 0:00:00:00.000
Uploading a new fileset took 0:00:00:02.484
```

1.602 + 0.878 + 0.000 against a phase of 2.484 leaves **four milliseconds** for everything else, so the two new timers account for the phase rather than merely sitting inside it.

One thing that becomes visible and is worth knowing when reading a report about waiting for an upload: `RemoteOperationPut` for that dlist finishes **after** the phase does, because uploads are asynchronous by default. The phase named `UploadNewFileset` does not wait for the upload.

## Does the slow query monitor make this unnecessary?

@ts678 asked this directly, and for #7127 the answer is essentially yes. The monitor named `LIST_FOLDERS_AND_SYMLINKS` and timed it at ~46,767 s. Nothing here was needed to find that, and I would not want this PR credited with it.

What these two timers add is the time that is **not** in a query:

- `filesetvolume.Close()` is compression and finalisation, not SQL, so the monitor cannot see it at all
- `WriteFilesetAsync` also contains the reader loop that builds the volume. As noted in #7147, the monitor's `QueryScope` is disposed when the reader is returned, so the loop is outside it

So this is worth having for a report where the phase is slow and the monitor stays quiet, which is the case it does not currently cover. If that is too narrow to be worth two lines of diff, I am happy for this to be closed — it is a small convenience, not a fix, and the issue it was written for turned out not to need it.

## The queries are deliberately left alone

`WriteFilesetAsync` runs its two queries with `writeLog: false`, which suppresses the profiling record. Turning that on would print each query's time, but:

- `SlowQueryMonitor.StartQuery` sits **outside** that flag in `ExtensionMethods.cs`, so every query is registered with the monitor either way
- `writeLog: false` is what 123 call sites in `Library/Main` do; changing two of them would make the convention harder to read, not easier
- the record it would print is `ExecuteReader: <the whole SQL>`, which is not what makes a profiling log readable

## Scope

This is a diagnostic change. **It does not fix #7127** and is not offered as one.

## Verification

- Built clean; the warnings present are the pre-existing `GoogleCredential` and `DBusConnection` ones
- The breakdown above is from a real run, not a constructed one
- `EmptyFileTests` green

### DIFF
--- a/Duplicati/Library/Main/Operation/Backup/UploadRealFilelist.cs
+++ b/Duplicati/Library/Main/Operation/Backup/UploadRealFilelist.cs
@@ -60,8 +60,15 @@ internal static class UploadRealFilelist
                 // We ignore the stop signal, but not the pause and terminate
                 await taskreader.ProgressRendevouzAsync().ConfigureAwait(false);
 
-                await db.WriteFilesetAsync(filesetvolume, filesetid, taskreader.ProgressToken).ConfigureAwait(false);
-                filesetvolume.Close();
+                // Named separately from the phase around them: a report that this phase is
+                // where a backup stops cannot say which of the five things inside it is the
+                // one that stopped, and the commit and the upload are the only two that
+                // already say so for themselves.
+                using (new Logging.Timer(LOGTAG, "WriteFileset", "Writing the fileset"))
+                    await db.WriteFilesetAsync(filesetvolume, filesetid, taskreader.ProgressToken).ConfigureAwait(false);
+
+                using (new Logging.Timer(LOGTAG, "CloseFilesetVolume", "Closing the fileset volume"))
+                    filesetvolume.Close();
 
                 // We ignore the stop signal, but not the pause and terminate
                 await taskreader.ProgressRendevouzAsync().ConfigureAwait(false);


ts678:
> the server profiling log puts it at UploadNewFileset. The reporter cannot say more than that

The forum original post had much more, including where it was in code, plus suspected query,
which was [confirmed by the slow query monitor](https://forum.duplicati.com/t/slow-or-stuck-in-sqlite-at-waiting-for-upload-to-finish/22597/2?u=ts678) whose PR also had [decompiler data provided](https://github.com/duplicati/duplicati/pull/7147#issuecomment-5281022898).
I tried debugging at `IL` level, but the debugger couldn't set breakpoints finely. Unknown why.

Does the (fixed, thanks for that) slow query monitor report remove need for this further detail?

As a side note, I'm seeing that after process kill, the next backup (even small Source) gets slow.
By attaching debugger to check the stack of busy thread, it might be `VerifyConsistencyInner`.

I'd like to find a test case that is super slow in a friendlier environment, such as SQLite browser.

JamBalaya56562:
@ts678 Thanks — you were right about the description, and I've corrected it above rather than softened it. I also answered the "does the monitor make this unnecessary" question there: for #7127 it essentially does, and I've said so.

Two things from chasing your side note.

## `VerifyConsistencyInner` and `UploadNewFileset` are the same query

Your guess about the busy thread looks right, and I think it is not a second problem. `LIST_FOLDERS_AND_SYMLINKS` is used in two places:

- `LocalDatabase.cs:2484` — `WriteFilesetAsync`, which is what `UploadNewFileset` reaches
- `LocalDatabase.cs:2096` — inside `VerifyConsistency`, wrapped in a `COUNT(*)` to compare the expanded fileset against `FilesetEntry`

So "stuck writing the fileset" and "next backup stuck before it does anything" are the same SQL from two callers. That also explains why killing the process doesn't help: nothing about the kill changes what the next run has to execute.

## The stale-statistics theory does not survive being tested

I want to report this as a negative result, because it looked very promising and it is wrong.

The query is extremely plan-sensitive. On a 12k-file database, with a deliberately degenerate `sqlite_stat1` (every stat forced to `1`), the plan picks up three `SCAN`s and the query goes from 0.015 s to not finishing in 90 s — a single row never came back. `ANALYZE` restores it to 0.016 s. Worth knowing: **no statistics at all is fine** (0.020 s) — only wrong statistics are fatal.

That made "small first backup writes tiny statistics, large second backup inherits them" look like the answer, especially with `PRAGMA optimize` running at connection close (`LocalDatabase.cs:3362`) and the hang happening long before any close.

It does not hold up:

- The statistics a real one-file backup writes are not degenerate. They are `Block "3 1 1"`, `FilesetEntry "2 2 1"` and so on — small, but with the *shape* intact.
- Transplanting those real statistics onto a 21k-file database gives **0.044 s against 0.037 s**, and no `SCAN`s. My degenerate injection was simply unrealistic.
- Running the actual sequence — one file, then 15,000 files in 6,000 folders — did not reproduce anything: `UploadNewFileset` took about a second and the slow query monitor stayed silent.

So the shape of the trigger is still unexplained, and I have stopped guessing again.

## The one thing that would settle it

You have a database that reproduces, and it reproduces in DB Browser, which means the answer is one statement away:

```sql
EXPLAIN QUERY PLAN
SELECT "G"."BlocksetID", "G"."ID", "G"."Path", "G"."Length", "G"."FullHash",
       "G"."Lastmodified", "G"."FirstMetaBlockHash", "H"."Hash" AS "MetablocklistHash"
FROM (
    SELECT "B"."BlocksetID", "B"."ID", "B"."Path", "D"."Length", "D"."FullHash",
           "A"."Lastmodified", "F"."Hash" AS "FirstMetaBlockHash",
           "C"."BlocksetID" AS "MetaBlocksetID"
    FROM "FilesetEntry" "A", "File" "B", "Metadataset" "C",
         "Blockset" "D", "BlocksetEntry" "E", "Block" "F"
    WHERE "A"."FileID" = "B"."ID" AND "B"."MetadataID" = "C"."ID"
      AND "C"."BlocksetID" = "D"."ID" AND "E"."BlocksetID" = "C"."BlocksetID"
      AND "E"."BlockID" = "F"."ID" AND "E"."Index" = 0
      AND ("B"."BlocksetID" = -100 OR "B"."BlocksetID" = -200)
      AND "A"."FilesetID" = 2
) "G"
LEFT OUTER JOIN "BlocklistHash" "H" ON "H"."BlocksetID" = "G"."MetaBlocksetID"
ORDER BY "G"."Path", "H"."Index";
```

If the output contains `SCAN` lines, the planner is the problem and `ANALYZE` on a copy should make it fast — which would also tell us the fix has to put statistics in a good state *before* this query runs, not at close. If there are no `SCAN`s, the planner is exonerated and the time is going somewhere else entirely, and I'd stop looking at query plans.

`SELECT * FROM sqlite_stat1;` on the same database would be useful alongside it. No rush, and no obligation — you have already done more measuring on this than anyone.

ts678:
> So "stuck writing the fileset" and "next backup stuck before it does anything" are the same SQL from two callers.

Agreed, and I put that in an edit to the forum topic before noticing that the PR here had been updated.

> SELECT * FROM sqlite_stat1;

This was described in the forum topic earlier today, commenting mostly on 1 (rows) as the first number.
Surprisingly, `optimize` or `analyze` didn't change that. I wasn't comparing all of the numbers carefully...

> EXPLAIN QUERY PLAN

This was also in the forum topic, however it was done on the larger UNION query. It shows lots of SCAN.
DB Browser for SQLite is "Busy (reading rows)" now. I can look at your query after letting it run overnight.
And I never did try a DB from scratch, just keeping on banging on this one from a process-killed backup.

I was considering posting either a debug log or a DIY version since your repro from scratch isn't working.
I did post what I think was the Perl script that made my tree, but it sounds like you created a similar tree.

JamBalaya56562:
@ts678 The "1" in `sqlite_stat1` is the part I got wrong, and the detail that fixes it is the one I had been ignoring: **the database came from a process-killed backup.**

I said earlier that the product does not create degenerate statistics. That was wrong, and here is the counter-example from my own machine — a backup I killed part-way through, then read:

| table | rows actually in it | what `sqlite_stat1` claims |
|---|---|---|
| `FileLookup` | 46,180 | **2** |
| `BlocksetEntry` | 81,314 | **3** |
| `Block` | 81,314 | **3** |

Those are the numbers from the one-file first backup, still describing tables that have since grown five orders of magnitude. The mechanism is not subtle once you look: `PRAGMA optimize` runs in `LocalDatabase.DisposeAsync`, just before the connection closes (`LocalDatabase.cs:3362`). Kill the process and the close never happens, so the statistics from the last clean shutdown are what the next run plans against. It is not the small first backup that poisons things — it is the kill.

So your database being "from a process-killed backup" is not incidental to it, and starting from scratch was never going to reproduce it. That is my mistake, and it cost a day of measurements aimed at the wrong thing.

## Where I still cannot follow you

Having reproduced the *state*, I still cannot reproduce the *plan*. On that killed database the UNION query has one `SCAN` — the subquery the UNION dedups through — and finishes:

```
FilesetID=2   SCAN=1 of 24 plan rows   COUNT=46178   0.489 s
```

I also built a set with `--blocksize=5kb` so files span ten blocks and `BlocksetEntry.Index` spreads 0–9 rather than being all zero, transplanted the first-backup statistics onto it, and got the same shape: one `SCAN`, 0.372 s. And a clean 1-file → 111,100-file run finished in 22 minutes with no slow-query warnings at all, the folders query taking 0.37 s over 133,323 rows.

Forcing every stat to literally `1` does break it — three `SCAN`s, and 90 seconds without a single row — but that is not what a kill produces, so I am reproducing the fault only by cheating.

Something about the data in yours is doing work that mine is not, and I have stopped guessing which.

## One thing worth checking on the `ANALYZE` result

> Surprisingly, `optimize` or `analyze` didn't change that.

That would be very odd for a plain `ANALYZE`, so before treating it as a finding: DB Browser for SQLite keeps changes in an uncommitted transaction until you press **Write Changes**. If the `ANALYZE` ran there and was never written, `sqlite_stat1` would read back unchanged, and reopening would discard it. Worth ruling out — if `ANALYZE` genuinely cannot update those numbers, that is its own bug and a more interesting one.

## On the offer

> I was considering posting either a debug log or a DIY version

Yes please, whichever is less work for you — the DIY version if the tree is what matters, the database if the state is. Given that the kill turns out to be part of the recipe, a DIY version probably needs to include it: small backup, start the large one, kill it, then run again.

And there is no hurry on running my query overnight. If the plan on the larger UNION already shows plenty of `SCAN`s, that is the answer to what I was asking; the folders query on its own would only confirm which half it is coming from.

ts678:
EDIT 4 [here](https://forum.duplicati.com/t/slow-or-stuck-in-sqlite-at-waiting-for-upload-to-finish/22597/2?u=ts678) answers some of the questions presented, however that's probably not the end of it.
There's a look at SQLite statistics compile options, results, PRAGMA use, EXAMINE QUERY PLAN,
continued timing tests on cut-down query (basically innermost query with big-comma-join cuts).

> So your database being "from a process-killed backup" is not incidental to it, and starting from scratch was never going to reproduce it.

I think this is mixing two scenarios of the slow query (corresponding to its two uses). The original issue test involved no kill until the situation was deemed unreasonably hopeless. I find it difficult to debug DB issues during the middle of a Duplicati run, however the kill prior to the close in second backup gives a chance to examine a similar situation either with a fresh Duplicati start, or with an external SQLite tool of one's choice, relying on the happy (and I hope not misleading) fact that the slow query is used again as part of a UNION.

The start from scratch has been run numerous times on the original issue. I want to show second reproduce, using exact known conditions instead of merely the conditions I think it was in on two prior runs, to be sure.

JamBalaya56562:
Closing this. The reason it was written no longer holds.

The description opened by saying the reporter could not say more than `UploadNewFileset`. That was wrong, as @ts678 pointed out — the forum thread had the suspected query and where it sat in the code, and the slow query monitor had already confirmed it by name. I corrected the text but not the conclusion that followed from it.

The cause has now been found, and none of it came through these timers: it is stale `sqlite_stat1`, written as `1` by `PRAGMA optimize` when the first backup is a single file given as a file path, and never refreshed while the second backup runs. Details and a from-scratch reproduction are in [#7127](https://github.com/duplicati/duplicati/issues/7127#issuecomment-5307802039).

That leaves this PR with only the argument I gave when @ts678 asked whether the slow query monitor made it unnecessary: `filesetvolume.Close()` is compression rather than SQL, so the monitor cannot see it. That is true, but it is a hypothetical gap — I have no report where it is the slow part, and I would rather not carry a change whose stated purpose has been refuted just because the diff is two lines.

If a report does turn up where `UploadNewFileset` is slow and the query monitor stays quiet, this is easy to write again.

Thanks @ts678 — the correction about the kill, and the `sqlite_stat1` dump, are what made the actual cause findable.

PR #7181, opened 2026-08-16
Rebuild the query statistics when they no longer describe the database

### LINKED ISSUE #7127: "Waiting for upload to finish" stuck in SQLite on large backup after small first

JamBalaya56562:
Fixes #7127.

A first backup small enough to leave the tables holding a handful of rows — a file or three, given as **file paths rather than a folder**, which would add its own row — has that written into `sqlite_stat1` by `PRAGMA optimize` at connection close. The second backup fills those tables with tens of thousands of rows, and nothing refreshes the numbers along the way: the pragma runs only when the connection closes, which the stuck backup never reaches.

Reproduction and the original measurements are in [the issue](https://github.com/duplicati/duplicati/issues/7127#issuecomment-5307802039).

## How small the first backup has to be

Forcing every recorded row count to a given value and running `LIST_FOLDERS_AND_SYMLINKS` against real data:

| recorded row count | 12,103 files | 52,069 files |
|---|---|---|
| 1 | **did not finish in 45 s** | **did not finish in 45 s** |
| 2 | **did not finish in 45 s** | **did not finish in 45 s** |
| 3 | **did not finish in 45 s** | 0.3244 s |
| 10 | 0.0950 s | 0.1933 s |
| accurate | 0.0434 s | 0.2181 s |

So it is not only the one-file case the issue was reported from — two or three files do it as well, and where it stops depends on the database.

Worth noting that the plan shape is not the tell: at 52,069 files a recorded count of 2 produces **no `SCAN` at all** and still does not finish. It is the join order that goes wrong, so counting `SCAN` lines is a misleading way to check this.

Nor is it one query. In the run below the one that would not finish was the length check inside `VerifyConsistency`, not the fileset query the issue was reported against — which is the argument for fixing the statistics rather than any particular statement.

## The mask is the part that matters

`0x10000` is what makes SQLite look at every table rather than only the ones this connection has touched. Without it the pragma **does nothing at all while a transaction is open**, and one always is here — `ReusableTransaction` opens a new one as soon as it commits.

```
inside a transaction:  PRAGMA optimize          0.0003 s   statistics unchanged
                       PRAGMA optimize=0x10002  0.3655 s   statistics updated
```

There is a test for the mask, because dropping it looks harmless in a diff and quietly disables all of this.

## Why setting it at connection open is not enough

@ts678 [suggested](https://github.com/duplicati/duplicati/pull/7181#issuecomment-5308786408) `CUSTOMSQLITEOPTIONS_DUPLICATI=optimize=0x10002`, which needs no code at all, and it does fix a backup that follows a killed one. It cannot fix the case the issue was reported from: when the connection opens, the tables still hold the handful of rows the statistics describe, so there is correctly nothing to do, and the growth happens afterwards.

Measured on master with that variable set — one file, then 20,000:

| | |
|---|---|
| completion | **none in 25 minutes** |
| slow query warnings | 16 |
| longest single query | still running after **975.9 s** |

## Where it is called

Twice: once when the database is opened, before anything reads a large table, and again after the main operation, when the tables have just grown by everything the backup added.

The first placement matters more than it looks. I originally put it just before the consistency check and measured six slow-query warnings anyway — the longest-path probe in `PreBackupVerify` reads the `File` view, and on a killed database **that query alone ran for over seven minutes**.

The second is what covers the reported case. `PRAGMA optimize` sees the rows the open transaction has added but not yet committed — measured by adding 52,069 rows inside a transaction and watching the recorded count go to 104,138 — so it describes the database as it stands.

## Cost

`PRAGMA analysis_limit=400` bounds the sampling, and SQLite decides for itself which tables have moved far enough to be worth re-analyzing, so there is no guard around the call:

| | |
|---|---|
| database whose statistics already fit | **0.0040 s** |
| database needing the rebuild, 58,115 rows | 0.0538 s |

## Verification

The same scenario that would not finish in 25 minutes on master:

| | slow query warnings | duration |
|---|---|---|
| master, `optimize=0x10002` at open | **16** | **did not finish in 25m** |
| this change | **0** | **5m52s** |

with the fileset upload taking 0.870 s and the consistency check 0.148 s, and the statistics ending accurate (`FileLookup` 40,002 rows, recorded as 40,002).

- `QueryStatisticsTests`, 2 cases. Red before the change, and red again if the mask is dropped (`The statistics still describe 1 row(s) where there are 441`)
- `SymLinkTests`, `EmptyFileTests`, `EmptyMetadataTests` — 14 cases green
- Build clean, warning list byte-for-byte identical to master

## Credit

@ts678 — the `sqlite_stat1` dump is what made the size of the first backup findable, the correction that the original case involved no kill is what stopped me reproducing the wrong thing, and the pragma mask replaced a worse implementation of mine: I had been running `ANALYZE` behind a hand-written staleness check, which cost 0.0523 s on a database that needed nothing done to it.

### DIFF
--- a/Duplicati/Library/Main/Database/Local/LocalDatabase.cs
+++ b/Duplicati/Library/Main/Database/Local/LocalDatabase.cs
@@ -1376,6 +1376,45 @@ namespace Duplicati.Library.Main.Database.Local
         }
 
         /// <summary>
+        /// The number of rows ANALYZE samples per index. SQLite recommends 400 as a value that
+        /// keeps the cost bounded while still producing usable estimates.
+        /// </summary>
+        private const long StatisticsAnalysisLimit = 400;
+
+        /// <summary>
+        /// Rebuilds the query statistics if the ones recorded no longer describe the database.
+        ///
+        /// A backup that starts from an almost empty database leaves "sqlite_stat1" saying the
+        /// tables hold a row or two, and then fills them with tens of thousands. The planner keeps
+        /// believing the recorded numbers, and picks join orders for the fileset and consistency
+        /// queries that do not finish in any reasonable time.
+        ///
+        /// The mask matters. 0x10000 is what makes SQLite look at every table rather than only
+        /// the ones this connection has touched, and without it the pragma does nothing at all
+        /// while a transaction is open -- which one always is here. 0x0002 is the ANALYZE step.
+        /// SQLite recommends exactly this combination for long-lived connections.
+        /// </summary>
+        /// <param name="token">Cancellation token to monitor for cancellation requests.</param>
+        /// <returns>A task that completes when the statistics have been refreshed, if they needed it.</returns>
+        public async Task RefreshQueryStatisticsAsync(CancellationToken token)
+        {
+            await using var cmd = m_connection.CreateCommand()
+                .SetTransaction(m_rtr);
+
+            using (new Logging.Timer(LOGTAG, "RefreshQueryStatistics", "Refreshing the query statistics"))
+            {
+                await cmd.ExecuteNonQueryAsync($"PRAGMA analysis_limit={StatisticsAnalysisLimit}", token)
+                    .ConfigureAwait(false);
+
+                // SQLite decides for itself which tables have moved far enough to be worth
+                // re-analyzing, so there is nothing to gain from guessing at that here. On a
+                // database whose statistics already fit, this costs about four milliseconds.
+                await cmd.ExecuteNonQueryAsync("PRAGMA optimize=0x10002", token)
+                    .ConfigureAwait(false);
+            }
+        }
+
+        /// <summary>
         /// Performs a VACUUM operation on the database to reclaim unused space.
         /// This operation can help optimize the database performance by defragmenting it.
         /// Note: This operation can take a significant amount of time depending on the size of the database.
--- a/Duplicati/Library/Main/Operation/Backup/BackupDatabase.cs
+++ b/Duplicati/Library/Main/Operation/Backup/BackupDatabase.cs
@@ -385,6 +385,15 @@ namespace Duplicati.Library.Main.Operation.Backup
             );
         }
 
+        public Task RefreshQueryStatisticsAsync(CancellationToken cancellationToken)
+        {
+            return RunOnMainAsync(async () =>
+                await m_database
+                    .RefreshQueryStatisticsAsync(cancellationToken)
+                    .ConfigureAwait(false)
+            );
+        }
+
         public Task RemoveRemoteVolumeAsync(string remoteFilename, CancellationToken cancellationToken)
         {
             return RunOnMainAsync(async () =>
--- a/Duplicati/Library/Main/Operation/BackupHandler.cs
+++ b/Duplicati/Library/Main/Operation/BackupHandler.cs
@@ -215,6 +215,15 @@ namespace Duplicati.Library.Main.Operation
 
                     result.Dryrun = options.Dryrun;
 
+                    // A backup that was killed before it could close leaves the query statistics
+                    // describing the database as it was when that run started, which is enough to
+                    // make the planner choose join orders that never finish. Everything below
+                    // reads tables that are affected, starting with the longest-path probe, so
+                    // this has to come before any of it.
+                    await database
+                        .RefreshQueryStatisticsAsync(result.TaskControl.ProgressToken)
+                        .ConfigureAwait(false);
+
                     // Check the database integrity
                     await Utility.UpdateOptionsFromDbAsync(database, options, result.TaskControl.ProgressToken)
                         .ConfigureAwait(false);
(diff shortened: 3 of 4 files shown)


ts678:
I don't know how method below compares to this way, but before I noticed this I was testing with:

[PRAGMA optimize(MASK);](https://www2.sqlite.org/pragma.html#pragma_optimize)

> 0x10000 	Check the size of all tables, not just tables that have been recently used, to see if any have grown and shrunk significantly and hence might benefit from being re-analyzed. Off by default.

> Applications with long-lived database connections should run "PRAGMA optimize=0x10002" when the database connection first opens

I add that with my RUN.bat script by adding `set CUSTOMSQLITEOPTIONS_DUPLICATI=optimize=0x10002`

Profiling log output of selected lines shows it fixes the second backup after kill due to first hanging:

```
2026-08-16 13:27:01 -04 - [Verbose-Duplicati.Library.SQLiteHelper.SQLiteLoader-CustomSQLiteOption]: Setting custom SQLite option 'optimize=0x10002'.
2026-08-16 13:27:01 -04 - [Profiling-Timer.Begin-Duplicati.Library.Main.Database.ExtensionMethods-ExecuteNonQueryAsync]: Starting - ExecuteNonQueryAsync: PRAGMA optimize
2026-08-16 13:27:01 -04 - [Profiling-Timer.Finished-Duplicati.Library.Main.Database.ExtensionMethods-ExecuteNonQueryAsync]: ExecuteNonQueryAsync: PRAGMA optimize took 0:00:00:00.000
2026-08-16 13:27:03 -04 - [Verbose-Duplicati.Library.SQLiteHelper.SQLiteLoader-CustomSQLiteOption]: Setting custom SQLite option 'optimize=0x10002'.
2026-08-16 13:27:19 -04 - [Profiling-Timer.Begin-Duplicati.Library.Main.Operation.Backup.UploadRealFilelist-UploadNewFileset]: Starting - Uploading a new fileset
2026-08-16 13:27:20 -04 - [Profiling-Timer.Finished-Duplicati.Library.Main.Operation.Backup.UploadRealFilelist-UploadNewFileset]: Uploading a new fileset took 0:00:00:00.253
2026-08-16 13:27:22 -04 - [Profiling-Timer.Begin-Duplicati.Library.Main.Database.ExtensionMethods-ExecuteNonQueryAsync]: Starting - ExecuteNonQueryAsync: PRAGMA optimize
2026-08-16 13:27:22 -04 - [Profiling-Timer.Finished-Duplicati.Library.Main.Database.ExtensionMethods-ExecuteNonQueryAsync]: ExecuteNonQueryAsync: PRAGMA optimize took 0:00:00:00.000
2026-08-16 13:27:23 -04 - [Verbose-Duplicati.Library.SQLiteHelper.SQLiteLoader-CustomSQLiteOption]: Setting custom SQLite option 'optimize=0x10002'.
2026-08-16 13:27:23 -04 - [Profiling-Timer.Begin-Duplicati.Library.Main.Database.ExtensionMethods-ExecuteNonQueryAsync]: Starting - ExecuteNonQueryAsync: PRAGMA optimize
2026-08-16 13:27:23 -04 - [Profiling-Timer.Finished-Duplicati.Library.Main.Database.ExtensionMethods-ExecuteNonQueryAsync]: ExecuteNonQueryAsync: PRAGMA optimize took 0:00:00:00.000
2026-08-16 13:27:23 -04 - [Verbose-Duplicati.Library.SQLiteHelper.SQLiteLoader-CustomSQLiteOption]: Setting custom SQLite option 'optimize=0x10002'.
2026-08-16 13:27:24 -04 - [Profiling-Timer.Begin-Duplicati.Library.Main.Database.ExtensionMethods-ExecuteNonQueryAsync]: Starting - ExecuteNonQueryAsync: PRAGMA optimize
2026-08-16 13:27:24 -04 - [Profiling-Timer.Finished-Duplicati.Library.Main.Database.ExtensionMethods-ExecuteNonQueryAsync]: ExecuteNonQueryAsync: PRAGMA optimize took 0:00:00:00.000
```

JamBalaya56562:
@ts678 That mask is the answer, thank you — I have replaced what I had with it.

I had been running `ANALYZE` behind a hand-written staleness check, which is a worse version of what `0x10002` already does: SQLite decides which tables have moved far enough, and it costs **0.0040 s** on a database that needs nothing, against **0.0523 s** for mine. The check is gone.

It also corrects something I asserted in this description. I had written that `PRAGMA optimize` does nothing inside a transaction, which is only true of the bare form — I never tried it with a mask.

```
inside a transaction:  PRAGMA optimize          0.0003 s   unchanged
                       PRAGMA optimize=0x10002  0.3655 s   updated
```

On setting it at connection open only, I tried it on master with `CUSTOMSQLITEOPTIONS_DUPLICATI=optimize=0x10002` and the case the issue was reported from — one file, then 20,000:

| | |
|---|---|
| completion | none in 25 minutes |
| slow query warnings | 16 |
| longest single query | still running after 975.9 s |

Which makes sense: when the connection opens the tables still hold the handful of rows the statistics describe, so there is correctly nothing to do, and the growth happens afterwards. It does fix the run *after* a killed one, which is what your log shows.

So the PR now runs `PRAGMA analysis_limit=400; PRAGMA optimize=0x10002` at both points, and there is a test that fails if the mask is ever dropped — that change looks harmless in a diff. Same scenario: **5m52s, no slow query warnings**.

Incidentally the query that would not finish in that run was the length check in `VerifyConsistency`, not the fileset query this was reported against.

PR #7190, opened 2026-08-20
Fix similarly named path inclusion

### LINKED ISSUE #7189: backup quietly skips source folder if path start is also a source file

---

## Environment info

-   **Duplicati version**: 2.3.0.4_stable_2026-07-09
-   **Operating system**: Windows 10 and RHEL 10.2
-   **Backend**: N/A

## Description

Duplicati studies source selection to possibly remove subfolders of folders that are in source.
This plan is led into an incorrect folder removal if there is a source file matching folder prefix.
There are several different cases (please check all) handled that all produce similar messages:

> `2026-08-19 09:41:18 -04 - [Verbose-Duplicati.Library.Main.Controller-RemovingSubfolderSource]: Removing source "C:\backup source\main\main.lrcat-data\" because it is a subfolder or subfile of "C:\backup source\main\main.lrcat"`

for example, and I'm not sure if a subfile can have the same error. Please review code cited at:

[Files/folders not being backed up when selected together](https://forum.duplicati.com/t/files-folders-not-being-backed-up-when-selected-together/22627)

There is another seeming case of this at

[Issue backing up files and directories in Linux when directory ends in .d](https://forum.duplicati.com/t/issue-backing-up-files-and-directories-in-linux-when-directory-ends-in-d/22601)

```
Example:

/etc/ssh/sshd_config.d/

/etc/ssh/sshd_config
```

## Steps to reproduce

1. Add to Source the path to a file
2. Add to Source the path to a folder which starts with above path, then has more
3. Backup

-   **Actual result**:
Folder is not backed up, with no clues in regular log. Verbose job log shows loss.
-   **Expected result**:
Folder is backed up.

kenkendk:
The code that checks for multiple inclusions of the same sub-folder was not verifying that the source was actually a folder.

When including a file AND a folder that were similar, the file path would be treated as a folder that includes the folder, and the folder would be omitted.

The problem was the duplicate-path detection logic, it was designed to capture this
```
/etc/ssh/ <- folder
/etc/ssh/stuff/ <- folder
```

Here, there is no need to include `etc/ssh/stuff`. The problem is this setup:
```
/etc/ssh <- file
/etc/ssh/stuff/ <- folder
```

This triggers the same logic, but naturally `/etc/ssh/stuff` cannot be inside the file, so it should not be removed from the source list.

This fixes #7189

### DIFF
--- a/Duplicati/Library/Main/Controller.cs
+++ b/Duplicati/Library/Main/Controller.cs
@@ -1352,7 +1352,10 @@ namespace Duplicati.Library.Main
             //Sanity check for multiple inclusions of the same folder
             for (int i = 0; i < sources.Count; i++)
                 for (int j = 0; j < sources.Count; j++)
-                    if (i != j && sources[i].StartsWith(sources[j], Library.Utility.Utility.ClientFilenameStringComparison) && sources[i].EndsWith(Util.DirectorySeparatorString, Library.Utility.Utility.ClientFilenameStringComparison))
+                    if (i != j
+                        && sources[i].StartsWith(sources[j], Library.Utility.Utility.ClientFilenameStringComparison)
+                        && sources[j].EndsWith(Util.DirectorySeparatorString, Library.Utility.Utility.ClientFilenameStringComparison)
+                        && sources[i].EndsWith(Util.DirectorySeparatorString, Library.Utility.Utility.ClientFilenameStringComparison))
                     {
                         if (filter != null)
                         {
--- a/Duplicati/Library/Main/Controller.cs
+++ b/Duplicati/Library/Main/Controller.cs
@@ -141,8 +141,9 @@ namespace Duplicati.Library.Main
 
             return await RunActionAsync(new BackupResults(), inputsources, inputFilter, false, static async config =>
             {
+                var (expandedSources, filter) = ExpandInputSources(config.Paths, config.Filter, config.Options);
                 using (var h = new Operation.BackupHandler(config.Options, config.Result))
-                    await h.RunAsync(ExpandInputSources(config.Paths, config.Filter, config.Options), config.BackendManager, config.Filter)
+                    await h.RunAsync(expandedSources, config.BackendManager, filter)
                         .ConfigureAwait(false);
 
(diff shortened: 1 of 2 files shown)


ts678:
Is the `subfile` part of `because it is a subfolder or subfile of` possible? If not, why say that?

kenkendk:
@ts678 I updated the description with an example.

ts678:
@kenkendk 

It's just a terminology thing to try to avoid obscure or incorrect words that muddy up the primary concept.

There is commonly a difference seen between a folder and a file, although at an OS internals level, a folder might be viewed as a type of file. I would still prefer not calling a folder a file. Subfile seems a rare concept, according to research (where it has a different meaning). Personally, I'd call this a subfile. Maybe I'm wrong:

`/etc/ssh/` is a folder
`/etc/ssh/sshd_config` is a file, but is it also a "subfile" of `/etc/ssh`? Arguably everything descends from /.
It might be bad if this sort of subfile was removed, even if it's sometimes redundant. Sometimes it's not...

I found a path to another issue by testing to see if my maybe-wrong "subfile" got removed by a backup.
It does not, but one has to fight the UI to even get it in, then when run it gets confused and reports this:

> `2026-08-20 09:52:31 -04 - [Warning-Duplicati.Library.Main.Operation.Backup.FileBlockProcessor.FileEntry-FileProcessingFailed]: Failed to process path: C:\backup source\ssh\sshd_config SqliteException: SQLite Error 19: 'UNIQUE constraint failed: FilesetEntry.FilesetID, FilesetEntry.FileID'.`

Citations:

`https://www.computerlanguage.com/results.php?definition=files+vs.+folders`

`https://www.merriam-webster.com/dictionary/subfolder`

`https://www.merriam-webster.com/dictionary/subfile`

kenkendk:
@ts678 yes, the error you found is exactly why the check is there in the first place.

If you somehow manage to get the same source multiple times, Duplicati will happily enumerate twice, but it cannot store the paths twice, so you get errors in that case.

I will update this PR to also filter out the forced inclusion of a file, if it exists in a folder that is included.

PR #7203, opened 2026-08-23
Replace missing metadata with empty metadata, not with another blockset

JamBalaya56562:
This is another way at what draft [#7052](https://github.com/duplicati/duplicati/pull/7052) is reaching for. **The aim there is right** — a repair should always have empty metadata to point at — and this arrives at it from the other end: rather than putting the block in every backup in advance, the one command that needs it writes it when it needs it.

## What a file is handed today

`purge-broken-files` keeps a file whose data survived but whose metadata is gone, and gives it replacement metadata. It describes that as empty metadata, and the error it throws when it cannot find any calls it "an empty metadata blockset". What a file actually gets is whatever metadata blockset happens to be smallest:

```csharp
// LocalDatabase.cs, GetEmptyMetadataBlocksetIdAsync
// No empty block found, pick the smallest one
    ...
    AND "Blockset"."Length" > 0
ORDER BY "Blockset"."Length" ASC
LIMIT 1
```

The XML doc says so outright: *"If no empty blockset is found, it returns the ID of the smallest blockset that is not in the given block volume IDs."*

**That step is not a rare fallback.** Measured on an ordinary backup, the database holds:

```
zero-length blocksets     = 0
empty-metadata blocksets  = 0
```

so the first two steps find nothing and the third one answers every time.

## What that does, end to end

A file that keeps its data and loses only the volume that held its metadata:

```
before purge   version 1 of "a" -> blockset 2   (its own metadata)
               version 2 of "a" -> blockset 4   (its own metadata, volume gone)
after purge    version 2 of "a" -> blockset 2
```

**The newest version came back wearing the previous version's metadata**, a timestamp it never had. With more than one file in the backup, `ORDER BY Length ASC` reaches another file's metadata just as easily — its timestamps, its attributes, and on Unix its ownership and permissions.

## The change

The empty metadata is a constant; the caller computes it two lines above and then throws it away. So:

- `GetEmptyMetadataBlocksetIdAsync` keeps its behaviour for its other caller, and the two steps that answer with *real* empty metadata are split out as `FindEmptyMetadataBlocksetIdAsync`.
- `purge-broken-files` asks the strict one, and when there is nothing to find it writes the five bytes to the destination as a volume of its own and points the file at that.

Two details of the write are deliberate, and both were found by measuring rather than by reading:

**It does not commit or wait for the backend.** The first version did, and the later `ReplaceMetadata` then matched **0 rows** — so the substitution would have quietly survived the change meant to remove it. It runs inside the transaction the purge already holds.

**It chooses the row id instead of letting the database pick.** Removing the missing volumes deletes `Block` rows and leaves the `BlocksetEntry` rows that pointed at them, and SQLite hands those ids out again:

```
metadataset 3  blockset 4  liveblocks 1  entries 1  blockids 65  sizes 5
```

Blockset 4 is the blockset whose block is gone, and it had reattached to the five-byte block that had just been inserted on the reused id — so it stopped looking broken. Deleting those dangling entries instead is not an option: they are how a file with missing data is recognised, and dropping them makes the consistency check report the wrong size for the file (`actual size 99332, dbsize 1024`).

## Tests

`ReplacementMetadataTests` covers what the file is handed, that its data and the older version are untouched, and that a dry run writes nothing. The first one fails without this change:

```
The file was handed a metadata blockset that belongs to something else
Expected: "qiXpeARtaA74dA2Dfm3lvB4qLcYInb2hASVEtTjVP2U="
But was:  "jKUnvj12LgA4Y9R295OyRLbhKbXqHPDIy7kWjnxYG3A="
```

The state it needs — data intact, metadata gone — is built by backing the file up, touching only its timestamp and backing it up again, which puts one metadata block in a volume of its own, then deleting that volume.

Checked: `dotnet build Duplicati.slnx -c Debug --no-incremental` — 0 errors and the same 3 warnings as master. `ReplacementMetadataTests`, `RepairWithMissingRemoteFiles`, `Issue5987`, `Issue6504` and `RepairHandlerTests` run together — **58 passed, 0 failed**. The whole suite was run on a fork before opening this: unit tests, integration tests and Playwright, **green on ubuntu, macOS and Windows**.

## Noticed, not changed

- **`FixEmptyMetadatasetsAsync` in the repair path uses the same third step.** It runs in `RunRepairCommonAsync`, which has no backend to upload to, so it cannot write the constant the way this does. Left alone; it needs a decision this PR should not make. In practice its precondition means a zero-length blockset exists, so the second step usually answers there.
- **Why not seed the block in every backup, as #7052 does.** Measured across a database recreate:
  ```
  after backup    Block rows for the empty-metadata hash = 1,  Metadataset = 4
  after recreate  Block rows for the empty-metadata hash = 0,  Metadataset = 3
  ```
  The seed is a `Metadataset` no file references, so nothing in the dlist brings it back and a recreate drops it, leaving the block orphaned in the dblock and the guarantee gone. It also fails 22 unit tests, because a block with no source file cannot be rebuilt by repair.

### DIFF
--- a/Duplicati/Library/Main/Database/Local/LocalDatabase.cs
+++ b/Duplicati/Library/Main/Database/Local/LocalDatabase.cs
@@ -3349,15 +3349,20 @@ namespace Duplicati.Library.Main.Database.Local
         }
 
         /// <summary>
-        /// Returns the ID of an empty metadata blockset. If no empty blockset is found, it returns the ID of the smallest blockset that is not in the given block volume IDs.
-        /// If no such blockset is found, it returns -1.
+        /// Returns the ID of a blockset that really is empty metadata, or -1 when the database
+        /// holds none that survives the given block volumes.
         /// </summary>
+        /// <remarks>
+        /// This answers only with an empty metadata blockset or a zero-length one. It never hands
+        /// back a blockset that belongs to some other file, which is what
+        /// <see cref="GetEmptyMetadataBlocksetIdAsync"/> falls back to.
+        /// </remarks>
         /// <param name="blockVolumeIds">The volume ids to ignore when searching for a suitable metadata block.</param>
         /// <param name="emptyHash">The hash of the empty blockset.</param>
         /// <param name="emptyHashSize">The size of the empty blockset.</param>
         /// <param name="token">A cancellation token to cancel the operation.</param>
-        /// <returns>A task that when awaited contains the ID of the empty metadata blockset, or -1 if no suitable blockset is found</returns>
-        public async Task<long> GetEmptyMetadataBlocksetIdAsync(IEnumerable<long> blockVolumeIds, string emptyHash, long emptyHashSize, CancellationToken token)
+        /// <returns>A task that when awaited contains the ID of the empty metadata blockset, or -1 if there is none</returns>
+        public async Task<long> FindEmptyMetadataBlocksetIdAsync(IEnumerable<long> blockVolumeIds, string emptyHash, long emptyHashSize, CancellationToken token)
         {
             await using var tmptable = await TemporaryDbValueList.CreateAsync(this, blockVolumeIds, token)
                 .ConfigureAwait(false);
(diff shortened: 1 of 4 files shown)


kenkendk:
I merged a rather large rewrite that addresses a similar issue. I *think* the underlying issue is now fixed so we should close this PR, but wanted to get your reading on it first.

JamBalaya56562:
Your reading is right, and thank you for checking first — this can be closed.

Everything here is in master, down to the names and the comments: `FindEmptyMetadataBlocksetIdAsync` (`LocalDatabase.cs:3365`), `StoreEmptyMetadataAsync` (`PurgeBrokenFilesHandler.cs:250`), `AddEmptyMetadataBlocksetAsync` (`LocalListBrokenFilesDatabase.cs:466`), and `ReplacementMetadataTests` with the same three cases.

I built master and ran them rather than reading only: 3/3 green. `TheReplacementIsEmptyMetadataAndNotAnotherBlocksetAsync` is the one that fails while the bug is present, so the underlying issue is confirmed fixed, not assumed.

The part I would have re-sent if it had been lost survived as well — the explicitly chosen block id in `AddEmptyMetadataBlocksetAsync`. `RemoveMissingBlocksAsync` deletes `Block` rows but leaves the `BlocksetEntry` rows pointing at them, and SQLite hands a deleted row id out again on the next insert, so a block landing on one of those ids silently reattaches a blockset that is supposed to look broken. That guard is there.

The third tier in `GetEmptyMetadataBlocksetIdAsync` is still at `LocalDatabase.cs:3451`, which is as intended: `purge-broken-files` no longer reaches it, and the repair path that also calls it keeps its old behaviour.

Closing.

PR #7204, opened 2026-08-23
Put the OneDrive folder name into the url encoded

JamBalaya56562:
The other half of what [#7202](https://github.com/duplicati/duplicati/pull/7202) reported and left alone. That one said this call needed a real account to settle. It has now been settled against one.

## The symptom

**A folder whose name contains a percent sign or a hash cannot be reached**, and a folder named `a b` and one named `a%20b` are the same request.

## Two things add up to it

**The path is read out of the destination url twice.** `RelaxedUri` decodes both halves as it parses:

```csharp
var h = UrlEncoding.UrlDecode(m.Groups["hostname"].Success ? m.Groups["hostname"].Value : "");
...
var p = UrlEncoding.UrlDecode(m.Groups["path"].Success ? m.Groups["path"].Value : "");
```

and `GetRootPathFromUrlAsync` decodes `HostAndPath` again.

**The result is then concatenated into the request url as it stands.** Graph reads the path in a url percent-decoded, so the name of the folder and the text in the url are not the same string. Measured on the backend, no account involved:

```
onedrivev2://a%20b     ->  root [a b]
onedrivev2://a%2520b   ->  root [a b]        two different folders, one request

built   .../root:/a#b:/children
sent    fragment [#b:/children]              the rest of the url never leaves
```

**Removing the second decode alone does not fix it.** `a%2520b` would then arrive as `a%20b` and still be concatenated unescaped, which Graph reads back as `a b`. Both halves are needed, which is why #7202 stopped here.

## Checked against a real account

Two folders were made in one drive, one named `a%20b` and one named `a#b`, and addressed four ways:

| url | | |
| --- | --- | --- |
| `.../root:/temp/a%2520b` | **200** | `"name":"a%20b"` |
| `.../root:/temp/a%20b` | **404** | names `a b`, which is not there |
| `.../root:/temp/a%23b` | **200** | `"name":"a#b"` |
| `.../root:/temp/a#b` | **404** | cut at the fragment |

So Graph decodes the path once. That is what `ToUrlPath` now writes: separators stay separators and every segment is escaped, the same shape `RelaxedUri.AsString` already uses.

Afterwards, the url this backend builds was sent to that account again and landed where it should:

```
onedrivev2://temp/a%2520b  ->  .../root:/temp/a%2520b  ->  200  "name":"a%20b"
onedrivev2://temp/a%23b    ->  .../root:/temp/a%23b    ->  200  "name":"a#b"
```

## Where the encoding goes

**`RootPath` stays decoded on purpose.** `CreateFolderAsync` splits it and uses each part as a `DriveItem` name:

```csharp
foreach (string folder in this.RootPath.Split(new[] { '/' }, StringSplitOptions.RemoveEmptyEntries))
    ...
    Name = folder,
```

Encoding it there would create a folder actually called `a%2520b`. So the encoding happens where the path becomes a url, and the eight places that built an item url by hand now go through one `RootItemUrl` — which is also the method the tests drive, so the seam is the one production uses.

The three places that address a path from the drive root rather than from the configured folder call `ToUrlPath` directly.

## What changes for a user

| destination url | before | after |
| --- | --- | --- |
| `onedrivev2://a%2520b` | folder `a b` | folder `a%20b` |
| `onedrivev2://a%23b` | request cut at the `#` | folder `a#b` |
| `onedrivev2://My%2BFolder` | folder `My Folder` | folder `My+Folder` |
| `onedrivev2://100%2525done` | folder `100%done` | folder `100%25done` |
| `onedrivev2://Backup/Sub` | unchanged | unchanged |
| `onedrivev2://My+Folder` | folder `My Folder` | **unchanged**, see below |

An encoded plus is fixed for the same reason as the Dropbox one: `%2B` became `+` on the first decode and a space on the second. **A plus written literally still becomes a space** — that happens in the shared parser, and this backend cannot move to `System.Uri` because the first segment of its url is a folder name, which is case sensitive. Same classification as [#7145](https://github.com/duplicati/duplicati/pull/7145).

**Existing destinations keep working.** The names Duplicati writes are made of characters that encode to themselves, so the request for a dblock is byte for byte what it was; there is a test holding that. A user whose folder name contains a percent sequence was reaching a different folder before, and will now reach the one they named.

## Tests

`OneDriveUrlTests` drives `RootItemUrl`, the method the operations use. It fails on the previous code in 8 cases, including the collision:

```
TheFolderNameReachesTheUrlEncoded("onedrivev2://a%2520b", ...)
  Expected: "/v1.0/me/drive/root:/a%2520b"
  But was:  "/v1.0/me/drive/root:/a b"
```

One test asserts directly that `a%20b` and `a%2520b` no longer share a url.

Checked: `dotnet build Duplicati.slnx -c Debug --no-incremental` — 0 errors and the same 3 warnings as master. `OneDriveUrlTests`, `DropboxPathTests`, `BoxEntryLookupTests`, `UriUtilityTests` and `UrlEncodingTests` run together — **122 passed, 0 failed**.

## Worth knowing

`LiveTests` has targets for Azure, B2, Box, CIFS, CloudStack, Drime and Dropbox, but **none for OneDrive**, so this path has not been exercised against a real drive by CI. That is part of why it went unnoticed.


kenkendk:
I discovered some of the same independently from another angle, so the two PRs were tangled.
I have updated with a consolidated version.

PR #7245, opened 2026-08-31
GCS should detect delete of missing file

kenkendk:
This adds handling for 404 messages on delete, similar to how other places handle it.


kenkendk:
Bug is real, but not an eventually consistent issue

PR #7282, opened 2026-09-05
Keep every copy of a duplicated Google Drive name, and list it once

### LINKED ISSUE #5846: Improve handling of duplicated filenames
In some backends, Google Drive in particular, it is possible to have multiple files with the same name. Internally, Google Drive uses a unique (permanent) id for each file and treats filenames as "labels".

The implementation for Google Drive is using a dictionary that "resolves" filenames, so they uniquely match one entry. But in the case where there are duplicated filenames, the results are unpredictable, as one of them will be returned, but which depends on calling order.

Also, when iterating the folder, it will return all filenames, including possible duplicates, which breaks the file verification code.

JamBalaya56562:
Google Drive keeps a permanent id for every file and treats the name as a label, so two files in
one folder can carry the same name. [#5846](https://github.com/duplicati/duplicati/issues/5846)
describes what that costs. Looking into it turned up two separate faults in `ListAsync`, both of
which the new tests pin down.

## 1. Every copy after the first was dropped from the cache

```csharp
Array.Resize(ref lst, lst.Length + 1);
lst[lst.Length - 1] = n;
```

`lst` is the local from `TryGetValue`. `Array.Resize` allocates a new array and assigns it **to the
local**; the dictionary slot still holds the one-element array it already had. The second copy was
written into an array nobody kept.

The cache is what the rest of the backend resolves names through, so:

- **`DeleteAsync` removed one copy and left the other on the destination.** It reads through
  `GetFileEntriesAsync`, which returns the cached array on a hit. The new test lists and then
  deletes, and before this change it recorded exactly one delete — of the **older** id:
  ```
  deleted: id-old
  ```
- **`GetAsync` downloaded the wrong copy.** It primes the cache from the listing and then takes
  `OrderByDescending(x => x.createdDate).First()`, but over a single surviving entry — so the
  newest-wins rule silently resolved to whichever copy Drive listed first.

## 2. The listing reported the name twice, which stops everything

`ListAsync` yielded every entry, so a duplicated name reached
`FilelistProcessor.VerifyRemoteListAsync`:

```csharp
if (doubles.Count > 0)
{
    var s = "Found remote files reported as duplicates, either the backend module is broken or you need to manually remove the extra copies.…";
    throw new RemoteListVerificationException(s, "DuplicateRemoteFiles");
}
```

Backup, restore and repair all stop, with a message that blames the backend module and asks the
user to delete the extra copies by hand. On Drive a repeated name is not a fault, so that is not an
answer.

## The change

Both are fixed the way two other backends already do it:

| backend | how it lists a duplicated name |
|---|---|
| `MegaBackend.cs:212` | `n.OrderByDescending(x => x.ModificationDate).First()` |
| `B2.cs:552` | `x.Value.OrderByDescending(y => y.UploadTimestamp).First()` |
| **GoogleDrive, now** | `entries.OrderByDescending(x => x.createdDate).First()` |

The cache accumulates every copy — that is what lets a delete take all of them — and the listing
reports one entry per name: the newest, which is the copy a read fetches, so the listing and a
subsequent `GetAsync` agree.

Folders are untouched and still go out as they are found; an ambiguous folder name is already
rejected outright by `GetFolderIdAsync`.

`FilelistProcessor` is deliberately left alone. On a backend where a repeated name really is a
fault, that error is the right answer, and `Issue1282` tests it — it stays green.

## Red, then green

`Duplicati/UnitTest/GoogleDriveDuplicateNameTests.cs` is new. It drives the backend against stubbed
Drive API responses through a small `HttpMessageHandler`, so it needs no network and no
credentials — the same shape as the existing `BoxEntryLookupTests`. That needed a seam:
`[assembly: InternalsVisibleTo("Duplicati.UnitTest")]` and an `internal` constructor taking an
`HttpClient`, mirroring `BoxBackend`. The public constructors are unchanged, so the
`BackendLoader`'s parameterless activation is untouched.

| test | before | after |
|---|---|---|
| a name held twice is listed once | **red** — listed twice | green |
| the copy that is listed is the newest one | **red** | green |
| a delete after a listing removes every copy | **red** — one delete, the older id | green |
| names that appear once are unchanged | green | green |

The last one is the regression guard: it is the shape where the current behaviour was already
right, and it shows the change does not reach past the fault.

Full solution build: 0 errors, and the same two pre-existing `MSB3246` warnings as before.

## Not measured

- **No real Google Drive.** I have no credentials, so the API responses are stubbed. The stub feeds
  the backend's own `ListFolder` parsing, so the shape it produces is the shape the code already
  expects, but a difference between my stub and the live API would not show up here.
- **Duplicates are now collapsed silently.** The extra copy stays on Drive until a `PutAsync` or
  `DeleteAsync` touches that name, and nothing tells the user it is there. This file has no logging
  at all today, so adding a log line meant introducing that; I left it out rather than widen the
  change, but say the word and I will add one.
- `IFolderEnabledBackend.EnumerateAsync` — the folder browser — still yields every entry. It feeds
  a file picker rather than the verification, so a repeated row there is cosmetic. Left alone for
  the same reason.

## About the issue

The other half of #5846 — "the results are unpredictable, as one of them will be returned, but
which depends on calling order" — is genuinely about this: `GetAsync` looked deterministic but was
not, because of the cache fault above. `PutAsync` and `DeleteAsync` already handled `N > 1`
correctly once they could see it. I have left the `Fixes` keyword off so you can judge whether the
issue is fully answered; the parts I could reach are covered.

Fixes #5846

### DIFF
--- a/Duplicati/Library/Backend/GoogleServices/GoogleDrive.cs
+++ b/Duplicati/Library/Backend/GoogleServices/GoogleDrive.cs
@@ -28,6 +28,8 @@ using System.Net.Http.Json;
 using System.Runtime.CompilerServices;
 using System.Text;
 
+[assembly: InternalsVisibleTo("Duplicati.UnitTest")]
+
 namespace Duplicati.Library.Backend.GoogleDrive
 {
     // ReSharper disable once UnusedMember.Global
@@ -68,6 +70,22 @@ namespace Duplicati.Library.Backend.GoogleDrive
             m_teamDriveID = options.GetValueOrDefault(TEAMDRIVE_ID);
             m_oauth = new OAuthHelperHttpClient(auth.AuthId!, this.ProtocolKey, auth.OAuthUrl) { AutoAuthHeader = true };
             m_filecache = new Dictionary<string, GoogleDriveFolderItem[]>();
+        }
+
+        /// <summary>
+        /// Creates a backend that talks through the supplied <see cref="HttpClient"/>,
+        /// so the Drive API responses can be decided by a test
+        /// </summary>
+        /// <param name="url">The backend url</param>
+        /// <param name="options">The options to use</param>
+        /// <param name="httpClient">The client to send the requests through</param>
+        internal GoogleDrive(string url, Dictionary<string, string?> options, HttpClient httpClient)
+            : this(url, options)
+        {
+            var auth = AuthIdOptionsHelper.Parse(options)
+                .RequireCredentials(TOKEN_URL);
+
+            m_oauth = new OAuthHelperHttpClient(auth.AuthId!, this.ProtocolKey, auth.OAuthUrl, httpClient) { AutoAuthHeader = true };
         }
 
         private async Task<string> GetFolderIdAsync(string path, bool autocreate, CancellationToken cancelToken)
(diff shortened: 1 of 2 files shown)


kenkendk:
Fixes https://github.com/duplicati/duplicati/issues/5846

Issue #6461, opened 2025-08-07
Backup can get stuck after requesting abort


kenkendk:
Suggested repro steps:

1. Create a non-trivial data set (say 50GB to get some processing time)
2. Start the backup
3. Wait a few seconds
4. Click “Stop now”
5. Backup hangs but server is responsive

duplicatibot:
This issue has been mentioned on **Duplicati**. There might be relevant details there:

https://forum.duplicati.com/t/stop-button-not-working-in-2-1-0-120-2-1-0-125/20974/4

ts678:
Since I had both .124 and .125 stuck for days, I tried sacrificing .124 with a TrayIcon menu Quit, and it wouldn't. Days later I killed it.

PR #7325, opened 2026-09-15
Stop waiting for a backend call when the operation is aborted

JamBalaya56562:
## What happens

"Stop now" on a backup whose upload is stuck in a backend call that does not observe its cancellation
token never returns. This is one mechanism behind #6461 ("Backup can get stuck after requesting
abort"); it is not claimed to be the only one.

Reproduced with a test backend that wraps `file://` and stalls every upload, in two variants: the stall
observes the token, or it ignores it (the way a transfer stuck in a socket write, or a stream copy that
was never handed the token, behaves). Two uploads in flight (`asynchronous-upload-limit=2`), 40 files
of 64 KB in 100 KB volumes, `AbortAsync()` once both slots are stuck:

| Upload | `BackupAsync` returns after the abort |
|---|---|
| observes the token | 0.2 s |
| ignores the token | **not within 2 minutes** |

### Why

A heap dump of the stuck process (`dotnet-dump`, `dumpasync`) shows the chain:

```
DataBlockProcessor  → BackendManager.PutAsync → QueueTaskAsync → CoCoL.Channel<PendingOperationBase>.WriteAsync
StreamBlockSplitter → DataBlock.AddBlockToOutputAsync → CoCoL.Channel<DataBlock>.WriteAsync
FileBlockProcessor  → StreamBlock.ProcessStreamAsync → CoCoL.Channel<StreamBlock>.WriteAsync
BackupHandler.RunMainOperationAsync → Task.WhenAll(...)
```

`BackendManager.Handler.ExecuteAsync<TResult>` awaits the backend call unconditionally
([BackendManager.Handler.cs:590-615](https://github.com/duplicati/duplicati/blob/df30f7032e032850eca2da3b9e51ecc255dfb352/Duplicati/Library/Main/Backend/BackendManager.Handler.cs#L590-L615)).
When the call does not end on cancellation, the upload task never completes, `EnsureAtMostNActiveTasksAsync`
keeps the handler loop waiting for a free slot, the request channel is not read, and every producer
upstream blocks on a channel write that has no token. The termination path that gives active transfers
one second and then abandons them (`WaitForPendingItemsAsync`) is in the loop's `finally`, which is only
reached once the loop ends - and it never does.

## The change

`ExecuteAsync<TResult>` awaits the backend call through `UntilCancelledAsync`: `Task.WhenAny(call,
cancellation)`. When the token fires first, the call is left to end on its own (its outcome is observed
so it does not surface as an unobserved exception), the backend instance is marked not-reusable and
disposed by the existing `using`, and an `OperationCanceledException` goes to `ExecuteWithRetryAsync`,
which already turns a cancelled token into `op.SetCancelled()`. The awaiting `PutAsync` is released,
the producers behind it unwind through the channels, and `BackupAsync` ends with the cancellation.

Nothing changes for a backend that observes the token: its call ends first and the result is returned
as before. Nothing changes without a stop: the token is never cancelled.

## Red to green

`--filter "FullyQualifiedName~AbortedBackupWithStalledUploads"`:

| Test | Before |
|---|---|
| `AbortedBackupTests.AbortedBackupWithStalledUploadsReturnsAsync(True)` (upload ignores the token) | fails: `The abort did not make the backup return within two minutes` |
| `AbortedBackupTests.AbortedBackupWithStalledUploadsReturnsAsync(False)` (upload observes the token) | passes before and after, in 0.2-0.3 s |

Red then green on Windows and on Linux (WSL). `TestCategory=Disruption`, the Drime/Filejump delete-retry
tests, `DeprecatedUploadOptionTests` and the restore abort/stop tests are green (61).

Once the backup has returned, the test releases the stalled uploads and waits until they have left the
backend before it ends. The released uploads carry on into the `file://` backend and write into the
target folder after `BackupAsync` has returned; the teardown removes that folder, and on Windows a file
that is still open cannot be deleted. Measured: two uploads are still inside the backend when the test's
`finally` runs and finish about 60 ms later.

## Not covered

- The other awaits that a stop does not reach (a synchronous source-file read, a running SQLite
  statement) are unchanged; a backup stuck there still does not return. This PR addresses the backend
  call only, which is the one the dump showed and the one a test could pin down.
- The abandoned call keeps running until the destination gives up or the disposed backend makes it
  fail; the manager logs it as an active transfer at termination, as it already did for the one-second
  grace period.


JamBalaya56562:
The `Unit tests (windows-latest)` failure on [run 35349354054](https://github.com/duplicati/duplicati/actions/runs/35349354054/job/105613480876) was in this PR's own test, in its teardown:

```
TearDown : System.IO.IOException : The process cannot access the file 'duplicati-...dblock.zip.aes' because it is being used by another process.
```

With `uploadIgnoresCancellation = true`, the two abandoned uploads are released at the end of the test and then carry on into the file backend, which creates the dblock files in the target folder after the backup has already returned. The teardown deletes that folder at the same time; on Windows a file that is still open cannot be deleted, on Linux and macOS it can, so only the Windows job could fail. Measured locally: when the test method's `finally` runs, two uploads are still inside the backend and finish about 60 ms later — nothing waited for them.

[4916bc3a](https://github.com/duplicati/duplicati/commit/4916bc3ad1017edc5e7227ffaef5e6bf20bdc969) makes the test wait until the released uploads have left the backend before it returns (5/5 green locally on Windows). The production change is untouched.

(The `Unit tests (macos-latest)` red on the previous run was `TestMultiExceptionOnDlistAsync(False,False,3,True)`, the known DisruptionTests flake, not this PR.)

